PR #8, opened 2016-02-11
AppVeyor CI

davidpodhola:
Basic appveyor continous integration setup.
Appveyor still needs to be created on the master repository (see https://ci.appveyor.com/project/davidpodhola/fabel)


davidpodhola:
appveyor is running tests too
ugly search&replace "fix" for https://github.com/fsharp/FSharp.Compiler.Service/issues/529

alfonsogarciacaro:
Fantastic, thanks a lot for that! I gave Appveyor access to the master repo and it's working :) About the bug in FSharp.Compiler.Service, it seems they're already taking care of it so hopefully we can remove the hack soon.

rneatherway:
This should be fixed in the just-released 2.0.0.6. Let me know how it goes.

alfonsogarciacaro:
I updated FSharp.Compiler.Service, removed the hack and the AppVeyor build still works so it seems the problem is fixed :) Thanks a lot for your work, Robin!

rneatherway:
Great!

PR #12, opened 2016-02-16
[WIP] Generate more from .d.ts

davidpodhola:
Work In Progress, the code is not optimal and is not DRY.
Going through files like jquery.d.ts I added more to be generated.
Could you have a look @alfonsogarciacaro please? Thanks!


tools/typescript2fabel.js
```diff
@@ -273,10 +286,28 @@ function visitModule(node) {
...
+                break;
+            case ts.SyntaxKind.VariableStatement:
+				if (modules.length == 0 ) { modules.push( emptyModule ); }
+                getVariables(node).forEach(x =>
+                    emptyModule.properties.push(x));
+                break;
+            case ts.SyntaxKind.FunctionDeclaration:
+				if (modules.length == 0 ) { modules.push( emptyModule ); }
+                emptyModule.methods.push(getMethod(node, true));
+                break;
```
alfonsogarciacaro: Besides `Import` attribute, there'll be a `Global` attribute. Maybe we can use that for elements defined outside modules.

tools/typescript2fabel.js
```diff
@@ -266,6 +273,12 @@ function visitModule(node) {
             case ts.SyntaxKind.FunctionDeclaration:
                 mod.methods.push(getMethod(node, true));
                 break;
+            case ts.SyntaxKind.ModuleDeclaration:
+                modules.push(visitModule(node, modules));
+                break;
+            case ts.SyntaxKind.ClassDeclaration:
+                mod.interfaces.push(visitInterface(node));
+                break;
```
alfonsogarciacaro: Classes are a bit more complicated that interfaces because they can have constructors and static methods (and for that yo need a explicit reference to the class). We have to think yet how to translate properly. Luckily, they don't appear that often in definition files (I think).

tools/typescript2fabel.js
```diff
@@ -266,6 +273,12 @@ function visitModule(node) {
             case ts.SyntaxKind.FunctionDeclaration:
                 mod.methods.push(getMethod(node, true));
                 break;
+            case ts.SyntaxKind.ModuleDeclaration:
+                modules.push(visitModule(node, modules));
+                break;
```
alfonsogarciacaro: I wonder why you make the nested modules siblings of their parent? Does this work that way? Can you point a sample of the kind of modules this is trying to deal with?

tools/typescript2fabel.js
```diff
@@ -44,7 +44,8 @@ constructor:
 function printParameters(parameters) {
     function printParameter(x) {
         if (x.rest) {
-            var type = /ResizeArray<(.*?)>/.exec(x.type)[1] + "[]";
+			var execed = /ResizeArray<(.*?)>/.exec(x.type);
+            var type = (execed == null) ? "unknown" : execed[1] + "[]";
```
alfonsogarciacaro: It may be better to use "obj" here instead of "unknown"

tools/typescript2fabel.js
```diff
@@ -240,9 +241,15 @@ function visitInterface(node) {
             case ts.SyntaxKind.PropertySignature:
                 ifc.properties.push(getProperty(node));
                 break;
+            case ts.SyntaxKind.PropertyDeclaration:
+                ifc.properties.push(getProperty(node));
+                break;
```
alfonsogarciacaro: What's the difference between Property Signature and Declaration? (Also for methods) I'm still puzzled by TypeScript SyntaxKind elements (they're so much more complicated than the symbols exposed by F# compiler services!). I wrote this script mostly by trial and error using astexplorer.net with typescript parser.

alfonsogarciacaro:
Looks good, thanks! I wrote some comments :) I'm trying to think in a way to test the generated code, something similar to the tests that are usually included in DefinitelyTyped.

davidpodhola:
Thanks for the comments, will fix tomorrow.

tools/typescript2fabel.js
davidpodhola: Elements outside modules were in `Unknown` module, will be in `Global` module

tools/typescript2fabel.js
davidpodhola: For sure.

PR #14, opened 2016-02-19
Update README.md

forki:


alfonsogarciacaro:
Well spotted, thank you! Now that I see that I noticed I should have said "the designer of the language" instead of "developer". Would you mean changing that so a separate commit is not necessary? Cheers!

alfonsogarciacaro:
I had to fix a conflict with another PR so I did it myself :)

PR #24, opened 2016-02-21
Add checking for current and reserved keywords - Fix #22

### LINKED ISSUE #22: Typescript2Fable - Function names
In typescript definition file there may exist functions with not valid F# name. For such functions name should be enclosed into double ticks.

Krzysztof-Cieslak:


alfonsogarciacaro:
Great work, thank you!

PR #28, opened 2016-02-22
Add brackets to fix name check

davidtme:


alfonsogarciacaro:
Nice catch, thank you!

PR #29, opened 2016-02-23
fix compilation on windows

rflechner:
Hi,

This is the workaround to fix issue https://github.com/fsprojects/Fable/issues/27

I tested it on macos and windows with the sample projects and new projects in another directory.

Regards.


alfonsogarciacaro:
Thanks for that!

PR #30, opened 2016-02-23
Fix typo.

jskripsky:


alfonsogarciacaro:
Wow, good catch, thank you! It seems the test wasn't good enough to detect the bug.

PR #31, opened 2016-02-23
Update build process

Krzysztof-Cieslak:
Moves calling `npm install` to `build.cmd` and `build.sh`

1. Makes setup easier for user doing fresh clone of repo - no need to run 2 separate commands
1. Makes downloading Fable using Paket v3 git support easier 
1. If developer add any new node dependency for Fable there is no need to run `npm install` separately


alfonsogarciacaro:
Makes lot of sense. Thanks, Krzysztof!

Issue #27, opened 2016-02-22
Windows: cmd.exe + quotes = Hell


fdcastel:
In Windows 8.1, with Node.js 5.4.0, running

`node tools\fable2babel.js --projFile "C:\Temp\MyProj\Core\MyProj.Core.fsproj"
`

gives me this

```
cmd /C C:\Temp\Fable\tools/../build/main/Fable.exe {"lib":".","outDir":".","symbols":[],"watch":false,"projFile":"MyProj.Core.fsproj"}
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: spawn cmd ENOENT
    at exports._errnoException (util.js:856:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:178:32)
    at onErrorNT (internal/child_process.js:344:16)
    at nextTickCallbackWith2Args (node.js:478:9)
    at process._tickCallback (node.js:392:17)
    at Function.Module.runMain (module.js:432:11)
    at startup (node.js:141:18)
    at node.js:1003:3

```

which is not a really useful error message :) -- But then, I tried

`cmd /C C:\Temp\Fable\tools/../build/main/Fable.exe {"lib":".","outDir":".","symbols":[],"watch":false,"projFile":"MyProj.Core.fsproj"}
`

which outputs a somewhat better description of the real problem

```
Unhandled Exception: Newtonsoft.Json.JsonReaderException: Input string '.' is not a valid number. Path 'lib', line 1, position 6.
   at Newtonsoft.Json.JsonTextReader.ParseNumber(ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ReadStringValue(ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ReadAsString()
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ReadForType(JsonReader reader, JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ResolvePropertyAndCreatorValues(JsonObjectContract contract, JsonProperty containerProperty, JsonReader reader, Type objectType)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters(JsonReader reader, JsonObjectContract contract, JsonProperty containerProperty, ObjectConstructor`1 creator, String id)
[...]

PR #36, opened 2016-02-25
Windows cmd quotes hell

davidpodhola:
Fixing #27 in tools/fable2babel.js only
Tests added too


davidpodhola:
maybe tests should be structured in folders later

alfonsogarciacaro:
Great, thanks a lot! Yes, AppVeyor tests shouldn't be different from the FAKE target, I'll fix that later. Also, I don't want to be picky but a commit history a bit cleaner would be appreciated ;)

PR #37, opened 2016-02-25
ts2f - React js

davidtme:
A lot of the code which was added to get react.d.ts converted had been removed... This gets it working again.

The main problem is mixing classes and interfaces. This now treats all class as interfaces.

(with converted file you still need to remove the JSX parts)

### DIFF
--- a/sample/node/react/App.config
+++ b/sample/node/react/App.config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
+  </startup>
+</configuration>
--- a/sample/node/react/Dom.fs
+++ b/sample/node/react/Dom.fs
@@ -1,0 +1,28 @@
+namespace Fable.Import
+open System
+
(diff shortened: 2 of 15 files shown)


alfonsogarciacaro:
Actually, classes must stay as they're now because they have callable constructors. But they're giving me a headache too as in TypeScript weird things like interfaces deriving from classes are allowed.
I've made more modifications to the script and parsed the react and react-dom definitions. I uploaded them [here](https://github.com/fsprojects/Fable/blob/master/lib/Fable.Import.React.fs), I hope you find them useful. I had to fix the import attributes by hand. I still have to figure out how to deal with TypeScript export statements.

davidtme:
Thank for the link to Fable.Import.React.fs but that file isn't valid

[Fable.Import.React.fs#L65](https://github.com/fsprojects/Fable/blob/master/lib/Fable.Import.React.fs#L65)

```'inherit' cannot be used on interface types. Consider implementing the interface by using 'interface ... with ... end' instead.```

alfonsogarciacaro:
Yeah, I still need to improve translation of class inheritance. Please feel free to edit the file as you see fit, either by implementing the interface correctly or just commenting the line.

PR #39, opened 2016-02-25
Make sure you completed the node samples before the browser sample

davidpodhola:


alfonsogarciacaro:
I think this is already implied by "the [server] we built above comes in handy". I'm not sure the change is necessary.

PR #40, opened 2016-02-25
Run samples from https://github.com/fsprojects/Fable/blob/master/docs/compiling.md#samples as part of the test

davidpodhola:
The idea is to make sure the [samples are buildable as described in the documentation](https://github.com/fsprojects/Fable/blob/master/docs/compiling.md#samples).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,7 @@ test_script:
   - node tools\fable2babel.js --projFile "C:\Temp\HelloWorld.fsproj"
   - npm install --prefix sample/browser/todomvc
   - node tools/fable2babel.js --projFile sample/browser/todomvc/app.fsx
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
I like the idea but still not sure the samples can be considered tests. Also, I'm not on Windows now and I don't have powershell. If possible, I wouldn't like that the instructions to run the tests on the CI server are that much different from my machine.

alfonsogarciacaro:
BTW, if you want to contribute with tests there're many things I have pending in the current test suite: pattern matching, active patterns, inheritance, overloads, secondary constructors... ;)

alfonsogarciacaro:
Closing for now. I'd prefer to keep a distinction between samples and tests to prevent breaking builds when modifying samples.

PR #41, opened 2016-02-25
npm install must be run in the sample directory

davidpodhola:
to create `node_modules` correctly (at least on Windows 10). If was run from the project folder it created another subfolder `fable` and the samples did not work.

### DIFF
--- a/docs/compiling.md
+++ b/docs/compiling.md
@@ -94,9 +94,9 @@ and run a sample follow these steps:
 Let's say you want to build the node static server. From the project root folder, type:
 
 ```
-cd sample\node\server
+cd sample/node/server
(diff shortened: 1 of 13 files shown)


alfonsogarciacaro:
Maybe it's necessary to create the `node_modules` in advance as stated in this [SO answer](http://stackoverflow.com/a/14469516/3922220). Can you please check if that's the case?

davidpodhola:
unfortunatelly not see the animated gif.
Sorry about the backslashes, I think Windows works with forward slashes too so I will fix it.
fable issue

alfonsogarciacaro:
I tested in Windows 10 and you were right, thanks for the fix!

Issue #34, opened 2016-02-25
Typescript2Fable - Enums


Krzysztof-Cieslak:
At the moment enums are ignored (no types for them is generated) but names of enums types are used as function parameters.

Option 1. Erase enums to obj ( make alias type ENUM_TYPE = obj)
Option 2. Erase enums to ints ( make alias type ENUM_TYPE = int) - enums in TS are just ints under the hood, we could use that
Option 3. Create F# enum types
Option 4. Create F# class with static members for each enum case. 

To be fair, I would be happy enough with Option 2 for start and build better solution later on. I should be able to PR it, just let me know which solution you want to use.

alfonsogarciacaro:
I think FunScript managed to translate TypeScript enums to F# `[<Enum>]`. It would be great to do that. I don't think it'll cause problems with Fable compiler as it's already converting .NET enums to int literals (this is actually done by F# compiler). It shouldn't be difficult, we only need to add a check for `EnumDeclaration` (see [sample](http://astexplorer.net/#/ZBT5MJbwEr)). If you want to send a PR that'd be awesome. I'll also try to have a look in the evening :)

Krzysztof-Cieslak:
I'll be happy to work on it, will try to PR it today. AST explorer will be very helpful, thanks for link!  :)

alfonsogarciacaro:
Yes, if it weren't for AST explore it would have been impossible for me to understand TypeScript AST ;) I wish it were as simple as F#'s!

PR #42, opened 2016-02-26
Add Enum support to binding generator

Krzysztof-Cieslak:
As discussed in #34


Krzysztof-Cieslak:
Also adding node.d.ts and typescript.d.ts files and referencing them in generator file - it enables having JS autocomplete, tooltips etc. in VS Code

tools/typescript2fable.js
```diff
@@ -310,9 +335,12 @@ function printModule(prefix) {
     return function(mod) {
         var template = prefix + templates.module
             .replace("[NAME]", escapeKeyword(mod.name));
-            
-        template = append(template, mod.interfaces.map(
-            printInterface(prefix + "    ", mod.name)).join("\n\n"));
+
+        var interfaces = mod.interfaces && mod.interfaces.length > 0
+                            ? mod.interfaces.map(printInterface(prefix + "    ", mod.name)).join("\n\n")
+                            : "";
```
alfonsogarciacaro: What are the edge cases this is trying to solve? `mod.interfaces` should never be `undefined` and executing `map` + `join` on an empty array should give an empty string so the check `mod.interfaces.length > 0` shouldn't be necessary

tools/typescript2fable.js
```diff
@@ -497,7 +536,8 @@ function getModule(name) {
       interfaces: [],
       properties: [],
       methods: [],
-      modules: []
+      modules: [],
+      enums: []
```
alfonsogarciacaro: We need to add a check for `enums` length in https://github.com/fsprojects/Fable/blob/master/tools/typescript2fable.js#L577. Or maybe we can do something generic like checking at least one array property (`Array.isArray(mod[k])`) satisfies `length > 0`.

alfonsogarciacaro:
Great work, thank you! I added a couple of comments, can you please check? Also, having typescript definitions is a great idea (I was too lazy to setting them up) but as the files are very big I think it'd be cleaner if we add them to npm dependencies instead.

tools/typescript2fable.js
Krzysztof-Cieslak: Not sure to be fair, when I was playing with this code, I've expected exception here... but that could be result of me messing something in other parts of code.

Issue #44, opened 2016-02-28
typescript2fable Inheritance 


davidtme:
I have an idea on how to get around inheritance problems especially an interface extending a class:

I've got some work in progress but before I finish I'll like some feedback

```
// interface FirstInterface {
//      FirstFunction(s: string): void
// }

// Would generate the following code:

// An interface prefixed with I
type IFirstInterface =
    abstract FirstFunction : string -> unit

// An abstract class which implements the interface
// This means any child class can inherit this class and not worry about implementing interfaces
[<AbstractClass>]
type FirstInterface =
    interface IFirstInterface with
        member x.FirstFunction(s): unit = failwith "Not implemented yet"

// class SecondClass implements FirstInterface {
//      SecondFunction(i: number): void
// }

// Would generate the following code:

// The interface in needed so any child interfaces can use this.
type ISecondInterface =
    inherit IFirstInterface
    abstract SecondFunction: int -> unit

// Same as the interface generated code but not an abstract class.
type SecondInterface =
    inherit FirstInterface
    interface ISecondInterface with
        member x.SecondFunction(i: int): unit = failwith "Not implemented yet"


// interface ThirdInterface extends SecondInterface {
//      ThirdFunction(b: bool): void
// }

// This interface can now inherit without any problems

type IThirdInterface =
    inherit ISecondInterface
    abstract ThirdFunction: bool -> unit

[<AbstractClass>]
type ThirdInterface =
    inherit SecondInterface
    interface IThirdInterface with
        member x.ThirdFunction(arg1: bool): unit = failwith "Not implemented yet"
```

alfonsogarciacaro:
Looks like a great idea! If you send a PR I'll test it. It'd be better if we can restrict this solution when the interface is being inherited (few cases) to avoid too much boilerplate, but I think I can do that. We may still have to fake some constructors, for example, `FirstInterface` abstract class needs a constructor so `SecondInterface` can inherit from it.

alfonsogarciacaro:
This has been fixed a bit differently, see [explanation](https://github.com/fsprojects/Fable/pull/48#issuecomment-190884373).

davidtme:
@alfonsogarciacaro I've found a new problem which this pr could help with solving... optional interface members

To create class in react it i should be able to do this:

```
type CC() =
    interface React.ComponentSpec<obj, obj> with
        member x.render(): ReactElement<obj> = 
            // Some code here
        member x.displayName
            with get (): string option = 
               // Some code here
[...]

PR #48, opened 2016-02-29
Typescript2fable - Inheritance

davidtme:
This tries to handle the odd inheritance model typescript has such as an interface inheriting from class 

[Issue #44](https://github.com/fsprojects/Fable/issues/44)

I have been using a react sample as my test definition file so I have also included this.

### DIFF
--- a/sample/node/react/App.config
+++ b/sample/node/react/App.config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
+  </startup>
+</configuration>
--- a/sample/node/react/Dom.fs
+++ b/sample/node/react/Dom.fs
@@ -1,0 +1,28 @@
+namespace Fable.Import
+open System
+
+[<AutoOpen>]
+module Dom =
+    type DataTransfer =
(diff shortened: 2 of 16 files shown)


davidtme:
@alfonsogarciacaro I see what you have done with the type cache but I still think every interface needs a class and interface generated (and vice-versa) because external d.ts files can inherit from them. Also a classes that inherit from an interface would have to implement all the ancestors, again they could be in external d.ts files.

alfonsogarciacaro:
I took inspiration from your work (thanks again for that!) but I solved things a bit differently. Now if a class is extending an interface, the intermediate abstract class will be created and if an interface extends a class, the interface itself will become an abstract class. I got the [react sample](https://github.com/fsprojects/Fable/blob/master/sample/browser/react/app.fsx) working, please have a look if you have time.

You're right that conflicts may occur between external typescript files, but my goal for the typescript parser was not to cover all corner cases (due to Typescript's flexibility, I'm not sure this is possible) but rather generate clean files that could be easily edited by hand (something that was not that simple in FunScript). Creating an abstract class and interface for each interface in TypeScript would pollute the namespace a bit too much in my opinion. And it can also affect compile times.

In any case, the parser and the import bindings are completely independent from Fable compiler, so it's still possible to have different versions of the parser to fit different tastes :)

PR #61, opened 2016-03-12
Specify minimal node and npm versions to fix #60

### LINKED ISSUE #60: Fable compiler should require specific Node.js engine version
[Specify required version of node.js](https://docs.npmjs.com/files/package.json#engines) to make sure the compile would not fail (screenshot from [Ubuntu 14.04 running the node.js installed using Ubuntu Package Manager](http://www.hostingadvice.com/how-to/install-nodejs-ubuntu-14-04/#ubuntu-package-manager))

fable-not-working-ubuntu

davidpodhola:
@alfonsogarciacaro could you please specify if the versions are correct? Maybe there are too strict. Thanks!

### DIFF
--- a/src/fable-js/package.json
+++ b/src/fable-js/package.json
@@ -42,5 +42,10 @@
   "bugs": {
     "url": "https://github.com/fsprojects/Fable/issues"
   },
-  "homepage": "https://github.com/fsprojects/Fable#readme"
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
The current node stable version is 4.4, so this has been set as the required version. Thanks for the report!

PR #64, opened 2016-03-16
Object expression from classes

davidtme:
This tries to address https://github.com/fsprojects/Fable/issues/63

This fits my needs as I have a class
```
type [<AbstractClass>] [<Import("./dummy?get=dummy")>] ARouteProps() =
    ....
```

and a dummy.js file
```
exports.dummy = null
```


alfonsogarciacaro:
Object expressions from classes are now supported :)

PR #75, opened 2016-03-27
Fix Fable.ObjExpr call in Fable.Plugins.Random.fsx plugin sample.

fdcastel:
Both documentation and plugin sample are outdated. Fable.ObjExpr now has 4 arguments.


alfonsogarciacaro:
I forgot to update that, thank you!

PR #80, opened 2016-03-29
Support for HashSet. Argument order for Assert.AreEqual(). Minor cleanup.

### LINKED ISSUE #78: HashSet?
Any ideas on how to map a HashSet (System.Collections.Generic.HashSet) to JavaScript?

Should it be part of fable-core.js?

fdcastel:
Initial support for HashSets (closes #78).

Also, during the tests I found the order of 'actual' and 'expected' arguments reversed(?!). Both in Utils.fs and in NUnit plugin (which makes the output to appear "correct" in Mocha -- But running them in F#/NUnit shows the values in the wrong order).

### DIFF
--- a/src/fable-fsharp/Replacements.fs
+++ b/src/fable-fsharp/Replacements.fs
@@ -437,6 +437,11 @@ module private AstPass =
             prop "length" i.callee.Value |> Some
         | _ -> None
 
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
Fantastic work, thanks a lot! Contributions like this is what makes open source so awesome :)

PR #82, opened 2016-03-30
Adds support to Enum comparison operators.

### LINKED ISSUE #81: Enums comparison
`Cannot find replacement for Microsoft.FSharp.Core.Operators.>=` when comparing enums.

The expression 

`1 >= 2`

Has the following ApplyInfo:

```
{methodName = ">=";
 ownerFullName = "Microsoft.FSharp.Core.Operators";
 callee = null;
 args =
  [Value (NumberConst (Case1 1,Int32)); Value (NumberConst (Case1 2,Int32))];
 returnType = PrimitiveType Boolean;
 decorators = [];
 calleeTypeArgs = [];
 methodTypeArgs = [PrimitiveType (Number Int32)];}
```

Which is fine. However, the same expression using enums

`MyEnum.A >= MyEnum.B`

where

```
type MyEnum = 
| A = 0x20uy
| B = 0x50uy
```

produces

```
{methodName = ">=";
 ownerFullName = "Microsoft.FSharp.Core.Operators";
 callee = null;
 args =
  [Wrapped
     (Value (NumberConst (Case1 32,UInt8Clamped)),DeclaredType MyEnum Class null);
   Wrapped
     (Value (NumberConst (Case1 80,UInt8Clamped)),DeclaredType MyEnum Class null)];
 returnType = PrimitiveType Boolean;
 decorators = [];
 calleeTypeArgs = [];
 methodTypeArgs = [DeclaredType MyEnum Class null];}
```

Which fails the 'compare' function of Replacements.fs

Suggestions? ;)

BTW: I'm adding more functions to support enums (https://github.com/fdcastel/Fable/commit/b7ab040beb3f6df776d86194536817d6143eb581), but not enough to make a PR, yet.

fdcastel:
Closes #81.

### DIFF
--- a/src/fable-fsharp/Replacements.fs
+++ b/src/fable-fsharp/Replacements.fs
@@ -176,13 +176,16 @@ module Util =
             
     let compare com (i: Fable.ApplyInfo) (args: Fable.Expr list) op =
         let op = Fable.BinaryOp op |> Fable.Value
-        match args.Head.Type with
-        | Fable.UnknownType
-        | Fable.PrimitiveType _  // TODO: Array comparison?
-        | FullName "System.TimeSpan"
-        | FullName "System.DateTime" ->
+
+        match args.Head with
+        | Type Fable.UnknownType
+        | Type (Fable.PrimitiveType _)  // TODO: Array comparison?
+        | Fable.Wrapped (Type (Fable.PrimitiveType _), _)  // For enums only ("Wrapped (_, DeclaredType <enumName> Class null)")
+        | Type (FullName "System.TimeSpan")
+        | Type (FullName "System.DateTime") ->
             Fable.Apply(op, args, Fable.ApplyMeth, i.returnType, i.range) |> Some
-        | Fable.DeclaredType ent ->
+
+        | Type (Fable.DeclaredType ent) ->
             match ent.Kind with
             | Fable.Class _ when ent.HasInterface "System.IComparable" ->
                 let comp =
@@ -190,6 +193,8 @@ module Util =
                     |> makeCall com i.range (Fable.PrimitiveType (Fable.Number Int32))
                 Fable.Apply(op, [comp; makeConst 0], Fable.ApplyMeth, i.returnType, i.range)
                 |> Some
+            | Fable.Class None ->  // For enums only ("DeclaredType <enumName> Class null")
+                Fable.Apply(op, args, Fable.ApplyMeth, i.returnType, i.range) |> Some  
             // TODO: Record and Union structural comparison?
             | _ -> None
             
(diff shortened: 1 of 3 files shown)


src/fable-fsharp/Replacements.fs
```diff
@@ -190,6 +193,8 @@ module Util =
                     |> makeCall com i.range (Fable.PrimitiveType (Fable.Number Int32))
                 Fable.Apply(op, [comp; makeConst 0], Fable.ApplyMeth, i.returnType, i.range)
                 |> Some
+            | Fable.Class None ->  // For enums only ("DeclaredType <enumName> Class null")
+                Fable.Apply(op, args, Fable.ApplyMeth, i.returnType, i.range) |> Some
```
alfonsogarciacaro: Why is this necessary? Are some enum types escaping `Fable.Wrapped (Type (Fable.PrimitiveType _), _)`?

fdcastel: Remove it and you will see in the tests :) 

Sorry. I made my comments in the issue https://github.com/fsprojects/Fable/issues/81 instead of here in the PR. Please review the issue.

But, in short, without this, `MyEnum.B >= a` will compile but `a >= MyEnum.B` won't.

alfonsogarciacaro: Sorry, I just read your comment in the issue report and I understand now why this is needed. Hmm, at the end I think we have to add enums to Fable's AST. It shouldn't be difficult though, I'll do it :) It's wonderful that you caught that with your tests :+1:

fdcastel: P.s.: I understand why you think it can be dangerous. And I fully agree. But every test passes until now. Could you think a way this could go wrong? If yes, please add a test to it. 

Anyway, This is only to fix the case when the first operand is not wrapped and the second is. If you there's a better way, please tell me :)

fdcastel: Great! If you could work on it, to add "real" enums to Fable would be nice. 

Unfortunately my Fable-fu is not enough to do it by myself ;) But if I could help you with anything, please don't hesitate to ask me!

alfonsogarciacaro:
I've finally [added `Enum` as another kind of `PrimitiveType` in Fable's AST](https://github.com/fsprojects/Fable/blob/master/src/fable-fsharp/AST/AST.Fable.fs#L14). This makes it compatible with primitive comparison operators in JS :) I think this is a great addition to Fable, thanks a lot for your work!

In a completely unrelated note, do you have a Twitter username?

fdcastel:
I did see it. Great! I will try the new version today, yet!




*> In a completely unrelated note, do you have a Twitter username?*

Hehehe. No. Neither Facebook :)

But you can reach me via email anytime. I'm old school :)

PR #83, opened 2016-03-30
Fix the react sample

davidtme:
Change the Emit to imports as extra () where being added:

(0, _express2.default)()() should be (0, _express2.default)()

### DIFF
--- a/import/express/Fable.Import.Express.fs
+++ b/import/express/Fable.Import.Express.fs
@@ -252,8 +252,11 @@ module express =
             [<Emit("$0($1...)")>] member __.Invoke(root: string, ?options: Options): Handler = failwith "JS only"
 
     type Globals =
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
The React sample is working fine as is for me, I've uploaded the result of compiling the sample with Fable's last version to a [gist](https://gist.github.com/alfonsogarciacaro/36464766b0707b83ab13038aca580d7c#file-server-js-L51). Could you please have a look and tell me if you're getting a different result? Thanks!

alfonsogarciacaro:
My bad, I was compiling using an older version. With the latest one is adding the extra braces indeed. Seems something got broken in one of the latest commits, I'm investigating this.

alfonsogarciacaro:
This has been fixed in the main code. I'll upload the new version to npm shortly, thanks for reporting this!

PR #85, opened 2016-03-30
remove unneeded parameters from transformTestFixture function

scitesy:


docs/compiling.md
```diff
@@ -205,8 +205,7 @@ the `fable` command in the sample directory. Just be sure to install the npm dep
 the first time.
 
 ```shell
-npm install
-fable
+npm install -g fable-compiler
```
alfonsogarciacaro: This was assuming Fable was already installed globally (as written at the beginning of the document): `npm install` here is to install the local dependencies of each sample and `fable` to compile the sample with the options from `fableconfig.json` (actually, right now only the second instruction would be necessary as `npm install` is included as a `rebuild` event in `fableconfig.json`).

Could you please remove this before merging the PR? Thanks!

alfonsogarciacaro:
Nice one! This will resolve the warning when compiling the Test plugin, I've been too lazy to do that ;)

I've added a comment, could you please check it. Thanks!

docs/compiling.md
scitesy: Thanks for the review! I'll do this later today.

alfonsogarciacaro:
Thanks for that!

PR #100, opened 2016-04-06
Change union internal format

davidtme:
Changing tags to Case and dataX to Fields[X] means that out the box fable can parse json produced by json.net


alfonsogarciacaro:
Nice addition to improve interoperability, thanks!

PR #102, opened 2016-04-07
Check method signature in NUnit plugin tests.

fdcastel:


src/plugins/Fable.Plugins.VisualStudio.UnitTests.fsx
```diff
@@ -47,7 +47,6 @@ module Util =
         | _ -> None
 
     // Compile tests using Mocha.js BDD interface
-    // TODO: Check method signature
     let transformTestMethod com ctx (testMethod: Fable.Member) name (decorator: Fable.Decorator) =
         let testName =
```
alfonsogarciacaro: Should we add the `test.Arguments.Length` check here too?

alfonsogarciacaro: Sorry, just read the note in the commit :) I'll merge the PR, thanks!

fdcastel:
My bad. I should have made one single commit. And the explanation why VS plugin doesn't need the check should be in the comments. I'll fix this.

PR #103, opened 2016-04-20
Type Namespace & FullName

davidtme:
Allows the use of typeof<x>.Namespace and typeof<x>.FullName


alfonsogarciacaro:
Nice one, thank you! Type naming in Fable's AST is a bit different from .NET reflection but it should be valid for most common cases. Also, it may be better to refactor the pattern matching to a different method but I can do that later :)

Issue #105, opened 2016-04-24
Seq.range is causing out of memory


lijunle:
### Description

This is a valid code `['1'..'9']` to generate a char list. However, it the Fable-Core's [Seq.range](https://github.com/fsprojects/Fable/blob/master/import/core/fable-core.js#L1236) function does not handle it correctly. Seq.range only handle the case the `first` and `last` are numbers.

### Expected behavior

Expect the code to generate a char list from '1' to '9'.

### Actual behavior

The build is passing, the webpack works fine. Finally, IE gives me out-of-memory error, Chrome hangs for seconds and finally be killed - because the steps continuously going up but never touch the `last` element - it is a char (a string actually after convert to JavaScript).

### Known workarounds

For 1 to 9 digit char, it is fine to write

```fs
[1..9] |> List.map char // CAUTE: this ONLY works in compiled JavaScript, not in F#
```

However, for a to z character, and for code works both in F# and JavaScript, I am writing a full list: 

```fs
let charsA2Z = ['a'; 'b'; 'c'; 'd'; 'e'; 'f'; 'g'; 'h'; 'i'; 'j'; 'k'; 'l'; 'm'; 'n'; 'o'; 'p'; 'q'; 'r'; 's'; 't'; 'u'; 'v'; 'w'; 'x'; 'y'; 'z']
```

Related: https://github.com/fsprojects/Fable/issues/106

fdcastel:
@lijunle Please check if #107 solves it for you.

Note that it requires a change into fable-core.js.

lijunle:
@fdcastel thanks for handling this. I find that not only integer and char can be seq. All type class implementing `IEnumable<T>` can be pass into seq.

So, it could be better to throw an exception for the unsupported type and maybe support then later.

fdcastel:
Yes. Any IEnumerable can be used as a sequence in F#. But I thought you were having problems only with the range constructors. The fix was essentially to solve that.

Could you share what are you trying to accomplish? Maybe we can find a better way :)

lijunle:
Yes, I'm only using char here. But I spent a lot of time to figure out which piece of code of causing the problem. In case someone use another enumerable type in seq, he will spend another couples of hours to find out which piece of code is causing problem. That is why the error message for unsupported type is important and useful.

alfonsogarciacaro:
[...]

PR #107, opened 2016-04-25
Fix Seq.range for chars.

fdcastel:
Proposed fix for #105. Please review.

I didn't push it into master because it requires changes to fable-core.js.

If you see a better way which doesn't need to change it, please let me know. ;)

### DIFF
--- a/import/core/fable-core.js
+++ b/import/core/fable-core.js
@@ -1233,6 +1233,11 @@
         ? [x, x + step] : null;
     }, first);
   };
+  Seq.rangeChar = function (first, last) {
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Yeah, I think it's OK to update `fable-core.js` here :) The only problem I've noticed is that unfortunately conversion to `int` is not working correctly for `char`. This is because at the moment Fable's AST doesn't distinguish between chars and strings (it did at the beginning but I removed it for simplicity). We may have to add the char type back to Fable in the future but for now I just tweaked the test to make it work ;) Thanks for your work!

PR #110, opened 2016-04-26
External References

davidtme:
At the moment if you reference an external f# file in your project the JavaScript output for the reference file is the original folder… Which is bad new if you have a few projects referencing it with different fable options.  Even worse, if you are using the outDir option the resulting path is relative to the outDir so you end up with unexpected locations for the js files.

This tries to solve the problem by placing these external files in a “.fabel.external” folder with a hash code (of the path) to avoid name collisions.


davidtme:
Need a bit more testing...

alfonsogarciacaro:
This is a very good idea! I would just add a compiler flag for activation as this may not be the desired behaviour for all users. I will bring it to a separate branch for now to add this and some testing. Thanks a lot for your work!

alfonsogarciacaro:
@davidtme, do you have any preference for the name of the compiler flag to activate this feature?

davidtme:
@alfonsogarciacaro I haven't really got a preference, maybe something along the lines of shadow external?

alfonsogarciacaro:
Ideally should be something short. It's difficult to convey the whole meaning with just the flag but then we have the help text for that. I was thinking of [--copyExt](https://github.com/fsprojects/Fable/blob/pr/110/src/fable-js/index.js#L35), what do you think?

alfonsogarciacaro:
Superseded by #111

PR #116, opened 2016-05-05
Add ICSSProp so extra props can be added.

davidtme:


alfonsogarciacaro:
This is a nice idea, thank you! Thanks to this I also realised I need to rename better the the interfaces here, I'll do it in a separate commit before republishing the package :+1:

BTW, not yet published but I implemented a way to cheat the compiler in these cases using tuples if necessary. For example:
```
    R.div [
        ClassName "comment"
        Style [
            Border "3px dotted blue"
            Margin "5px 0"
            Padding "0 5px"
            unbox("backgroundColor", "yellow")
        ]
    ]
```

davidtme:
While we are talking about cheating I've added this to my helpers but it maybe worth adding into the core?

``` F#
[<Emit("Object.assign({}, $0, $1)")>]
let ( ++ ) (a:'a list) (b:'a list) : 'a list = failwith "JS Only"
```

So you can do the follow (not only styles but all key value lists):
``` f#
let niceBorder = [ Border "1px solid blue" ]
let style = [ Display "inline-block" ] ++ niceBorder
```

davidtme:
If you are going to clean up that file you may want to remove Height and Color from HTMLAttr - They conflict with CSSProp Color & Height and in the html5 world you shouldn’t being using them. 

At the moment you have to use:

```f# 
[ Display "inline-block"; CSSProp.Color "red" ]
```

alfonsogarciacaro:
I was thinking about adding the `(++)` to the core library. However, users should be aware `KeyValue` lists behave differently from normal lists. Trying to adapt some or all the operators for this particular case, will pose too heavy a burden to the core lib.

In any case, this is still very valuable so I've added the info to the [docs](https://github.com/fsprojects/Fable/blob/master/docs/interacting.md#keyvaluelist-attribute) instead. I hope you're OK with that :)

PR #117, opened 2016-05-08
Change react children from List to Seq

davidtme:
This means you can still use lists but you can pass in an array - This means Fable doesn't have to create an array, convert it into a list the back into array, giving a little performance boost.


alfonsogarciacaro:
Actually, this optimisation should be done at compile time (see [here](https://github.com/fsprojects/Fable/blob/master/src/fable-fsharp/Replacements.fs#L115-L117)) but thanks to your report I just realised it wasn't being triggered because an active pattern wasn't working correctly after I changed the way imports work :/

It should be fixed now, so I'm closing the PR. Please give it a try with version 0.2.9 and tell me what you think :)

Cheers!

davidtme:
Looks good, thanks

PR #126, opened 2016-05-15
Webpack mario

Krzysztof-Cieslak:
Use webpack for Mario demo (example for  @tpetricek)

### DIFF
--- a/samples/browser/mario/Readme.md
+++ b/samples/browser/mario/Readme.md
@@ -1,0 +1,3 @@
+## Mario with Fable
+
+Showing how to use Fable with CommonJS, NPM modules and Webpack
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
Fantastic work, thank you! Only thing is that Tomas is currently working in the Fable web page and we're considering using require.js to load the samples (so we don't have to bundle `fable-core.js` with every sample). Can you please give me a couple of days before accepting the PR to see how this turns out and check the best method to automate sample compilation while building the web pages? Cheers!

tpetricek:
I suppose it could be done either way - I got something to work using `require.js`, but we could equally switch to this model and just have one JS file for each sample. I should have a PR tomorrow.

alfonsogarciacaro:
Closing this for now. Probably it's better to use `requirejs` with the samples so they can all share the same copy of common files (`core-js`, `fable-core`).  Anyways, thanks a lot for your work Krzysztof!

PR #131, opened 2016-05-17
New Fable website!

tpetricek:
For now, you can see the result at: http://tpetricek.github.io/Fable.

All documentation related stuff is in the `docs` folder. There is:

 - `docs.fsx` together with `liquid.fs` and `helpers.fs` generates the documentation 
 - `templates` contains DotLiquid templates for the page look 
 - `source` is the source for the documentation. This can contain static files (just copied to the output directory), Markdown files (formatted and passed to the `content.html` template) and HTML files (that can specify formatting using DotLiquid header - see `index.html` for example).
 - `source/content` contains some JS for tooltips and CSS styles
 - `source/docs` contains the original Fable docs (Markdown) with some tweaks

In `paket.dependencies`, I added a separate group for the docs, because this might need different version of `FSharp.Compiler.Service.dll` than the one Fable compiler needs. For this reason, `docs.fsx` is a separate build script. It is invoked from main `build.fsx` by spawning a new copy of FAKE. There are two targets:

 - `PublishDocs` processes all documentation and pushes it to gh-pages
 - `GenerateDocs` processes all documentation but does not push it
 - `BrowseDocs` starts a server on localhost and opens the docs. It supports neat live reloading :). This does not recompile all samples, because that can take quite long.

They are stand-alone, but I suppose you can add `"PublishDocs" ==> "Publish"` to release docs on every release!

In the samples and in the documentation, the file always starts with a Markdown list that specifies some meta data about the item. For samples, this is CSS for the embedded JS app, title & tagline for the page (passed to DotLiquid via "model" object) etc. For documentation, this is again the title & tagline that appear in the purple box.

I only did this for a couple of the samples (Mario, Hokusai, Ozmo, Pacman for the browser and Express and Server from the nodejs samples), but it would be really nice to add more "ordinary" things like the TodoMvc or event the Electron demo. But I guess others can contribute now too :-)


tpetricek:
Oh, also, the URL for the video needs to change once Mark uploads the video to FSharp.TV Vimeo or Youtube.

tpetricek:
As @ReedCopsey pointed out we say people need F# 4.4 :) that's a tough requirement. I suppose Fable requires just F# 4.0?

tpetricek:
Also, the code font I picked is odd - it does ugly `fl` ligatures. Will change that.

alfonsogarciacaro:
This is wonderful Tomas, thanks a lot! It looks really, really, really nice. I couldn't have come to something even close to this. I just merged your PR (btw, you have also rights if you need to push directly) and I'll go deeper through your recommendations later. You're right about F# 4.4. That was mainly for development as one of the first contributors had problems if FSharp.Core 4.4 wasn't installed. But I guess for users it should be enough with F# 4.0 :+1:

forki:
Xoxo. Just wonderful!

dsyme:
Could you add a link to the website in the description at the top of http://github.com/fsprojects/Fable please?

forki:
done

PR #132, opened 2016-05-18
Updating gh-pages

tpetricek:
`gh-pages` are automatically generated, but this PR removes `.gitignore` from the `gh-pages` branch. Without that, the auto-generation skips some files.


tpetricek:
(To be merged after #131 but before running `build PublishDocs`)

PR #138, opened 2016-05-18
Fix margins in small mobile view

tpetricek:
Hmm, is this some git history rewriting magic here?

How should I send the PR to get rid of my previous commits which are already in?


alfonsogarciacaro:
Since Github made this "Squash and merge" button available I don't care anymore about git history in the PRs ;) But probably it's because I used it before that your previous commits are still visible in this PR too.

In any case, thanks a lot for the fix!

tpetricek:
Oh, I see :) new git trickery!

PR #150, opened 2016-05-20
Add F# events support

Krzysztof-Cieslak:
~~Do not merge, work in progress.~~

- [x] Basic `Event` object implementation
- [x] `Event` module functions 
- [x] 'IEvent` methods
- [x] Testing
- [x] Docs

### DIFF
--- a/import/core/fable-core.js
+++ b/import/core/fable-core.js
@@ -1987,4 +1987,217 @@
       }, w),
     ];
   };
+  
+  var Event = exports.Event = function(sbscrb, delegates) {
+    var _this = this;
+    this.delegates = delegates || new Array()
+    
+    this.publish = function () {
+       return _this 
+    };
+    
+    this.trigger = function (value) { 
+      Seq.iter(function(f) { f(value) }, _this.delegates ) 
+    };
+    
+    var addHandler = function (f) {
+      _this.delegates.push(f)
+    };
+    
+    var removeHandler = function (f) {
+      var index = _this.delegates.indexOf(f);
+      if (index > -1) {
+        _this.delegates.splice(index, 1);
+      }
+    }
+    
+    this.subscribe = sbscrb || function (observer) {
+      var f = observer.onNext; 
+      addHandler(f); 
+      
+       var disp = {
+        dispose: function () {
+          removeHandler(f);
+        }
+      };
+      disp[FSymbol.interfaces] = ["System.IDisposable"];
+      return disp;
+    }
+
+  };
+  
+  Event.add = function (f, w) {
+    w.subscribe(new Observer(f));
+  };
+  
+  Event.map = function(f, w) {
+    var s = function(observer) {
+      w.subscribe(new Observer(
+        function (v) {
+          Obs.__protect(
+            function () { 
+              return f(v)
+            },
+            observer.onNext,
+            observer.onError
+          );
+        },
+        observer.onError,
+        observer.onCompleted));
+    }
+    return new Event(s, w.delegates) 
+  }
+  
+  Event.choose = function (f, w) {
+    var s = function (observer) {
+      return w.subscribe(new Observer(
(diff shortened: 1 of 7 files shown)


Krzysztof-Cieslak:
@alfonsogarciacaro, I think I have small problem with workflow - changes I've done to `import/core/fable-core.js` are not copied to test folder. So all my tests are failing. Am I doing anything wrong?

alfonsogarciacaro:
I wanted the tests in the CI server to run against the `fable-core` version in npm, so I'm checking the [environment variable `DEV_MACHINE`](https://github.com/fsprojects/Fable/blob/master/build.fsx#L163) before replacing `fable-core` with the development version. You can comment that line for now, I will uncomment it again when the new version of `fable-core` is published :+1:

Thanks a lot for your work!

alfonsogarciacaro:
And sorry, I forgot, can you please also edit [the part about Events in the docs](https://github.com/fsprojects/Fable/blob/master/docs/source/docs/compatibility.md#event-and-observable) now that you made `Events` work too :)

Krzysztof-Cieslak:
It's not ready yet - as for now I've get `Event` module functions working. Still need to add methods from `IEvent` (`AddHandler` etc.).

Will update docs when I'm done :)

Krzysztof-Cieslak:
Looks like I'm green on all tests, docs updated. From my point of view it's ready to merge. But code review and any comments appreciated.

import/core/fable-core.js
```diff
@@ -1987,4 +1987,240 @@
...
+       var disp = {
+        dispose: function () {
+          _removeHandler(f);
+        }
+      };
+      disp[FSymbol.interfaces] = ["System.IDisposable"];
+      return disp;
+    }
+
+  };
```
alfonsogarciacaro: We could add here `Util.setInterfaces(Event.prototype, ["Microsoft.FSharp.Control.IEvent"]);` so it can be type tested.

src/fable-fsharp/Replacements/Replacements.fs
```diff
@@ -1092,6 +1092,9 @@ module private CoreLibPass =
             fsharp + "Collections.Seq" => ("Seq", Static)
             fsharp + "Collections.Set" => ("Set", Static)
             fsharp + "Core.Choice" => ("Choice", Both)
+            fsharp + "Control.Event" => ("Event", Both)
```
alfonsogarciacaro: Not entirely sure and actually it won't make any real difference in this case, but I think we can use `Static` here.

alfonsogarciacaro:
This is an amazing work! The code looks great, I just left a couple of comments but they're not very important. Only, if we're really interested in making the size of `fable-core.js` smaller, we could try to unify the code for `Event` and `Observable` static functions. But probably when we implement tree shaking this won't be very important.

I'm really surprised you got the code for the second replacement pass (the CoreLibPass) right as it's not very intuitive :) I'll leave some comments for future contributors. Also, I forgot to tell there's a `readCoreLibMethods.js` script in the `src/tools` folder to automatically update `CoreLibMethods.fs`. Did you use it?

I'll merge the code as is. I'm also giving you edit rights in case you need to change something later.

Thanks a lot for this, Krzysztof!

src/fable-fsharp/Replacements/Replacements.fs
Krzysztof-Cieslak: I think `Both` is needed for `Event.Publish` and `Event.Trigger` functions which are class methods not static functions on module... ( I think I  `Static` was my 1st choice and it failed on those functions)

Krzysztof-Cieslak:
Thanks. :)

Yes, unification between `Event` and `Observable` sounds like good idea.

No, I haven't used `readCoreLibMethods.js`, i just modified this file I shouldn't modify by hand ;)

PR #152, opened 2016-05-23
Add an extension to allow let! bindings of promises in async

7sharp9:


tpetricek:
I don't think having this in general for all Fable projects is a good idea - for the same reason for which F# core library does not have an overload for `let!` on tasks. It makes overloading messy, because you cannot write:

    let foo some = async {
      let! res = some ()
      return 1  }

I'd be in favor leaving `AwaitPromise` as an explicit operation. Those who prefer to have `let!` overload can always add it in their projects (or perhaps there could be some standardized set of extensions for this), but I think the core should follow the design style usual in the F# core libraries.

7sharp9:
No worries, I already have three different permutations of promising:

```
    //overloaded async
    let data = Observable.create() 
    async {
       let! req = fetch "http://az664292.vo.msecnd.net/files/ZjPdBhWNdPRMI4qK-colors.json"
       let! json = req.json ()
       do (data.value <- json) } |> Async.Start 

    // concrete promise
    promise {
       let! req = fetch "http://az664292.vo.msecnd.net/files/ZjPdBhWNdPRMI4qK-colors.json"
       let! json = req.json ()
       do (data.value <- json) } |> ignore
        
    //pipe oriented
    fetch "http://az664292.vo.msecnd.net/files/ZjPdBhWNdPRMI4qK-colors.json"
    |> Promise.success (fun resp -> resp.json())                                                                       
    |> Promise.success (fun result -> data.value <- result)
    |> ignore
```

alfonsogarciacaro:
Ok, I'll close the PR then but probably we should work in the direction pointed by Tomas and also discussed with Dave: building a more opinionated set of extensions and functions (a framework if you wish) to make it an easier start for users who don't know well the JS ecosystem.

There's also a tricky thing about `Fable.Core` (the F# version), as it's distributed as a `dll` (this makes compilations faster) the AST cannot be read and Fable won't generate any code directly from it, so all calls [must be replaced during compilations](https://github.com/fsprojects/Fable/blob/master/src/fable-fsharp/Replacements/Replacements.fs#L1052) the same way `FSharp.Core` calls are. This is a bit cumbersome but, again, the idea of Fable.Core is providing a minimal functionality and basic definitions (JS, Browser and Node) and not a comprehensive framework for JS development (which could be distributed separately).

Cheers!

tpetricek:
:+1:  for a more opinionated take on F#-friendly client-side library! Looks like @7sharp9 has some great starting points already :-)

Krzysztof-Cieslak:
I'm also working on really opinionated React <-> F# abstraction layer, if we talk about opinionated libraries ;)

7sharp9:
It should be possible to get the AST from Fable.Core, I was making some headway doing something similar with FSharp.Core because I needed its AST available too.

7sharp9:
Essentially a sub compilation of the referenced dll is needed, Fable.Core is available in the project at a known location so that should be easy to figure out //cc @alfonsogarciacaro

7sharp9:
Actually you can read in the AST entities for a reference too.

alfonsogarciacaro:
Good to know, I just assumed the AST of compiled dlls couldn't be read :) If it's possible that will open interesting possibilities.

forki:
ok I would love to have this in some specialized module, but directly in fable.

like `open Fable.Core.AsyncPromiseExtensions` and things appear magically!?

forki:
could be sued like this: https://github.com/fsprojects/fable-react_native-demo/commit/3f73f4d995a1c42b7c886d129c9d60d82b2505ec

alfonsogarciacaro:
In general I'd prefer no to add nice-to-have features to Fable.Core to keep it as lean as possible and let users include the utility functions they need in their own projects (I know that I've already several utilities to Fable.Core and it's debatable if they should be there according to that guideline).

However, I acknowledge there's a good reason to add feature to core libs: making them idiomatic. And, as many items async workflows in Fable will be used to deal with JS promises, users will likely appreciate to be spared of typing `|> Async.AwaitPromise` every time. My only concern about this is it could lead developers to believe F# Async and JS Promise are exactly the same thing, which can cause some confusion in certain scenarios.

forki:
After thinking about it. I think the fetch bindings should do the mapping.
I have specific ideas for that and need to discuss that with Dave

forki:
In other words I agree that the mapping from promise to async should be
explicit. But the bindings could provide "overloads" that are already
mapped.

PR #160, opened 2016-05-25
Add implementation of fetch api

7sharp9:
See: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

### DIFF
--- a/import/core/Fable.Import.Fetch.fs
+++ b/import/core/Fable.Import.Fetch.fs
@@ -1,0 +1,86 @@
+namespace Fable.Import
+open System
+open Fable.Core
+open Fable.Import.JS
+
+module Fetch =
+
+    type [<AbstractClass; Import("*","Body")>] Body() =
+        abstract bodyUsed: bool with get, set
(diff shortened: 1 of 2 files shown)


import/core/Fable.Import.Fetch.fs
```diff
@@ -0,0 +1,86 @@
...
+        U2<Headers, ResizeArray<string>>
+
+    and BodyInit =
+        U3<Browser.Blob, Browser.FormData, string>
+
+    and RequestInfo =
+        U2<Request, string>
+        
+    type Globals =
+        static member fetch (url: U2<string,Request>, ?init: RequestInit) =  failwith "JS only" :Promise<Response>
```
alfonsogarciacaro: To make `fetch` globally accessible you can decorate this member with `[<Global>]`, but I'm not sure if this will work when using fetch.js as a dependency.

alfonsogarciacaro:
Thanks! For now, I've moved Fetch to its own package. I'll have a look later so we can decide the easiest way to make the interaction with user code :1+:

PR #164, opened 2016-05-25
Use an erased union as the input to fetch

7sharp9:
So syntax for a normal fetch would be:
```fsharp
GlobalFetch.fetch (Url “http://blah”)
```


alfonsogarciacaro:
This makes much more sense, thanks!

PR #174, opened 2016-05-28
fix of escaping reserved words and keywords in ts2fable

llRandom:


alfonsogarciacaro:
Ups, I missed that. I'm always expecting the compiler to fix these mistakes for me, as with F# 😅 Thanks!

PR #176, opened 2016-05-29
Fixing website links

evelinag:
I changed the `.md` extensions in links to `.html` so that they work on the Fable website.


alfonsogarciacaro:
Good point, thank you!

Issue #144, opened 2016-05-19
Running fable with -w gives file is being used exception


ritcoder:
### Description

Running `fable -w` run the first time. Subsequent times fail with `file in use by another process` exception

### Repro steps

Run `fable -w`

### Expected behavior

Files are compiled and webpack-dev-server runs

### Actual behavior

````
u:\Work\PlayGround\Fable\CreatingVSCodePlugin\test2>fable -w
Start compilation...
Compiled Fable.Helpers.React.fs at 7:36:31 PM
Compiled models.fs at 7:36:31 PM
Compiled reactTest.fsx at 7:36:31 PM
Compiled main.fsx at 7:36:31 PM
Watching u:\Work\PlayGround\Fable\CreatingVSCodePlugin\test2\src
Cannot read project options: The process cannot access the file 'u:\Work\PlayGround\Fable\CreatingVSCodePlugin\test2\src\main.fsx' because it is being used by another process.
Compiled main.fsx at 7:38:01 PM
Cannot read project options: The process cannot access the file 'u:\Work\PlayGround\Fable\CreatingVSCodePlugin\test2\src\main.fsx' because it is being used by another process.
Cannot read project options: The process cannot access the file 'u:\Work\PlayGround\Fable\CreatingVSCodePlugin\test2\src\main.fsx' because it is being used by another process.
````

### Known workarounds

Run `fable` manually when there is a change

### Related information 

* Operating system: Window 10
* Branch: Fable 0.2.14
* .NET Runtime, CoreCLR or Mono Version
* Performance information, links to performance testing scripts

alfonsogarciacaro:
Thanks for the report. Can you please post the `fableconfig.json` you're using? Also, if you have the copy in Github or somewhere public a link would be very helpful to reproduce the issue. Cheers!

ritcoder:
Hi,
I'm migrating an application from `elm` to `f#` so cannot post it online. One observation is that it affects only the `projFile`. When I edit any other file, it compiles just fine. It is only the `main.fsx` file that gives that error.


`fableconfig.json` below

````
{ "module": "commonjs",
  "sourceMaps": true,
  "projFile": "./src/main.fsx",
  "outDir": "build",
  "scripts":{
  },
  "targets":{
    "watch":{
      "scripts":{
        "postbuild": "webpack-dev-server --content-base public/"
      }
    }
  }
}
````

alfonsogarciacaro:
[...]

PR #179, opened 2016-06-02
Fix blocking project file

irium:
Fix for #144.


alfonsogarciacaro:
Nice, thanks a lot for that!

irium:
You're welcome! It also would be awesome to fix #180 (that prevents from making Fable webpack loader).

PR #187, opened 2016-06-09
Fable bindings for Redux along with react-redux and react-router-redux.

psfblair:
These were generated automatically from Typescript declarations; hopefully they're ok.

### DIFF
--- a/import/react-redux/Fable.Import.ReactRedux.fs
+++ b/import/react-redux/Fable.Import.ReactRedux.fs
@@ -1,0 +1,41 @@
+namespace Fable.Import
+open System
+open System.Text.RegularExpressions
+open Fable.Core
+open Fable.Import.JS
+
+module ``react-redux`` =
+    type [<Import("ElementClass","react-redux")>] ElementClass() =
+        interface Component<obj, obj>
+
+
+    and ClassDecorator =
+        [<Emit("$0($1...)")>] abstract Invoke: ``component``: 'T -> 'T
+
+    and MapStateToProps =
+        [<Emit("$0($1...)")>] abstract Invoke: state: obj * ?ownProps: obj -> obj
+
+    and MapDispatchToPropsFunction =
+        [<Emit("$0($1...)")>] abstract Invoke: dispatch: Dispatch * ?ownProps: obj -> obj
+
+    and MapDispatchToPropsObject =
+        [<Emit("$0[$1]{{=$2}}")>] abstract Item: name: string -> ActionCreator with get, set
+
+    and MergeProps =
+        [<Emit("$0($1...)")>] abstract Invoke: stateProps: obj * dispatchProps: obj * ownProps: obj -> obj
+
+    and Options =
+        abstract ``pure``: bool with get, set
+
+    and Property =
+        abstract store: Store option with get, set
+        abstract children: Function option with get, set
+
+    and [<Import("Provider","react-redux")>] Provider() =
+        interface Component<Property, obj>
+
+
+    type [<Import("*","react-redux")>] Globals =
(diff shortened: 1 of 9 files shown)


alfonsogarciacaro:
Thanks! I'll have a look :+1:

Do you want to publish the package to npm? If you do, you can also make a collaborator in case there's a need to make some change. Also if you have any sample of Fable+Redux that you'll like to share, that'll be highly appreciated :)

psfblair:
I'd like to publish to npm, but haven't done that before so I'll have to look it up. Before doing that, though, I'm trying to see if I can get it to work. I'm creating a brunch skeleton with some toy application code, which will be useful as a sample.

Having some trouble with the typing, though: I want to create a Redux reducer, which is just a function. The type declaration in Fable.Import.Redux.Reducer makes it a  
```
Reducer =
        inherit Function
        [<Emit("$0($1...)")>] abstract Invoke: state: obj * action: obj -> obj
```

I'm not sure I understand how to make a function implement one of these callable interfaces. I finally got it to work like this:
```
type ActionType = Increment | Decrement
type Action = { Type : ActionType }

let inner = function
    | (state, { Type = Increment }) -> state + 1
    | (state, { Type = Decrement })-> state - 1

let reducer =
    { new Redux.Reducer with
        member this.Invoke(state, action) = inner (state :?> int, action :?> Action) :> obj }
```
but this seems like a crazy amount of casting, and takes away the type safety. I haven't done F# in awhile; what's the right way to do this?

alfonsogarciacaro:
Publishing in npm is very easy, as you already have a `README.md` and a `package.json` you just have to type `npm publish` in the package directory (before that, you need to register in [npm's site](https://www.npmjs.com)), but I can also do it if you want.

Sometimes to get the definitions work properly in an F# idiomatic way. Here, I think you can replace the `Reducer` definition with:

```fsharp
and Reducer = Func<obj, obj, obj>
```

Then the last lines in your example would become:

```fsharp
let reducer = Func<obj,obj,obj>(fun state action -> inner(unbox state, unbox action) |> box)

// Either box/unbox or casting would work (they're both erased in JS)
//let reducer = Func<obj,obj,obj>(fun state action -> inner(state :?> int, action :?> int) :> obj)
```

Still not very beautiful. It could be possible to use generics to have more type safety, but this could cause problems if it's need to combine reducers of different types is lists, etc. I have read about Redux but I haven't worked that much with the library so I'm not sure yet what's the best solution. Sometimes it makes sense to create a thin layer or a helper to ease the interaction with the library from F# [as it was done for React](https://github.com/fsprojects/Fable/blob/master/import/react/Fable.Helpers.React.fs). Is there a specific Redux sample you want to port? I could try to have a look and give it a try.

psfblair:
Thanks for the help; turns out all I needed was an interface definition of:
```
Reducer<'T, 'U> = Func<'T, 'U, 'T>
```
which lets me do:
```
let inner = function
    | (state, { Type = Increment }) -> state + 1
    | (state, { Type = Decrement }) -> state - 1

let reducer = Func<int,Action,int>(fun state action -> inner (state, action))
```
which is reasonable.

I'm working on a Brunch recipe for Fable with React/Redux/React Router. If I can get it off the ground I'll consider the interface definitions good enough to push to npm, though I've had to make a number of changes to get things to work so far. I'll send you another pull request when I'm comfortable with the results--or at least when I'm ready to stop.

I'm still struggling with newbie setup; I've never had so much trouble with F# before. Ionide and Visual Studio Code both hit me with long pauses fairly often and Ionide has crashed at least once. Xamarin still seems to have a pretty buggy UI and suffers from the fact that you can only see/edit files you've actually added to the project (although it does give a slightly better experience when dealing with .fsproj files).

alfonsogarciacaro:
Interesting, Visual Studio Code+Ionide works very well for me. Please make sure you're using the latest version. But it's true `Find All References` takes too long. I usually use `Find` or `Find in Files`. You can also give it a try to Atom+Ionide. There're also F# plugins for Sublime and emacs but I think they don't as many features.

Issue #196, opened 2016-06-15
Option Type instance method crashes


robkuz:
### Description

It looks like on option types and their erasure some weird things are happening.

### Repro steps
the following code 
```fsharp
let fromOption<'T> (defaultVal:'T) (o: 'T option) : 'T = if o.IsSome then o.Value else defaultVal

let x = fromOption 2 (Some(1))
let z = fromOption 2 None
````

generates this javascript
```javascript
"use strict";

var fromOption = function (defaultVal, o) {
     return (o != null)() ? o : defaultVal;
};

var x = fromOption(2, 1);
var z = fromOption(2);
```

which has the following issues

    return (o != null)()
                      ^
                      why is there a function application here? probably from the call to .IsSome. 
                      but it crashes the process

### Expected behavior

- no function application after the check
- maybe also check for `undefined` so that 
  `if (o != null && o != undefined)``

### Actual behavior

- crashes as `(o != null)` yields no callable value
- the same is true for calls to `IsNone`

### Known workarounds

- none. Makes Option types more or less unusable
- in my specific case I created the following workaround but I dont know how to explain it in more generic terms

```fsharp
[<Emit("($1 != null && $1 != undefined) ? $1 : $0")>]
let fromOption<'T> (defaultVal:'T) (o: 'T option) : 'T = failwith "JS only"
```

### Related information 

* Operating system OSX
* Branch Fable 0.3.20
* most probable fix should be here https://github.com/fsprojects/Fable/blob/41ec7b9a7b182b6f0be706e731caca4a1e085997/src/fable-fsharp/Replacements/Replacements.fs#L587-L588

alfonsogarciacaro:
Thanks a lot for the detailed report. I noticed the same a few days ago and found out it was out to some strange interpretation of the instance `IsSome`/`IsNone` members in F# AST. This should be fixed in [`react-todomvc` branch](https://github.com/fsprojects/Fable/blob/react-todomvc/src/fable-fsharp/Replacements/Replacements.fs#L592-L594). I'll merge it shortly, sorry for the delay!

alfonsogarciacaro:
This should be fixed in v0.3.22, can you please check?

robkuz:
the build failed it seems

alfonsogarciacaro:
Sorry, I had some problems to publish v0.3.22 to npm (weak internet connection at the moment). It should be published now. Could you please reinstall fable: `npm install -g fable-compiler` and try again?

[...]

PR #200, opened 2016-06-15
Add React TodoMVC sample

alfonsogarciacaro:
And fix some errors: getters & setters with indexers and #196


tpetricek:
It would be fantastic if we could turn this into a page in the docs... Is the tutorial using any server, or is it just client-side?

alfonsogarciacaro:
Yes, my intention is to make this a sample for the web. It is just I'm traveling this week and don't have much time for the project, but the branch contained also a couple of fixes for the compiler so I wanted to merge before it diverged too much from master :)

alfonsogarciacaro:
The codes just client side. It's basically a port of the React sample in todomvc.com

PR #201, opened 2016-06-16
Add SameGame sample

battermann:
I added the SameGame sample and docs for the web page.

The UI needs its own style sheet to display correctly on the samples web page.

It all looks pretty good locally so I hope that no fixes are needed.


alfonsogarciacaro:
Fantastic, thanks a lot for this @battermann! I'm closing the PR because I merged it locally to make a couple of changes before (usually Github detects that, but it seems it didn't this time):

- To simplify the doc build process I've adapted SameGame to be similar to the other samples: use `requirejs` (amd JS modules) to load the sample and dependencies and put the whole code in a single file.
- Shorten the longest lines so they look better in the tutorial.
- Fable doesn't support `System.Int32.Parse` but it does support the F# `int` operator. So I've removed the custom function `strToInt` and replaced it by `int` calls.

I hope this is OK with you. The sample is live now, please tell me if there's anything you would like to change.

Thanks again for your contribution!

alfonsogarciacaro:
I just noticed your name is not in the sample. I'll put it together with a link to your blog 👍

battermann:
That's perfect, thanks for helping me out and improving the code!

PR #212, opened 2016-06-22
Fix Array Sort

### LINKED ISSUE #197: Array.sort not sorting ints correctly
### Description

Sorting an array of int sorts by string rather than int

### Repro steps

``` fsharp
[<EntryPoint>]
let main argv =
    let items =
        [| 1;2;10;20 |]
        |> Array.sort

    printf "%A" items
    0
```
### Expected behavior
1, 2, 10, 20

### Actual behavior
1, 10, 2, 20

davidtme:
This is to fix #197 

*fable-core.min.js seems to have a much bigger change that I was expecting but tests (with SET DEV_MACHINE=1) pass*


davidtme:
I'm not sure how to go about editing fable-core so the tests pass on CI?

alfonsogarciacaro:
Sorry, I had some work on this before seeing your PR. I just pushed a fix that resolves this at compile time and should cover more situations. I've also expanded the tests.

Yes, adding `fable-core.min.js` unfortunately means that any change to the core JS library creates a big diff in `fable-core.min.js`. BTW, now `import/core/es2015.js` should be considered the base file to make changes to the library. Running FAKE "CoreLib" target propagates changes to `commonjs.js` (this file was necessary for Fuse) and `fable-core.js` (the version with universal UMD modules).

The CI server always takes the last `fable-core` package from npm. This is a way for me to remember to publish the package before updating `fable-compiler` :) The problem is, as in your case, PRs fail tests but we need to live with that for now :/

davidtme:
Thanks for the info.

PR #215, opened 2016-06-22
Correctly determine type of if-then-else and try-with expressions

i-p:
In Fable AST, the type of if-then-else was incorrectly determined to be Unit, when 'then' branch was a throw expression.

As a result, function like this 
```
let f () =
   if false then failwith "error" else 7
```
didn't have a return statement after translation to JavaScript.

The same applies to try-with expression.

### DIFF
--- a/src/fable-ast/AST.Fable.fs
+++ b/src/fable-ast/AST.Fable.fs
@@ -210,17 +210,27 @@ and Expr =
     | Wrapped of Expr * Type
 
     member x.Type =
+        let combineTypes (e1: Expr) (e2: Expr) =
+            // The expressions can have different types
+            // if one of them is a throw expression (it has always type Unit)
+            match e1.Type with
+            | Type.PrimitiveType(PrimitiveTypeKind.Unit) -> e2.Type
+            | _ -> e1.Type
+
         match x with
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Sorry, I went too quickly through this and thought it was an issue instead of a PR 😅 I agree with you solution, and I particularly appreciate you're including tests. I was just thinking that it may be safer to add the type directly to `Throw` expression as the original F# type is accessible [here](https://github.com/fsprojects/Fable/blob/master/src/fable-fsharp/Replacements/Replacements.fs#L406). Would you agree with that?

i-p:
Sure, your solution is more elegant.

By the way, when looking for the files to change, I found out that there is file
"src\fable-fsharp\AST\AST.Fable.fs" that is not included in fsproj. It looks like an old version of "src\fable-ast\AST.Fable.fs".

alfonsogarciacaro:
True, I forgot to delete it when I split the Fable AST (the purpose of this is to make it more easily available for Type Providers). I removed it :+1:

Closing the PR as I merged your code with the change mentioned above, locally. Thanks a lot for your contribution!

PR #217, opened 2016-06-26
Counter and todomvc with Elm architecture

mastoj:
The Elm architecture implemented in F# and Fable. It is probably a lot more to add to this, performance wise and maybe in the API as well. I do think this is a great starting point and something to build on.

For this PR to be complete the sample need to render two divs with specific id right before each example. Didn't get the sample to render at all, so didn't manage to render a div with the markdown, if event possible. It is possible to try the samples out with just running fable and open the index.html file. The sample contains a simple counter that also modifies the html and an implementation of todomvc.

I haven't had the time to proof read the sample text and describe everything in detail. Hopefully it is good enough anyway.

### DIFF
--- a/samples/browser/virtualdom/Fable.Helpers.Virtualdom.fsx
+++ b/samples/browser/virtualdom/Fable.Helpers.Virtualdom.fsx
@@ -292,7 +292,7 @@ module App =
 
     type AppState<'TModel, 'TMessage> = {
             Model: 'TModel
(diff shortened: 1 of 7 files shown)


alfonsogarciacaro:
Thanks for all your work! I think it's a very nice addition to Fable. I'll make sure the sample is working fine with the and update the docs :+1:

mastoj:
Just glad I could help. Not sure if all the helpers and the dsl should be under `Import.Virtualdom`, but they could probably go there as a start.

PR #219, opened 2016-06-28
Update virtualdom sample

mastoj:
The sample was update so the sample applications are where they should be in the document. There seems to be some kind of issue with safari, but I will register that as an issue.


alfonsogarciacaro:
Thanks for updating the package! Seems `docs.fsx` sneaked into the PR, I'll merge it locally to remove it :+1:

alfonsogarciacaro:
Superseded by #224

mastoj:
Good catch. I commented out the other to build the doc page faster :).

alfonsogarciacaro:
No problem, I guessed that 😉  Thanks again for the fix, I'll be updating the web shortly. I want to also add the TodoMVC sample with React so people can compare both :+1:

PR #241, opened 2016-07-08
Fix build FableCore on Windows

tpetricek:


tpetricek:
I added one more thing - I want to reference Fable via Paket Git dependency (so that I can get the latest version without waiting for a release). For this, it would be useful to have a target that builds `All` and `FableCore`, so that I can just say in my `paket.dependencies`:

    git https://github.com/fsprojects/Fable.git master build: "build.cmd AllAndCore", OS: windows
    git https://github.com/fsprojects/Fable.git master build: "build.sh AllAndCore", OS: mono

And get everything built by Paket :)

tpetricek:
Looks like you need to run `All` before `FableCore`, but I didn't want to make it a hard dependency - because I guess you often just want to run `FableCore` by hand once you've built everything.

This FAKE magic seems to be doing the trick - it says that if you are running both targets, `All` should run before `FableCore`:

    "All" ?=> "FableCore"

alfonsogarciacaro:
Thanks!

Issue #242, opened 2016-07-10
"Please upgrade fable-compiler package"


fdcastel:
`build GenerateDocs` is failing due to some weird behavior of `semver` npm package.

When the build process tries to run

`node build/fable samples/browser/mario -o ../../../temp`

Fable returns the following error:

```
Fable version: 0.4.0
Required: ^0.3.24
Please upgrade fable-compiler package
```

which doesn't make sense, since `0.4.0` should satisfy `^0.3.24` (it should fail only for `~0.3.24`)

The failing test is [here in `fable.js`](https://github.com/fsprojects/Fable/blob/master/src/fable/Fable.Client.Node/js/fable.js#L500)

Bug in `semver` package? Or I am missing something?

alfonsogarciacaro:
Argh, this is due to some semver oddity I recently read about: If I'm not mistaken, when using `^`, for versions previous to `1.0.0` it disallows updates to the `minor` version as _there're usually breaking changes between `0.x` and `0.y` versions_. I'll try to fix that, for now you can run the `UpdateSampleRequirements` build target. Hopefully we'll move soon to `1.0.0` so this is not a problem anymore :+1:

alfonsogarciacaro:
Samples have been already updated so this shouldn't be a problem anymore :+1:

PR #243, opened 2016-07-10
Plugins fix galore

fdcastel:
Fix for all current plugins build problems. Also updates the plugins tutorial in docs.

Also, please see #242.


alfonsogarciacaro:
Thanks for that!

PR #247, opened 2016-07-11
Update virtual dom and fix performance

mastoj:
* Update the virtualdom wrapper to only render with 60 fps. The first version rendered after every event causing some serious performance issues. 
* Change the definition of a `View` for virtualdom so it doesn't take in the handler as an argument, instead HTML event can only create new actions, instead of actually doing the function call.
* Update the sample.


alfonsogarciacaro:
Fantastic work, thank you!

PR #253, opened 2016-07-12
Fable core refactors

fdcastel:
Collection of refactors and changes to strong typing. All test passing.

What do you think?

### DIFF
--- a/src/fable/Fable.Core/npm/README.md
+++ b/src/fable/Fable.Core/npm/README.md
@@ -1,4 +1,4 @@
-# fable-import
+# fable-core
 
 Fable core lib & bindings for native JS objects, browser and node APIs
 
@@ -52,3 +52,27 @@ requirejs.config({
 requirejs(["app"]);
 </script>
 ```
+
+## Development
+
+The source is written in TypeScript. The use of a TypeScript-aware editor 
+(like [VSCode](https://code.visualstudio.com) or [ALM](http://alm.tools/))
+is highly recommended.  
+
+Output files are generated as shown below:
+
+- `fable-core.ts` (source)
+  - `es2015.js` (Created from `fable-core.ts |> tsc --target ES2015`)
+    - `fable-core.js` (Created from `es2015.js |> babel --plugins ...umd`)
+      - `fable-core.min.js` (Created from `fable-core.js |> uglifyjs`)
+    - `commonjs.js` (Created from `es2015.js |> babel --plugins ...commonjs`)
+
+### To build
+
+Simply run
+
+```
+build fablecore
+``` 
+
+from Fable root project folder.
--- a/src/fable/Fable.Core/npm/fable-core.ts
+++ b/src/fable/Fable.Core/npm/fable-core.ts
@@ -4,13 +4,29 @@ const FSymbol = {
 };
 export { FSymbol as Symbol };
 
-export class Choice {
-  public Case: any;
-  public Fields: any;
+export class Choice<T1, T2> {
+  public Case: "Choice1Of2" | "Choice2Of2";
+  public Fields: Array<T1 | T2>;
 
-  constructor(t: any, d: any) {
+  constructor(t: "Choice1Of2" | "Choice2Of2", d: T1 | T2) {
     this.Case = t;
     this.Fields = [d];
+  }
+
+  static Choice1Of2<T1, T2>(v: T1) {
+    return new Choice<T1, T2>("Choice1Of2", v)
+  }
+
+  static Choice2Of2<T1, T2>(v: T2) {
+    return new Choice<T1, T2>("Choice2Of2", v)
+  }
+
+  get valueIfChoice1() {
+    return this.Case == "Choice1Of2" ? this.Fields[0] : null;
+  }
+
+  get valueIfChoice2() {
+    return this.Case == "Choice2Of2" ? this.Fields[0] : null;
   }
 }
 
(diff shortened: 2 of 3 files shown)


fdcastel:
P.S.: I'm very sorry for the change in the coding style and I recognize this is very bad etiquette from my part. If this new style for `.ts` code is not ok, I would gladly reformat the code to the old one (or another one you wish to adopt).

I did these refactors piece by piece and I have been using the change in style as a way to know what parts of the code I already covered (there's yet some parts untouched, as you can see).

alfonsogarciacaro:
If the tests pass and this is useful for you to interact with `fable-core` from TypeScript I'm OK with that :+1: If, after merging lots of hidden bugs appear, I'll send the angry mob to you :wink:

Don't worry about the style, if we're making the move to TypeScript, I agree to match a more idiomatic style in the language, so that's fine 😄 It'd be nice if you could write a tutorial about how to interact with F#/Fable/fable-core from with TypeScript.

Do you want to merge this now? Or will you be adding more commits to the PR?

fdcastel:
Well.. Alpha means Alpha... :sweat_smile: 

Hehehe. Just kidding. I understand your worries but I took the utmost care in trying to keep everything as before. About 95% of the changes were just syntactical ones. The output code is the same.

However, there's some parts (those 5%...) where I replaced an imperative code by its functional counterpart (like most of the rest). E.g.: https://github.com/fsprojects/Fable/pull/253/commits/f4b0ea5c4a8b55a2392804560357e3ebe9b95e01 (replacing `forEach` and `fold`/`forEach` mix with pure `fold`)

> It'd be nice if you could write a tutorial about how to interact with F#/Fable/fable-core from with TypeScript.

Sure! Added to the task list. -- I'll just have to fix the `integration-typescript` sample before (it is _very_ broken as of now... I'll get to this later)

Finally: Please hold the merge. I have some future improvements yet to make. I just wanted to hear from you about all this ;) :+1:

alfonsogarciacaro:
Hhhm, actually I think that using the native `forEach` would be more performant that calling `Seq.fold` but I don't have actual numbers and it seems I wasn't very consistent about that anyways so probably it shouldn't be much problem.

fdcastel:
>  I think that using the native forEach would be more performant that calling Seq.fold

In principle, I agree with you. But without hard numbers we are both just assuming this. So, for now, let's just keep the code simpler and let the compilers do their job ;)

When (if*?) we got a case when it hits a performance bump, we can surely come back and optimize it.

(*) [TCOs](http://www.2ality.com/2015/06/tail-call-optimization.html) _should_ already be finding its way into engines by now.

fdcastel:
I don't know how a bug with FRegExp.replace escaped my tests yesterday. Pushed a hotfix just now. 

Working on remaining classes, yet.

fdcastel:
Merge at your discretion. Comments / Reviews are welcome.


I think I'll go take a break from all these strongly typed languages and write something in good ol' C. :smile:

tpetricek:
`FObservable` sounds like pretty weird name to me...

alfonsogarciacaro:
The main reason to add the `F` before other modules is to avoid collision with native JS objects (Symbol, String, Set, Map, etc...). `Observable` is not yet a native JS object but I think it is planned for ES7. My reason to use `Obs` before was mainly to type less ;) In any case, it must be noted these names are private to the module and the public name is indeed `Observable` (imported JS modules are usually fully qualified so there're less chances of collision).

tpetricek:
Would

    [<CompiledName("FableObservable")>]
    type Observable = (...)

work?

alfonsogarciacaro:
`fable-core.ts` is now written in TypeScript so that won't work ;) About `CompiledNameAttribute`, as of now it's not being considered by Fable (besides `StringEnum` union cases). I was thinking about it at the beginning, but I was afraid it could create some confusion when calling the method/types from JS so I'm just directly using the `DisplayName`. In any case, it's not needed to avoid conflicts with native JS objects as types, modules, members have [both a public and a private name](https://github.com/fsprojects/Fable/commit/53be707ce5442a14001621933df6fdf1ec8be39f). If I have a `Promise` module for example, it will compile to something like this (which is the same `fable-core.ts` is doing with the exported objects):

```js
var Promise_1 = /* Module code */
{ export Promise_1 as Promise }
```

tpetricek:
Aah, right! I would still go for `FableObservable` rather than `FObservable`, even if it means more bytes :)

fdcastel:
BTW: Sorry. I forgot to add here the rationale for those last renames. 

1) I created the `TTuple` type as a way to "lend a hand" to the type inference system to distinguish a tuple from an `[ArrayWith, TwoElements]`. I originally put the T in front of the type to distinguish from the constructor function `Tuple()` but then I realized the compiler is smart enough to distinguish them.

2) As for `Disposable` I first followed the commonly adopted "_Do not use "I" as a prefix for interface names_" for Typescript. But _in this specific case for Fable_ it is better to keep the original .NET name which this interface is mirroring (just like the others `IObservable`, `IEvent` etc)

3) Finally, the change from `Obs` to `FObservable` was just to keep up with the same convention used by the other types which @alfonsogarciacaro already cited.

P.S.: @tpetricek : I would also prefer the `Fable` prefix instead of `F` ;) -- If Alfonso agrees with the change I can make the "hard work" (it's just a F2 in VSCode, actually :wink: )

alfonsogarciacaro:
I'd actually appreciate it more if the contribution efforts were more towards fixing bugs or adding samples and features, rather than renaming internal variables :wink: But anyways, if `FMap`, `FSet`, etc... are a problem let's prefix them with `Fable` instead (actually `Observable` is the less problematic of all because even if it eventually becomes a native JS object, it will only conflict if we wanted to use it from within `fable-core`).

alfonsogarciacaro:
I'm merging this :+1: We can do minor changes (like renaming) later :)

PR #255, opened 2016-07-12
unify item update functionality

mexx:


alfonsogarciacaro:
The code is definitely much nicer this way 👍 Thank you!

PR #256, opened 2016-07-13
Add withInit function to allow bootstrapping app on load

devcrafting:
It is similar to what can be done with ELM: http://guide.elm-lang.org/architecture/effects/
In ELM, basically, you use App.beginnerProgram that take only a Model (as Fable virtualdom for now), but to add "effects", you use App.program that take an init (Model, Cmd Msg) tuple (ELM example: http://elm-lang.org/examples/http).

I thought it would be better integrated with a "withInit" function applied on App type in Fable virtualdom implementation.

### DIFF
--- a/import/virtualdom/Fable.Helpers.Virtualdom.fs
+++ b/import/virtualdom/Fable.Helpers.Virtualdom.fs
@@ -338,6 +338,7 @@ module App =
     type App<'TModel, 'TMessage> =
         {
             AppState: AppState<'TModel, 'TMessage>
+            Init : ('TMessage -> unit) -> unit
             JsCalls: (unit -> unit) list
             Node: Node option
             CurrentTree: obj option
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Looks good! 👍  @mastoj I gave you write rights so you can review the PR and merge it if it gets your approval 😉

mastoj:
Awesome that you're helping out @devcrafting !

Could we change the signature to `Init : 'TMessage option` and then have it set to `None` as default? No need for the init function to know about how things is post to the application. It should be necessary to have a function that might create an initial model and might kick off a long running process.

I'm thinking that setting the initial model is done directly when you call `createApp` where the model is part of the `appState` argument. The `Init` property should be used to kick off the first action basically and then it should be enough to have it return `'TMessage` and we handle the call inside the agent so the user doesn't have to think about it.

Does it make sense?

alfonsogarciacaro:
@mastoj, as you've merged #269, can this be closed now?

mastoj:
Forget about this one, but yes. Closing in 1... 2... 3.

Merged with: https://github.com/fsprojects/Fable/pull/269

PR #261, opened 2016-07-14
Fix for Typescript sample

fdcastel:
With the `fable-core.ts` and `.d.ts` files in npm package now we don't need the dummy `index.d.ts` anymore.

Also, I added a `typings` key in project's `package.json` pointing to `fable-core.d.ts`.

Finally: I believe the files in the current  fable-core@0.2.0-alpha were compiled with TypeScript 2.0 :cry:. 

E.g. The line 102 of `fable-core.d.ts`

    readonly elapsed: number;

fails since TypeScript 1.8 doesn't recognize the `readonly` modifer.

Pushing a new version of the package with the changes in this PR should solve all these problems and make the sample build again.


fdcastel:
Quick reminder: we also need to publish a new release of fable-core to fix this ;)

fdcastel:
Everything working now! (with `fable 0.4.3` and `fable-core 0.2.2`) :tada:

Issue #266, opened 2016-07-16
build GenerateDocs fails on Windows


tpetricek:
One more build failure - I tried running `build GenerateDocs` and it failed because it could not find `fable-core.min.js`, which is probably due to #264. After changing the `docs.fsx` to use `fable-core.js` instead of `fable-core.min.js`, I got another error:

    Deleting contents of C:\Tomas\Public\fsprojects\Fable\temp
    node build/fable samples/node/nunit -o ../../../temp --symbols TUTORIAL
    npm install
    (...)
    Start compilation...
    Compiled index.fsx at 10:18:08 PM
    npm test

    > @ test C:\Tomas\Public\fsprojects\Fable\samples\node\nunit
    > mocha out

    ERROR: C:\Tomas\Public\fsprojects\Fable\samples\node\nunit\node_modules\mocha\lib\utils.js:634
        throw new Error("cannot resolve path (or pattern) '" + path + "'");

alfonsogarciacaro:
I have changed the npm "test" script from `mocha out` to `mocha ./out`. I'll check later if this has fixed the problem on Windows.

fdcastel:
Windows 8.1:

```
mkdir C:\AVeryLargeFolderNameLifeUniverseAndEverything
pushd C:\AVeryLargeFolderNameLifeUniverseAndEverything
git clone https://github.com/fsprojects/Fable.git
cd Fable
build
build FableCore
```

all fine in Windows 8.1. However,

`build GenerateDocs`

is failing with:

```
Generating sample page: mario
Compiling sample: mario
Deleting contents of C:\AVeryLargeFolderNameLifeUniverseAndEverything\Fable\temp
node build/fable samples/browser/mario -o ../../../temp --symbols TUTORIAL
module.js:442
    throw err;
    ^

Error: Cannot find module 'C:\AVeryLargeFolderNameLifeUniverseAndEverything\Fable\build\fable'
    at Function.Module._resolveFilename (module.js:440:15)
    at Function.Module._load (module.js:388:25)
    at Module.runMain (module.js:575:10)
    at run (node.js:348:7)
    at startup (node.js:140:9)
    at node.js:463:3
 Running build failed.
Error:
System.Exception: 'C:\AVeryLargeFolderNameLifeUniverseAndEverything\Fable> node build/fable samples/browser/mario -o ../
../../temp --symbols TUTORIAL' task failed
[...]

PR #267, opened 2016-07-16
Fix tryWith in async, tweak docs

tpetricek:
This fixes an issue with `tryWith` - the function to handle exceptions returns another async, rather than just a value so it needs to be called with `ctx` as an argument.

I added a test but I was not able to run it, because it seems that no matter what I do `build MochaTests` uses the latest version of `fable-core` from `npm`. Is that the case? If so, it would be good to change how this works...

I also wanted to tweak the samples page - make all images the same size & use nicer NUnit logo, but I was not able to check how it looks due to #266. 

I'm not sure what to do to make things run - but I'd be happy to test the docs & run the tests if I can get some help with running things :)


alfonsogarciacaro:
Great, thank you! Interesting, it seems the TypeScript compiler was already [giving a hint about this](https://github.com/fsprojects/Fable/pull/253/commits/b81953278b9fc71ff5efd01336bb5c82936e3cf6#diff-edf454fa124698028edfb062fd433671R2068) ;)

About the tests, [in the development machine](https://github.com/fsprojects/Fable/blob/master/build.fsx#L220) I'm using the development version of `fable-core.js` while in the CI server I'm using the last version in npm. This was a way to remind me to update fable-core in npm, but it's true it's causing confusion among contributors so probably I should find another way (like publishing together fable-compiler and fable-core).

Issue #257, opened 2016-07-13
Align JSON serialisation with json.net conventions


alfonsogarciacaro:
http://www.newtonsoft.com/json/help/html/serializetypenamehandling.htm

PR #280, opened 2016-07-25
Use compiled names, add comparison methods and persistent data structures

alfonsogarciacaro:
This PR is another mixed bag. Originally thought to align JSON serialisation conventions with Newtonsoft.Json (#257) at the end it includes several other features.

* Use compiled names (instead of display names) to represent entities and member in JS. This is because Newtonsoft.Json uses the compiled type names (e.g. `FSharpList`) and also gives more flexibility to developers (who can now use `CompiledNameAttribute` when necessary). However, this also means that some methods in `fable-core` are a bit different than the display names (e.g. `mapi` vs `mapIndexed`).
* Align case of .NET interface and type members. The method names of some particular interfaces are not lowered any more like before (except for `toString`), this simplifies dealing with interfaces from .NET BCL. Now most instance method in `fable-core` start with upper case (static methods still use lower case though).
* Add comparison methods to records and unions without `ReferenceEqualityAttribute`.
* Add persistent data structures: Sets and Maps (so when deserialising they can be told apart from JS native mutable Set and Map).
* Allow Null literal for Fable.Import types in Fable.Core

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -293,11 +293,11 @@ Target "FableCore" (fun _ ->
     Npm.script fableCoreNpmDir "tsc" ["fable-core.ts --target ES2015 --declaration"]
     // Compile Es2015 syntax to ES5 with different module targets 
     setEnvironVar "BABEL_ENV" "target-commonjs"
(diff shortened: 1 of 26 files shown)


fdcastel:
Nice! :+1: :smile:

PR #284, opened 2016-07-26
Add a Array.choose

davidtme:
I'm doing a bit of work with canvas and arrays so just trying to speed up a few bits.

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -865,8 +865,8 @@ module private AstPass =
               "replicate"; "reverse"; "singleton"; "unzip"; "unzip3" ]
 
     let implementedArrayFunctions =
-        set [ "copyTo"; "partition"; "permute"; "sortInPlaceBy"; "unzip"; "unzip3" ]
-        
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
This is a nice addition, thanks! The only thing that worries me is that `Array.choose` can produce a numeric array and with this change the new array won't become a typed array anymore (actually my intention for using typed arrays actually was to have a performance boost when using canvas). This could be fixed but of course, if you create a typed array there's no need to build the intermediate array anymore.

Hmmm, I need to think about this. In any case, please remember you can create a plugin to replace `Array.choose` (I can help you with that if you need it) so you'll get the desired result in your compilations (plugins always have precedence over Fable's replacements). I also added recently code injection plugins (I haven't documented them yet though), so you can now inject an arbitrary function and replace `Array.choose` with calls to this function instead of inlining code every time.

davidtme:
Thanks for the tip about the plugins. I’ll give that a shot while you're have a think.  One thing I did notice is I *think* Array.groupBy is returning a list (I've also trying to speed up array groupBy and concat).

alfonsogarciacaro:
As the issue has been fixed using plugins I'm closing the PR for now. I hope this is OK for you, please tell me otherwise.

Cheers!

PR #301, opened 2016-07-29
Adding LSystems demo by Andrea and Ross

tpetricek:
This is LSystems demo from yesterday's London meetup by @andrea and @pezipink. I still need to clean this a bit & add some web page, but to make sure I do not forget to do this, here is a work-in-progress PR :-)


tpetricek:
I wonder if this test failure is some error?

alfonsogarciacaro:
I adjusted a bit too much the times in the test and sometimes it's failing in one of the two CI servers (I think it never did on my machine). A new build usually fixes the problem but I'll try to have a deeper look the next time it fails... I don't even remember now which test exactly was failing 😅

tpetricek:
This is now ready to merge :+1: 

I sadly cannot release the docs myself because I'm getting #310, but when I comment out the Nunit sample, everything works perfectly. If you can publish new docs, that would be splendid! :tada: :shipit:

alfonsogarciacaro:
Fantastic! :clap: This is done with master, right? Then it may be better to release the new version before updating the docs, I'll probably do it tomorrow 👍

tpetricek:
Yes, using master. Maybe that's why i had issues with List.rev (there is a workaround for that). Tomorrow sounds good!

alfonsogarciacaro:
Merge time! Which problems did you have with `List.rev` btw?

tpetricek:
I was getting some JavaScript error in `fable-core.js`. I thought this might be because the project was just getting `fable-core` from NPM and so it was getting an older version.

I added a workaround for this:

 * https://github.com/fsprojects/Fable/blob/master/samples/browser/lsystem/lsystem.fsx#L21

I guess this can now be safely removed....

alfonsogarciacaro:
Ok, it'll be nice if you could remove it and check that everything is working. Also, the sample is now live and I'm still getting `Maximum call stack size exceeded` for iterations close to 10. This is not related with the trampoline you added to Async, right?

Thanks a lot for your help!

tpetricek:
Will have a look! 

No, the stack size exceeded is a normal stack overflow exception that you get for a lot of iterations - just because it involves very deep recursion.

PR #304, opened 2016-07-29
Remove [was Bring back] AllAndCore target

tpetricek:
[This change](https://github.com/fsprojects/Fable/commit/37b4e195f39a3b5af32ff6afe2674cfbb906ee58#diff-6074983490330a0aac1bd23c958232d7) broke the `AllAndCore` target that I added (it removed the lines that specify its requirements, so it is no-op). Was that intentional?

I think it is good idea to have a target that builds everything - plus I really need it so that I can reference latest Fable using Paket's git dependencies. Please please please :pray:

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -330,7 +330,6 @@ Target "PublishDocs" (fun _ ->
 )
 
 Target "All" ignore
-Target "AllAndCore" ignore
 
 // Build order
 "Clean"


tpetricek:
Oh, I see...

> Cyclic dependency between FableCore and FableCompilerRelease

tpetricek:
Looks like `All` now builds everything, including `FableCore`, so I changed the PR just to remove obsolete target that's no longer used :) 

This makes my life even easier :+1:

alfonsogarciacaro:
Yeah, since we changed `fable-core` to typescript it must be compiled anyways most of the times, so I just added the `FableCore` target to the default build to make sure it works for any user cloning it for the first time :+1: (I still have `FableCompilerDebug` for quick development builds).

Besides that, probably it's because I'm dumb, but I don't have a complete understanding of how FAKE dependencies work, so I didn't know how to manage the `?==>` operator :wink:

tpetricek:
I also had only a vague idea about `?==>` :baby:

PR #305, opened 2016-07-29
Add trampoline to avoid StackOverflow in asyncs

tpetricek:
I'm using `async` too much in Fable - when you have too many `async` calls that are not actually doing anything asynchronously, you get a StackOverflow exception pretty soon.

This can be avoided by counting (guessing) how deep we are and when we are too deep, resuming the continuation in a new context with empty stack - e.g. using `setTimeout`. This is what this PR is doing.

### Potential issues

 - Is `setTimeout(f, 0)` going to work on server-side JS? (We are already using it in `Async.Sleep`, so if this is an issue, we should probably figure out what to do in `Async.Sleep` too)

 - Is 1000 the right number? It solved the problems I had in Chrome in the web context, but I really do not have idea how much is right in general. I guess we can tweak this as needed.

 - When you have async operations in your code, this change will still use `setTimeout`  every 1000 calls, because we do not know when async operation happens - this would be somehow specified by the user (when they await something).

### Alternative options

Do not do anything - we can just tell the users they have to insert `do! Async.Sleep(0)` when they run lots of async operations that are not in fact asynchronous.

### DIFF
--- a/src/fable/Fable.Core/npm/fable-core.ts
+++ b/src/fable/Fable.Core/npm/fable-core.ts
@@ -3552,12 +3552,26 @@ export interface CancellationToken {
   isCancelled: boolean;
 }
 
+const maxTrampolineCallCount = 1000;
+
+export class Trampoline {
+  private callCount = 0;
+  incrementAndCheck() {
+    return this.callCount++ > maxTrampolineCallCount;
+  }
+  hijack(f:() => void) {
+    this.callCount = 0;
+    setTimeout(f, 0);
+  }
+}
+
 export interface IAsyncContext<T> {
   onSuccess: Continuation<T>;
   onError: Continuation<any>;
   onCancel: Continuation<any>;
 
   cancelToken: CancellationToken;
+  trampoline: Trampoline;
 }
 
 export type IAsync<T> = (x: IAsyncContext<T>) => void;
@@ -3567,6 +3581,8 @@ const AsyncImpl = {
     return (ctx: IAsyncContext<T>) => {
       if (ctx.cancelToken.isCancelled)
         ctx.onCancel("cancelled");
+      else if (ctx.trampoline.incrementAndCheck()) 
+        ctx.trampoline.hijack(() => f(ctx))
       else
         try {
           return f(ctx);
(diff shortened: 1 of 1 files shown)


src/fable/Fable.Core/npm/fable-core.ts
```diff
@@ -3567,6 +3581,8 @@ const AsyncImpl = {
     return (ctx: IAsyncContext<T>) => {
       if (ctx.cancelToken.isCancelled)
         ctx.onCancel("cancelled");
+      else if (ctx.trampoline.incrementAndCheck()) 
+        ctx.trampoline.hijack(() => f(ctx))
```
alfonsogarciacaro: Maybe we should wrap the callback also in a try/catch as [it was done in FunScript](https://github.com/ZachBray/FunScript/blob/master/src/main/FunScript/Core/Async.fs#L38-L39).

alfonsogarciacaro:
I think this makes a lot of sense, thank you! Actually, this was already [taken into account in FunScript](https://github.com/ZachBray/FunScript/blob/master/src/main/FunScript/Core/Async.fs#L35) but it seems I missed when translating the code to `fable-core` 😞 

> Is setTimeout(f, 0) going to work on server-side JS? (We are already using it in Async.Sleep, so if this is an issue, we should probably figure out what to do in Async.Sleep too)

`setTimeout` also works in node so I guess this shouldn't be a problem.

> Is 1000 the right number? It solved the problems I had in Chrome in the web context, but I really do not have idea how much is right in general. I guess we can tweak this as needed.

There's also a note in FunScript saying the number is pretty arbitrary. However, I'm not sure setting it as a parameter will solve the problem because if we don't know which is the right number the user probably won't either :wink: I think 1000 is a good number if it fixes the stack overflow problems. If we're worried about the performance hit of `setTimeout(f, 0)` we can set it a bit higher, see [here some tests of stack sizes in different browsers](http://stackoverflow.com/a/7828803/3922220) (I did it for node and got `20183` on my Macbook), but I guess it shouldn't be necessary.

> When you have async operations in your code, this change will still use setTimeout every 1000 calls, because we do not know when async operation happens - this would be somehow specified by the user (when they await something).

I'm actually not sure about this. Anyways, it shouldn't be harmful to insert the `setTimeout(f, 0)` every 1000 calls even if there're actual async calls in between, right?


In any case, I think this is very needed and I'm happy to merge it. I just left a comment, please have a look and tell me what you think.

Thanks a lot for the fix!

tpetricek:
Good catch with the exception handler!

 - The last commit adds `try .. catch` in the hijacked lambda
 - I also incremented 1000 to 2000, which seems to work fine in my scenario (it might be easier to start with higher number and decrement it if someone complains :)). I know for sure 10000 was too much for me.

alfonsogarciacaro:
Perfect, thanks for the modification!

Issue #279, opened 2016-07-24
Make the (?) operator optional


tpetricek:
The `?` operator is great for JS interop, but as it is defined as a global value, it takes precedence over `?` that you might define for your own types.

For example, I'm experimenting with a little DSL for building HTML nodes:

    type DomNode = 
      | Text of string
      | Element of tag:string * attributes : (string * string)[] * children : DomNode[]

    type El() = 
      static member (?) (_:El, n:string) = fun a b ->
        Element(n, Array.ofList a, Array.ofList b)

    let text s = Text(s)
    let h = El()

You can write:

    h?div ["class","test"] [
      h?h1 [] [ text "Heading" ]
      h?p [] [ text "Some long text" ] 
    ]

Sadly, this only works if I do not open `Fable.Core`, because otherwise the `?` in `Fable.Core.Operators` hides the one that my `El` type defines.

Would it make sense to either not make the `Operators` module `AutoOpen` (and perhaps rename it to `JsInterop`, which sounds like a more suitable name)?

alfonsogarciacaro:
I guess this makes sense. It will be kind of a breaking change but it should be OK at this point. Only thing is I didn't want to force the user to open many namespaces, it's a pity `?` is not overriden by the custom implementation in this case.

PR #309, opened 2016-08-01
Add Fable.Import.Test.fsproj to build

forki:


forki:
all the errors are unrelated and already there. what do we do?

alfonsogarciacaro:
That's because this breaking change: #279. Let me finish v0.5.0, update the bindings and the samples and then we can merge the PR :+1:

forki:
cool. just let me know if I have to do something.

alfonsogarciacaro:
New version is about to be released, so let's start fixing these things :+1:

PR #315, opened 2016-08-02
Add mapping for Array.truncate

tpetricek:
Working on a fix...

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -867,7 +867,7 @@ module private AstPass =
               "replicate"; "reverse"; "singleton"; "unzip"; "unzip3" ]
 
     let implementedArrayFunctions =
-        set [ "copyTo"; "partition"; "permute"; "sortInPlaceBy"; "unzip"; "unzip3" ]
+        set [ "copyTo"; "partition"; "permute"; "sortInPlaceBy"; "unzip"; "unzip3"; "truncate" ]
 
     let nativeArrayFunctions =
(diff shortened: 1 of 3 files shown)


tpetricek:
I'm not sure if this is the nicest way of doing it, because we can almost map `Array.truncate` to JavaScripts `slice` function - but not quite, because `Array.truncate n a` should become `a.slice(0, n)`. So instead of adding ad-hoc case, I added a function to `fable-core.ts` - but that makes the core library larger... so perhaps ad-hoc case would be better?

(I'm happy to change this, thought I'd add something simple to learn a bit about the internals :))

tpetricek:
I'm getting some failures here - did I break anything?

 1. Comparison Record equality works:  
     TypeError: Cannot read property 'length' of undefined
 2. Seqs Seq.sum with non numeric types works:  
     TypeError: Cannot read property 'length' of undefined
 3. Seqs Seq.sumBy with non numeric types works:  
     TypeError: Cannot read property 'length' of undefined

alfonsogarciacaro:
Yeah, it might be better an add hoc case like [getSubArray](https://github.com/fsprojects/Fable/blob/master/src/fable/Fable.Compiler/Replacements.fs#L1086-L1089). Actually I'm betraying myself, fable-core will almost double in the next release 😮 But after minification and gzipping is still a very reasonable size (around 20 kb). Anyway, I'll merge this and modify it myself. Don't worry about the tests, I'm about to release a new _major_ version (before 1.0) so it's the perfect moment to change everything, that's what hipsters do :wink:

Thanks for the feature and the test!

PR #319, opened 2016-08-03
Do not delete obj dirs in node_modules

tpetricek:
On Windows, deleting `src/fable/**/obj` causes error because "path too long". I suspect this is because FAKE tries looking into `src/fable/Fable.Core/npm/node_modules` to see if there are any `obj` folders there.

AFAIK, this change means that only `obj` folders immediately under `fable`, e.g. `fable/Fable.Core/obj`, but not in more deeply nested folders - which seems to fix the issue for me!

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -144,8 +144,9 @@ module Fake =
 // Targets
 Target "Clean" (fun _ ->
     // Don't delete node_modules for faster builds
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Interesting, I didn't know about that trick, thanks for that!

tpetricek:
I also didn't know about it, but I tried and it seems to do what I was hoping for :laughing:

PR #320, opened 2016-08-03
Feature/htmlmap

mastoj:
The html map feature expanded somewhat. This adds the following concepts and changes: 

* Producers - async actions that produces messages to the application triggered outside of the application, a timer is a simple example.
* Actions - is async actions triggered from the update function.
* The state of the application has been simplified
* Event handling has been simplified, mostly due to the large amount of event types in js that is a mess. Now there is only one event handling type, and it is up to the user to define how the event should be decoded. In time we could add helpers for the most common scenarios.

### DIFF
--- a/samples/browser/virtualdom/virtualdom.fsx
+++ b/samples/browser/virtualdom/virtualdom.fsx
@@ -9,8 +9,8 @@
 *)
 (*** hide ***)
 #r "node_modules/fable-core/Fable.Core.dll"
-//#load "node_modules/fable-import-virtualdom/Fable.Helpers.Virtualdom.fs"
-#load "../../../import/virtualdom/Fable.Helpers.Virtualdom.fs"
+#load "node_modules/fable-import-virtualdom/Fable.Helpers.Virtualdom.fs"
+//#load "../../../import/virtualdom/Fable.Helpers.Virtualdom.fs"
 (**
 ##Architecture overview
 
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
Nice! 👏 

Just one thing: it seems `samples/browser/lsystem/lsystem.fsproj` sneaked into the PR (probably it was modified by Paket or Forge). Could you please remove it, please?

samples/browser/virtualdom/virtualdom.fsx
```diff
@@ -117,19 +117,80 @@ You define a `svg` the same way as you do with any other html element.
...
+
+type NestedModel = { Top: int; Bottom: int}
+
+type NestedAction = 
+    | Reset
+    | Top of CounterAction
+    | Bottom of CounterAction
+
+let nestedUpdate model action = 
+    let mapCounterAction tag action =
```
MangelMaxime: This function is declared but never used.

samples/browser/virtualdom/virtualdom.fsx
```diff
@@ -56,16 +56,16 @@ type CounterAction =
...
     else window.setTimeout((fun _ -> h (message)), 2000) |> ignore
 
-let counterUpdate model action =
-    match action with
+let counterUpdate model command =
+    match command with
     | Decrement x -> model - x
     | Increment x -> model + x
-    |> (fun m -> m,[],[fakeAjaxCall model])
+    |> (fun m -> m, (fakeAjaxCall model) |> action)
```
MangelMaxime: If I understand well, this `action` function is a wrapper to transform the left into a list of action.

I would change the named of the function to something like: `toActionList`
Just to understand what we are doing with this function.

mastoj: Good comment, changed to `toActionList`.

samples/browser/virtualdom/virtualdom.fsx
mastoj: Nice catch, updated the PR and removed the function.

alfonsogarciacaro:
Merged!

PR #326, opened 2016-08-07
Implement fetchAs for Fetch bindings

forki:
This adds a fetchAs<'T> method to the Fetch npm package.

This allows us to download a json file and driectly unwrap it in a object of type 'T.
I also added async versions. This allows us to avoid explict Async.AwaitAsPromise in suer code.
I also added RequireQualified access everywhere, because of name clashes.

/cc @tpetricek @7sharp9 @alfonsogarciacaro


alfonsogarciacaro:
Only problem with this is that, by convention, `Fable.Import` prefixed files include only bindings and Fable will ignore them so the utility functions will be lost in the compiled JS code. This is the case even if the functions are inlined.

In order for this to work, we would have to change Fable's behaviour (not sure if that's desirable) or  put the helper functions in a `Fable.Helpers.Fetch.fs` file as with React or Virtualdom.

forki:
Are you sure? I mean I use it like that!?

alfonsogarciacaro:
You're using it like that and the code compiles to JS? That's interesting, import files should be ignored by [this line](https://github.com/fsprojects/Fable/blob/master/src/fable/Fable.Compiler/FSharp2Fable.fs#L960). I'll try to have a look and see if Fable isn't working as (I 😉 ) expected.

forki:
ok fixed

alfonsogarciacaro:
Thx!

PR #330, opened 2016-08-08
[WIP] DotNet Core 1.0 projects

ncave:
Work in progress (projects build manually, tests are passing, no fake build yet).


samples/browser/virtualdom/lib/virtual-dom.js
```diff
@@ -14,21 +14,21 @@ var h = require("./virtual-hyperscript/index.js")
 module.exports = h
```
ncave: No change, line endings only.

samples/other/fable-helpers-sample/src/MyLib.fs
```diff
@@ -1,11 +1,11 @@
-module Fable.Helpers.Sample.MyLib
```
ncave: No change, line endings only.

alfonsogarciacaro:
Oh my, this great! I've been wanting to tackle this for a while but I'm no dotnetcore expert. Please tell me if there's anything I can do to help :+1:

Some questions.
- This is necessary to compile F# dotnetcore projects (with project.json, netcore APIs...), right?
- Users will need to have dotnetcore installed (and in PATH) to run this. Would it be possible/practical to distribute the runtime together with Fable as a standalone crossplatform app?
- In a comment you say .NET Core MSBuild can't successfully build .fsproj yet, does this mean we'll have to keep two Fable versions (.NET/Mono and dotnetcore) for a while?

ncave:
As always, this is only an opinion so take it with 2 grains of salt:

1) No, I don't think you can target .NET Core F# projects (project.json) with Fable (yet), unless you only need the list of files from it, which is easy to do. For now, it's only .fsx that should work (but needs testing).

2) Yes, users will need to have [.NET Core](https://www.microsoft.com/net/core#macos) installed (and in PATH) to run Fable, and it definitely can be distributed as a single standalone cross-platform app (but it won't have the run-time embedded, which is native to each platform).

It does seem possible to embed the run-time in the app by compiling it to a native app, either through [.NET Native](https://github.com/dotnet/corert/blob/master/Documentation/intro-to-corert.md) or IL-to-Cpp or something like that, but it will create a different binary for each platform. For a single cross-platform binary (or source) to run, you need something native to be already installed there, or at least a compiler, same as with java/javascript/node/python etc. so the question is more of an app distribution (easier these days with docker and package managers).

3) Yes, if you need to be able to resolve dependencies in .fsproj using the FCS.ProjectCracker (which uses MSBuild), it would seem that you'll have to keep two Fable versions (.NET/Mono and dotnetcore) for a while, until .NET Core MSBuild can handle older .fsproj files. Another option is to use a different project system altogether, as discussed [here](https://github.com/fsprojects/Fable/issues/210).

alfonsogarciacaro:
Perfect, thanks a lot for your answers! I guess we could release an experimental package like `fable-dotnet` for dotnet core support :+1:

ncave:
That would be nice. I'm still trying to figure out how to generate and run all the MOCHA (JS) tests, but the F# tests all pass.

alfonsogarciacaro:
This is the [FAKE target](https://github.com/fsprojects/Fable/blob/master/build.fsx#L225-L236) that compiles the tests to mocha and runs them. Isn't it working with the netcore build? I'll try to give it a try later but basically the steps are (not sure about the commands to compile for netcore so I put them in brackets):

1. <Build src/tests/DllRef/Fable.Tests.DllRef into src/tests/DllRef/bin/>
2. node build/fable src/tests/DllRef
3. node build/fable src/tests/Other
4. node build/fable src/tests
5. <copy src/tests/package.json to build/tests/>
6. <in build/tests/ dir> npm install
7. <copy src/fable/Fable.Core/npm/fable-core.js to build/tests/node_modules/fable-core/>
8. <in build/tests/ dir> npm run test

alfonsogarciacaro:
I just tested and it's working, fantastic job! I'll be soon publishing the experimental package :+1:

PR #334, opened 2016-08-08
New post API for Fetch

forki:
This will make it easier to post data 

    async { 
        let! response =
            fetchAsyncWithInit(
                RequestInfo.Url "http://www.server.com/data.json" , 
                [ RequestProperties.Method HttpMethod.POST
                  RequestProperties.Body (unbox "hello world3!!")])
    
            // ...              
    }

Open questions:

* how to make body nicer?
* why is the first parameter of fetch a 

    type RequestInfo =
        | Url of string
        | Req of Request

it looks to me that Request can't be changed!?

/cc @7sharp9 

I also would like to see the inverse of

    let inline fetchAs<'T> (req: RequestInfo) : Async<'T> = async {
        let! fetched = GlobalFetch.fetch(req) |> Async.AwaitPromise
        let! json = fetched.text() |> Async.AwaitPromise
        return ofJson<'T> json    
    }

how would that look like?
Currently I have:

    /// Sends a HTTP post with the record serialized as JSON  
    let inline postRecord<'T> (req: RequestInfo,record:'T) : Async<Response> =
        fetchAsyncWithInit(
          req, 
          [ RequestProperties.Method HttpMethod.POST
            RequestProperties.Body (unbox (toJson record))])

### DIFF
--- a/import/fetch/Fable.Helpers.Fetch.fs
+++ b/import/fetch/Fable.Helpers.Fetch.fs
@@ -51,3 +51,10 @@ let inline fetchAs<'T> (req: RequestInfo) : Async<'T> = async {
     let! json = fetched.text() |> Async.AwaitPromise
     return ofJson<'T> json    
 }
+
+/// Sends a HTTP post with the record serialized as JSON  
+let inline postRecord<'T> (req: RequestInfo,record:'T) : Async<Response> =
+    fetchAsyncWithInit(
+      req, 
+      [ RequestProperties.Method HttpMethod.POST
+        RequestProperties.Body (unbox (toJson record))])
--- a/import/fetch/Fable.Helpers.Fetch.fs
+++ b/import/fetch/Fable.Helpers.Fetch.fs
@@ -7,17 +7,40 @@ open Fable.Core
 open Fable.Import
 open Fable.Core.JsInterop
 
+type [<StringEnum; RequireQualifiedAccess>] HttpMethod =
+| [<CompiledName("GET")>] GET 
+| [<CompiledName("HEAD")>] HEAD 
+| [<CompiledName("POST")>] POST 
+| [<CompiledName("PUT")>] PUT 
+| [<CompiledName("DELETE")>] DELETE 
+| [<CompiledName("TRACE")>] TRACE
+| [<CompiledName("CONNECT")>] CONNECT
+
+[<KeyValueList>]
+type IRequestProperties =
+    interface end
+
+[<KeyValueList>]
+type RequestProperties =
+    | Method of HttpMethod
+    | Headers of U2<HeaderInit, obj>
+    | Body of BodyInit
+    | Mode of RequestMode
+    | Credentials of RequestCredentials
+    | Cache of RequestCache
+    interface IRequestProperties
+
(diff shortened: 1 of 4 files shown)


MangelMaxime:
For the body we probably could take inspiration from [Http.Fs](https://github.com/relentless/Http.fs/blob/master/HttpFs.SamplePostApplication/Program.fs)

This version is only for form-data.

```fsharp
    // Set up parameters for small pizza with bacon and onions.
    let bodyParams :KeyValuePair list = [
        {Key = "custname"; Value = "John Doe"}
        {Key = "custtel"; Value = "123456789"}
        {Key = "custemail"; Value = "example@example.com"}
         // Radio button
        {Key = "size"; Value = "small"}
         // Checkboxes
        {Key = "topping"; Value = "bacon"}
        {Key = "topping"; Value = "cheese"}

        {Key = "delivery"; Value = "4:00"}
        {Key = "comments"; Value = ""}
        ]

    /// Function to convert parameter list to Body.
    let bodyFunc (acc:string) (tArg:KeyValuePair) = 
        // Encode characters like spaces before transmitting.
        let rName = tArg.Key |> HttpUtility.UrlEncode
        let rValue = tArg.Value |> HttpUtility.UrlEncode

        let mix = HttpUtility.UrlEncode rName + "=" + rValue
        match acc with
        | "" -> mix // Skip & for first one.
        | _ -> acc + "&" + mix  // application/x-www-form-urlencoded
    
    // Create body.
    // custname=John+Doe&custtel=123456789&custemail=example%40example.com&size=small&topping=bacon&topping=cheese&delivery=4%3a00&comments=
    let body = bodyParams |> Seq.fold bodyFunc ""
```

7sharp9:
@forki I made it a union type as the js API you can use either a `Request` or a `Url`

forki:
Yes but request has only get properties. Do you have a sample how to use
it?

forki:
Ok I looked at https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

    var myHeaders = new Headers();
    
    var myInit = { method: 'GET',
                   headers: myHeaders,
                   mode: 'cors',
                   cache: 'default' };
    
    var myRequest = new Request('flowers.jpg', myInit);
    
    fetch(myRequest, myInit)
    .then(function(response) {
      return response.blob();
    })
    .then(function(myBlob) {
      var objectURL = URL.createObjectURL(myBlob);
      myImage.src = objectURL;
    });

so we're createing a Request object by using url and myInit and then call fetch with myInit again? that doesn't make too much sense for me.

I think I want to simplify this in the helpers.

forki:
async {
        try
            let! response = postRecord("http://requestb.in/1123z4f1", x.state)
            // ...
        with
        | error -> //...
      } |> Async.StartImmediate

forki:
ready for review.

forki:
Added HTTP headers:

    async { 
        let! response = postRecord("http://www.server.com/data.json", myRecord,
                        [ RequestProperties.Headers [ 
                            HttpRequestHeaders.Accept "application/xml" ]
                        ])
        // ...              
    }

PR #335, opened 2016-08-09
[ON HOLD] Compile F# lambdas as delegates (functions with multiple arguments) 

alfonsogarciacaro:
- [ ] Write application tests
- [ ] Update docs

So far it seems the biggest source of confusion for developers when interacting with native JS code is how F# lambdas are represented in JS. At the moment, Fable doesn't do any special treatment to lambdas and their representation is very faithful to what can be seen in the AST generated by the F# compiler. One of the trickiest parts is that type and module methods have a very "classic" representation (`__.Foo(x, y)` and `__.Bar x y` compile both to an uncurried method with two arguments) and the F# compiler takes care of adding the necessary manipulations to make the methods work as expected in F#. For example, if you pass a tuple to `Foo`, it will be destructured before calling the method, and if you apply partially `Bar`, the method call will be wrapped in a lambda.

``` fsharp
typ A() =
  member __.Foo(x, y) = x + y   // function Foo(x, y) { return x + y }
  member __.Bar x y = x + y      // function Bar(x, y) { return x + y }

let a = A()
let t = 2, 3
a.Foo t       // var x = t[0], y = t[1]; a.Foo(x, y)
a.Bar 5       // var x = 5; function (arg00) { return a.Bar(x, arg00); }
```

However, lambdas do compile to what you expect in F#: anonymous function with always a single argument (that can be tupled) and which return another function in the case of curried arguments. I've tried to explain all of this and how to interact with JS by converting the functions to delegates here:
http://fsprojects.github.io/Fable/docs/interacting.html#Calling-F-code-from-JavaScript

``` fsharp
fun x y -> x + y      //  function(x) { return (y) { return x + y } }
fun (x, y) -> x + y  //  function(tupleArg) { var x = tupleArg[0], y = tupleArg[1]; return x + y }
```

It'd be ideal if Fable could do with lambdas the same thing as it does with type and module methods: represent them as "classic" delegates and let the compiler make the conversions when needed (wrapping in a lambda in partial applications, etc). First I tried flattening lambdas with `CurriedParameterGroups` as methods are represented in F# AST. So lambdas would compile as follows and in applications Fable would destruct tuples or wrap the call in another lambda for partial application:

``` fsharp
fun x y -> x + y                                       //  function(x, y) { return x + y }
fun (x, y) -> x + y                                   //  function(x, y) { return x + y }
fun (t: (int*int)) -> let x, y = t in x + y   // function(t) { var x = t[0], y = t[1]; return x + y }
```

Flattening lambdas:
https://github.com/fsprojects/Fable/blob/bf576bab16f65efe00ca98bb67ff2e305d014f87/src/fable/Fable.Compiler/FSharp2Fable.Util.fs#L308-L322

Applications:
https://github.com/fsprojects/Fable/blob/bf576bab16f65efe00ca98bb67ff2e305d014f87/src/fable/Fable.Core/AST/AST.Fable.Util.fs#L209-L242

But this fails in many situations where the lambda accepts a tuple argument and Fable thinks it must be destructured.

``` fsharp
Seq.fold (fun (a, b) x -> if a then (a, x::b) else (a, b))
// flableCore.Seq.fold(function (a, b, x) { ... })
```

So I tried then to ignore tupled arguments and transform curried lambdas as if they all were converted to delegates:

``` fsharp
fun x y -> x + y
// function(x, y) { return x + y }

fun (x, y) -> x + y
// function(tupleArg) { var x = tupleArg[0], y = tupleArg[1]; return x + y }

fun (t: (int*int)) -> let x, y = t in x + y
// function(t) { var x = t[0], y = t[1]; return x + y }
```

Then I only need to deal with partial applications:
https://github.com/fsprojects/Fable/blob/3a2b4c8b1c7781c0911a4589369c845409ef8350/src/fable/Fable.Core/AST/AST.Fable.Util.fs#L202-L210

With this, all Fable tests are passing (we've more than 800) but one:

``` fsharp
    let a = [| "a"; "b"; "c" |]
    let concaters1 = a |> Array.map (fun x y -> y + x)
    concaters1.[0] "x" |> equal "xa"
```

Compiled as follows (in the experimental branch):

``` js
    var a = ["a", "b", "c"];
    var concaters1 = a.map(function (x, y) {
      return y + x;
    });
    (0, _Util.equal)("xa", concaters1[0]("x"));
```

The solution to this problem is adding an [arity check](https://github.com/fsprojects/Fable/blob/flatten-lambdas/src/fable/Fable.Core/AST/AST.Fable.Util.fs#L192) for all method calls and applications, so when you pass a function with a higher arity than expected, Fable wraps it to prevent the unexpected behaviour. The previous sample code `[| "a"; "b"; "c" |] |> Array.map (fun x y -> y + x)` now compiles to:

``` js
  ["a", "b", "c"].map(function ($var1) {
    return function ($var2) {
      return function (x, y) {
        return x + y;
      }($var1, $var2);
    };
  });
```

Now all tests pass! 🎉 I just hope there're no other plausible scenarios where this can fail 🙏 If someone can build the branch and give it a try with their own projects to be sure everything works fine that'd be much appreciated.

### DIFF
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -305,6 +305,14 @@ module Patterns =
             else None
         | _ -> None
 
+    let (|FlattenedApplication|_|) fsExpr =
+        // Application args can be empty sometimes so a flag is needed
+        let rec flattenApplication flag args = function
+            | Application(callee, _, args2) ->
+                flattenApplication true (args2@args) callee
+            | callee -> if flag then Some(callee, args) else None 
+        flattenApplication false [] fsExpr
+
     let (|FlattenedLambda|_|) fsExpr =
         let rec flattenDestructs destructs = function
             | Let ((var, TupleGet(_,i,Value arg)), body) -> flattenDestructs ((var,i,arg)::destructs) body
@@ -316,10 +324,11 @@ module Patterns =
                     then flattenDestructs destructs body
                     else destructs, body
                 flattenLambda (arg::args) destructs body
-            | e -> args, destructs, e
-        match flattenLambda [] [] fsExpr with
-        | [], _, _ -> None
-        | args, destructs, body -> Some(List.rev args, List.rev destructs, body)
+            | body ->
+                if List.isEmpty args
+                then None
+                else Some(List.rev args, List.rev destructs, body)
+        flattenLambda [] [] fsExpr
 
     /// This matches the boilerplate F# compiler generates for methods
     /// like Dictionary.TryGetValue (see #154)
@@ -449,6 +458,13 @@ module Types =
         | args -> List.concat args |> List.map (fun x -> makeType com Context.Empty x.Type)            
 
     and getMembers com (tdef: FSharpEntity) =
+        let getArgTypes com (args: IList<IList<FSharpParameter>>) =
+            // FSharpParameters don't contain the `this` arg
+            match args |> Seq.map Seq.toList |> Seq.toList with
+            | [] -> []
+            | [[singleArg]] when isUnit singleArg.Type -> []
+            // The F# compiler "untuples" the args in methods
+            | args -> List.concat args |> List.map (fun x -> makeType com Context.Empty x.Type)
         let isOverloadable =
             // TODO: Use overload index for interfaces too? (See overloadIndex below too)
             not(tdef.IsInterface || isImported tdef || isReplaceCandidate com tdef)
@@ -791,6 +807,7 @@ module Util =
                      (typArgs, methTypArgs) callee args =
         let argTypes = getArgTypes com meth.CurriedParameterGroups                     
         let args =
+            let args = ensureArity argTypes args
             if hasRestParams meth then
                 let args = List.rev args
                 match args.Head with
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -800,7 +800,7 @@ module Util =
                      (typArgs, methTypArgs) callee args =
         let argTypes = getArgTypes com meth.CurriedParameterGroups                     
         let args =
-            let args = ensureArity argTypes args
+            let args = ensureArity com argTypes args
             if hasRestParams meth then
                 let args = List.rev args
                 match args.Head with
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -458,13 +458,6 @@ module Types =
         | args -> List.concat args |> List.map (fun x -> makeType com Context.Empty x.Type)            
 
     and getMembers com (tdef: FSharpEntity) =
-        let getArgTypes com (args: IList<IList<FSharpParameter>>) =
-            // FSharpParameters don't contain the `this` arg
-            match args |> Seq.map Seq.toList |> Seq.toList with
-            | [] -> []
-            | [[singleArg]] when isUnit singleArg.Type -> []
-            // The F# compiler "untuples" the args in methods
-            | args -> List.concat args |> List.map (fun x -> makeType com Context.Empty x.Type)
         let isOverloadable =
             // TODO: Use overload index for interfaces too? (See overloadIndex below too)
             not(tdef.IsInterface || isImported tdef || isReplaceCandidate com tdef)
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -305,6 +305,22 @@ module Patterns =
             else None
         | _ -> None
 
+    let (|FlattenedLambda|_|) fsExpr =
+        let rec destruct args = function
+            | Let ((arg, TupleGet _), body) -> destruct (arg::args) body
+            | e -> args, e
+        let rec flattenLambda args = function
+            | Lambda(arg, body) ->
+                let args, body =
+                    if arg.FullType.IsTupleType && arg.IsCompilerGenerated && arg.CompiledName = "tupledArg"
+                    then let args', body = destruct [] body in (List.rev args')::args, body
+                    else ([arg]::args), body
+                flattenLambda args body
+            | e -> args, e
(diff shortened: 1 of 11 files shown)


alfonsogarciacaro:
Another possibility to solve partial applications and other cases, instead of injecting code at compile time like this PR does, is to wrap lambdas with a runtime check that reacts accordingly if the number of passed arguments is not the same as the lambda arity.

rmunn:
I'm inclined to say "no" to the runtime arity check in lambdas, and "yes" to this PR's solution. The runtime check would add a usually-unnecessary check to *every* lambda, whereas this PR's compile-time solution only adds the partial-application code to the *call site*, rather than the call target (the lambda itself). That means that only the parts of the Javascript code that actually *need* the partial-application code will get it, and the rest becomes simple.

In other words, I want `fun x y -> x + y` to compile to `function (x, y) { return x + y; }` instead of `function () { if (arguments.length == 2) { return arguments[0] + arguments[1]; } else return function ($var2) { return ... /* some nested function-inside-function that I can't write off the top of my head */ }; }`. The latter is **not** "JavaScript you can be proud of", but the former **is**.

alfonsogarciacaro:
Thanks a lot for sharing your opinion @rmunn. I was thinking that we could create an utility method  to avoid creating the boilerplate for every function, but unfortunately it's not possible to use JS `f.apply(this, arguments)` without changing the `this` context and this will cause problems in some scenarios. So, as you say, the runtime checks would have to be created ad hoc for **every** lambda 😞

alfonsogarciacaro:
The only problem we've found so far is that sometimes side effects may be delayed. For example, consider:

```fsharp
let test (f: 'a->'b->'c) x = f x
test (fun x -> printfn "hello"; (fun y -> x + y)) 2
```

In order to match the expected arity (=number of arguments) and apply arguments correctly, the compiler does the following transformations in JS:

```js
function test(f, x) {
  return function ($var1) {
    return f(x, $var1);
  };
}

test(function ($var2, $var3) {
  return function (x) {
    _fableCore.String.fsFormat("hello")(function (x) {
      console.log(x);
    }, true);

    return function (y) {
      return x + y;
    };
  }($var2)($var3);
}, 2);
```

Unfortunately this means "hello" won't be printed (it does in F# interactive). It will happen when a second argument is applied though.

We'll raise a warning about the issue (_Side effects may be delayed_) whenever a function passed to another function is wrapped because it has less arguments than expected.

tpetricek:
I do not think you can detect arity of a function at compile-time - when you have code in a generic function, you do not know if the application will be partial or not.

Would the _application_ in the `call` function need to be compiled in two different ways to make the following work?

    let call f = f 42

    call (fun x y -> x + y)
    call (fun x -> x + 1)

The same is the case with _function declaration_ in the following `call` function - it may or may not be partial:

    let call f g = f (fun x -> g)

    call (fun f -> f 1)
    call (fun f -> f 1 2)

So, I think to make this work, injecting some dynamic JS check either on the call-site or on the declaration-site is needed. I think doing that when lambdas are created (using some helper function so that it is not ugly) is probably the sensible solution.

alfonsogarciacaro:
I think dynamic checks are going to be difficult, because of the performance penalty (though I'm not sure how much would that be) and because of the problem mentioned above ([in JS you cannot do `f.apply(this, arguments)` without changing the `this` context](http://stackoverflow.com/questions/5247060/in-javascript-is-there-equivalent-to-apply-that-doesnt-change-the-value-of-thi)) so the checks would probably have to be added on site, and that'd be too much boilerplate.

The good news however is that the compiler can actually resolve most of the situation both in the declaration site and the call site. Your sample code with the flatten-lambda branch compiles as follows:

```js
// let print (i: int) = System.Console.WriteLine(i)

// let call f = f 42

// call (fun x y -> x + y) 3 |> print
// call (fun x -> x + 1) |> print

function call(f) {
  return f(42);
}

print(call(function ($var7) {
  return function ($var8) {
    return function (x, y) {
      return x + y;
    }($var7, $var8);
  };
})(3));

print(call(function (x) {
  return x + 1;
}));

// let call2 f g = f (fun x -> g)

// call2 (fun f -> f 1) 3 |> printf

function call2(f, g) {
  return f(function (x) {
    return g;
  });
}

print(call2(function (f) {
  return f(1);
}, 3));

// call2 (fun f -> f 1 2) (fun x -> x) |> print

print(call2(function (f) {
  return f(1, 2);
}, function (x) {
  return x;
}));
```

However, it's true the last sample fails (`call2 (fun f -> f 1 2) (fun x -> x)`). I don't know how many times this situation will occur in real code. May we could prevent it by forbidding to pass functions to generic arguments (it would still be possible if the parameter is marked as `obj`), what do you think?

There may still be some edge cases that can fail and we haven't thought of yet. So I guess the question is whether the benefit of easy interop with JS code using F# lambdas directly is worth the risky, or we should aim for safety and keep forcing users to convert functions to delegates whenever they want to send them to JS.

tpetricek:
I'm afraid the edge cases would come up more often then we might think, so I think doing anything unsafe is not a good solution (after all, that's why people should use F# over JavaScript :-)). 

As for disallowing certain uses of generics, I think that would have to disallow too much - in the end, you'd have to disallow even simple things like `list<int -> string>` which is quite useful pattern.

I do not particularly mind using delegates (the compiler lets you use normal function as an argument most of the times anyway).

It is a very tricky problem... What is the primary motivation? Avoiding the use of delegates in imported JS libraries?

alfonsogarciacaro:
Yeah, after writing the comment I realised it's not a good idea making restriction on generic arguments. And yes, there'll probably be edge cases we may miss one way or another... 😕 

About the motivation, it seems that most people find the conversion to delegates very cumbersome. Although the compiler makes it automatically in some cases, this doesn't work in several others, particularly when using dynamic programming, which confuses some users. And apparently nobody managed to understand [this section of the docs](https://fsprojects.github.io/Fable/docs/interacting.html#Calling-F-code-from-JavaScript) but that's probably my bad writing 😅 

The community seems to be happy to adopt this change in spite of the risks. But I also agree that's a pity to _intentionally_ leave holes in Fable implementation, so I'm not sure...

Maybe we could add a compiler flag to activate this option. I think it's not a very good idea because it may cause confusions too and make it difficult to share code written using different compiler options. But we could do it during a transition phase to get feedback from users.

 ...or we could forget about this PR and try to improve JS interop in other ways. For example, converting lambdas to delegates automatically in dynamic applications or when creating a JS object.

PR #342, opened 2016-08-10
Add Promise.reject<T> into PromiseConstructor

kirill-gerasimenko:
Hi, 
I'd like to add missing reject<T> to Promises. I'm using it in Ionide project (PR i'm working on now). 
Please let me know if I missed something.

Thanks!

### DIFF
--- a/import/vscode/Fable.Import.VSCode.fs
+++ b/import/vscode/Fable.Import.VSCode.fs
@@ -18,6 +18,7 @@ and PromiseConstructor =
     abstract all: values: ResizeArray<U2<'T, Thenable<'T>>> -> Promise<ResizeArray<'T>>
     abstract race: values: ResizeArray<U2<'T, Thenable<'T>>> -> Promise<'T>
     abstract reject: reason: obj -> Promise<unit>
+    abstract reject: reason: obj -> Promise<'T>
     abstract resolve: value: U2<'T, Thenable<'T>> -> Promise<'T>
     abstract resolve: unit -> Promise<unit>
 
--- a/src/fable/Fable.Core/Import/Fable.Import.JS.fs
+++ b/src/fable/Fable.Core/Import/Fable.Import.JS.fs
@@ -1134,6 +1134,7 @@ module JS =
         abstract all: values: Iterable<U2<'TAll, PromiseLike<'TAll>>> -> Promise<ResizeArray<'TAll>>
         abstract race: values: Iterable<U2<'T, PromiseLike<'T>>> -> Promise<'T>
         abstract reject: reason: obj -> Promise<unit>
+        abstract reject: reason: obj -> Promise<'T>
         abstract resolve: value: U2<'T, PromiseLike<'T>> -> Promise<'T>
         abstract resolve: unit -> Promise<unit>
 


alfonsogarciacaro:
Hi @kirill-gerasimenko, thanks a lot for your PR! Could you please provide some sample code of how you would use this? I'm concerned that having two overloads with the same argument types can confuse the compiler, but I may be missing something.

kirill-gerasimenko:
Sure, 

Here is start() method from LanguageServices.ts in Ionide: 

```
   let start () =
        promise {
            let path = (VSCode.getPluginPath "Ionide.Ionide-fsharp") + "/bin/FsAutoComplete.Suave.exe"
            let child = 
                if Process.isMono () then 
                    let mono = workspace.getConfiguration().get("FSharp.monoPath", "mono")
                    child_process.spawn(mono, [| path; string port|] |> ResizeArray)
                else
                    child_process.spawn(path, [| string port|] |> ResizeArray)

            return!
                if JS.isDefined child?pid then
                    Promise.resolve child
                else
                    Promise.reject<child_process_types.ChildProcess> ()
        }
        |> Promise.``then``  
            (fun suaveProcess ->
                service <- Some suaveProcess
                suaveProcess?on ("error", fun e -> Browser.console.error (e.ToString())) |> ignore
                suaveProcess.stderr?on $ ("data", fun n -> Browser.console.error (n.ToString())) |> ignore
                suaveProcess.stdout?on $ ("data", fun n -> Browser.console.log (n.ToString())) |> ignore
                Promise.resolve())
            (fun error -> 
                if Process.isMono () then
                    "Failed to start language services. Please check if mono is in PATH, or correct path is used in FSharp.monoPath setting"
                else
                    "Failed to start language services. Please check if Microsoft Build Tools 2013 are installed"
                |> vscode.window.showErrorMessage
                |> ignore) 
```

Using `Promise.reject<ChildProcess>` - makes this code compile, since types of the result in `promise {...}` fit.
The code doesn't directly use 'reject<T>' method, but rather one from Ionide helper, which is a simple wrapper which doesn't have other 'reject: obj -> unit' method. Something like this would be needed if both method are needed (but the first 'reject<T>' is more generic than 'rejectUnit', no need to have latter at all). 

```
    let reject<'T> reason : Promise<'T> =
        Promise.reject<'T> reason

    let rejectUnit (reason: obj -> Promise<unit>) : Promise<unit> =
        Promise.reject reason
```

Does it make sense?

alfonsogarciacaro:
Perfect, thanks a lot for the explanation! I sometimes get confused with F# overload resolution rules 😅

alfonsogarciacaro:
Shall I republish fable-import-vscode after merging this, @Krzysztof-Cieslak? For fable-core, the change will be included in the next release.

kirill-gerasimenko:
@alfonsogarciacaro Thanks!

kirill-gerasimenko:
Hi again, 

Now I realize that in order to use this change in another PR I need to have this released first :) So the question is - when is the next fable-core release? 

Thanks!

alfonsogarciacaro:
I have to add another couple of fixes, I'll try to publish the new release 0.5.3 this weekend. Will that work for you?

alfonsogarciacaro:
I've release fable-compiler & fable-core 0.5.3 with the fix :+1: I'll soon publish another release with a couple of more fixes.

kirill-gerasimenko:
Great, thanks!

PR #349, opened 2016-08-12
Show broken HashSet

forki:


alfonsogarciacaro:
Thanks! Now I have to fix this ;)

forki:
Will add some more. for remove and stuff ;-)

alfonsogarciacaro:
I think remove is already being tested as it also returns true/false in JS. Not sure which other methods have return values different from .NET though.

forki:
yes just saw that. lol

PR #351, opened 2016-08-12
Adding a complex sudoku test

forki:
I'm proposing to add this as a test.

@alfonsogarciacaro please take a look at the last line. is this expected to work?

### DIFF
--- a/src/tests/SudokuTest.fs
+++ b/src/tests/SudokuTest.fs
@@ -1,0 +1,104 @@
+﻿/// This module tests a couple of collection functions in a more complex setting.
+[<NUnit.Framework.TestFixture>] 
+module Fable.Tests.Sudoku
+
+open NUnit.Framework
+open Fable.Tests.Util
+
+open System
+
+open System.Collections.Generic
+
+type Box = int
+type Sudoku = Box array array
+
+let rows = id
+let cols (sudoku:Sudoku) =
+    sudoku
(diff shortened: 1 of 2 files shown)


src/tests/SudokuTest.fs
```diff
@@ -0,0 +1,104 @@
...
+         [2; 1; 6;  4; 3; 8;  7; 5; 9]
+         [4; 8; 5;  7; 9; 1;  3; 2; 6]
+         [3; 9; 7;  5; 2; 6;  4; 8; 1]
+
+         [7; 6; 2;  9; 1; 4;  8; 3; 5]
+         [9; 4; 3;  8; 5; 7;  1; 6; 2]
+         [8; 5; 1;  2; 6; 3;  9; 7; 4]]
+            |> toSudoku
+
+    solution |> equal expectedSolution
```
alfonsogarciacaro: This won't work because it defaults to `Assert.AreEqual` which translates to `assert.equal` and only tests JS equality. However, the fix it's easy: `solution = expectedSolution |> equal true`. If that doesn't work it means we need to fix comparison functions too ;)

alfonsogarciacaro:
This is nice! :+1: I just left a comment to reply your question. Ideally I'd like to have a pool of projects or scripts with their own tests and check all of them work with every major release. Right now I'm doing something like that manually by compiling the web samples and checking they sill work :)

forki:
fixed that assertion. hope I manually solved it correctly :P

alfonsogarciacaro:
Thanks!

PR #356, opened 2016-08-14
Update deps in react-todo and add fable-compiler

forki:
This made it work for me


alfonsogarciacaro:
Sorry, I've updated automatically the dependencies of all samples for 0.5.4 release, so this shouldn't be needed anymore. Thanks anyways!

PR #360, opened 2016-08-16
Fixed plugin loader

ncave:
I only tested with plugins compiled with .NET Core, YMMV


alfonsogarciacaro:
Something weird has happened, I merged the PR but the commit is not appearing 😕 Probably is due to the organisation change. I've included your change directly in [this commit](https://github.com/fable-compiler/Fable/commit/f94b3345f82d0287cc3f65fc459b293a92dadd02), I hope that's OK with you.

I can confirm it's working, even with dlls compiled for Mono :tada: Thanks a lot for the incredible quick fix and all your help!

ncave:
@alfonsogarciacaro 
You're very welcome, thank you for maintaining such an important project for the community, I hope it drives further adoption for the language. And thank you for the invitation!

Yes, I did see the merge commit, it was there, then I joined the org and it disappeared... very strange.

PR #362, opened 2016-08-18
Fixed netcore xplat tests

ncave:
There were a few tests not passing on Linux.
All tests are passing now.


src/tests/AsyncTests.fs
```diff
@@ -200,18 +200,22 @@ open Fable.Core
...
+        async { res := true }
+        #if FABLE_COMPILER
+        |> Async.StartAsPromise
+        |> Async.AwaitPromise
+        #else
+        |> Async.StartAsTask
+        |> Async.AwaitTask
+        #endif
+        |> Async.StartImmediate
+        // some wait is needed on Linux for some reason
```
tpetricek: This is a bit odd - the basic idea of `StartImmediate` is that it starts the work and runs it synchronously until first "async" operation happens inside it - there is no "async" operation in `res := true`, so it would be good to understand what's going on here....

src/tests/tests.fsx
```diff
@@ -0,0 +1,37 @@
+#r "System.Threading.dll"
```
alfonsogarciacaro: If we're using this script to compile tests in netcore, we can remove src/tests/files.fsx, right?

src/tests/AsyncTests.fs
alfonsogarciacaro: The semantics of `Async.Start/StartImmediate` is translated very loosely in Fable (actually both methods have the same effect). As in JS you cannot define _true_ async methods to run in a separate thread (unless you use web workers), it's difficult to know if an async workflow will actually run in a separate thread. This only depends if native APIs (HTTP requests, timers, events...) are used.

The purpose of this test is just to check that converting async to promise back and forth doesn't throw any error.

> Note that `Async.RunSynchronously` is forbidden except for unit tests.

src/tests/tests.fsx
ncave: Sure, I just didn't want to remove it myself, in case you were incorporating it in a build somewhere. But yes, there should be only one, ideally, and it's a temporary measure until proper project file support appears.

src/tests/AsyncTests.fs
```diff
@@ -200,18 +200,22 @@ open Fable.Core
...
+        do! Async.Sleep 50
```
ncave: Technically I should have wrapped the Async.Sleep like that, cause it's only happening in clr, not js:
```
#if !FABLE_COMPILER
do! Async.Sleep 50
#endif
```
For some reason the issue is not happening under Windows.
There were some recent changes in Async.AwaitTask but I don't think that's it.

src/tests/AsyncTests.fs
ncave: Technically I should have wrapped the Async.Sleep like that, cause it's only happening in clr, not js:

#if !FABLE_COMPILER
do! Async.Sleep 50
#endif

For some reason the issue is not happening under Windows.
There were some recent changes in Async.AwaitTask but I don't think that's it.

PR #363, opened 2016-08-18
Ease of use change

voronoipotato:
I didn't have to run npm install explicitly as that gets run when I did `fable -d`.  Running fable a few times thinking I had done something wrong, and then having to read through the whole document to find out the correct way to build this was a pain point for something which is otherwise impressively painless. Fable.ReactHelper.fs is not currently in this directory, so I also removed references to it in the readme.

I tried to put an emphasis on getting the developer up and running before explaining why it's running.


alfonsogarciacaro:
Thanks!

PR #367, opened 2016-08-20
Support all event like (add, addDir, etc.) for watcher

MangelMaxime:
:warning:  **Carefull** tested only under linux.


alfonsogarciacaro:
Thanks!

PR #368, opened 2016-08-22
[WiP] Bring back Suave client

tpetricek:
I'm trying to build a prototype of something like a new version of Try F# using Fable, so I needed Fable as a Service (FaaS!) This is very much work-in-progress and I may be updating this over the next week a bit. But I just wanted to send a WiP request to mention what I'm doing :-)

It mostly just adapts the existing code in the repo to work again. The only thing that I added is that the result also includes a list of declarations with their pretty printed original F# type (I'm wondering if I can use this to keep "fsi-like session" - maybe, maybe not...).


src/fable/Fable.Compiler/FSharp2Fable.Util.fs
```diff
@@ -433,6 +434,14 @@ module Types =
         // The F# compiler "untuples" the args in methods
         | args -> List.concat args |> List.map (fun x -> makeType com Context.Empty x.Type)            
 
+    and makeFullType com (args: IList<IList<FSharpParameter>>) returnType =
```
alfonsogarciacaro: The name of this function may be a bit confusing as there is already `makeType` which doesn't produce partial types. I'd suggest something like `makeTypeFromParams` (there's also `makeTypeFromDef`).

src/fable/Fable.Core/AST/AST.Fable.fs
```diff
@@ -104,12 +104,13 @@ and MemberKind =
...
     | Field
 
-and Member(name, kind, argTypes, returnType, ?genParams, ?decorators,
+and Member(name, kind, argTypes, returnType, ?fullType, ?genParams, ?decorators,
            ?isPublic, ?isMutable, ?isStatic, ?hasRestParams, ?overloadIndex) =
     member x.Name: string = name
     member x.Kind: MemberKind = kind
     member x.ArgumentTypes: Type list = argTypes
     member x.ReturnType: Type = returnType
+    member x.FullType: Type = defaultArg fullType (Function(argTypes, returnType))
```
alfonsogarciacaro: As commented above, `FullType` seems to indicate there're "partial types" in Fable. To avoid that but still convey the fact that this is not the actual type as parameters are not curried in the compiled JS, I'd suggest something like: `OriginalType`, `FSharpType` or `CurriedType` ...maybe the best option is the last one with a comment explaining that arguments are not actually curried in compiled code.

alfonsogarciacaro:
Fantastic! An online Try F#/Fable is sorely needed, please tell me if you need any help 👍

src/fable/Fable.Client.Suave/Main.fs
```diff
@@ -188,14 +259,17 @@ open Suave.Successful
...
+     
     let setCORSHeaders =
         setHeader  "Access-Control-Allow-Origin" "*"
         >=> setHeader "Access-Control-Allow-Headers" "content-type"
 
     let handle (req : HttpRequest) = 
         let getString rawForm = System.Text.Encoding.UTF8.GetString(rawForm)
-        req.rawForm |> getString |> makeCompiler |> compile checker projOpts |> OK
+        let code = req.rawForm |> getString
+        let result = compile checker code projOpts (makeCompiler())
```
alfonsogarciacaro: Recently I was thinking that maybe the reason the previous Try Fable site failed was because the compiler was trying to satisfy several requests in parallel. Beforehand it was a more serious problem because I was writing to disk with every request. Now the Suave client uses the virtual file system but even with that we may need to protect `compile` calls against parallelisation (I'm not a Suave expert so I'm not sure if this is needed at all) with a `lock` or a `MailboxProcessor`.

Besides that, there are some static values we may have to reset, like the store for inline functions (I know, I know, that's a very bad functional programming practice :wink:): https://github.com/fable-compiler/Fable/blob/master/src/fable/Fable.Compiler/FSharp2Fable.fs#L836-L838

alfonsogarciacaro:
I have changed the code to [remove the dependency on static values](https://github.com/fable-compiler/Fable/commit/7ec1f6263c26c41cba26d83cbd24e44f088341f1), you may want to rebase your branch with master. Please tell me if you have any question :)

alfonsogarciacaro:
To make it easier to share the custom Babel plugins among different Fable clients, I've put them in a [separate file](https://github.com/fable-compiler/Fable/blob/master/src/fable/Fable.Client.Node/js/plugins.js) with UMD format so it can also be loaded with require.js. Please feel free to use it: I had to modify one of them to fix #377.

> Note that `removeUnneededNulls` must come after `transformMacroExpressions`.

alfonsogarciacaro:
We may also consider putting this directly in an Azure Function so we don't even need to set up the server... Yes, I just read Hanselman's post about serverless computing, is that obvious? :wink:

alfonsogarciacaro:
What's the status of this PR @tpetricek? If possible, I'd like to merge at least the changes to the core files (FSharp2Fable, AST.Fable) before releasing v0.6 so we can keep the AST as stable as possible during the life of this before-1.0 major version. Thanks!

alfonsogarciacaro:
Superseded by #395

PR #372, opened 2016-08-26
Added .fsproj support for .NET Core

ncave:
Using a subset of the Forge project thanks to its generous public domain license.

The .fsproj support is a bit brittle, you may have to adjust the references for some projects, but at least it's a start, until a better project system gets here.

### DIFF
--- a/src/fable/Fable.Client.Node/Main.fs
+++ b/src/fable/Fable.Client.Node/Main.fs
@@ -85,6 +85,68 @@ let loadPlugins (pluginPaths: string list) =
         | ex -> failwithf "Cannot load plugin %s: %s" path ex.Message)
     |> Seq.toList
 
+#if DOTNETCORE
+type private TypeInThisAssembly = class end
+
+let forgeGetProjectOptions projFile =
+    let projPath = Path.GetDirectoryName(projFile)
+    let projParsed = Forge.ProjectSystem.FsProject.load projFile
+    let sourceFiles =
+        projParsed.SourceFiles.AllFiles()
+        |> Seq.filter (fun fileName -> fileName.EndsWith(".fs") || fileName.EndsWith(".fsx"))
+        |> Seq.map (fun fileName -> Path.Combine(projPath, fileName))
+        |> Seq.toArray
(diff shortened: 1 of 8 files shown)


alfonsogarciacaro:
Fantastic work, thank you! 👏 Now we only need a netcore-only IDE to have a complete netcore dev environment ;)

forki:
so what is this doing?

alfonsogarciacaro:
It's not working with `Fable.Tests.fsproj`. How must the references be adjusted?

cloudRoutine:
@ncave if you're gonna take a whole bunch of stuff, it'd be nice if you threw a link to the repo in a comment at the top :stuck_out_tongue:

alfonsogarciacaro:
@ncave, have Forge files being edited? If not, maybe we could link directly directly to Forge repo using Paket Github dependencies (locked to a specific commit to avoid breaking our builds). Also, @cloudRoutine suggests there're some parts missing which may be causing the problems.

ncave:
@cloudRoutine By all means attribution can (and should) be added to the source files if they are going to stay, which is not certain yet, perhaps they can be references directly. Yes, the Forge.Core source files are exact copies with two small exceptions:
- Prelude.Core.fs is a trimmed-down version of the Prelude.js so we don't have to add too many #if s
- There is a minor implementation change in the ResizeArray.fs' map function to get around a missing method.

@alfonsogarciacaro I was debating (with myself, this way I always win :) what is the best option to reference the Forge project and opted for a copy as a quick test of viability, as we're only using a small part of it. It would be definitely better if we port all of it to .net core and reference it, or even as you suggest to directly reference those files through Paket http references.

Here are again the minor changes above that can be submitted to the Forge repo:
- break the Prelude.js into two files so it can be reused without having to port all of it.
- apply the ResizeArray.fs change, it should be backwards compatible.

ncave:
@alfonsogarciacaro Yes, the assembly resolution certainly can be improved, currently it only checks in .NET Core folder and the folder the Fable.Client.Node.dll is in.

alfonsogarciacaro:
Perfect, thanks a lot for the info :+1: This is a good base to add support for `.fsproj` in netcore!

Issue #231, opened 2016-07-06
Different representations of Option.None as a function argument


i-p:
### Description

`Option.None` value can be represented in JavaScript as `null` or `undefined`, depending on the way it is passed to a function. As a result, values that are equal in F# are not equal in JavaScript.

### Repro steps

Following two tests reproduce this issue:

```
type OptTest = OptTest of int option

[<Test>]
let ``Different ways of providing None to a union case should be equal``() =                      
    let value = None
    equal true ((OptTest None) = (value |> OptTest))    

[<Test>]
let ``Different ways of providing None to a function should be equal``() =        
    let f x = x
    let value = None
            
    equal true ((f None) = (value |> f))    
```

### Expected behavior

Both tests should pass in F# and in JavaScript.

### Actual behavior

Both tests fail in JavaScript.

### Known workarounds

Don't pass `Option.None` value to a function directly, use a variable instead.

### Related information 

* Windows 10 64bit
* master branch

alfonsogarciacaro:
This is my only fault ;) I was [removing null arguments](https://github.com/fsprojects/Fable/blob/master/src/fable-fsharp/Fable2Babel.fs#L51-L58) in function calls (mainly for aesthetic reasons to comply with our _JS you can be proud of_ motto) but this can cause conflicts as you've noticed so I've removed the removing (pun intended).

Thanks a lot for reporting and providing the tests!

alfonsogarciacaro:
Unfortunately my fix was causing some problems with other programs. Option types are used to represent optional arguments in TypeScript definitions and some methods failed when passed `null` to represent `None`, like `CanvasRenderingContext2D.fill: ?fillRule: string -> unit`. I had to revert the fix. Another solution is needed 😕

fdcastel:
Instead of trying to solve this problem with `null`s, what about creating a "real" `Option` type [like what was done with `Choice`](https://github.com/fsprojects/Fable/blob/master/src/fable/Fable.Core/npm/fable-core.ts#L14-L22)? Simple. And gets the job done.

One of the beauties of F# is that an `int option` is a completely different beast than an `int`. Not just "an `int?`".

If you are worried with the performance..., let's make it _right_. THEN, make it _faster_ ;)

alfonsogarciacaro:
[...]

PR #377, opened 2016-08-27
Problem with WebSocket constructor

bhandfast:
I was testing websockets, essentially porting the code to Fable from https://github.com/SuaveIO/suave/blob/master/examples/WebSocket/index.html

I got this error in Chrome 52.0.2743.116:
"Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received"

And the cause seems to be calling the websocket constructor as: new WebSocket(address, null);

If I manually changed it to new WebSocket(address) then it worked correctly.

This PR adds an overload to the constructor to prevent emitting null as the second parameter.

### DIFF
--- a/src/fable/Fable.Core/Import/Fable.Import.Browser.fs
+++ b/src/fable/Fable.Core/Import/Fable.Import.Browser.fs
@@ -9563,6 +9563,7 @@ module Browser =
         abstract CLOSING: float with get, set
         abstract CONNECTING: float with get, set
         abstract OPEN: float with get, set
+        [<Emit("new $0($1)")>] abstract Create: url: string -> WebSocket
         [<Emit("new $0($1...)")>] abstract Create: url: string * ?protocols: U2<string, ResizeArray<string>> -> WebSocket
 
     and [<AllowNullLiteral>] WheelEvent =


alfonsogarciacaro:
Nice catch, thanks a lot for spotting this! Setting optional arguments to null was also causing problems with other APIs (like Web Canvas) and was being dealt with [here](https://github.com/fable-compiler/Fable/blob/9edaf0730140445d40a80a50d61b42824ce0f2b4/src/fable/Fable.Compiler/Fable2Babel.fs#L60-L63) by removing null args at the end of method and constructor calls (see also [this issue](https://github.com/fable-compiler/Fable/issues/231)). However, as you have noticed, this solution wasn't enough for macros as they're transformed on the JS side of Fable by a custom Babel plugin.

Instead of patching the issue by adding an overload to the bindings, it may be better to solve the issue for all macros, so I've moved the null arg cleaning operation to a plugin happening after the macro resolution: https://github.com/fable-compiler/Fable/commit/f162062f42267b41f9ccdc7fe5e38e9701c466f6

I'll publish a new version with the fix shortly. Please give it a try and we can use your PR if it's still not working for you.

Thanks a lot for your help!

bhandfast:
Thanks, version 0.5.10 solves the issue for me without the overload.

PR #384, opened 2016-09-04
Adding rewrite plugins with documentation

tpetricek:
This adds `IRewritePlugin` which lets the author of a plugin perform various transformations on the Fable AST. I think this has various nice uses:

 - I added documentation with a sample that uses this to insert logging into every top-level declaration (so every function will start with `console.log("<funcName>").

 - I was able to use this for a prototype of a plugin that turns tail-recursive functions into while loops. You can see the plugin [here](https://github.com/tpetricek/fable-tailcalls/blob/master/tailcalls.fs). It turns the [sample code](https://github.com/tpetricek/fable-tailcalls/blob/master/test.fsx) into the [following JS](https://github.com/tpetricek/fable-tailcalls/blob/master/test.js). It's work-in-progress and breaks easily, but I think it will work nicely eventually!

Because of the tail-call plugin, I also had to add `continue`, `break` (not strictly needed) together with label declarations (Goto considered awesome!) and explicit return (which is needed to explicitly make Fable generate `return` from inside a loop).


alfonsogarciacaro:
Splendid job, thank you! :clap:

PR #386, opened 2016-09-05
Int64 / UInt64 support (kind of)

ncave:
- added int64 / uint64 (to the extent possible)
- added sprintf for netcore


alfonsogarciacaro:
Thanks!

Issue #385, opened 2016-09-05
Cannot find replacement for Microsoft.FSharp.Collections.ListModule.Except


damonmcminn:
### Description

`List.except` is apparently not supported by Fable. I could not find on the compatibility page where this is explicit. Am I missing something?

I'd love the opportunity to implement this and make a PR if this is needed.

### Repro steps

1. Use `List.except` in your code

2. Compile

### Expected behavior

Project compiles.

### Actual behavior

Project fails to compile with error `Cannot find replacement for Microsoft.FSharp.Collections.ListModule.Except`

### Known workarounds

Use `List.filter`

### Related information 

* Fable Version `0.5.5`
* .NET Runtime, CoreCLR or Mono Version `Mono JIT compiler version 4.4.0`
* Operating system `OS X 10.11.4`

alfonsogarciacaro:
Ups, seems I missed that function. We're striving to support everything in the collection modules so it should be added indeed. If you want to implement it yourself, that'd be fantastic! Replacements of .NET BCL/FSharp.Core are in the aptly called [Replacements module](https://github.com/fable-compiler/Fable/blob/master/src/fable/Fable.Compiler/Replacements.fs). The code is a bit messy so I'll try to explain the process to add a replacement:

1. Decide if the replacement will be inline (e.g. `"".Length` > `"".length`) or with a call to `fable-core` (e.g. [Seq.filter](https://github.com/fable-compiler/Fable/blob/master/src/fable/Fable.Core/npm/fable-core.ts#L1591-L1599)). For collections, usually `Array` functions are replaced with native JS functions for performance.
2. For collections, if the function is common to `Seq`, `List` & `Array`, unless there're performance reasons, I usually implement it only for `Seq` and then build a list or an array with the result if necessary.
3. Implement tests (e.g. tests for [List.filter](https://github.com/fable-compiler/Fable/blob/master/src/tests/ListTests.fs#L154-L158))

Probably in this case you can add the method to `fable-core`, and then it should work just by adding `except` to [this list](https://github.com/fable-compiler/Fable/blob/master/src/fable/Fable.Compiler/Replacements.fs#L921-L927). If you need more help to implement the replacement, just tell me. At any rate, implementing the tests would already be tremendously helpful :)

damonmcminn:
Champion! I'll get on this then make a PR. I'll yell if I need help :D

damonmcminn:
[...]

PR #389, opened 2016-09-08
Implement `Seq.except`

damonmcminn:
Addresses https://github.com/fable-compiler/Fable/issues/385

Tests pass.

But am I missing anything?

- Have I appropriately tested collection types?
- Does `===` satisfy `generic hash and equality comparisons to compare values`?

### DIFF
--- a/src/fable/Fable.Core/npm/fable-core.ts
+++ b/src/fable/Fable.Core/npm/fable-core.ts
@@ -1575,7 +1575,7 @@ export class Seq {
 
   static except<T>(itemsToExclude: Iterable<T>, source: Iterable<T>) {
     const exclusionItems = Array.from(itemsToExclude);
-    const testIsNotInExclusionItems = (element: T) => !exclusionItems.some(item => item === element);
+    const testIsNotInExclusionItems = (element: T) => !exclusionItems.some(excludedItem => Util.equals(excludedItem, element));
 
     return Seq.filter(testIsNotInExclusionItems, source);
   }
--- a/src/fable/Fable.Core/npm/fable-core.ts
+++ b/src/fable/Fable.Core/npm/fable-core.ts
@@ -82,7 +82,7 @@ export class Util {
     const fullName = Util.getTypeFullName(cons);
     const i = fullName.lastIndexOf('.');
     return i > -1 ? fullName.substr(0, i) : "";
-  } 
+  }
 
   static getTypeName(cons: any): string {
     const fullName = Util.getTypeFullName(cons);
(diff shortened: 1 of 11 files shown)


damonmcminn:
I have added further tests for non-primitive types to ensure the build fails and to answer my own question regarding the suitability of `===` for comparing values.

Any guidance beyond using `Util.equals` would be most helpful.

Thanks!

src/fable/Fable.Core/npm/fable-core.ts
```diff
@@ -1572,6 +1572,13 @@ export class Seq {
     return fst.value;
   }
 
+  static except<T>(itemsToExclude: Iterable<T>, source: Iterable<T>) {
+    const exclusionItems = Array.from(itemsToExclude);
+    const testIsNotInExclusionItems = (element: T) => !exclusionItems.some(item => item === element);
```
alfonsogarciacaro: Sorry, I should have told you about this. To get structural equality for unions and records, `Util.equals(item, element)` must be used instead of the `===` operator, which only compares by reference.

src/tests/SeqTests.fs
```diff
@@ -667,13 +667,25 @@ let ``Seq.truncate works``() =
...
-    |> equal 3
\ No newline at end of file
+    |> Seq.length
+    |> equal 3
+
+[<Test>]
+let ``Seq.except works``() =
+    Seq.except [2] [1; 3; 2] |> Seq.last |> equal 3
+    Seq.except [2] [2; 4; 6] |> Seq.head |> equal 4
+    List.except [1] [1; 1; 1; 1] |> List.isEmpty |> equal true
```
alfonsogarciacaro: Just for the sake of consistency, would you mind putting the `List.except` assertions in a different tests in the `ListTests.fs` file. It'd be great if you can also add a test for `ArrayTests.fs` (just copying the same test and replacing `List` with `Array` should be fine).

alfonsogarciacaro:
Great job, thanks a lot! I wrote a couple of comments, could you please check them? Also, if you want you can run `build NUnitTest` to be sure the tests also pass in .NET.

Cheers!

damonmcminn:
Comments addressed and thanks for the help. I have inadvertently introduced noise in `SeqTests.fs` from editor trailing whitespace pruning. Is this OK?

p.s. I forgot the most important thing from the get-go: thanks for Fable! It's great.

alfonsogarciacaro:
Thanks to you for the PR! Contributors are always welcome 😄 About the noise in `SeqTests.fs`, for the core files in the compiler it's probably better to try to keep a clean git history but I guess we don't need to be so strict for the unit tests :+1:

damonmcminn:
My pleasure :D

I won't let the noise creep in in the future. Chur!
👍

PR #392, opened 2016-09-08
Mithril

musheddev:
resyncing fable base


alfonsogarciacaro:
Nice work! I've read that Mithril is a really fast and easy-to-use DOM lib, so it'd be really interesting to have Fable bindings for it. Since Fable org was created, we're moving the important bindings to their own repos so maintaners can evolve them faster. I can invite you to Fable org, create a fable-mithril repo and make you admin, would you be OK with that? Of course, we can also add a Mithril sample to Fable's website :)

musheddev:
This PR was actually an accident, but the mithril bindings are almost ready to be released. I will welcome a fable-mithril repo to push too.

alfonsogarciacaro:
Done! I sent you an invitation to join Fable org, created a fable-mithril repo and gave you Admin access :+1:

PR #399, opened 2016-09-10
[WIP] Fix #398 ResizeArray.Find

### LINKED ISSUE #398: ResizeArray.Find // wrong code generation
PR with failing test case to follow.
~~~~
...
let f = System.Predicate<_> (fun x -> x = 1.)
equal 1. <| li.Find f
...
~~~~
generates
~~~~
_fableCore.Seq.find(f);
~~~~
for the call to `Find` - note the argument for the list itself is missing. A run-time exception occurs.
`build.sh NUnitTests` pass.

* Fable Version 0.5.11
* .NET Runtime 4.5
* Operating System win7

markek:


alfonsogarciacaro:
Nice catch! Could you please rebase your PR with master? I pushed a wrong commit by mistake. I removed it from master but it's remaining in your branch 😅 

This can be fixed by adding the a code similar to this, right before [this line](https://github.com/fable-compiler/Fable/blob/c6d81a8442a215a5e697e7fbee83bc11067fda89/src/fable/Fable.Compiler/Replacements.fs#L1064):

```fsharp
        | "find" when Option.isSome c ->
            // JS Native Array.prototype.find, returns `undefined` if not found
            icall "find" (c.Value, args) |> Some 
            // Alternative: fable-core/Seq.find, throws exception if not found
            // ccall "Seq" meth (args.Head::[c.Value]) |> Some
```

Seems `ResizeArray.Find` returns a default value (e.g. 0 for integers) if no element is found, but maybe it's better not to implement this behaviour as it can cause silent errors. Your call ;)

markek:
1. I don't understand how the guard `when Option.isSome c` helps, but when I leave it out, I get the same wrong code. I would expect the guard to reduce the cases where the fix would be applied, not increase it.

2. I'd like to provide the default instead of undefined. Something like
~~~~
  function asBool(x) {
    return x === undefined ? false : x;
  }
  function asNumber(x) {
    return x === undefined ? 0 : x;
  }
  function asRef(x) {
    return x === undefined ? null : x;
  }
  var c = exports.c = asRef(a.find(f));
~~~~
using asRef, asBool, or asNumber as appropriate, or
~~~~
  function define(def, x) {
    return x === undefined ? def: x;
  }
  var c = exports.c = define(null, a.find(f));
~~~~
using null/false/0 as appropriate. What do you think?

I'll be back Monday. Thanks.

alfonsogarciacaro:
`c` in `Option.isSome c` stands for callee, which basically means whether it's an instance method, to tell it apart from `Array.find`, `List.find`...

Because the default value of number/bool/others was already calculated for `Unchecked.defaultof` I merged the PR and refactored a bit the code to avoid duplications.

Thanks a lot for the fix! :+1:

PR #403, opened 2016-09-10
Pixi sample: fix compile error

MaxWilson:
? operator is no longer automatically supported, will cause error when compiling

    bunny <- PIXI.Sprite(unbox resources?bunny?texture)

unless JsInterop is included.


alfonsogarciacaro:
Thanks!

Issue #198, opened 2016-06-15
Option.iter evaluates input to many times


mastoj:
### Description

When calling `Option.get f x` the javascript that is being generated resolved `x` twice. 

### Repro steps

```
#r "node_modules/fable-core/Fable.Core.dll"
open Fable.Core

let mutable x = Some "Hello"
let tryGet = function
    | Some y ->
        x <- None
        Some y
    | None -> None

Option.iter (fun s -> printfn "%s" s) (tryGet x)
```

### Expected behavior

Output: `Hello`

### Actual behavior

Output: `null`

### Known workarounds

Change

```
Option.iter (fun s -> printfn "%s" s) (tryGet x)
```

To 

```
let something = tryGet x
Option.iter (fun s -> printfn "%s" s) something
```

### Related information 

* OSX
* Master
* Mono: 4.2.1

PR #407, opened 2016-09-11
Merge integer conversions from bitwise plugin into main

ncave:
Although the bitwise plugin is a nice example, these integer conversions really are basic functionality and belong in the main.

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -127,19 +127,39 @@ module Util =
             CoreLibCall ("Util", Some "toString", false, [arg])
             |> makeCall com i.range i.returnType
 
-    let toInt, toFloat =
-        let toNumber com (i: Fable.ApplyInfo) typ (arg: Fable.Expr) =
-            match arg.Type with
-            | Fable.String ->
-                GlobalCall ("Number", Some ("parse"+typ), false, [arg])
-                |> makeCall com i.range i.returnType
-            | _ ->
-                if typ = "Int"
-                then GlobalCall ("Math", Some "floor", false, [arg])
-                     |> makeCall com i.range i.returnType
-                else wrap i.returnType arg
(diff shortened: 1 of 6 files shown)


src/fable/Fable.Compiler/Replacements.fs
```diff
@@ -127,19 +127,39 @@ module Util =
...
+            | Fable.Number kind -> 
+                match kind with
+                | Int8 -> Some "($0 + 0x80 & 0xFF) - 0x80"
+                | UInt8 -> Some "$0 & 0xFF"
+                | Int16 -> Some "($0 + 0x8000 & 0xFFFF) - 0x8000"
+                | UInt16 -> Some "$0 & 0xFFFF"
+                | Int32 -> Some "($0 + 0x80000000 >>> 0) - 0x80000000"
+                | UInt32 -> Some "$0 >>> 0"
+                | Int64 -> Some "Math.trunc($0)" // only 53-bit (still better than nothing)
+                | UInt64 -> Some "($0 > 0) ? Math.trunc($0) : ($0 >>> 0)" // 53-bit positive, 32-bit negative
```
alfonsogarciacaro: Care is needed when emitting inline JS with multiple arg references (like `$0` in this case): if `$0` is a function call, a template like `($0 > 0) ? Math.trunc($0) : ($0 >>> 0)` will call the function three times (see #198). These cases should be replaced by a function in fable-core or by a [let-wrapping](https://github.com/fable-compiler/Fable/commit/bc101df191d637dbe0eae062b57aece0273dbcc7).

alfonsogarciacaro:
The bitwise plugin was separated from the core because it's probably not worth to wrap all number operations with bound checks, however I agree we can do it for explicit conversions. I just left a comment, could you please check it? It also be nice to have some tests.

ncave:
- replaced multiple args pattern with a function
- added tests
- tried to minimize conversions

alfonsogarciacaro:
Awesome, thank you!

PR #412, opened 2016-09-13
Implement ResizeArray.FindAll

markek:
Problem - the test that uses
~~~
Seq.length
... producing js:
    var xs = Array.from(_fableCore.List.ofArray([1, 2, 3, 4]));
    (0, _Util.equal)(3, _fableCore.Seq.count(function (arg00) {
      return _fableCore.Seq.filter(arg00, xs);
    }(function (x) {
      return x <= 3;
    })));
~~~
to count the result passes, but the test that uses
~~~
(fun l -> l.Count)
... producing js:
    var xs = Array.from(_fableCore.List.ofArray([1, 2, 3, 4]));
    (0, _Util.equal)(3, function (l) {
      return l.length;
    }(function (arg00) {
      return _fableCore.Seq.filter(arg00, xs);
    }(function (x) {
      return x <= 3;
    })));
~~~
fails with:
~~~
  1) ResizeArrays ResizeArray.FindAll using ResizeArray.Count works:
     AssertionError: 3 == undefined
 ~~~
The call to (ResizeArray) Count is compiled as js property .length, which seems to be undefined for the output js array. There are tests that show .Count works for ResizeArrays that are constructed by literals and by Add, so it must be that my implementation is not producing a well-formed output array. I am using an approach similar to ResizeArray.Find, but apparently I'm missing something.

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -1129,6 +1129,8 @@ module private AstPass =
         | "find" when Option.isSome c ->
             let defaultValue = defaultof i.calleeTypeArgs.Head 
             ccall "Seq" "tryFind" [args.Head;c.Value;defaultValue] |> Some
+        | "findAll" when Option.isSome c ->
+            ccall "Seq" "filter" [args.Head;c.Value] |> Some
         | "findLast" when Option.isSome c ->
             let defaultValue = defaultof i.calleeTypeArgs.Head 
             ccall "Seq" "tryFindBack" [args.Head;c.Value;defaultValue] |> Some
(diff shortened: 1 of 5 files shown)


src/fable/Fable.Compiler/Replacements.fs
```diff
@@ -1121,6 +1121,8 @@ module private AstPass =
         | "find" when Option.isSome c ->
             let defaultValue = defaultof i.calleeTypeArgs.Head 
             ccall "Seq" "tryFind" [args.Head;c.Value;defaultValue] |> Some
+        | "findAll" when Option.isSome c ->
+            ccall "Seq" "filter" [args.Head;c.Value] |> Some
```
alfonsogarciacaro: `FindAll` produces a new ResizeArray so you need make one from the result of `Seq.filter` (see [this](https://github.com/fable-compiler/Fable/blob/6f44e97c1268b2707e73aef22009f237c3a2ee1d/src/fable/Fable.Compiler/Replacements.fs#L1217-L1222)). It should be fixed by doing:

```fsharp
ccall "Seq" "filter" [args.Head;c.Value] |> toArray com i |> Some
```

markek:
Thanks

alfonsogarciacaro:
Superb, thank you!

Issue #239, opened 2016-07-07
Decimal literals don't compile


tastyeggs:
Sample script:

`let x = 1.0m`
`printfn "%A" x`

Produces the following output:
Cannot find replacement for Microsoft.FSharp.Core.LanguagePrimitives.IntrinsicFunctions.makeDecimal (L1,8-L1,12)

Guessing [this](https://msdn.microsoft.com/visualfsharpdocs/conceptual/intrinsicfunctions.makedecimal-function-%5bfsharp%5d) function needs to be implemented.

alfonsogarciacaro:
Hi there and thanks for the report! It's a bit more complicated than that, because decimals are a type of its own (with several numerical components) and all the arithmetic operations must be implemented in order to translate it to JS with the full semantics.

This is related to #205 were support for BigInteger was requested and it was discussed how to implement it (adding the type to `fable-core` or using an existing JS library). In any case, implementing decimals and big integers would be nice but requires some work and it has low priority at the moment. Of course, if you want to give it a try at implementing I'll give you as much support as I can.

Cheers!

alfonsogarciacaro:
I wouldn't do it in the standard library but if for some reason you're happy to use JS numbers for decimals, it could be easier to write a plugin to just do that.

tastyeggs:
Got it, I figured as much. I don't really need decimals right now, but I was working on sharing some domain models between my backend and frontend and hit upon this issue. Yes, plugin seems the way to go.

Thanks.

PR #413, opened 2016-09-13
Added Decimal support

ncave:


alfonsogarciacaro:
Thanks for this! We had a conversation about decimals in the past (See #239). The only problem I see with this PR is it won't produce _true_ decimals (that's more precise values than usual floats), but I guess it's useful to compile some code that uses decimals in F# (like FSharp.Data providers).

Maybe we could a compiler flag like `decimalAsFloat` or similar so the user has to enable this behaviour consciously. What do you think?

ncave:
Well, we don't have true 64-bit integers either (using floats instead). IMO it's better to have it out of the box (without flags) as it's one of the basic numeric types.

ncave:
Technically what we probably want (down the road) is to take a dependency on a full implementation like [decimal.js](https://github.com/MikeMcl/decimal.js), or even the whole [Bridge.NET](https://github.com/bridgedotnet/Bridge) for all the missing BCL types, instead of reinventing the wheel. I'll open a discussion issue for it.

alfonsogarciacaro:
You're right, I'll add it to the documentation :+1: Maybe in the future we can add a more strict decimal type :)

PR #414, opened 2016-09-13
Allow trailing carriage return in Fable console output

tpetricek:
On Windows, you get `\r\n` and so we need to allow `\r` at the end of `[SIGSUCCESS]`.


tpetricek:
Oops, this is already fixed from yesterday :)

alfonsogarciacaro:
Yeah, it was reported very soon after the release. I need some tests for the watch mode so it can be tested in AppVeyor. Thanks a lot for the fix in any case!

PR #415, opened 2016-09-13
Nicer homepage design

tpetricek:
I'm not a big fan of the blue twitter link - I think the blue on purple looks pretty bad! Here is an alternative version, also with a gitter link :-)

fable


alfonsogarciacaro:
He, he, you're right as usual. Thanks!

PR #417, opened 2016-09-14
Fix to support recursive records (do not StackOverflow)

tpetricek:
(Looking)


tpetricek:
Okay, this is now fixed.

I followed the example of members, which are also delayed - we need to delay the kind too, because evaluating it too early causes evaluation of the members (and their types) before the type is stored in the cache.

PR #419, opened 2016-09-15
[WIP] Remove need for json $type

davidtme:
At the moment to support Json you (realistically) need to use Newtonsoft.Json with ```TypeNameHandling=TypeNameHandling.All``` which adds ```$type``` everywhere.  There are a few problems with this:

1. Larger json
2. Need to keep client and server code in sync
3. 3rd parties can’t/won’t allow this option or don’t use json.net

I have a WIP branch which tries remove this dependence by added extra info into ```setInterfaces```.  At the moment it doesn’t provided all the support the normal ```ofJson``` does so I have created a ```ofJsonSimple```

### DIFF
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -537,9 +537,15 @@ module Types =
             // It's ok to use an empty context here, because we don't need to resolve generic params
             |> Seq.map (fun x -> x.Name, makeType com Context.Empty x.FieldType)
             |> Seq.toList
+
+        let makeUnionCases (tdef: FSharpEntity) =
+            tdef.UnionCases
+            |> Seq.map(fun x -> x.Name, x.UnionCaseFields |> Seq.map(fun f -> makeType com Context.Empty f.FieldType) |> Seq.toList)
+            |> Seq.toList
+
         let getKind () =
             if tdef.IsInterface then Fable.Interface
-            elif tdef.IsFSharpUnion then Fable.Union
+            elif tdef.IsFSharpUnion then makeUnionCases tdef |> Fable.Union
             elif tdef.IsFSharpRecord then makeFields tdef |> Fable.Record
             elif tdef.IsFSharpExceptionDeclaration then makeFields tdef |> Fable.Exception
             elif tdef.IsFSharpModule || tdef.IsNamespace then Fable.Module
--- a/src/fable/Fable.Compiler/FSharp2Fable.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.fs
@@ -804,7 +804,7 @@ let rec private transformEntityDecl
         // Unions, records and F# exceptions don't have a constructor
         let cons =
             match fableEnt.Kind with
-            | Fable.Union -> [makeUnionCons()]
+            | Fable.Union _ -> [makeUnionCons()]
             | Fable.Record fields
             | Fable.Exception fields -> [makeRecordCons fields]
             | _ -> []
--- a/src/fable/Fable.Compiler/Fable2Babel.fs
+++ b/src/fable/Fable.Compiler/Fable2Babel.fs
@@ -243,6 +243,19 @@ module Util =
         |> List.map (fun x -> Babel.StringLiteral x :> Babel.Expression |> U2.Case1 |> Some)
         |> Babel.ArrayExpression :> Babel.Expression
 
+    let buildFields com ctx (ent: Fable.Entity)  =
+        match ent.Kind with
+        | Fable.Record fields -> fields
+        | _ -> []
+
+        |> List.map(fun (n,t) -> 
+            [ Babel.StringLiteral n :> Babel.Expression |> U2.Case1 |> Some
+              Babel.StringLiteral t.FullName :> Babel.Expression |> U2.Case1 |> Some ]
+            |> Babel.ArrayExpression :> Babel.Expression
+            |> U2.Case1 |> Some
+        )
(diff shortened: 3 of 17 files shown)


alfonsogarciacaro:
This is a very interesting idea. If we're to add type fields info maybe we should go down this road and unify it with `ofJson`/`toJson`. I need to take a deeper look to this but it should be doable :)

tpetricek:
One subtle issue with this is that (unless I'm missing something) it won't work if you use `ofJsonSimple` in generic code. For example:

```fsharp
type Wrapper<'T> = { thing : 'T }
let parseAndUnwrap (json) : 'T = (ofJsonSimple<Wrapper<'T>> json).thing
```

When compiling `parserAndUnwrap`, we do not know the actual type - we just know it is a wrapper of some sort. 

So, I think `ofJsonSimple` should give warning if the type argument involves generics.

davidtme:
@alfonsogarciacaro while it's still a WIP I wanted to keep the different functions but I think the ultimate goal would be to replace  ```ofJson/toJson```.  I'll draw up a list of to-dos the current toJson does which needs to happen.

davidtme:
@tpetricek You’re right about generic code.  It works for simple types as JSON.parse handles that.  For complex types I have added the option for ```$type``` back in - For generics in json.net I think you would need that too (```TypeNameHandling.Auto``` maybe?). When I get around to doing the toJson function I’ll will double check this interaction with json.net

alfonsogarciacaro:
@tpetricek Yeah, I was thinking a bit about this when I was offline and reached the same conclusion. @davidtme I haven't checked, but I think Json.net can handle generics without TypeNameHandling by using reflection.

It may be a bit confusing for the users to have two JSON serialization methods, as there's already the native `JSON.parse/stringify` for simple cases. There may still be a way to avoid the need of `$type` by passing the type schema (with the generics resolved) directly to `ofJson`. I'll try to upload something on this direction a bit later :+1:

davidtme:
I’ve made a bit of progress on this. At the moment it only works for fields but I’ll move onto properties next, also it respects the ```$type``` Json.net adds in.
Also it now works with lists, maps, sets, dictionaries, options and tuples (all using the default json.net settings)

tpetricek:
Aside from generics, another tricky case might be class hierarchies, or cases where you cannot infer type from the context. For example I used this pattern recently - not in Fable - but it would be nice if this worked:

```fsharp
type IData = interface end

type Text = 
  { kind:string; text:string }
  interface IData

type Numbered = 
  { kind:string; number:int }
  interface IData

type Things = { name:string; data:IData }
```

Then you can read/parse JSON using things like:

```json
[ { name = "one"; data = { kind = "number"; number = 4 } },
  { name = "two"; data = { kind = "number"; number = 3 } },
  { name = "three"; data = { kind = "text"; text = "yo!" } } ]
```

Just trying to come up with cases that would be worth testing/covering, because there are a couple of places where type-driven things tend to be tricky in my experience.

alfonsogarciacaro:
I'm afraid if won't be possible to deserialize JSON with interface fields as in @tpetricek sample and have proper type instances (instead of plain JS objects) if there's no type info stored in the JSON 😕 

I've something working and I'll upload it shortly. The code's becoming a bit complicated and I just hope it won't happen as with #335, that couldn't be merged because it didn't cover all cases. I'm still not sure whether it's a good idea to offer two JSON serialize options: one full-proof and the other simpler but half-baked.

davidtme:
@alfonsogarciacaro I'm going to get rid of that simple json function once I've got all the current test green and I've copied all the $type stuff out.

alfonsogarciacaro:
Ok, I've finally pushed the alternative approach to simplify JSON serialization. You can see it here:

https://github.com/fable-compiler/Fable/compare/json-serialize

I still need to fix the old tests and maybe add a couple more but you can build the json-serialize branch and see how JsonTests.fs is compiled, but basically it does something like this:

```js
  it("Simple json - Unions", function () {
    var u = new U("CaseB", [_fableCore.List.ofArray([new Simple("Sarah", new Child("John", 14))])]);
    (0, _Util.equal)(true, function (y) {
      return u.Equals(y);
    }(_fableCore.Serialize.ofJson(_fableCore.Serialize.toJson(u), 2, [{
      $type: "Fable.Tests.Json.Child"
    }, {
      $type: "Fable.Tests.Json.Simple",
      Child: 0
    }, {
      $type: "Fable.Tests.Json.U",
      $cases: {
        CaseB: [3],
        CaseA: [-1]
      }
    }, {
      $type: "Microsoft.FSharp.Collections.FSharpList",
      $genArg0: 1
    }])));
  });
```

That is, it passes an optimized schema of the target type to `toJson`. The main advantages of this approach is that's not necessary to keep reflection info in a global value and that generic info can be resolved at compile time.

It's not possible to properly deserialize interfaces as in @tpetricek's sample above. However, I don't think another deserializer like Json.NET can do that without type info encoded in the JSON as it's necessary to know which constructor must be called. @davidtme Would this solution be OK for you?

davidtme:
@alfonsogarciacaro Look good to me.  I've add a few extra test including a deserialize with interfaces which uses a ```$type``` fallback.

Only 6 tests are failing for me, some of which I have added on my branch:
image

alfonsogarciacaro:
Great! We can still decide which implementation to take. I like your approach but I think we still need to resolve and send a schema at compile time, because if I'm not mistaken in this PR generics can only be known at the root level of the object.

alfonsogarciacaro:
Closing this as it's been superseded by #463. Thanks a lot @davidtme for starting the work that has brought a new reflection system to Fable and removed the need to include `$type` info in the JSON!

PR #420, opened 2016-09-16
Write errors and warnings to stderr as fsc and fsi do

markek:


alfonsogarciacaro:
Does this cause the process to fail if there's an error or warning? I like the idea but I'm a bit concerned we end up with a confusion with the output channel (process.stderr/stdout vs console.log) and probably it's better to unify everything. Do you have a use case where you need the errors and warnings to be written to process.stderr?

markek:
I don't have a requirement for this, just taking a slow approach to what might be improvements to error reporting. Shall I proceed to unify things to stdio in fable.js?

alfonsogarciacaro:
Sorry, I've been  busy and I didn't have the time to test this. I'm concerned that, when using Fable in a chid process, it will crash if writing to stderror, even for a mere warning.

markek:
I went ahead and changed console.log to stdoutLog and used stderrLog as needed.

I think the script interface to fable should be as much as possible just like that of fsc and fsi,

alfonsogarciacaro:
Thank you and sorry for the delay to merge this!

PR #421, opened 2016-09-16
Fixing the vuejs example 

hoonzis:
very minor fix, just adding the Fable.Core.JsInterop namespace needed for the VueHelper module

### DIFF
--- a/samples/browser/vue-todomvc/index.fsx
+++ b/samples/browser/vue-todomvc/index.fsx
@@ -7,6 +7,7 @@
 
 open Fable.Core
 open Fable.Import
+open Fable.Core.JsInterop
 
 // Use this dummy module to hold references to Vue and Router objects
 // exposed globally by loading the corresponding libraries with HTML script tags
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks! I haven't been developing anything else with Vue because Fable uses ES6 arrows functions to translate lambdas and this creates some problems with the special uses of `this` in Vue. But I still think is an interesting library, especially if you prefer to put the template in the HTML.

hoonzis:
I have used Knockoutjs quite a lot before and Vue is kinda similar: two-way binding with not much code, nice templating in html for simple projects it works. I agree that for bigger projects and specially in functional context React or something else with simple flow is better. Btw I have developed a small web app with Fable to show of my options pricing lib (https://github.com/hoonzis/Pricer). The web app is available here: http://www.payoffcharts.com/, it uses nvd3 for charting. I will try to write a short blog post to sum up what I have learned, but in general converting it so that it would be consumable by Fable was much less pain than I thought - so thnx for the great work.

alfonsogarciacaro:
Nice, thanks to you for trusting Fable! Please tell me when the post is ready so we can tweet it add a link in the website 😄

hoonzis:
ok, I will!

tpetricek:
@hoonzis Nice work :) 

@alfonsogarciacaro Looks like we'll need to add testimonials page to Fable web site pretty soon!

PR #423, opened 2016-09-19
Use reference hint paths for netcore

ncave:
Changes:
- now uses reference hint paths for netcore
- fixed small typo
- updated FCS reference to latest
- updated Forge references


src/fable/Fable.Compiler/Replacements.fs
```diff
@@ -169,11 +169,13 @@ module Util =
         | Fable.String ->
             GlobalCall ("Number", Some "parseInt", false, args)
             |> makeCall com i.range i.returnType
+            |> wrap i.returnType
         | Fable.Number kindFrom ->
             match i.returnType with
             | Fable.Number kindTo ->
                 let pattern = if needToCast kindFrom kindTo then patternFor kindTo else "$0"
                 emit i pattern [args.Head]
+                |> wrap i.returnType
```
alfonsogarciacaro: Why are these wrappings needed? The proper return type should be set by `makeCall` and `emit`. Did you notice a problem?

ncave: You're right, I took them out, I only added wrapping because it sounded right, there was no difference in the output either way.

PR #425, opened 2016-09-20
[WIP] Uniqueness records

alfonsogarciacaro:
This is me trying to look smart at ICFP :) I was already thinking of something similar beforehand but one of the talks has inspired me to implement [uniqueness types](https://en.wikipedia.org/wiki/Uniqueness_type) (sort of) in Fable. The idea is to compile updates of records with `UniquenessAttribute` as mutations. For example:

```fsharp
[<Uniqueness>]
type A = { a: int; b: int; c: int }

let transform a =
    printfn "%i" a.a
    let a2 = {a with b = -2; c = -3}
    // printfn "%i" a.a
    a2

{a=1; b=2; c=3} |> transform
```

Compiles to this:

```js
function transform(a) {
  _fableCore.String.fsFormat("%i")(function (x) { console.log(x); })(a.a);

  var a2 = (a.b = -2, a.c = -3, a);

  return a2;
}

transform(new A(1, 2, 3));
```

Which prevents allocation and helps performance (particularly in games with frequent updates of multiple objects) while keeping our code functional and elegant.

Of course, this means it's not safe to use the original value after update (uniqueness types can only be consumed once) and this will be checked by the Fable compiler. In the sample above, if we remove the comment from the second `printfn` we'll get a compilation error :+1:

> Note: at the moment, the PR is considering a value being consumed when being updated or passed as an argument to a function.

I'd love to hear some feedback before merging this. Maybe @tpetricek could give some advice?

### DIFF
--- a/RELEASE_NOTES_COMPILER.md
+++ b/RELEASE_NOTES_COMPILER.md
@@ -1,3 +1,6 @@
+### 0.7.0
+
+
 ### 0.6.12
 
 * Now fableconfig.json can be in JSON5 format (comments!)
--- a/RELEASE_NOTES_COMPILER.md
+++ b/RELEASE_NOTES_COMPILER.md
@@ -1,5 +1,8 @@
 ### 0.7.0
 
+* Add type info to JS constructors: cases (unions) and properties (records and classes)
+* Extend type references with generic info when calling `typeof`
+* Add `GenericParamAttribute` to implicitly pass type info of generic parameters
 
 ### 0.6.12
 
--- a/RELEASE_NOTES_CORE.md
+++ b/RELEASE_NOTES_CORE.md
@@ -1,3 +1,6 @@
+### 0.7.0
+
+
 ### 0.6.8
 
 * Fix #459: Seq.filter doesn't blow the stack with long sequences
--- a/RELEASE_NOTES_CORE.md
+++ b/RELEASE_NOTES_CORE.md
@@ -1,5 +1,6 @@
 ### 0.7.0
 
+* Serialize.ofJson doesn't need JSON to contain `$type` info any more
 
 ### 0.6.8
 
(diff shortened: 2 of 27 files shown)


davidtme:
Not sure it’s possible but could the uniqueness be controlled by variable use rather than an attribute.  Take you "Uniqueness records work" test, it would be great if fable spotted that x isn’t used after ```with``` statement and use the mutation.

alfonsogarciacaro:
You mean do this automatically for every record? This can be complicated as it will force Fable to do two passes to the F# AST. But in any case I'm not sure it's a good idea, as I was thinking to add a experimental feature that should be used with care by the programmer. Actually, in order for uniqueness types to be 100% types you need to capture the scope of the value (as Rust does) something that's not possible in F# at the moment.

tpetricek:
This looks like a cool idea :-). I do not have strong opinion on the current implementation.... 

I'm using mutable records quite heavily in [my new artisanal hand-crafted parser](https://github.com/the-gamma/thegamma-script/blob/master/src/thegamma/parser/parser.fs#L20), because I know that I never share the value (except in one place, where it is explicitly cloned). So I guess this would be the kind of example where using this feature would make sense. But than, thanks to the fact that I went for mutation, I do not have to worry about propagating the context back via function results, which makes the code a bit easier to write... 

Performance was definitely part of my reasoning when I went for mutation, so feature like this would be nice - but I'd worry that without compiler warnings when you break the "uniqueness laws", this might sometimes confuse people.

alfonsogarciacaro:
@tpetricek Yes, the compiler gives you an error if the value has already been "consumed". At the moment, a value is considered consumed when one of these two things happen:

- The value is passed as reference argument
- The record is updated

I know for sure this is not 100% safe and there'll be edge cases Fable won't cover, as to properly implement Uniqueness you need _Linear types_ (or so Academia guys told me), but I wanted to ask you if you can think of another common case that'll be worth checking.

The good thing is that, as this is controlled by an attribute, we can add the feature with the _Experimental_ tag so developers know they need to have a bit of care when using it.

PR #426, opened 2016-09-20
Code fixes for new Json Serialize

davidtme:
$type override
Added a check for back and forth between json.net when running NUnit tests


davidtme:
I've pointed my game at the update code and all seem to work well.  Generates quite a big schema to test against and doesn't seem to impact the performance.

```
function onMessageRecived(e) {
    (function (arg00) {
        serverMessageRecived.Trigger(arg00);
    })(_fableCore.Serialize.ofJson(e.data, 10, [{
        $type: "Microsoft.FSharp.Core.FSharpOption"
    }, {
        $type: "Models.Client.ActionItem"
    }, {
        $type: "Models.Client.GameMap",
        Points: 15
    }, {
        $type: "Models.Client.Player",
        Direction: 5,
        Status: 8
    }, {
        $type: "Models.Client.State",
        Map: 2,
        Players: 13,
        Weather: 9
    }, {
        $type: "Models.Common.Direction",
        $cases: {
            West: [],
            SouthWest: [],
            SouthEast: [],
            South: [],
            NorthWest: [],
            NorthEast: [],
            North: [],
            East: []
        }
    }, {
        $type: "Models.Common.Item",
        Type: 0,
        Values: 16
    }, {
        $type: "Models.Common.MapPoint",
        Items: 14
    }, {
        $type: "Models.Common.PlayerStatus",
        $cases: {
            Walking: [],
            Idle: []
        }
    }, {
        $type: "Models.Common.Weather",
        $cases: {
            Sunny: [],
            Raining: []
        }
    }, {
        $type: "Models.Messaging.Responce",
        $cases: {
            StateUpdate: [4],
            ActionList: [12]
        }
    }, {
        $type: "Tuple",
        $tupleLength: 2
    }, {
        $type: "[]",
        $genArg0: 1
    }, {
        $type: "[]",
        $genArg0: 3
    }, {
        $type: "[]",
        $genArg0: 6
    }, {
        $type: "[]",
        $genArg0: 7
    }, {
        $type: "[]",
        $genArg0: 11
    }]));

    return {};
}
```

davidtme:
@alfonsogarciacaro I wondered if there is anything I can do to help get any new json features into the main build?

src/fable/Fable.Compiler/FSharp2Fable.Util.fs
```diff
@@ -540,7 +540,8 @@ module Types =
         let makeProperties (tdef: FSharpEntity) =
             tdef.MembersFunctionsAndValues
             |> Seq.choose (fun x ->
-                if x.IsPropertyGetterMethod
+                // TODO: Remove anymore unneeded props?
+                if x.IsProperty
```
alfonsogarciacaro: Why this? If you have properties with setters you'll get duplicated properties.

src/fable/Fable.Compiler/Replacements.fs
```diff
@@ -400,17 +400,10 @@ module private AstPass =
...
-                        | Some t -> true, t::acc
-                        | None -> inflate, t::acc)
-                    |> function
-                    | true, ts -> Some(Fable.Tuple ts)
-                    | false, _ -> None
+                    Some(Fable.Tuple ts)
                 | Fable.DeclaredType(ent, [genArg])
                     when ent.FullName = "Microsoft.FSharp.Core.FSharpOption" ->
-                    needsInflate genArgsMap genArg
+                    Some(Fable.DeclaredType(ent, [genArg]))
```
alfonsogarciacaro: This and the removal of code for tuples above prevents optimization of the schema (for example, a string option or an int*int tuple don't need any special JSON deserialization treatment so they don't need to be included in the schema).

alfonsogarciacaro:
@davidtme Sorry, I didn't have time yet to check the PR. I've left a couple of comments but I need to check the rest of the code. In general, I don't understand well the purpose of the changes so having some explanation would be very helpful. Apparently some optimizations of the schema are being removed, is this because tests were failing?

Also, I'm not completely sure about the representation of options and tuples. If we're doing this change to make JSON lighter and not dependant of JSON.net it's probably better to keep the option and tuple representation as it is in runtime, as that's easier to serialize and helps interaction with other JS/TypeScript modules.

Please tell me what you think.

davidtme:
@alfonsogarciacaro I was hoping to be able to use json.net out of the box without any extra setup.  To get tuples working you would need a Tuple converter (such as this one: http://gorodinski.com/blog/2013/01/05/json-dot-net-type-converters-for-f-option-list-tuple/) which is an extra thing to set up.

With the current ofJson I needed 3 converters in the end:

- OptionConverter
- TupleArrayConverter
- UnionTypeConverter - this was to add $type as json.net omits this.

Writing the 3rd converter motivated me to rewrite the ofJson as I didn’t want to battle against the defaults every time I started a new project.

Also trying to sell this to my work colleague became harder as we needed another json.net configuration - 

1. An older json.net 7 with converters for fsharp types (it’s in our large backlog to fix this)
2. Json.net 8 with the f# defaults - This performs better for the ETL phase of our software
3. Fable *but* with alternative of json.net 8 defaults if the request comes from a knockout client

For me “other JS/TypeScript modules” isn’t a problem. We currently have a knockout website (I want to replace this with Fable) that talks to out F# web service.  As we just used json.net 8 out of the box for this part we saw the tuple json from the web service as {Item1: “”, Item2: “”} so while writing the typescript we just work with that.  Most of the time you can't use complex json (types, classes etc) with typescript without some sort of converter or 3rd party library (+1 for Fable).

I don’t think we will be able to break the dependency of json.net as if you want to use class hierarchies you will still need the $type

Shouldn’t the default ofJson work with the defaults (as close as possible) of json.net - if you customise json.net settings shouldn’t you have to customise ofJson?

src/fable/Fable.Compiler/FSharp2Fable.Util.fs
davidtme: I'm not sure about the duplicates but the property used for the test doesn't seem to work if you ```IsPropertyGetterMethod```

```
type PropertyJson() =
    member val Prop1 = {a="";b=0} with get,set
```

Also the schema (and test) with just ```IsProperty``` for the following code seems ok
```
type PropertyJson2() =
    let mutable prop1 = {a=""; b=0}
    member this.Prop1 with get () = prop1 and set value = prop1 <- value
```

schema:
```
    var result = _fableCore.Serialize.ofJson(json, 1, [{
      $type: "Fable.Tests.Json.Child"
    }, {
      $type: "Fable.Tests.Json.PropertyJson2",
      Prop1: 0
    }]);
```

src/fable/Fable.Compiler/Replacements.fs
davidtme: We need to go into the tuple to extract/output the length of the tuple as it's not just an array, also mark it as such so we can ```if (expected === "Tuple")``` later on.

As options in json.net are not just simple values, we need to output that it's an FSharpOption.  We could just check for ```Case == "Some"``` but I don’t think that is as safe

alfonsogarciacaro:
Hmm, I need to think about this. I understand that it'd be very convenient to support json.net out of the box, but it strikes me we're this change was supposed to make the JSON lighter and then it becomes bigger again (and the schema as well) by transforming options and tuples. Maybe we could ask the opinion of the community about this.

About the json.net converters, we could put the option and tuple ones together in a single converter and then distribute it through Nuget to make the setup easier. Would that work for you?

> I don’t think we will be able to break the dependency of json.net as if you want to use class hierarchies you will still need the $type

I should check again, but in the current json-serialize branch it should be possible to (de)serialize nested classes without the `$type` field. Isn't that working for you? Could you give an example?

alfonsogarciacaro:
Superseded by #463. Thanks again @davidtme for your work and for adding the tests!

PR #427, opened 2016-09-20
Add missing encodeURI and encoreURIComponent

MangelMaxime:


MangelMaxime:
Oh yes that's possible. I was thinking it related to the Browser and the Window. Because in JS we can do `window.encodeURI` for example.

But you are saying that there more missing for JS global. 

Is the global js imported to window or this is just two different ways to access the sample function in JS ?

alfonsogarciacaro:
[MDN lists those functions as global JS functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects), which means they should work in the browser and other JS environments (Node, etc). But we can add them also to the `Window` object for developer convenience if they're also available from there, but I cannot find `encondeURI` in the [documentation for Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) so maybe it only works in a few browser and it's not standard.

MangelMaxime:
OK no problem. 

I will add them to JS global 😊

MangelMaxime:
I just updated the PR.

I have added the Function Properties as you mentions them.


For `isFinite` and `isNaN` I took sample from the existing one in Number.
However, in the docs we could pass anything as string, number, etc. 
But inside the imports with limit it to numbers. I just wanted to mention it :) 

For me number is more interesting than anything because why would you want to test if a string is a NaN ^^

MangelMaxime:
This PR should used jsNative as proposed by #443

alfonsogarciacaro:
Thanks!

alfonsogarciacaro:
Sorry! There were some conflicts with PR #443 and at the end I had to revert this and commit it by hand. I hope you don't mind.

MangelMaxime:
No that's fine for me :).

The important is it's in Fable now :)

PR #428, opened 2016-09-21
Add test for List.unfold

tpetricek:
I think there is a bug - `List.unfold` returns sequence and not a list. Will send fix once CI finishes :)


tpetricek:
Looks like both the test & the fix work now, so this is good to merge, I think :)

PR #434, opened 2016-09-25
Add ownerType to ApplyInfo

ncave:
Expose owner type to plugin writers.


alfonsogarciacaro:
I think this should be useful. As it changes the API, it may be better to merge it when publishing 0.7.0 (to be released soon). Out of curiosity, what do you need it for?

ncave:
It's an additive, non-breaking change, so it's backwards-compatible.
It's useful for examining properties of the enclosing entity for example.

alfonsogarciacaro:
Thanks!

PR #443, opened 2016-09-27
Add jsNative

7sharp9:
This provided a more friendly way of specifying something is
implemented in native JS and as such has no implementation in F# code.
AKA fail with “JS only”


alfonsogarciacaro:
Thanks!

PR #445, opened 2016-09-29
Test and explain "index out of range" behavior

markek:


src/tests/ArrayTests.fs
```diff
@@ -786,4 +786,11 @@ let ``Array.except works``() =
...
-    Array.except [|{ Bar= "test" }|] [|{ Bar = "test" }|] |> Array.isEmpty |> equal true
\ No newline at end of file
+    Array.except [|{ Bar= "test" }|] [|{ Bar = "test" }|] |> Array.isEmpty |> equal true
+
+[<Fable.Core.Emit ("$0 === undefined")>]
+let isUndefined _ = failwith "JS Only"
+[<Test>]
+let ``Array.[i] is undefined when i is out of range`` () =
+    let xs = [|0|]
+    xs.[1] |> isUndefined |> equal true
```
alfonsogarciacaro: At the moment it's not included in the default build target, but there's another target to compile tests for .NET and run them with NUnit, so this will make it fail. I'd suggest something like this (also to avoid the need of `isUndefined`):

```fsharp
[<Test>]
let ``Array.[i] is undefined in JS when i is out of range`` () =
   let xs = [|0|]
   #if FABLE_COMPILER
   box x.[1] = null
   #else
   (try xs.[1] |> ignore; false with | _ -> true)
   #endif
   |> equal true
```

> This would apply also to the ResizeArray test below.

markek: Thanks. l will finish this tomorrow.

markek:
@alfonsogarciacaro - ready to merge.

alfonsogarciacaro:
Thanks!

PR #448, opened 2016-09-29
Include NUnitTest during CI on Travis and AppVeyor

markek:
Guards against unit tests that are not written for the NUnitTest build activity.


alfonsogarciacaro:
This was indeed needed, thank you!

Issue #439, opened 2016-09-27
Fable -w gives "Cannot read project options" "used by another process."


ericharding:
### Description

When using fable -w if I edit the source and save twice fairly quickly I the message

> "ERROR: Cannot read project options: The process cannot access the file 'C:\Users\Eric\Documents\hg\Inventory\client\fs\client.fsx' because it is being used by another process."

The fable compiler continues to run but my script remains out of date until I make another change.

### Repro steps

1. run "npm run build" which runs "fable -w"
2. (probably not important) run npm run start in another window
3. Edit client.fsx twice within a few seconds

### Expected behavior

Fable should eventually compile the last change.

### Actual behavior

Fable compiles the first change and prints an error:

> ERROR: Cannot read project options: The process cannot access the file '.../client.fsx' because it is being used by another process.

The last change is not reflected int he javascript but fable keeps working and when the file is changed again it generally works.

### Known workarounds

I looked through the code and it seems like the file is still locked by the first build when it gets to Fable.Client.Node\Main.fs at getProjectOpts.  checker.GetProjectOptionsFromScript throws and it prints an error.  

I added a function to wait for the file to unlock and that *does* appear to work around the problem.
```
let terribleWaitForUnlock path =
  let rec loop count = 
    if count > 0 then
      try
        use f = File.OpenRead(path)
        f.ReadByte() |> ignore
      with
      | _ -> 
        System.Threading.Thread.Sleep(600)
        loop (count-1)
  loop 50
```
I'm sure there's a more elegant solution.

### Related information 

* Fable Version: 0.6.7 and also what I built from source control (at change: 27def8199a1c6e4626453030e69d8b7a2b7a57fe)
Fable is installed locally in node_modules.
* .NET Runtime
* Windows 10
* This is basically the 'elmish counter' sample from https://github.com/fable-compiler/fable-elmish/tree/master/samples/react/counter.  

### Without the "fix"

change: C:\Users\Eric\Documents\hg\Inventory\client\fs\client.fsx at 8:15:34 PM
ERROR: Cannot read project options: The process cannot access the file 'C:\Users\Eric\Documents\hg\Inventory\client\fs\client.fsx' because it is being used by another process.
change: C:\Users\Eric\Documents\hg\Inventory\client\fs\client.fsx at 8:15:35 PM
Compiled client.fsx at 8:15:36 PM
change: C:\Users\Eric\Documents\hg\Inventory\client\fs\client.fsx at 8:15:39 PM
[...]

PR #449, opened 2016-09-30
Bug #439 - "Cannot read project options" error caused by editor locking

ericharding:
"Cannot read project options" error caused by editor locking files.

fsautocomplete and possibly other editor services also watch for file
changes and briefly lock the file when it is changed.  This causes a
race condition where Fable can fail to read and thus fail to recompile.


alfonsogarciacaro:
Yeah, that was a weird error in AppVeyor, I was closed and reopened the PR to force a new build. Everything worked after that, thanks a lot!

Issue #440, opened 2016-09-27
Seq.groupBy different values order for each key


xiety:
### Description

Seq.groupBy different values order for each key

### Repro steps

`Seq.groupBy fst [(1, "A"); (1, "B"); (1, "C")]`

### Expected behavior

`[[1,[[1,"A"],[1,"B"],[1,"C"]]]]`

### Actual behavior

`[[1,[[1,"C"],[1,"B"],[1,"A"]]]]`

### Known workarounds

`Seq.map (fun a -> (fst a, snd a |> Seq.rev))`

### Related information 

* Fable Version 0.6.7

alfonsogarciacaro:
Thanks! Again, the unit test wasn't strong enough to catch this 😕 It'll be fixed in fable-core 0.6.6.

alfonsogarciacaro:
`fable-core@0.6.6` has already been published :+1: Please update the dependency to get the fix.

PR #453, opened 2016-10-01
List.groupBy should return valid list

tpetricek:
It seems `groupBy` now returns an array, not a list - and so we cannot pattern match on it.


tpetricek:
Okay, so the issue is that `Seq.toList(Seq.groupBy ...)` turns the outer sequence to list, but not the inner sequences containing individual groups.

Before fix for #440, the groups were lists and the fix changed them to return arrays, which broke my tests :-). The fix:

 1. adds `List.groupBy` which does proper conversion (and add replacement record for it)
 2. adds test for `List.groupBy` checking the group is a list
 3. adds test for `Array.groupBy` checking the group is an array

The current implementation of `Seq.groupBy` happens to return array, so (3) should pass. If the `Seq.groupBy` implementation eventually changes, the test should catch it. We do not need explicit `Array.groupBy` at this point.

Note - are there any other functions returning `seq<seq<..>>` that could suffer from the same issue?

tpetricek:
✔️ woohoo

alfonsogarciacaro:
Ups, that was a regression bug :/ Thanks a lot for fixing it and sorry for breaking your tests. I'll check about other potential conflicts :+1:

tpetricek:
No problem, it was easy to find & fix :-) I'm glad my tests helped to discover this!

PR #456, opened 2016-10-04
Add CNAME file to gh-pages generator

tpetricek:
I think this (and re-publishing docs) should do the trick :)

### DIFF
--- a/docs/docs.fsx
+++ b/docs/docs.fsx
@@ -116,10 +116,9 @@ type Page =
     Tagline : string
     Content : string }
 
-
 // Copy static files from the 'source' folder to 'output' folder (add more extensions!)
 let copyFiles force =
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
So you were expecting me to be too lazy to do this myself... If that's the case you were right ;) Thanks!

alfonsogarciacaro:
It's working! :tada: http://fable.io/

tpetricek:
whome

I forgot one thing though - #457 - sorry about that!

avsm:
I wonder how GitHub authenticates these -- what if some other project also commits a fable.io to their gh-pages?

tpetricek:
@avsm This has been puzzling me for some time - and surprisingly didn't see that documented anywhere. It seems that forking a repo with CNAME does not "steal" the domain, so I suppose it might be "first come first serve" (which is a bit funny method...)

PR #458, opened 2016-10-06
Try to reproduce union equals error

forki:


alfonsogarciacaro:
Thanks!

MaxWilson:
Is the underlying bug still active? I hit this same bug in 0.6.15.

alfonsogarciacaro:
Sorry, I don't actually remember what was the underlying bug of this, I think it was a talk on Gitter. It's possible that it's fixed in the 0.7.x-alpha.x version but not in 0.6.x. What's your problem exactly, @MaxWilson?

PR #460, opened 2016-10-06
Fix the name of the package.

MangelMaxime:


alfonsogarciacaro:
Ups, this means it's been published with the wrong name to npm. I'll republish it. Thanks for that!

PR #461, opened 2016-10-08
allow running redux-webpack sample on an empty computer

davidpodhola:
If you do clone the repo and are not familiar with [webpack](https://webpack.github.io/), trying to run `samples/browser/redux-webpack` will be not easy.

Missing (?) packages added. Also updated according to the latest webpack requirements.


alfonsogarciacaro:
Thanks!

PR #463, opened 2016-10-10
Add type info to classes

alfonsogarciacaro:
The purpose of this PR is to pass types with generic info around, so there's no need to inline functions. This will it make it possible to build libraries which are referenced as `dll`s but still can call methods that need concrete types (like `Fable.Core.Serialize.ofJson`). This is achieved by two main additions:

- Types now include info about their cases (Unions) or properties (Records and Classes).
- When we use `typeof<A<string>>`, if the type has generics, Fable will call `fableCore.Util.makeGeneric` which extends the type with generic info:

```js
// fable-core.ts
export class Util {
  static makeGeneric(typeDef: FunctionConstructor, genArgs: any): any {
    return class extends typeDef { [FSymbol.generics]() { return genArgs; } };
  }
}

// In your code, `typeof<A<string>>` becomes:
fableCore.Util.makeGeneric(A, { T: "string" })
```

This makes it possible to pass type info **explicitly** through several functions without having to inline them to resolve generics as before.

> **Attention**: Fable cannot emit generic info when using `x.GetType()`! 

For convenience, `GenericParamAttribute` has also been added. When it decorates an optional argument of type `System.Type`, if omitted by the caller, Fable will fill it with a type reference of the generic parameter of the specified name. For example:

```fsharp
open Fable.Core
type GenericParamTest =
    static member Foo<'T>([<GenericParam("T")>] ?t: Type) = t.Value

GenericParamTest.Foo<TestType>()
// The last line translates to: GenericParamTest.Foo(TestType)
```

> Note: There're three possible representations of types at runtime:
>
> - As a **string**, for JS primitives: "number", "boolean", "string" (plus "function").
> - As a **([TypeKind](https://github.com/fable-compiler/Fable/blob/7abaf46b5c18d08b0ea15be5ec58e7b01559a9e1/src/fable/Fable.Core/AST/AST.Fable.fs#L43-L53) * obj) tuple** (JS array), for other non-declared types: options, tuples, interfaces, generic parameters...
> - As a **constructor reference**, for declared types.

### DIFF
--- a/RELEASE_NOTES_COMPILER.md
+++ b/RELEASE_NOTES_COMPILER.md
@@ -1,5 +1,8 @@
 ### 0.7.0
 
+* Add type info to JS constructors: cases (unions) and properties (records and classes)
+* Extend type references with generic info when calling `typeof`
+* Add `GenericParamAttribute` to implicitly pass type info of generic parameters
 
 ### 0.6.12
 
--- a/RELEASE_NOTES_CORE.md
+++ b/RELEASE_NOTES_CORE.md
@@ -1,5 +1,6 @@
 ### 0.7.0
 
+* Serialize.ofJson doesn't need JSON to contain `$type` info any more
 
 ### 0.6.8
 
--- a/src/fable/Fable.Client.Node/AssemblyInfo.fs
+++ b/src/fable/Fable.Client.Node/AssemblyInfo.fs
@@ -2,10 +2,10 @@
 namespace System
 open System.Reflection
 
-[<assembly: AssemblyVersionAttribute("0.6.12")>]
-[<assembly: AssemblyMetadataAttribute("fableCoreVersion","0.6.8")>]
+[<assembly: AssemblyVersionAttribute("0.7.0")>]
+[<assembly: AssemblyMetadataAttribute("fableCoreVersion","0.7.0")>]
 do ()
 
 module internal AssemblyVersionInformation =
-    let [<Literal>] Version = "0.6.12"
-    let [<Literal>] InformationalVersion = "0.6.12"
+    let [<Literal>] Version = "0.7.0"
+    let [<Literal>] InformationalVersion = "0.7.0"
--- a/src/fable/Fable.Client.Suave/Main.fs
+++ b/src/fable/Fable.Client.Suave/Main.fs
@@ -242,8 +242,8 @@ let rec formatFSharpTypeSig = function
     | FableType.Unit -> "unit"
     | FableType.Boolean -> "bool"
     | FableType.String -> "string"
-    | FableType.Regex -> "System.Text.RegularExpressions.Regex"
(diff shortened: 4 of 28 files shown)


michaelsg:
Is it still going to be possible to produce and consume JSON with $type in 0.7.0?

alfonsogarciacaro:
@michaelsg I've been thinking about this. I wanted to keep this feature but it relied on a global dictionary to relate type references and names I didn't like very much, so for the moment I removed it. It should still be possible to achieve a similar effect by adding the type tag by yourself and keeping a dictionary yourself of the types you want to serialize. This will suppose a bit of boilerplate but it should be more performant (I think going through all the types in your assembly as you've to do now is very costly), would that be an option for you? I can try to write some sample code if needed.

michaelsg:
I'm not concerned at all about the assembly scanning. I plan to add a `<FableTypes>` assembly attribute so I can weed out 99% of the assemblies and once cached, it won't be an issue.

It's not feasible to keep a list of types I want to exchange with the server. My design is based on the ability to recognize a type by name and using the information to enable message exchange through JSON serializataion. Easy on the server. A headache on the client. Only Fable made is possible to go forward with a design like this in the browser. 

I won't be able to continue my project without substantial redesign if you remove this single feature from Fable and 2 month's work gone. :-(

*Please* don't remove this unique feature of Fable. I think its an interesting opportunity for extending a system _onto_ the client/browser. I seem to be the only one who's noticed, so far, but who knows what this could enable. 

Is there a Fable organization? Do you need a donation? I'm serious.

michaelsg:
Are you trying to hurt me all for removing one global variable? ;-)

I've been considering ways that you might be able to support capturing such a type registry without supporting it directly in the compiler. The best idea so far involves the unfortunate topic of _decorators_. 
If an attribute could be added to a type that would emit a decorator call at compile time, the decorator could easily register the type name of the function/ctor being decorated. On the server, the same attribute could be used in reflection to identify types needed for Fable serialization. Sorta solves 2 problems.

Getting the $type into produced JSON and consuming it during deserialization isn't easy without support from Fable. I think coming up with a type registry solution without compiler support is easier than getting $type (recursively) into the produced JSON. Being able to consume the $type while deserializing sounds like another ugly problem that is auto-magical with Fable 0.6.9.

__Please__ consider continuing to support the ability to send and receive strongly-typed JSON. It's a unique feature of Fable that enables a kind of communication between browser and server that is normally only used between servers.

alfonsogarciacaro:
Haha, no worries! I was just concerned about having to keep a global dictionary with type names. But I agree this feature can be useful for some use cases. I've added old serialization back (with `ofJsonWithTypeInfo`/`toJsonWithTypeInfo`) :+1: https://github.com/fable-compiler/Fable/commit/375d26ef4bbfd72fae3435dc4a6103236efb7649

michaelsg:
You're a prince! Thank you!

PR #467, opened 2016-10-11
add missing location properties

### LINKED ISSUE #466: Location is missing password and username
According to the docs here: https://developer.mozilla.org/en-US/docs/Web/API/Location the location also has password and username. This is missing in fable.

mastoj:
Fixes #466 

Please double check that I didn't miss something.

### DIFF
--- a/src/fable/Fable.Client.Node/AssemblyInfo.fs
+++ b/src/fable/Fable.Client.Node/AssemblyInfo.fs
@@ -2,10 +2,10 @@
 namespace System
 open System.Reflection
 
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Cool, thank you! :clap: I'll publish a new fable-core version right away!

alfonsogarciacaro:
This change should be included in fable-core@0.6.9. Thanks again for the fix 👍

mastoj:
It was a no-brainer ;). Found it when I began looking into navigation for arch.

PR #471, opened 2016-10-14
Add String.Remove method

MangelMaxime:
@alfonsogarciacaro I am not sure of the arguments handling in `fable-core.ts`.

I did this way because if there is only one argument then the index is 0, and remove X characters.

I used `slice` method because it's more elegant than `substring` and performance are the sames.

### DIFF
--- a/src/fable/Fable.Core/npm/fable-core.ts
+++ b/src/fable/Fable.Core/npm/fable-core.ts
@@ -943,6 +943,18 @@ class FString {
     return FString.padLeft(str, len, ch, true);
   }
 
+  static remove(str: string, args1: number, args2?: number) {
+    var start = 0;
+    var len = 0;
+    if (args2 === undefined) {
+      len = args1;
+    } else {
+      start = args1;
+      len = args2;
+    }
+    return str.slice(start, -len);
+  }
+
   static replace(str: string, search: string, replace: string) {
     return str.replace(new RegExp(FRegExp.escape(search), "g"), replace);
   }
--- a/src/fable/Fable.Core/npm/fable-core.ts
+++ b/src/fable/Fable.Core/npm/fable-core.ts
@@ -944,23 +944,15 @@ class FString {
   }
 
   static remove(str: string, startIndex: number, count?: number) {
-    if (count === undefined) {
-      if (startIndex >= str.length) {
-        throw "startIndex must be less than length of string";
-      }
-    } else {
-      if (startIndex >= str.length || count > str.length) {
-        throw "Index and count must refer to a location within the string."
-      }
+    if (startIndex >= str.length) {
+      throw "startIndex must be less than length of string";
     }
-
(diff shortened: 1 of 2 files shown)


MangelMaxime:
I just saw the failure on Travis etc.
I will see it tomorrow.

src/fable/Fable.Core/npm/fable-core.ts
```diff
@@ -943,16 +943,22 @@ class FString {
...
+  static remove(str: string, startIndex: number, count?: number) {
+    if (count === undefined) {
+      if (startIndex >= str.length) {
+        throw "startIndex must be less than length of string";
+      }
+    } else {
+      if (startIndex >= str.length || count > str.length) {
+        throw "Index and count must refer to a location within the string."
+      }
+    }
```
alfonsogarciacaro: Maybe the warnings could cover more cases as follows:
```js
if (startIndex >= str.length) {
    throw "startIndex must be less than length of string";
}
if (typeof count === "number" && (startIndex + count) > str.lenght) {
    throw "Index and count must refer to a location within the string."
}

**Comment by MangelMaxime** [comment]
Ok I made a complete rewrite because I was actually not doing the right thing...

Now the tests pass and we have the same behavior has with .Net framework. I mimic the exceptions too but I don't know how to add them in the test.

@alfonsogarciacaro Can you please tell if the exception need to be added in the test? And if yes, guide me on this way :) ?

File: src/fable/Fable.Core/npm/fable-core.ts [hunk 2]
```diff
@@ -943,16 +943,22 @@ class FString {
...
+
+    if (count === undefined) {
```
alfonsogarciacaro: I usually use `count == null` because it covers both `null` and `undefined` (note it's using non-strict equality `==`) which are semantically the same for Fable. Here it would be even safer `typeof count === "number"`, but see below.

src/fable/Fable.Core/npm/fable-core.ts
```diff
@@ -943,16 +943,22 @@ class FString {
...
+      return str.slice(0, startIndex);
     } else {
-      start = args1;
-      len = args2;
+      return str.slice(0, startIndex) + str.substring(startIndex + count);
```
alfonsogarciacaro: We can put the two operations in the same line a là JS:
```js
return str.slice(0, startIndex) + (typeof count === "number" ? str.substr(startIndex + count) : "");
```

> I usually use JS `.substr` because it has the same semantics as .NET `.Substring` (the second argument is the length not an index). But here it doesn't really matters because there's only one argument.

MangelMaxime:
Thanks for the review. I made all the changes to match your comments.

And thanks for the one line return I didn't think about doing it this way :)

alfonsogarciacaro:
Superb, thanks again!

Issue #482, opened 2016-10-18
Bad identifier escaping


Pauan:
### Description

```
let ``foo$5EA`` = "bar"

let ``foo^A`` = "bar"

let ``fooת`` = "bar"
```

### Expected behavior

```
export var foo$24$5EA = "bar";
var foo_A = "bar";
export { foo_A as foo$5E$A };
var foo_ = "bar";
export { foo_ as foo$5EA$ };
```

### Actual behavior

```
export var foo$5EA = "bar";
var foo_A = "bar";
export { foo_A as foo$5EA };
var foo_ = "bar";
export { foo_ as foo$5EA };
```

### Related information 

[Fable uses `$` to encode invalid characters in identifiers.](https://github.com/fable-compiler/Fable/blob/0ad9b8fd5e4bbcc0da8554d27515ae55d08a9879/src/fable/Fable.Core/Util.fs#L54-L59)

That's fine, but there are two situations where this breaks:

1) Using `$` in an F# identifier (`foo$5EA` and `foo^A` get encoded to the same JS identifier).

2) Invalid character followed by a hexidecimal character (`foo^A` and `fooת` get encoded to the same JS identifier).

----

For the first case, I see two possible solutions:

1) When an F# identifier contains `$`, encode it as `$$`. This is similar to [double quoting](https://en.wikipedia.org/wiki/Delimiter#Padding_quoting_delimiters) in strings.

2) Encode `$` as `$24$`

----

For the second case, I see two possible solutions:

1) Add a `$` at both the start and end of the code point (e.g. `$5E$`)

2) Add in padding so that the hexadecimal is always the same number of characters (e.g. `$00005E`)

----

This is simple enough to fix that I'll probably do this myself and send you a pull request.

alfonsogarciacaro:
Related issue: #207

Pauan:
Also related: https://github.com/fable-compiler/Fable/commit/451deb954a64a5f5250d8032aeb0fb5182f6f59a#commitcomment-18059236

PR #484, opened 2016-10-19
Invalid identifiers are now escaped correctly

Pauan:
Fixes issue #482

### DIFF
--- a/src/fable/Fable.Core/Util.fs
+++ b/src/fable/Fable.Core/Util.fs
@@ -2,26 +2,26 @@ namespace Fable
 
 module Patterns =
     let (|Try|_|) (f: 'a -> 'b option) a = f a
-    
+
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Nice, thanks a lot for this!

About the temp variables: Fable doesn't worry about too much about the invalid characters in private local variables, it just replaces them with `_` (and then checks there's no other variable with the same name in scope). See [here](https://github.com/fable-compiler/Fable/blob/0ad9b8fd5e4bbcc0da8554d27515ae55d08a9879/src/fable/Fable.Core/Util.fs#L99-L106). So now that `$` is a forbidden char, it will just replace it by `_`.

About the removed whitespace: I sometimes use blame to blame myself :wink: and find the commit that introduced a change. But I guess those lines should be fine. Actually, I should have set my editor to remove trailing space too. I just did it :+1:

PR #486, opened 2016-10-19
Fix description of module command line parameter

inosik:
`commonjs` is the default now.


alfonsogarciacaro:
True, I missed that. Thank you!

Issue #483, opened 2016-10-19
Babel: "You must pass a scope and parentPath unless traversing a Program/File."


fdcastel:
### Description

The following program is failing to compile with the following error:

```
ERROR: C:/D/temp/fable-issue/Main.js: You must pass a scope and parentPath unless traversing a Program/File. Instead of that you tried to traverse a "Identifier" node without passing scope and parentPath.
Error: C:/D/temp/fable-issue/Main.js: You must pass a scope and parentPath unless traversing a Program/File. Instead of that you tried to traverse a "Identifier" node without passing scope and parentPath.
    at traverse (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\index.js:73:13)
    at File.buildCodeFrameError (C:\D\temp\fable-issue\node_modules\babel-core\lib\transformation\file\index.js:441:35)
    at Scope.checkBlockScopedCollisions (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\scope\index.js:397:27)
    at Scope.registerBinding (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\scope\index.js:579:16)
    at Scope.registerDeclaration (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\scope\index.js:483:14)
    at Object.BlockScoped (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\scope\index.js:185:28)
    at Object.newFn (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\visitors.js:318:17)
    at NodePath._call (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\path\context.js:76:18)
    at NodePath.call (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\path\context.js:44:14)
    at NodePath.visit (C:\D\temp\fable-issue\node_modules\babel-traverse\lib\path\context.js:105:12)
```

### Repro steps

Command line: `fable --projFile ./Main.fsx --verbose`

```fsharp
// Main.fsx
open LanguagePrimitives

type Instruction = 
| Jmp = 0x40uy // 0100 0000
| Jsr = 0x60uy // 0110 0000

type Register =
| R0 = 0x00uy // 0000 0000

let InstructionMask    = 0b11110000uy
let RegisterMask       = 0b00000111uy

type Operand =
| NoOp
| Reg of Register
| Addr of uint16

type InstructionRegister = {
    mutable Data: byte array
[...]

PR #491, opened 2016-10-20
Simpler test case for #483.

fdcastel:
Smaller test case for #483.

Like the original, it fails to compile when using the old version of `Fable2Babel.transformBranch`. But passes when using the new one.

### DIFF
--- a/src/tests/EnumTests.fs
+++ b/src/tests/EnumTests.fs
@@ -107,62 +107,20 @@ let ``EnumOfValue works``() =
     EnumOfValue 2 |> equal Fruits.Banana
     EnumOfValue 3 |> equal Fruits.Coconut
 
-open LanguagePrimitives
-
-type Instruction =
-| Jmp = 0x40uy // 0100 0000
-| Jsr = 0x60uy // 0110 0000
-
-type Register =
-| R0 = 0x00uy // 0000 0000
-
-let InstructionMask    = 0b11110000uy
-let RegisterMask       = 0b00000111uy
-
-type Operand =
-| NoOp
-| Reg of Register
-| Addr of uint16
-
-type InstructionRegister = {
-    mutable Data: byte array
-    mutable SourceOperand: Operand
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks for that, Fabio! Would it be possible to modify the test so it makes an actual assertion (like `equal a b`) instead of just checking it compiles?

alfonsogarciacaro:
Also, it may be a good thing if the test is complicated as the original because then we may catch a different error if it fails in a future refactoring ;)

fdcastel:
> Would it be possible to modify the test so it makes an actual assertion 

I tried. And I failed 😄 . Whenever I used constants to reproduce the path the problem didn't occur. (due to some FSharp compiler optimizations/short circuit evaluations, I believe).

This is a "meta" test. To use assertions here I believe we would need to invoke the compiler infrastructure (which would be overkill, i suppose).


> it may be a good thing if the test is complicated as the original 

I see your point but I don't know if it is a good path to follow. Tests should be deterministic. We should know what can go wrong and we put a guard against that in the code.

Imagine if we did the same to every other (simple) test from Fable. To make them "more complex" just for the sake of something "unknown" which "maybe" they can got. It would be a highway to hell :)

I know there is some cases when the tests were the simplest and failed to catch an error. But then (and only then) we get back to the test and improve it, this time knowing what may broke or not. 😉 

What do you think?

alfonsogarciacaro:
I completely agree with you. However, sometimes I'm concerned that tests contain very simple cases and there're more complicated patters that we're missing (like in your case that you were nesting pattern matching with the new switch cases). If in a later refactoring we find an unexpected error in the test code we can isolate it in another test.

Anyways, don't worry about it. I'll try to create an assertion myself :+1:

alfonsogarciacaro:
I've simplified the test in 02d676c0d8ed0e51126621b24ad2eb996e6480ed so I'm closing this, thank you! :+1:

PR #501, opened 2016-10-24
Added Enumerable support

ncave:


alfonsogarciacaro:
Cool! So you've managed to translate the `IEnumerable` implementation in F# to a JS iterator :+1: I was thinking to forbid this but it's true it may still be necessary in some cases. Thanks a lot!

BTW, can you have a look at [this PR](https://github.com/fable-compiler/Fable/pull/500/files#diff-f5e78fc9f99d5329ebcd13a8ace1cd5bL97)? Removing the ProjectCracker dependency will be very positive for the netcore version (it still needs work though, I need to decide if project references should be ignored and force the user to specify the compilation order, or try to infer it somehow). We could even unify the versions: compile Fable as a netstandard library and then let JS bootstrapper check if the `dotnet` runtime is installed and default to Mono to run the app otherwise.

ncave:
@alfonsogarciacaro AFAIK the netcore Fable was already using Forge instead of ProjectCracker, and FCS also did not have dependency on ProjectCracker, what was removed was the dependency on MSBuild.

I'm not sure if we can have a single binary before [.NET Standard 2.0](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/), it may or may not be possible.

ncave:
@alfonsogarciacaro Or, if Fable is only used as part of a tooling, you can fully switch to .NET Core for all platforms, at least it's a single framework.

PR #506, opened 2016-10-26
Fix Fable.JsonConverter failing on reading 'null' JSON token for erased option of union

valery-vitko:
Currently Fable.JsonConverter fails with "invalid token" for { "a": null }, when "a" is an erased option of a union. The fix propagates Null JSON token further as a null reference.


alfonsogarciacaro:
Nice, thanks a lot for this! Sorry, there were some conflicts and I had to merge this with the command line. I'll republish the JsonConverter on nuget soon :+1:

PR #507, opened 2016-10-26
Json conversion case insensitive union cases

valery-vitko:
In some places union case names appear in lowercase, e.g. cases of unions marked as StringEnum and initialized js-side. This patch makes first letter of union case names case insensitive in Fable.JsonConverter (it fails with an exception otherwise) and in Serialize.inflate/ofJson (mostly for uniformity with server-side). Not quite sure if it's the right way to go, but considering DU's case names must always be uppercase in F#, this should not introduce any ambiguities at least.

### DIFF
--- a/src/fable/Fable.Core/ts/Serialize.ts
+++ b/src/fable/Fable.Core/ts/Serialize.ts
@@ -74,6 +74,9 @@ export function inflate(val: any, typ: any): any {
             ? arr.map((x: any) => inflate(x, enclosing))
             : arr;
   }
+  function firstCharUpper(str: string) {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+  }
   function inflateMap(obj: any, keyEnclosing: List<any>, valEnclosing: List<any>): [any, any][] {
     const inflateKey = keyEnclosing.head !== "string";
     const inflateVal = needsInflate(valEnclosing);
@@ -135,7 +138,7 @@ export function inflate(val: any, typ: any): any {
     if (typ.prototype[FSymbol.cases]) {
       let u: any = { Fields: [] };
       if (typeof val === "string") {
-        u.Case = val;
+        u.Case = firstCharUpper(val);
       }
       else {
         const caseName = Object.getOwnPropertyNames(val)[0];
--- a/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
+++ b/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
@@ -31,6 +31,9 @@ type JsonConverter() =
                 read (index + 1) (acc @ [value])
         advance reader
         read 0 List.empty
+
+    let firstCharToUpper (str: string) = str.[0].ToString().ToUpper() + str.Substring(1)
+
     override x.CanConvert(t) =
         t.Name = "FSharpOption`1"
         || t.FullName.StartsWith("System.Tuple")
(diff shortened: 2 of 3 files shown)


alfonsogarciacaro:
Thanks for looking into this and sorry for the late reply. I think this makes sense but I have some doubts about `StringEnum`. They're supposed to be just a _hack_ to pass options to JS methods, and the user should be aware they can behave unexpectedly if they try to use them somewhere else. What worries me about introducing small fixes to support `StringEnum` in other situations is that at the end users will expect these types to work as any other union type, which will be a burden for maintenance. Do you have other use cases in mind besides `StringEnum` for this?

valery-vitko:
You know, when `StringEnum` put under perspective of just a hack for JS interop, it seems that it would be better to simply stick with `CompiledName` attribute as stated in docs. I just liked how F# `StringEnum` DU's match TypeScript string literal union types. Having a UI layer written in TypeScript with F#/Fable underneath, it just comes really handy. Thus sprinkling unions with some attributes shouldn't be a deal breaker.

By the way, just for the sake of completeness, proposed patch doesn't fix the problem completely. And even addresses it in a wrong way. I'm not really aware of any other use cases besides `StringEnum` where compiler does a name conversion, and `Fable.JsonConverter` knows nothing about it. Instead of ignoring case mismatch, we could've align `Fable.JsonConverter` behavior with compiler to convert `StringEnum` lowercase on writing, and expect lowercase on reading only for `StringEnum`.

alfonsogarciacaro:
Thank you for the explanation. Now I understand your use case much better. I need to think about this. I haven't tested, but a possible solution would be to have a fake interface that masks the string property as a `StringEnum` union:

```fsharp
[<StringEnum; RequireQualifiedAccess>]
type Options = A | B

type MyType = { num: int; option: string }

type IMyType =
    abstract member num: int
    abstract member option: Options

let x = ofJson<MyType> json |> unbox<IMyType>

printfn "%b" (x.option = Options.A)
```

valery-vitko:
@alfonsogarciacaro, thanks for the workaround, but case mismatch still stays there.

And it surfaces another problem with `StringEnum` :). Equality op translates to `x.option.Equals("a")`, and since x.option is a string, it fails. When I had stumbled on this I slapped an `.Equals(...)` to `string` prototype and pretended it's not a problem anymore.

As you've said, patching Fable here and there for `StringEnum` support is not desirable, maybe compiler should issue a warning for this and similar cases then?

alfonsogarciacaro:
@valery-vitko Hmm, I need to think how to tackle this. We could make the type of `StringEnum` translate to `string` in Fable AST, which would solve the problem with equality, but I'm afraid not other cases. We could add the warnings but it wouldn't be very elegant because again we have to add checks here and there and we may forget some. But I'll try to think about it, maybe there're some entry points where we can put the check and be safe for the majority of cases.

Again, my hope was the attribute would only be used for the specific case of passing options to a JS API, but I understand it's tempting to stretch it to other use cases :)

alfonsogarciacaro:
In the runtime, `StringEnum` are just plain strings, so I think it's better not to make Fable guess if every string is comming from a DU on the server side or not. However, Fable.JsonConverter has been updated so DUs decorated with `StringEnum` are converted into lowered-case strings in the JSON (see 23ff8089f6ba14a8fcfa237d3f6d2576c38e5fc7). I hope this makes it possible to you to use them in your interaction with the server.

valery-vitko:
I have gravitated towards normal DU's as you've recommended, but in a few cases where `StringEnum` is essential, it did make things better. Thanks!

PR #510, opened 2016-10-26
Add PassGenericsAttribute

alfonsogarciacaro:
This is basically an amendment to PR #463. To solve the problem of resolving generic arguments when inlining is not possible, PR #463 added `GenericParamAttribute` which would be automatically filled by Fable with a resolved generic argument. However this has raised justified concerns that we're deviating too much from standard F# and in some sense _forking_ the language.

To prevent that undesirable consequence, this PR replaces `GenericParamAttribute` with `PassGenericsAttribute`. When calling a function or method decorated with this new attribute, Fable will automatically add an extra argument with the generic arguments so you can do `typeof<'T>` within the method body.

This still forces the library author to explicitly add the attribute (so we avoid polluting with generic info the whole code base), but the code can be compiled normally with the F# compiler and you can access they type info with `typeof<'T>` as in standard F# (this [unit test](https://github.com/fable-compiler/Fable/compare/0.7.0...generics?expand=1#diff-8bc41db32192fad3fed112592e44f8b1L38) doesn't require the FABLE_COMPILER directive any more).

Fable will emit a warning when calling a method with `PassGenericsAttribute` if generics cannot be resolved.

Example:

```fsharp
let [<PassGenerics>] foo<'T>() = typeof<'T>
foo<TestType>()
```

becomes:

```js
function foo(x, _genArgs) {
    return _genArgs.T;
}
foo(5, { T: TestType });
```


et1975:
This looks better than `inline`ing!

PR #512, opened 2016-10-27
Optimizing equal and compare when their arguments are referencially equal

Pauan:
The GitHub diff is bad. Look at the diff in something like [meld](https://en.wikipedia.org/wiki/Meld_(software)) instead (with whitespace ignored).

### DIFF
--- a/src/fable/Fable.Core/ts/List.ts
+++ b/src/fable/Fable.Core/ts/List.ts
@@ -121,30 +121,42 @@ export default class List<T> implements IEquatable<List<T>>, IComparable<List<T>
   }
 
   Equals(x: List<T>) {
-    const iter1 = this[Symbol.iterator](), iter2 = x[Symbol.iterator]();
-    for (let i = 0; ; i++) {
-      let cur1 = iter1.next(), cur2 = iter2.next();
-      if (cur1.done)
-        return cur2.done ? true : false;
-      else if (cur2.done)
-        return false;
-      else if (!equals(cur1.value, cur2.value))
-        return false
+    // Optimization if they are referencially equal
+    if (this === x) {
+      return true;
+
+    } else {
+      const iter1 = this[Symbol.iterator](), iter2 = x[Symbol.iterator]();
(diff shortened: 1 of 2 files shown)


jims:
You can also add `?w=1` to the url to ignore whitespace :)

Pauan:
@jims That's super useful! That should be a checkbox on the "Files changed" page.

src/fable/Fable.Core/ts/List.ts
```diff
@@ -121,30 +121,42 @@ export default class List<T> implements IEquatable<List<T>>, IComparable<List<T>
...
-        return cur2.done ? true : false;
-      else if (cur2.done)
-        return false;
-      else if (!equals(cur1.value, cur2.value))
-        return false
+    // Optimization if they are referencially equal
+    if (this === x) {
+      return true;
+
+    } else {
```
jims: Isn't this `else` superflous?

tpetricek:
I was thinking the other day that having this optimization would be nice too! It seems to be breaking some tests though - can you look into that and figure out what's going on there?

alfonsogarciacaro:
This has actually surfaced a [bug in Date comparison](https://github.com/fable-compiler/Fable/commit/2d8342859bdef6742aca3a605d98493cf5798e6f#diff-91d3a972bb20dc75c94fd935b1f58fb6R154), thank you! :+1:

et1975:
Can this be back ported to 0.6.x?

Pauan:
@et1975 It's possible, but the structure of `fable-core` has changed between `0.6` and `0.7`, so it would require a non-trivial amount of work: I don't think it's as simple as using `git cherry-pick`. Is there any particular reason you can't use `0.7`?

et1975:
0.7.x entails a large amount of unknown, and I don't know when we'll be able to move, but fair enough. I thought it might be non-trivial (hey, look at that record equality difference between the two!) figured I'll ask before backporting it myself.

PR #514, opened 2016-10-27
[ON HOLD] Attempt to create a façade for interfaces

### LINKED ISSUE #502: Interface method conflicts with normal method
### Description

I ran into this problem when I was trying to create a `View` abstract type:

```
open System

type TransactionId = uint32


[<AbstractClass>]
type View<'A>() =
    abstract member Subscribe : (TransactionId -> unit) -> IDisposable

    interface IObservable<'A> with
        member this.Subscribe out =
            this.Subscribe (fun _ -> ())


type Map<'A>(f, parent: View<'A>) =
    inherit View<'A>()

    override this.Subscribe fn =
        parent.Subscribe fn
```

I trimmed down the code a lot, so that it only shows the broken parts.

As you can see, `View` has a `Subscribe` method, but it also implements `IObservable` which also has a `Subscribe` method.

In F# this is not a problem, because interface methods cannot conflict with normal methods: you must disambiguate by using `:>`
### Expected behavior

```
export class View {
    constructor() {}

    Subscribe(out) {
        return this.Subscribe_0(_arg1 => {});
    }
}
declare(View);


export class Map extends View {
    constructor(f, parent) {
        super();
        this.parent = parent;
    }

    Subscribe_0(fn) {
        return this.parent.Subscribe_0(fn);
    }
}
declare(Map);
```
### Actual behavior

```
export class View {
    constructor() {}

    Subscribe(out) {
        return this.Subscribe_0(_arg1 => {});
    }
}
declare(View);


export class Map extends View {
    constructor(f, parent) {
        super();
        this.parent = parent;
    }

    Subscribe(fn) {
        return this.parent.Subscribe_0(fn);
    }
}
declare(Map);
```

`Map.prototype.Subscribe` should be `Map.prototype.Subscribe_0`
### Related information
- Fable Version `0.7.5-alpha.15`

alfonsogarciacaro:
This will probably be another frustrated PR. I've been thinking how to fix #502 which is about name conflicts of implemented interface methods and methods of the type. At the moment, Fable does name mangling with overloaded class methods, but it's very complicated to modify the name of interface methods for a number of reasons.
- In TypeScript declarations, _fake_ interface overloads are allowed.
- Even if we added a `NoMangling` attribute for imported interfaces, interfaces cannot foresee which types are going to implement them so they know nothing about potential name conflicts.
- Even if we modify the name the names of the class methods to avoid the conflict, a type can implement two interfaces with the same method so these types would still be conflicting with each other.
- A type cannot know which interfaces an inhering type is going to implement.

> Also note overloading class methods is not easy either as there're cases, like statically resolved type parameters (lenses in Aether rely on this, for example), where the F# AST doesn't give much information about the method calls and it's difficult to resolve overloads.

The only solution would be to implement some kind of virtual table as other runtimes have. But this will mean all interface calls much be checked at runtime and it'll probably have a noticeable performance hit.

The proposed solution in this PR takes advantage on the fact that F# forces coertion when you want to use an interface method. The idea is that interface methods are in a separate object (which captures a reference to the implementing object) as a façade that gets activated when the object is coerced. If we need the full class back we recover the captured reference. Example:

``` fsharp
type IFoo =
    abstract Foo: unit->unit

type MyType() =
    member __.Foo() = printfn "Hello"
    interface IFoo with
        member this.Foo() =
            this.Foo()
            printfn "world!"

let test() =
    let x = MyType()
    x.Foo()
    let ifc = x :> IFoo
    ifc.Foo()
    match ifc with
    | :? MyType as y -> y.Foo()
    | _ -> ()
```

becomes:

``` js
export class MyType {
  constructor() {}

  Foo() {
    fsFormat("Hello")(x => {
      console.log(x);
    });
  }

  [_Symbol.interfaces]() {
    return {
      ["Test.IFoo"](_this) {
        return {
          [_Symbol.this]: _this,
          Foo() {
            _this.Foo();
            fsFormat("world!")(x => {
              console.log(x);
            });
          }
        };
      }
    };
  }
}
export function test() {
  const x = new MyType();
  x.Foo();
  // This is the same as x[_Symbol.interfaces]()["Test.IFoo"](x)
  const ifc = upcast(x, "Test.IFoo");
  ifc.Foo();

  if (ifc[_Symbol.this] instanceof MyType) {
    // Same as ifc[_Symbol.this]
    const y = downcast(ifc);
    y.Foo();
  }
}
```

The problem is that interfaces can also be used to interact with JS objects so it's necessary to check also whether the interface method is located directly in the object. Because Fable is designed to make interaction with JS easy, many times the compiler doesn't know where the types are coming from and you end needing a lot of casting and checks at runtime.

When I run the tests many of them were failing. I've managed to reduce the fails to less than 10, but I don't like that I have to deal with each case individually (as this means there may be other corner cases not covered by the tests) and that I'm introducing too much complexity for too little gain 😕 

I'll left this here so it can be reviewed by others. But for now, I'll just implement a compile time check that will throw an error if the user tries to implement an interface conflicting with a method of the class.

### DIFF
--- a/src/fable/Fable.Client.Node/Main.fs
+++ b/src/fable/Fable.Client.Node/Main.fs
@@ -59,8 +59,6 @@ let readOptions argv =
         declaration = def opts "declaration" false (un bool.Parse)
         symbols = def opts "symbols" [] (li id) |> List.append ["FABLE_COMPILER"] |> List.distinct
         plugins = def opts "plugins" [] (li id)
-        refs = Map(def opts "refs" [] (li (fun (x: string) ->
-            let xs = x.Split('=') in xs.[0], xs.[1])))
         extra = Map(def opts "extra" [] (li (fun (x: string) ->
             if x.Contains("=")
             then let xs = x.Split('=') in xs.[0], xs.[1]
--- a/src/fable/Fable.Client.Node/js/constants.js
+++ b/src/fable/Fable.Client.Node/js/constants.js
@@ -16,7 +16,7 @@ exports.FSHARP_PROJECT_EXTENSIONS = [".fsproj", ".fsx"];
 /** Set of options compatible with Fable's .NET process */
 exports.FABLE_BIN_OPTIONS = new Set([
     "projFile", "outDir", "coreLib", "symbols", "plugins",
-    "refs", "watch", "dll", "clamp", "extra", "declaration", "noTypedArrays"
+    "watch", "dll", "clamp", "extra", "declaration", "noTypedArrays"
 ]);
 
 /** Accepted JS modules an the alias for Rollup */
--- a/src/fable/Fable.Client.Node/js/fable.js
+++ b/src/fable/Fable.Client.Node/js/fable.js
@@ -20,7 +20,6 @@ var optionDefinitions = [
   { name: 'babelPlugins', multiple: true, description: "Additional Babel plugins (without `babel-plugin-` prefix). Must be installed in the project directory." },
   { name: 'loose', type: Boolean, description: "Enable “loose” transformations for babel-preset-es2015 plugins (true by default)." },
   { name: 'babelrc', type: Boolean, description: "Use a `.babelrc` file for Babel configuration (invalidates other Babel related options)." },
-  { name: 'refs', multiple: true, description: "Specify dll or project references in `Reference=js/import/path` format (e.g. `MyLib=../lib`)." },
   { name: 'dll', type: Boolean, description: "Generate a `dll` assembly." },
   { name: 'noTypedArrays', type: Boolean, description: "Don't compile numeric arrays as JS typed arrays." },
   { name: 'clamp', type: Boolean, description: "Compile unsigned byte arrays as Uint8ClampedArray." },
@@ -354,8 +353,6 @@ function resolvePath(optName, value, workingDir) {
             case "plugins":
             case "babelPlugins":
                 return resolveArray(value, resolve);
-            case "refs":
-                return resolveKeyValuePairs(value);
         }
     }
     return value;
@@ -559,14 +556,6 @@ function readOptions(opts) {
         opts.module = opts.rollup
             ? "iife"
             : (opts.ecma != "es2015" && opts.ecma != "es6" ? "commonjs" : "es2015");
-    }
-
-    // If refs is set in fableconfig.json, convert them to an array
-    if (typeof opts.refs == "object" && !Array.isArray(opts.refs)) {
-        var refs = [];
-        for (var k in opts.refs)
-            refs.push(k + "=" + opts.refs[k]);
-        opts.refs = refs;
     }
 
     // Check version
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -21,7 +21,7 @@ type ThisAvailability =
     // Object expressions must capture the `this` reference and
     // they can also be nested (see makeThisRef and the ObjectExpr pattern)
     | ThisCaptured
-        of currentThis: FSharpMemberOrFunctionOrValue
+        of currentThis: FSharpMemberOrFunctionOrValue option
         * capturedThis: (FSharpMemberOrFunctionOrValue option * Fable.Ident) list
 
 type MemberInfo = {
@@ -158,6 +158,16 @@ module Helpers =
             if isUnit args.[0].[0].Type then 0 else 1
         else args |> Seq.map (fun li -> li.Count) |> Seq.sum
 
+    let getMemberLoc (meth: FSharpMemberOrFunctionOrValue) =
+        if not meth.IsInstanceMember && not meth.IsImplicitConstructor
+        then Fable.StaticLoc
+        elif not meth.IsExplicitInterfaceImplementation
+        then Fable.InstanceLoc
+        else try
+                meth.ImplementedAbstractSignatures.[0].DeclaringType.TypeDefinition
+                |> sanitizeEntityFullName |> Fable.InterfaceLoc
+             with _ -> Fable.InstanceLoc // TODO: Throw error?
+
     let getMemberKind (meth: FSharpMemberOrFunctionOrValue) =
         let argCount = getArgCount meth
         if meth.EnclosingEntity.IsFSharpModule then
@@ -242,13 +252,15 @@ module Patterns =
             ||> List.forall2 (fun arg (ident, _) ->
                 if ident.IsMutable then false else
                 match arg with
-                | Coerce(_, Value arg) | Value arg -> ident = arg
+                // | Coerce(_, Value arg)
+                | Value arg -> ident = arg
                 | _ -> false)
         let checkArgs2 lambdaArgs methArgs =
             (lambdaArgs, methArgs)
             ||> List.forall2 (fun larg marg ->
                 match marg with
-                | Coerce(_, Value marg) | Value marg -> marg = larg
+                // | Coerce(_, Value marg)
+                | Value marg -> marg = larg
                 | _ -> false)
         let rec visit identAndRepls = function
             | Let((letArg, letValue), letBody) ->
@@ -556,15 +568,14 @@ module Types =
         with _ ->
             None
 
-    and makeMethodFrom com name kind argTypes returnType originalTyp overloadIndex
+    and makeMethodFrom com name kind loc argTypes returnType originalTyp overloadIndex
                        (meth: FSharpMemberOrFunctionOrValue) =
-        Fable.Member(name, kind, argTypes, returnType,
+        Fable.Member(name, kind, loc, argTypes, returnType,
             originalType = originalTyp,
             genParams = (meth.GenericParameters |> Seq.map (fun x -> x.Name) |> Seq.toList),
             decorators = (meth.Attributes |> Seq.choose (makeDecorator com) |> Seq.toList),
             isPublic = (not meth.Accessibility.IsPrivate && not meth.IsCompilerGenerated),
             isMutable = meth.IsMutable,
-            isStatic = not meth.IsInstanceMember,
             ?overloadIndex = overloadIndex,
             hasRestParams = hasRestParams meth)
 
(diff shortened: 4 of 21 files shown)


Pauan:
I feel like a `NoMangle` attribute is still the best option.

F# interfaces would use a separate sub-object in `Symbol.interfaces`, and TypeScript interfaces would use `NoMangle`, which does two things:

1. It causes the methods to be placed directly in the class, rather than in `Symbol.interfaces`

2. It causes the method names to be used as-is: no mangling

With this approach, F# interfaces will never conflict with other F# interfaces, and they will never conflict with TypeScript interfaces.

So you would only need to deal with conflicts between TypeScript interfaces. But TypeScript already has problems with interface conflicts, so Fable wouldn't be doing any worse than TypeScript itself. In fact, TypeScript developers may *expect* interface conflicts, because that's how TypeScript works.

alfonsogarciacaro:
It'll be a step forward but I'm not sure it's still a perfect solution. There are many corner cases I can think of at the moment (which means there're many others I _cannot_ think of 😉):

- Developers would have to remember to use the attribute 
- An interface could be used for both JS objects and F# objects.
- An interface can be used on the opposite direction, for example to represent F# types in JS, as `props` in React components.
- This affects also interfaces in the BCL (`IDisposable`, `IComparable`...), which we cannot modify. I tried to adapt types in fable-core to the method described above but it was very verbose and I foresee problems in the future if we want to get contributes to extend classes in fable-core.
- Not all the coertion cases in the AST are easy to interpret. When disposing a resource, for some reason F# first boxes the element and the coerces it to `IDisposable` so the original type info is lost at compile time. Or if you coerce a type to an interface, then box it and then try to type test it...

Pauan:
@alfonsogarciacaro 

> Developers would have to remember to use the attribute when writing a quick interface to represent an imported JS object.

Sure, but that's unavoidable. F# has certain behavior, and JS has different behavior. When the JS behavior differs from F#, you need some way of telling the compiler, "hey, use the JS behavior instead". The same is true with `Import`, `Emit`, `Erase`, `StringEnum`, `KeyValueList`, etc.

And the ts2fable project would of course insert `NoMangle` automatically, so that covers 90% of the use cases.

----

> An interface could be used for both JS objects and F# objects.

If it's an F# interface, then JS would have to use `Symbol.interfaces` to access it.

If it's a TypeScript/NoMangle interface, then JS can access it normally.

Yes this is a bit sucky, but if you want to guarantee good JS interop then you'll just have to use `NoMangle` and deal with conflicts. At least it's not any worse than using TypeScript, and it's not any worse than the current behavior in Fable.

Also keep in mind that in F#, classes and interfaces are rarely used except for interop, so it's pretty rare for an F# interface to be used in both JS and F#

----

> An interface can be used in the opposite direction: to represent F# types in JS (as props in React components).

I don't see the problem? You could just use a `NoMangle` interface for that.

----

> This affects also interfaces in the BCL (IDisposable, IComparable...), which we cannot modify. I tried to adapt types in fable-core to the method described above but it was very verbose and I foresee problems in the future if we want to get contributes to extend classes in fable-core.

Yes this can cause some pain when dealing with Fable internals (like `fable-core`). That's probably the biggest downside to it. But thankfully this only affects Fable internals, code written in Fable won't have any problems.

----

> Not all the coertion cases in the AST are easy to interpret. When disposing a resource, for some reason F# first boxes the element and the coerces it to IDisposable so the original type info is lost at compile time. Or if you coerce a type to an interface, then box it and then try to type test it...

Why does it need the original type information? If it knows that `IDisposable` is located at `foo[_Symbol.interfaces]()["System.IDisposable"].Dispose` then it can just call that method regardless of the type of `foo`, and regardless of whether `foo` is boxed or unboxed, right?

alfonsogarciacaro:
> Also keep in mind that in F#, classes and interfaces are rarely used except for interop, so it's pretty rare for an F# interface to be used in both JS and F#

That's what concerns me, classes and interfaces are not so common so maybe we're going through lot of trouble to avoid a rare case when you use classes and interfaces and there're name conflicts. I'd like to hear more opinions on this, but maybe it's easier for everybody if we just _forbid_ this rare case :wink:

> Why does it need the original type information? If it knows that IDisposable is located at foo[_Symbol.interfaces]()["System.IDisposable"].Dispose then it can just call that method regardless of the type of foo, and regardless of whether foo is boxed or unboxed, right?

I didn't explain the proposal of this PR clear enough. Actually, when you _cast_ an object to an interface you're creating a new object with the interface methods attached directly. This way you can send the object to any code expecting the interface and it will work normally. The downside is you need to cast the object back (basically accessing the `[_Symbol.this]` property) to access the class methods again.

Your proposal (keeping always the interface methods in a place where they won't conflict) makes sense, but then the compiler and all code interacting with these interfaces needs to know the distinction between hidden and not-hidden interfaces as you say.

We should the community's opinion about this 👍

Pauan:
I forgot to mention, there's also the option of using direct methods on the class, but with string keys:

```
export class MyType {
  constructor() {}

  Foo() {
    fsFormat("Hello")(x => {
      console.log(x);
    });
  }

  ["Test.IFoo.Foo"]() {
    this.Foo();
    fsFormat("world!")(x => {
      console.log(x);
    });
  }
}
export function test() {
  const x = new MyType();
  x.Foo();
  const ifc = x;
  ifc["Test.IFoo.Foo"](ifc);

  if (ifc instanceof MyType) {
    const y = ifc;
    y.Foo();
  }
}
```

This should have the same performance as the current Fable output, and the JS interop is good too, even when not using the `NoMangle` attribute.

If using `[<NoMangle>]`, it would remove the `Test.IFoo` prefix, so the method would just be called `Foo`

The only major downside of this approach is that it's no longer as easy to dynamically test whether an object implements an interface or not: you would need to check whether all of the methods exist or not.

In my opinion, this is a very nice, clean, and fast approach, with good JS interop. It's still technically possible to have conflicts, but it's very unlikely to happen by accident. This is my preferred approach.

ncave:
IMO what @Pauan suggests looks very reasonable. Something like this can help a lot in compiling existing F# code without needing to do additional attributing.

Pauan:
I just tested out my example code in TypeScript.

It works fine, but you lose out on static typing because it's using strings.

So instead, this is now my preferred solution:

```
export class MyType {
  constructor() {}

  Foo() {
    fsFormat("Hello")(x => {
      console.log(x);
    });
  }

  Test_IFoo_Foo() {
    this.Foo();
    fsFormat("world!")(x => {
      console.log(x);
    });
  }
}
export function test() {
  const x = new MyType();
  x.Foo();
  const ifc = x;
  ifc.Test_IFoo_Foo();

  if (ifc instanceof MyType) {
    const y = ifc;
    y.Foo();
  }
}
```

The method name includes the namespace and module name, but `.` is replaced with `_`

alfonsogarciacaro:
> Something like this can help a lot in compiling existing F# code without needing to do additional attributing.

@ncave Sorry, but so far adding features to Fable usually means increasing the maintenance burden on me (e.g. if I add a TS declaration parser, I become responsible to ensure it parses _any_ declaration). I'm a bit wary of promoting Fable can compile existing F# code as there may be many other hurdles (replacements, inline expressions in external projects, etc). For example, there've been several requests to compile FCS itself to JS with Fable, using Fable to create a "Try F#" web, etc, but I've seen little contributions on this regard.

I'm not completely against this, and I understand the reasons for your request. But I need to balance the benefits against the implementation/maintenance cost. This will also force republishing of all TS bindings (some of them private), break existing examples (e.g., when you use an interface to pass `props` to a React component) and probably code of some of most active Fable users. Because of this, I think it'll be better to ask the opinion of more Fable users should the change be added.

Pauan:
@alfonsogarciacaro I agree, it would be a good idea to get more opinions from the Fable community. Perhaps you should post another announcement in Gitter?

Because this thread (and other related threads) have gotten rather long, it may make sense to point to specific comments (like [this one](https://github.com/fable-compiler/Fable/pull/514#issuecomment-260556592) or [this one](https://github.com/fable-compiler/Fable/issues/502#issuecomment-260558894)) rather than the thread as a whole, so that people who are interested can find the relevant information faster.

Or perhaps we can add a new comment summarizing the issue, something like this:

----

In F#, if two different interfaces both use the method name `Foo`, there is no conflict. And classes can also use the method name `Foo` without conflicting with interfaces.

But in Fable, there is a conflict because all of the interfaces and classes compile to the same `Foo` method in JavaScript. That makes it impossible to use interfaces if there are method name conflicts, unlike in F#.

Assuming that there is an interface `IFoo` in the module `My.Module`, one solution is to use `My_Module_IFoo_Foo` for the interface method name, rather than just `Foo`

That solves the conflict problem. However, that is a breaking change, and so if anybody is using F# interfaces from within JavaScript/TypeScript, they will need to change the JavaScript/TypeScript code to use the full method name (including the namespace and module).

----

That should make it easier for people to decide whether they think it is worth it to add this feature (given the migration costs for existing JavaScript/TypeScript code).

Also, you should make it clear that this is basically a poll for the community, and that if people don't actively support this change, then the change won't be made.

Pauan:
@alfonsogarciacaro 
> Another tricky method is `Equals` which is inherited from `System.Object` and also part of `System.IEquatable`. This would mean that most of the types should implement both `Equals` and `System_IEquatable_Equals`.

Yes, [that's just how C# works.](https://msdn.microsoft.com/en-us/library/ms131187(v=vs.110).aspx#Remarks) You also need to overload the `op_Equality` operator (though perhaps that's handled automatically by F#)

On the other hand, I don't think `System.Object.Equals` is particularly important for Fable? So perhaps you could get away with only implementing `System_IEquatable_Equals`

ncave:
@alfonsogarciacaro I apologize if the above sounded like a request, it was more like a discussion. I absolutely agree, every major (or even smaller) feature like this does increase the maintenance burden and has to be balanced against that cost increase. Now that the project is entering a more stable phase, perhaps it can have a list of features that people can propose and vote on and contributors can grab and implement.

For the record, Fable is already an awesome tool and I'm always surprised how well it works (e.g. a sizeable portion of FCS already compiles with very little change). What is left is in the long tail of features, optional in their nature as they are not functional to begin with (as Scott Wlaschin always says, the best interface is a function). Obviously the ability to compile any large codebase that is using a big subset of BCL should not be a goal for Fable if we want it to stay lean.

In this particular case (and it is a relatively rare case where type members have the same names as interface methods), the already implemented `MangleAttribute` works for custom interfaces. A work-around for BCL interfaces that cannot be attributed can be to overwrite them with a custom definition that can be attributed. It may be a little cheesy, and it would be nice to have a more out-of-the-box way of dealing with it, but I agree it should go through the approval process first (and hopefully be implemented by a contributor).

alfonsogarciacaro:
@Pauan Thanks for the text! I opened a new issue #546, let's see what the community says. `System.Object.Equals` is very important for Fable in fact, as that's the usual way to implement custom equality for classes in .NET. In fact the compiler warns you if you implement `IEquatable` but don't override `Object.Equals` (actually, it's possible to have different implementations for both methods, but Fable is ignoring that atm).

@ncave Don't worry, I was probably too harsh too but it's just I don't seen the light at the end of the tunnel to release 0.7 ;) And I'm a bit reluctant to just delay this, because the purpose of 0.7 is to make it easier for devs to create libraries for Fable, and if (hopefully) Fable ecosystem grows it'd be much more difficult to make a change like this later.

Pauan:
@ncave 
> In this particular case (and it is a relatively rare case where type members have the same names as interface methods),

Actually, I ran into this issue because I had an interface method called `Subscribe` (from `IObservable`), and a class method also called `Subscribe`.

From what I understand, it's common in F# to have a class method which simply delegates to an interface method with the same name:

```
type IFoo =
    abstract member Foo : unit -> string

type Foo() =
    member this.Foo() =
        (this :> IFoo).Foo()

    interface IFoo with
        member this.Foo() =
            "foo"
```

This common idiom would essentially require `[<Mangle>]` on every interface (except the interfaces which are exposed to JavaScript/TypeScript).

----

@alfonsogarciacaro I understand the desire to implement this now rather than later, but I think version 0.8 is the right time to implement this.

If we delay it to 0.8, we can [put in deprecation warnings which will help ease the transition from 0.7 to 0.8](https://github.com/fable-compiler/Fable/issues/546#issuecomment-260694848)

If we try to push it in 0.7, we don't get deprecation warnings, making it harder for people to transition from 0.6 to 0.7

As a general principle, deprecation warnings are good for breaking changes.

(Also, 0.7 has already been delayed enough as-is, in my opinion)

ncave:
+1 for releasing 0.7 sooner rather than later.

PR #525, opened 2016-10-30
Fixed netcore tests

ncave:
The Regex not being replaced is still an issue, so those tests are disabled for now.

### DIFF
--- a/src/fable/Fable.Client.Node/js/fable.js
+++ b/src/fable/Fable.Client.Node/js/fable.js
@@ -386,8 +386,9 @@ function readFableConfigOptions(opts) {
         var projDir = fs && fs.statSync(fullProjFile).isDirectory()
                         ? fullProjFile
                         : path.dirname(fullProjFile);
-        cfgFile = fableLib.pathJoin(projDir, constants.FABLE_CONFIG_FILE);
-
+        if (fs && !fs.existsSync(cfgFile)) {
+            cfgFile = fableLib.pathJoin(projDir, constants.FABLE_CONFIG_FILE);
+        }
         // Delete projFile from opts if it isn't a true F# project
         if (!fableLib.isFSharpProject(fullProjFile)) {
             delete opts.projFile;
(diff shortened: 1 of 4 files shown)


src/tests/fableconfig.json
```diff
@@ -10,9 +10,8 @@
     "plugins": ["../plugins/nunit/Fable.Plugins.NUnit.dll"],
     "targets": {
         "netcore": {
-            "projFile": "project.fsx",
+            "projFile": ["Main/project.fsx"],
             "ecma": "es2015",
-            "module": "commonjs",
```
alfonsogarciacaro: Why is this deleted? The default module system now is `es2015`. It won't be possible to run the tests on Mocha if `commonjs` is not used.

src/fable/Fable.Client.Node/js/fable.js
```diff
@@ -386,8 +386,9 @@ function readFableConfigOptions(opts) {
         var projDir = fs && fs.statSync(fullProjFile).isDirectory()
                         ? fullProjFile
                         : path.dirname(fullProjFile);
-        cfgFile = fableLib.pathJoin(projDir, constants.FABLE_CONFIG_FILE);
-
+        if (fs && !fs.existsSync(cfgFile)) {
+            cfgFile = fableLib.pathJoin(projDir, constants.FABLE_CONFIG_FILE);
+        }
```
alfonsogarciacaro: Not sure this is a good idea. For example in [fable-powerpack](https://github.com/fable-compiler/fable-powerpack) there's a fableconfig file in the root dir and in the tests dir. If I call (from root) `fable tests` I expect the fableconfig file in `tests` directory to be read, not the one in the current (root) dir. That will fail with this change, can you please revert it?

alfonsogarciacaro:
Why is Regex not being replaced? Those tests worked before, right?

ncave:
@alfonsogarciacaro Ok updated with your suggestions.
Regex replacement works in master, but not 0.7.0, that's why it's disabled for now so
`build FableCompilerNetcore` can pass.

alfonsogarciacaro:
Thanks! Strange, I think I haven't touched the Regex tests during 0.7.0, I'll need to have a look.

ncave:
@alfonsogarciacaro I don't think it's the tests themselves, perhaps it has something to do with references, not sure, anyway, if you uncomment the tests in project.fsx you'll see the issue.

PR #526, opened 2016-10-30
Fix solution file and UseScriptResolutionRules=false

tpetricek:
This fixes the Visual Studio solution file and sets `UseScriptResolutionRules` to `false`.

With `UseScriptResolutionRules=true`, you cannot use one of my favorite tricks: https://github.com/the-gamma/thegamma-script/blob/master/tests/thegamma-tests/tokenizer.fs#L1. I have `fsproj` file with `fs` files that compile normally as part of the project, but have `#if INTERACTIVE` that also lets me run them interactively:

```
#if INTERACTIVE
#r "../../src/thegamma/bin/Debug/thegamma.dll"
#r "../../packages/NUnit/lib/net45/nunit.framework.dll"
#endif
```

Setting `UseScriptResolutionRules=true` also sets the `INTERACTIVE` symbol (see [source](https://github.com/fsharp/FSharp.Compiler.Service/blob/53d10d6e280301124dc2a3421ac270e3adf20ae2/src/fsharp/vs/IncrementalBuild.fs#L1785)). Aside from this, it also changes search paths order for MSBUILD (see [source](https://github.com/fsharp/FSharp.Compiler.Service/blob/9aed86819e50af08b455ae6902c0fe57ba30ae6a/src/fsharp/MSBuildReferenceResolver.fs#L283)) (according to comment, nobody knows why) and I think that's all - so I think we should be able to safely set this to `false`.

### DIFF
--- a/src/fable/Fable.Client.Node/Main.fs
+++ b/src/fable/Fable.Client.Node/Main.fs
@@ -188,7 +188,7 @@ let forgeGetProjectOptions (opts: CompilerOptions) projFile =
         OtherOptions = allFlags
         ReferencedProjects = [| |] // TODO: read from projParsed.ProjectReferences
         IsIncompleteTypeCheckEnvironment = false
-        UseScriptResolutionRules = true
+        UseScriptResolutionRules = false
         LoadTime = DateTime.Now
         UnresolvedReferences = None
     }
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
@ncave, could you please confirm if is OK to set use script resolution rules to false?

ncave:
I wouldn't question the mighty @tpetricek if I were you :)

But yes, [ProjectCracker.GetProjectOptionsFromProjectFile](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/FSharp.Compiler.Service.ProjectCracker/ProjectCracker.fs#L40) is already using `UseScriptResolutionRules=false`. Also, tests are passing with it.

ncave:
@alfonsogarciacaro About the solution change though, you may want to re-add the Fable.Tests.fsproj after you merge (cause you moved it one folder down).

PR #528, opened 2016-11-01
Add fetch to jsKeywords

davidtme:
The fable-powerpack fetch current complies to:

```
export function fetch(url, init) {
    return fetch(url, init);
}
```

Which creates a stack overflow


alfonsogarciacaro:
Ups, thanks for this!

alfonsogarciacaro:
I should write tests for the Fetch part too :/

alfonsogarciacaro:
I've published a new version of fable-powerpack with the fix. Thanks again for this!

PR #532, opened 2016-11-04
pacman sample : use arrow instead keyboard letter

cboudereau:


cboudereau:
thanks for the feedback. I have completed the change by updating docs.

alfonsogarciacaro:
Thanks!

PR #538, opened 2016-11-10
Fix tuple array in json converter

davidtme:
Currently serializing a tuple array doesn't work:

```F#
let json =
    [| "foo", 1. |]
    |> Json.serialize
```

It passes the `CanConvert` because ```System.Tuple`2[System.String,System.Double][]``` starts with `System.Tuple` but is really an array


alfonsogarciacaro:
Ups, thank you!

PR #542, opened 2016-11-13
0.7.0 Added missing netcore dependencies

ncave:
Fixed the netcore build so `build FableCompilerNetcore` is passing again.

### DIFF
--- a/src/tests/fableconfig.json
+++ b/src/tests/fableconfig.json
@@ -13,7 +13,7 @@
             "projFile": ["Main/project.fsx"],
             "ecma": "es2015",
             "module": "commonjs",
-            "coreLib": "../../build/fable-core/umd",
+            "coreLib": "../../build/fable-core",
             "symbols": ["DOTNETCORE"],
             "plugins": ["../plugins/nunit/bin/Release/netstandard1.6/nunit.dll"]
         }
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -11,6 +11,15 @@ open Fable
 open Fable.AST
 open Fable.AST.Fable.Util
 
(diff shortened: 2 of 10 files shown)


src/tests/fableconfig.json
```diff
@@ -13,7 +13,7 @@
             "projFile": ["Main/project.fsx"],
             "ecma": "es2015",
             "module": "commonjs",
-            "coreLib": "../../build/fable-core/umd",
+            "coreLib": "../../build/fable-core",
```
alfonsogarciacaro: Is this necessary to pass the tests? The default fable-core distribution now is in ES2015 module format, which is not understandable by node. That's why the tests use the UMD (universal module definition) distribution.

ncave: Yes, it's about location, when fable-core is compiled it goes to /build/fable-core/, not /build/fable-core/umd.

alfonsogarciacaro: Ah, ok! I didn't remember that for the `FableCompilerNetcore` target, fable-core TS files were compiled directly in commonjs format. Thanks for the explanation! :+1:

alfonsogarciacaro: Great! I added you as collaborator to the `fable-compiler-netcore` pkg so now you've publish rights :+1: Don't worry about the output of fable-core, I think both versions of fable-compiler can share the same fable-core dependency, so building fable-core in `FableCompilerNetcore` FAKE target is just to pass the tests.

There's already a `PublishCompilerNetcore` target that should work (once you add your npm credentials to your machine, check the npm site for that), it just runs `npm publish` from the build/fable dir (adding `--tag next` if needed). For fable-compiler I do the builds in AppVeyor and there I remove the *.pdb and *.xml files to make the package lighter (check `MakeArtifactLighter` target), you may want to do that as well.

Please tell me if you've any question, and thanks a lot for your help!

PR #549, opened 2016-11-17
Optionally output FSharp AST and/or Fable AST

ncave:
Optionally outputs FSharp AST and/or Fable AST to your outDir.
Add either or both of these to your fableconfig.json to enable:
```
    "extra": ["saveFSharpAst", "saveFableAst"],
```


alfonsogarciacaro:
Über cool, thank you!!

ncave:
@alfonsogarciacaro As extra seems to have changed to object, here are the updated options:

Add either or both of these to your fableconfig.json to enable:
`"extra": { "saveFSharpAst": "", "saveFableAst": "" },`

alfonsogarciacaro:
@ncave It's actually been a multiple argument of `key=value` pairs from the beginning, but if the value is missing it'll be assumed to be `true` ;)

PR #550, opened 2016-11-18
float to int - match .net

davidtme:
Make the float to int cast and convert match .net

`int(x)` - floors
`Convert.ToInt32(x)` - rounds

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -158,7 +158,8 @@ module Util =
         let patternFor = function
             | Int8 -> "($0 + 0x80 & 0xFF) - 0x80"
             | Int16 -> "($0 + 0x8000 & 0xFFFF) - 0x8000"
-            | Int32 -> "($0 + 0x80000000 >>> 0) - 0x80000000"
+            | Int32 when i.ownerFullName = "System.Convert" -> "Math.round($0)"
+            | Int32 -> "~~$0"
             | Int64 -> "Math.trunc($0)" // only 53-bit (still better than nothing)
             | UInt8 -> "$0 & 0xFF"
             | UInt16 -> "$0 & 0xFFFF"
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -158,7 +158,11 @@ module Util =
         let patternFor = function
             | Int8 -> "($0 + 0x80 & 0xFF) - 0x80"
             | Int16 -> "($0 + 0x8000 & 0xFFFF) - 0x8000"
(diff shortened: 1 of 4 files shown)


src/fable/Fable.Compiler/Replacements.fs
```diff
@@ -158,7 +158,12 @@ module Util =
         let patternFor = function
             | Int8 -> "($0 + 0x80 & 0xFF) - 0x80"
             | Int16 -> "($0 + 0x8000 & 0xFFFF) - 0x8000"
-            | Int32 -> "($0 + 0x80000000 >>> 0) - 0x80000000"
+            | Int32 when i.ownerFullName = "System.Convert" -> 
+                // TODO: When the fable round matches .net replace the following:
+                let x = (com : ICompiler).GetUniqueVar()
+                let r = com.GetUniqueVar()
+                "(("+x+") => { var "+r+" = Math.round("+x+"); return (((("+x+">0)?"+x+":(-"+x+"))%1)===0.5)?((0===("+r+"%2))?"+r+":("+r+"-1)):"+r+"; } )($0)"
```
alfonsogarciacaro: If the inline code is getting so complicated, maybe it's worth to add the function to fable-core and just call it. What do you think?

davidtme: My hope/guess is no one in the F# world will be using the `Convert.ToInt32`, this is just a fallback? I'm more than happy to make this a function if you want but I'll let you decided.

alfonsogarciacaro:
@ncave Would you mind having a look at this? You're the expert in numeric conversions in Fable :)

ncave:
Yes, technically the banker's rounding needs to be added to all Convert.ToIntXX functions in order to match the .NET implementation. The reason they were added without it was it was a cheap way (reusing the integer casts) to expand surface coverage with good enough conversion for now.

There are other areas in numeric conversion that are also in the "good enough for now" category, like proper 64-bit integers, proper decimals, etc. which are all easy to add, there are already JS implementations on github with compatible licenses that just need to be added and used.

@davidtme Creating a function for bankers rounding in fable-core would definitely help as it can probably then be composed with the integer casts for a quick win. Whether to do that in 0.7 or after, is a different question, if not we can keep it around as an issue, but it would be nice to have it before 1.0.

davidtme:
I've added a bankers rounding and hooked it up to `Convert.ToInt32`, `round` and `Math.Round`

davidtme:
@alfonsogarciacaro should I recreate this PR but to the master now 0.7 has shipped?

alfonsogarciacaro:
@davidtme Yeah, please. Sorry I couldn't have a look at this before 0.7 but the latest version looks good to merge :+1:

davidtme:
Superseded by #554

PR #555, opened 2016-11-22
Fix Array.indexed

davidtme:
`.entries` returns an iterator not an array and `Object.entries` isn't in node so I've gone for a map


alfonsogarciacaro:
Cool, thank you!

PR #559, opened 2016-11-24
Fix Array.rev

davidtme:
I also added a few other test to `truncate` and `sortDescending` just to be on the safe side

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -1228,8 +1228,10 @@ module private AstPass =
         | "reverse" when kind = Array ->
             match i.returnType with
             | Fable.Array _ ->
+                // Arrays need to be copied before sorted in place.
                 emit i "$0.slice().reverse()" i.args |> Some
             | _ ->
+                // ResizeArray should be sorted in place without copying.
                 icall "reverse" (instanceArgs c i.args) |> Some
         // Conversions
         | "toSeq" | "ofSeq" ->
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -1226,7 +1226,11 @@ module private AstPass =
         | "removeAt" ->
             icall "splice" (c.Value, [args.Head; makeConst 1]) |> Some
         | "reverse" when kind = Array ->
-            icall "reverse" (instanceArgs c i.args) |> Some
+            match i.returnType with
+            | Fable.Array _ ->
+                emit i "$0.slice().reverse()" i.args |> Some
+            | _ ->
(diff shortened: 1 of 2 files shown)


src/fable/Fable.Compiler/Replacements.fs
```diff
@@ -1226,7 +1226,11 @@ module private AstPass =
         | "removeAt" ->
             icall "splice" (c.Value, [args.Head; makeConst 1]) |> Some
         | "reverse" when kind = Array ->
-            icall "reverse" (instanceArgs c i.args) |> Some
+            match i.returnType with
+            | Fable.Array _ ->
+                emit i "$0.slice().reverse()" i.args |> Some
```
alfonsogarciacaro: We're already checking this is an array in the guard above (`when kind = Array`) so it's probably not necessary to check "returnType", we just use `emit i "$0.slice().reverse()" i.args |> Some` directly :+1:

alfonsogarciacaro:
Thanks for spotting this! Interesting, so `reverse` is mutating and returning a copy (by reference I assume)... javascript /shrug

src/fable/Fable.Compiler/Replacements.fs
davidtme: I tried that but stops the ResizeArray working:

```
BABEL ERROR: Failed to parse macro: $0.slice().reverse()
MACRO ARGUMENTS: trailingComments,leadingComments,innerComments
Property object of MemberExpression expected node to be of a type ["Expression"] but instead got undefined
[ERROR] C:/Dev/Fable/build/tests/Main/ResizeArrayTests.js: opts is not defined
```

alfonsogarciacaro: Ok, I guess if you use ` emit i "$0.slice().reverse()" (instanceArgs c i.args) |> Some` instead it should work. That's equivalent to `defaultArg i.callee i.args.Head`.

davidtme: That results in the `ResizeArray` getting `.slice().reverse();` which we don't want.

alfonsogarciacaro: Ah, ok! I finally got it, sorry 😅 So for `ResizeArray` this is mutating but not for for `Array.rev`. Maybe we could add a comment explaining this to avoid some head scratching to my future me? ;)

alfonsogarciacaro:
Thank you!

Pauan:
@alfonsogarciacaro [`Array.prototype.reverse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) reverses the input array in place and then returns the input array. Yes it's stupid, yes it's confusing. It really should return `undefined` to indicate that it mutates the array.

PR #561, opened 2016-11-24
Added UnboxFast intrinsic / List.truncate

ncave:
Just because they're cheap to add and intrinsic, and so it doesn't make sense to keep 'em in a plugin.


alfonsogarciacaro:
Cool, thanks! :clap:

PR #567, opened 2016-11-25
Speed up JsonConverter

davidtme:
CanConvert was impacting my program a lot (I've pulled out the code so I could test it):

image

There is still more that could be done.

### DIFF
--- a/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
+++ b/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
@@ -14,7 +14,7 @@ open System
 open FSharp.Reflection
 open Newtonsoft.Json
 open Newtonsoft.Json.Converters
-open System.Collections.Generic
+open System.Collections.Concurrent
 
 /// Converts F# options, tuples and unions to a format understandable
 /// by Fable. Code adapted from Lev Gorodinski's original.
@@ -36,7 +36,7 @@ type JsonConverter() =
         advance reader
         read 0 List.empty
 
-    let typeCache = Dictionary<Type, bool>()
+    let typeCache = ConcurrentDictionary<_, _>()
     override x.CanConvert(t) =
             match typeCache.TryGetValue(t) with
             | false, _ ->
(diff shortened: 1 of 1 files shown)


src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
```diff
@@ -34,10 +35,18 @@ type JsonConverter() =
                 read (index + 1) (acc @ [value])
         advance reader
         read 0 List.empty
+
+    let typeCache = Dictionary<Type, bool>()
```
alfonsogarciacaro: I think this is very sensible. As we don't have control over when `CanConvert` will be called, shall we make this a `ConcurrentDictionary` to be on the safe side?

src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
```diff
@@ -36,15 +36,15 @@ type JsonConverter() =
...
             match typeCache.TryGetValue(t) with
             | false, _ ->
                 let result = 
                     t.Name = "FSharpOption`1"
                     || FSharpType.IsTuple t
                     || (FSharpType.IsUnion t && t.Name <> "FSharpList`1")
-                typeCache.Add(t, result)
+                typeCache.TryAdd(t, result) |> ignore
                 result
             | _, result -> result
```
alfonsogarciacaro: Sorry, I forgot to tell you but `ConcurrentDictionary` has a `GetOrAdd` method which allows to do this in a more expressive way and I think is more performant (because you don't need to get the lock on the dictionary twice:

```fsharp
typeCache.GetOrAdd(t, fun t ->
    t.Name = "FSharpOption`1"
    || FSharpType.IsTuple t
    || (FSharpType.IsUnion t && t.Name <> "FSharpList`1"))
```

alfonsogarciacaro: @ncave Thanks for the clarification! :+1: One question though, the purpose of `&& t.Name <> "FSharpList`1"` is to make Json.NET deal with F# lists. If the guard is omitted, the converted will try to parse them as other union types giving unexpected results. Are you sure it's safe to remove it?

ncave: Sorry, you're right of course, it's needed.

alfonsogarciacaro:
Thanks!

PR #575, opened 2016-12-01
Speed up the compiler

davidtme:
Same sort of thing as #567 

Before and after:
image

image

(I'm not sure `LocationEraser` is being used but I've done it for completeness)


forki:
💋

alfonsogarciacaro:
Nice one, thank you!

PR #576, opened 2016-12-01
Try to reproduce union parse bug

forki:
@alfonsogarciacaro as discussed on gitter. Here is the corresponding unit test


forki:
closed because things are moved to ofJsonWithTypeInfo

PR #580, opened 2016-12-02
Handle sprintf "%X"

### LINKED ISSUE #504: sprintf "%X" does not work
### Description

code "sprintf "%X" 255" compiled with fable returns "255" in JS
### Expected behavior

returning "FF", as it does in FSI

davidtme:
fixes #504


alfonsogarciacaro:
I cannot accept this PR because the build is broken... Hahaha, just kidding ;) This is cool, thank you!

PR #582, opened 2016-12-02
Fixed Json.Converter for netcore

ncave:
- Fixed Json.Converter for netcore
- added missing Printf.sprintf signature
- added type signature output for methods


alfonsogarciacaro:
Thanks!

PR #594, opened 2016-12-07
Added char to int

ncave:
- Properly converts char to int32
- Fixed printfn "%A" to use .ToString() overwrite

### DIFF
--- a/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
+++ b/src/fable/Fable.Compiler/FSharp2Fable.Util.fs
@@ -796,7 +796,8 @@ module Types =
         else
         match fullName with
         | "System.Boolean" -> Fable.Boolean
-        | "System.Char" | "System.String" | "System.Guid" -> Fable.String
+        | "System.Char" -> Fable.Char
+        | "System.String" | "System.Guid" -> Fable.String
         | "Microsoft.FSharp.Core.Unit" -> Fable.Unit
         | "Microsoft.FSharp.Core.FSharpOption`1" ->
             let t = Seq.tryHead genArgs |> Option.map (makeType com ctx)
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
Originally, I wanted to represent chars as integer as you're doing here. But I discarded the idea because then enumerating strings didn't work well. I need to check if I aded tests for that. For example, something like this needs to work:

```fsharp
    "foo"
    |> Seq.rev
    |> Seq.map string
    |> String.concat ""
```

alfonsogarciacaro:
Sorry, I was mistaken. I see that you still represent chars as strings at runtime. I guess it's fine then. But please let me have a better look at the code later, probably it's better to add the tests for enumerating strings anyway.

ncave:
@alfonsogarciacaro The example above is passing, there is already a test for it [here](https://github.com/fable-compiler/Fable/blob/master/src/tests/Main/UnionTypeTests.fs#L396).

ncave:
@alfonsogarciacaro Also added a fix to properly call the .ToString() overwrite in printfn "%A".

alfonsogarciacaro:
But `printfn "%A"` doesn't always call `.ToString()`, for example, the code below doesn't print `foo`:

```fsharp
type R =
    { a: string }
    override __.ToString() = "foo"

let r = { a = "bar" }
printfn "%A" r
```

ncave:
Not always, only when .ToString() is defined. This matches the behavior of printfn "%A" in FSharp, which is to let objects define how they print themselves if they want to.

TL:DR printfn "%A" in FSharp calls .ToString()

ncave:
@alfonsogarciacaro If you need the purely internal structure output, there is another format specifier "%+A" that we can repurpose for that, cause it's probably used less often. The printfn output is not an exact match to F# anyway due to the json serialization.

alfonsogarciacaro:
I guess we don't need that for the moment :) Thanks for the PR!

PR #598, opened 2016-12-11
Correct the docs regarding default code generation target

latkin:
Installed Fable today for the first time and noticed that commonjs is **not** the default codegen target, contrary to docs.

Line 241 correctly states that umd is the default target.

### DIFF
--- a/docs/source/docs/compiling.md
+++ b/docs/source/docs/compiling.md
@@ -32,7 +32,7 @@ Besides the default argument (`--projFile`), the following options are available
 Option                  | Short     | Description
 ------------------------|-----------|----------------------------------------------------------------------
 `--outDir`              | `-o`      | Where to put compiled JS files. Defaults to project directory.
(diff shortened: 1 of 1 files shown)


inosik:
Actually, `es2015` is the default at the moment, I think @alfonsogarciacaro made it the default with the integration of Rollup, because Rollup works with ES2015 modules. Prior to that, it was `commonjs`, I adjusted the command line in #486.

alfonsogarciacaro:
Yes, @inosik is right. I still need to update the docs for Fable 0.7 sorry. I had no time for it in the past weeks, but hopefully they should be ready in the next few days.

alfonsogarciacaro:
Ok, finally the docs have been update so I don't feel shame of myself by closing this :wink: (as it no longer applies to latest version). Thanks a lot for the PR in any case, and sorry for taking so long to deal with it!

PR #600, opened 2016-12-12
Added 64-bit integers

ncave:
- proper 64-bit integers
- proper unsigned 32-bit integers
- proper 32-bit floats
- BitConverter support


alfonsogarciacaro:
Awesome! :clap: Thanks a lot for this and sorry for taking a bit to accept it. I've been away for some days :wink:

I'll check why Babel is failing when converting fable-core. Anyways, now that Typescript can compile ES6 down to ES5 probably we can remove that step.

PR #602, opened 2016-12-13
Add additional interface test

tpetricek:
This seems to be working on `master` but it is not working in http://github.com/the-gamma/thegamma-script. I created a test while debugging this, which might be useful for Fable anyway.


alfonsogarciacaro:
I'll check this, thank you!

> Not sure if this only affects `IDisposable`, but this interface is a bit tricky because, for example, JS iterators don't implement it. So I'm planning to add a check always before calling `Dispose` in the next version.

PR #603, opened 2016-12-13
Mark watch parameter as Boolean

tpetricek:
I was running fable using:

    fable --target dev -w src

which did not work because the command line options parser parsed `src` as a parameter of `watch` rather than as the default `projFile` option, i.e. `{"watch":[ "src" ]}`. Marking the `watch` parameter as `Boolean` fixes the problem! 🎉


tpetricek:
Hmm, it seems that the purpose of the `watch` flag has changed, so now the correct use is

    fable --watch src --target dev src

Where `--watch` takes a list of paths to watch? Because with my change, I later get an error from rollup (which is trying to watch the list of things passed to `--watch`).

If this is the new behaviour, then that's probably OK, though I guess the following is ambiguous:

    fable --watch src

😕 ❓

alfonsogarciacaro:
Yeah, I didn't think of this case 😕 I always put the source path at the beginning (`fable src --watch`).

I still need to document it, but `--watch` has become more flexible because some users needed to specify the directories to be watched. Now, besides a boolean, `--watch` also accepts a [string or a string array](https://github.com/fable-compiler/Fable/blob/213e0bcd26a0d012b7651bd426d26bf68609f405/src/fable/Fable.Client.Node/js/fable.js#L104-L113).

I should update the documentation quickly. Babies and OSS are not a good combo 🍼

PR #611, opened 2016-12-23
Fix toJson to serialize { 'Case': undefined } as { 'Case': null } instead of {}

valery-vitko:
In the wild the issue occurs when a non-Fable js passes an `undefined` for an option of a DU field. `Serializer.toJson` converts this to `{ Case': undefined }` object, which actually means and serializes to `{ }`. 

I've considered null/undefined interchangeable in this concrete case. If it is crucial to preserve `undefined`, I'd propose serializing it as `[undefined]` (needs checks in both `toJson` and `inflate`).

### DIFF
--- a/src/fable/Fable.Core/ts/Serialize.ts
+++ b/src/fable/Fable.Core/ts/Serialize.ts
@@ -39,7 +39,8 @@ export function toJson(o: any): string {
           return v.Case;
         }
         else if (v.Fields.length === 1) {
-          return { [v.Case]: v.Fields[0] };
+          const fieldValue = v.Fields[0];
+          return { [v.Case]: typeof fieldValue === 'undefined' ? null : fieldValue };
         }
         else {
           return { [v.Case]: v.Fields };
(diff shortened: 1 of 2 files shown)


src/fable/Fable.Core/ts/Serialize.ts
```diff
@@ -39,7 +39,8 @@ export function toJson(o: any): string {
           return v.Case;
         }
         else if (v.Fields.length === 1) {
-          return { [v.Case]: v.Fields[0] };
+          const fieldValue = v.Fields[0];
+          return { [v.Case]: typeof fieldValue === 'undefined' ? null : fieldValue };
```
alfonsogarciacaro: Maybe in the future a maintainer could wonder why this is necessary. Could you please add a comment or maybe a reference to this PR? (e.g. `// See #611). Thanks!

alfonsogarciacaro:
Thanks for covering this case! I just added a little comment, could you please have a look? Fable is not making a distinction between `undefined` and `null` (this is important for example, for compatibility with Typescript optional members) and it always makes non-strict null checks (e.g. `x == null`) to cover also `undefined`, so I think your solution is fine :+1:

src/fable/Fable.Core/ts/Serialize.ts
valery-vitko: Done :ok_hand:

alfonsogarciacaro:
Thanks a lot for the quick amendment!

PR #612, opened 2016-12-25
Fixed failing tests

ncave:
No need to fix the radix to 10, it already defaults to 10 if not passed, and if fixed at 10 will prevent the ability to pass other bases, like 2 or 16.

e.g. see Convert.ToIntXX tests in StringTests.fs that were failing.


alfonsogarciacaro:
Ah, ok, sorry! The [MDN docs](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/parseInt) recommended to add always the radix parameter as some JS environments may not default to 10, but I guess that's a very rare case. Thanks!

PR #613, opened 2016-12-26
Fable.JsonConverter DateTime fixes

valery-vitko:
### Fix `DateTime` parsing failing on some cultures
Deserialization was failing on my local machine due to en-US culture. `serializer.Deserialize(reader, typeof<string>)` is culture specific with no easy API to override on per-call basis and could deserialize to something like `"12/20/2016 13:56:00"`. Therefore, I made a short circuit for already parsed dates to `reader.Value`, and preserved the current parsing for manual parsing fallback for non-recognized dates.

### Do not convert `DateTime.Kind = Unspecified` to UTC on serialization
[.ToUniversalTime()](https://msdn.microsoft.com/en-us/library/system.datetime.touniversaltime(v=vs.110).aspx) assumes `Kind = Unspecified` values as `Local` ones and converts them to UTC. But it breaks roundtrip - serialized and then deserialized value will have UTC offset applied twice. Thus I've overridden this behavior to match JsonConverter  pre-0.1.0. If you think it's better to stick with default BCL behavior, I'd be happy to revert this change.

### DIFF
--- a/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
+++ b/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
@@ -93,7 +93,8 @@ type JsonConverter() =
             | true, Kind.DateTime ->
                 let dt = value :?> DateTime
                 // Make sure the DateTime is saved in UTC and ISO format (see #604)
-                serializer.Serialize(writer, dt.ToUniversalTime().ToString("O"))
+                let universalTime = if dt.Kind = DateTimeKind.Local then dt.ToUniversalTime() else dt
+                serializer.Serialize(writer, universalTime.ToString("O"))
             | true, Kind.Option ->
                 let _,fields = FSharpValue.GetUnionFields(value, t)
                 serializer.Serialize(writer, fields.[0])
--- a/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
+++ b/src/nuget/Fable.JsonConverter/Fable.JsonConverter.fs
@@ -136,8 +136,11 @@ type JsonConverter() =
         | false, _ ->
             serializer.Deserialize(reader, t)
         | true, Kind.DateTime ->
-            let json = serializer.Deserialize(reader, typeof<string>) :?> string
-            upcast DateTime.Parse(json)
+            match reader.Value with
+            | :? DateTime -> reader.Value // Avoid culture-sensitive string roundtrip for already parsed dates
+            | _ ->
+                let json = serializer.Deserialize(reader, typeof<string>) :?> string
+                upcast DateTime.Parse(json)
         | true, Kind.Option ->
             let innerType = t.GetGenericArguments().[0]
             let innerType =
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
I always have problems with the machine region settings and DateTime, so I really appreciate your efforts to make Fable.JsonConverter more robust, thank you! :clap:

ncave:
I know this is passing CI somehow, but locally the DateTimeKind.Unspecified roundtrip is now failing on TZ=UTC-8:
 ```
1062 passing (3s)
  1 failing

  1) Json Date Kind Unspecified roundtrip:

      Error: Expected: 11 - Actual: 19
...
      at equal (C:\Projects\Fable\build\fable-core\Assert.js:13:15)
      at equal (Main\Util\Util.js:22:21)
      at Context.it (Main\JsonTests.js:404:22)
```

**Comment by alfonsogarciacaro** [comment]
Let's see what the CI server says, but this should be fixed with commit 5623bc388306266cd6240949e08b0969106067d5. I had the same problem in my machine so I took the chance to make `DateTimeKind.Utc` the default. This diverges a bit from BCL behaviour but I think is more sensible and makes it easier to deal with Dates coming from JS.

**Comment by ncave** [comment]
CI looks ok but now 5 tests are failing locally for a different reason:
```
1058 passing (3s)
  5 failing

  1) Json Nested generics:
     TypeError: Class constructor A cannot be invoked without 'new'
      at new class_1 (C:\Projects\Fable\build\fable-core\Util.js:56:27)
      at inflate (C:\Projects\Fable\build\fable-core\Serialize.js:205:34)
      at inflate (C:\Projects\Fable\build\fable-core\Serialize.js:203:26)
      at inflate (C:\Projects\Fable\build\fable-core\Serialize.js:203:26)
      at ofJson (C:\Projects\Fable\build\fable-core\Serialize.js:216:12)
      at ofJson (Main\JsonTests.js:37:34)
      at Context.it (Main\JsonTests.js:223:16)

  2) Json Nested options:
     TypeError: Class constructor A cannot be invoked without 'new'
...
  3) Json Nested arrays:
     TypeError: Class constructor A cannot be invoked without 'new'
...
  4) Json Nested tuples:
     TypeError: Class constructor A cannot be invoked without 'new'
...
  5) Json generic:
     TypeError: Class constructor Wrapper cannot be invoked without 'new'
...
```

alfonsogarciacaro:
Weird, tests are passing now on my machine. @ncave I guess you're running the tests with the netcore version (though this should have not changed). Can you please try removing the `build` and `node_modules` folders in advance from your repo?

ncave:
I did, that's not it. It could be because netcore compiles the Fable.Core to commonjs, but it has to work either way, right?

alfonsogarciacaro:
Hmmm, I tried changing the `umd` modules to `commonjs`, but still tests run fine for me. And I cannot run the netcore version on Windows either because I downloaded the latest dotnet SDK and it's not reading project.json anymore 😬

PR #614, opened 2016-12-27
Added String.Compare

ncave:


alfonsogarciacaro:
Wow, this is cool, thanks! :tada: :clap: I guess it's not needed to modify the FSharp.Core generic `compare` with strings, as it should be equivalent to an ordinal comparison, right?

Now it probably makes sense to allow `StringComparison` argument with `System.String.Equals` too.

PR #618, opened 2016-12-27
Swapped commonjs for umd

ncave:
Just so tests are compiled the same way for netcore.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -284,7 +284,7 @@ Target "FableCompilerNetcore" (fun _ ->
         // Compile Fable.Core TypeScript
         let fableCoreSrcDir = "src/fable/Fable.Core/ts"
         Npm.install __SOURCE_DIRECTORY__ []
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Are tests passing on your machine with this change?

ncave:
No, that's just to make the Fable.Core compile the same way between platforms.
CI tests are passing, I'm still trying to figure out what's causing the local tests to fail.

ncave:
Ok it works now, looks like it was the `"ecma": "es2015"` setting in tests/fableconfig.json

alfonsogarciacaro:
Cool, thank you! :+1: I'll try to check why compiling to ES2015 didn't work after today's changes anyway.

alfonsogarciacaro:
@ncave I have fixed the tests when compiling to ES2015 and I've also updated the build so netcore tests are closer to the .NET/Mono version (e.g., `project.fsx` has been removed). Could you please check and confirm if the tests work for you?

Issue #584, opened 2016-12-05
Fable support in dotnet sdk


alfonsogarciacaro:
@enricosada

I'm opening this issue to follow the discussion [here](https://github.com/enricosada/Fable/blob/use-dotnetsdk/NOTES.md). It's interesting to keep a conversation using git commits but probably this is a better medium ;)

I'm trying to summarize the three main parts of the integration and the tasks we need to accomplish:

## Project initialization

User just types `dotnet new -t fable` as for other dotnet projects. More templates for different Fable projects (React, etc) will be available in the future.

**TODO**: Will you create the template @enricosada?

## Package restore

We'll try to make things easier for the user, but not replace/hide npm completely. There should be a command like `dotnet fable install fable-powerpack` that will install a package both in `package.json` and the `.fsproj` (with a `<FableReference Include="Fable.Core" />` item).

**TODO**: Decide if the user should type `fable-core` (the npm package name) or `Fable.Core` (the F# project name).

**TODO**: Add the references to `package.json` and the `.fsproj`. I can add this capability to Fable but, @enricosada, do you know what's the best way to edit the `.fsproj` XML?

**TODO**: The `<FableReference Include="Fable.Core" />` must be translated to a `./node_modules/fable-core/Fable.Core.dll` (the tool should be also able to detect if `package.json` is in a parent dir). I guess @enricosada can do the MSBuild trickery, but I can write a script for path resolution if needed.

## Building

User will type `dotnet fable` (or `dotnet fable build`, or `dotnet build fable`) and the project will be compiled normally.

**TODO**: MSBuild should generate a response file with the F# compiler arguments (@enricosada? :wink:) and then call Fable with a reference to this file.

-----------------------

## Other considerations

Right now it looks like Fable.Sdk and fable-compiler will be two different things. Fable.Sdk will only contain the Fable.targets, while fable-compiler remains an npm package. The main problem with this is it may be difficult to keep the versions in sync, so Fable.Sdk must pass the version to the compiler and warn the user if they differ. Probably Fable.Sdk should contain also the dotnet sdk tool that would redirect the `dotnet fable` calls to the locally installed fable.

enricosada:
> TODO: Will you create the template @enricosada?

Yes, waiting for preview5, atm in preview4 the `dotnet-new` is not extensibile, but will (ref https://github.com/dotnet/netcorecli-fsc/issues/31 ). if fable is ready before preview5, we can just use an clitool.

[...]

PR #515, opened 2016-10-28
Restored netcore tests

ncave:
@alfonsogarciacaro Trying to make `build FableCompilerNetcore` pass again.

Almost works, the only outstanding issue is `System.Text.RegularExpressions.Regex` is not being replaced properly with `fable_core/RegExp` in RegexTests and TypeTests, I can't figure out why, for some reason it's being excluded from replacing.

PR #619, opened 2016-12-28
Fixed netcore build

ncave:
- Added Long.Equals
- Fixed netcore build

@alfonsogarciacaro Unless we're ready to migrate to the new .fsproj format that can build for all targets, we can't yet abandon the tests/Main/project.fsx as it serves an important purpose: to compile the tests with dotnetcore (before compiling with Fable). I'm as eager as anybody else to start the migration, so please let me know if your plan is to do an early adoption or wait until netstandard 2.0 and/or tooling are RTM.


alfonsogarciacaro:
Do you mean compiling the tests with dotnetcore and running them with NUnit or just the F# compilation to get the AST? In the first case, `project.fsx` is not used, [but `src/netcore/tests/project.json`](https://github.com/fable-compiler/Fable/blob/e3d6718fee140eb03975a7e38a906182c3ba9304/build.fsx#L281-L282). In the second case, the `.fsproj` file is being _cracked_ with Forge and the arguments are being passed to the F# compiler, so I thought it'd be a nice way to check the netcore version can also compile `.fsproj` files and also to avoid the maintenance of the separate `project.fsx`. I didn't run the `FableCompilerNetcore` build target because of the problems I mentioned before I'm having to compile FCS with netcore, but I just compiled the tests with the published `fable-compiler-netcore` and it was working.

About the migration, as I have little time available at the moment, I'd prefer to wait until the new dotnet SDK with MSBuild 15 is closer to release to avoid unnecessary work (relying on features that change in the last moment and such). But @enricosada has already started some work (see #584) so we could create a `0.8.0` branch and start releasing alpha versions for testing.

ncave:
@alfonsogarciacaro I am all in for project cracking (see #372 :), but you're making a wrong assumption, project.fsx is being used both for dotnetcore compilation and fable compilation, precisely to minimize the maintenance to just 2 project files for 3 platforms. Yes, we can move around the test project files from project.fsx to project.json, but that's not going to decrease maintainability in any way, it will instead increase it, as now the dotnetcore build will have to require installing mono to compile all the auxiliary dlls just for testing, or create new netcore builds for them.

We already had this [conversation](https://github.com/fable-compiler/Fable/pull/515#issuecomment-256955948) before, so if your preference is to wait until RTM for the new tooling that will allow for a unified .fsproj that can compile for netcore, my suggestion is to continue having the tests/main/project.fsx until that time. The effort to maintain test parity is fairly minimal and something that I do with every netcore release.

alfonsogarciacaro:
Ah, ok! Sorry, I just noticed that `src/netcores/tests/project.json` actually only contains a reference to `project.fsx`, I thought it already contained references to all the test files 😅 You're taking care of the netcore build and I really appreciate that so if keeping `project.fsx` makes maintenance easier it's absolutely the way to go :+1: Sorry for the confusion!

PR #631, opened 2017-01-05
Build fix

ncave:


ncave:
Not sure why the build's still failing, builds fine locally. The only difference I see is the older node version on CI.

alfonsogarciacaro:
Ups, I forgot to check the CI, it was also working locally for me. I think I now what's happening here. I forgot to run `npm i` on the root folder first (it works for us locally because we already downloaded typescript to the root dir). Thanks for the reminder!

PR #636, opened 2017-01-07
Fixed IEnumerable iterator

ncave:
Without knowing, at the same time as you, I literally just checked in a bug fix for the iterator, just a hair too late :)

Oh well...

This is great, thanks for adding better support, what I had in place was a bit brittle as it was relying on GetEnumerator methods being in a specific order to properly call each other.

If you don't mind me asking, what pushed you to do such a big change in the IEnumerable support, was it the iterator bug fixed in this PR?

### DIFF
--- a/src/fable/Fable.Compiler/Replacements.fs
+++ b/src/fable/Fable.Compiler/Replacements.fs
@@ -1634,15 +1634,13 @@ module private AstPass =
                 | [max] -> makeConst 0, max
                 | [min; max] -> min, max
                 | _ -> failwith "Unexpected arg count for Random.Next"
-            "Math.floor(Math.random() * ($1 - $0)) + $0"
-            |> emit info <| [min; max]
-            |> Some
+            ccall com info "Util" "randomNext" [min; max] |> Some
         | _ -> None
 
     let enumerable com (info: Fable.ApplyInfo) =
-        match info.methodName with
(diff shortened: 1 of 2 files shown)


ncave:
Still some things to iron out in the new enumerable support, see issue #637

alfonsogarciacaro:
Hehe, I think our minds are synced now :wink: I was reviewing some of the weak spots of `Replacements.fs` (e.g. emits with duplicated placeholders, entailing the risk of double evaluation) and among them, I didn't like my cheap trick to emulate .NET Enumerator with a JS iterator. Also, some auto-generated methods were being added in the FSharp2Fable step and others in the Fable2Babel step, so I thought it'd probably be better to unify them. One thing led to the other and at the end the commit came with quite an amount of refactoring.

Thanks for spotting the problem in #637! It should be fixed now.

It seems this PR is not compatible with latest master, sorry! Do you think we still need to integrate the change somehow or can we close it?

BTW, I just realised you weren't made yet a collaborator of the project. So I fixed that as well :smile:

ncave:
No need for this PR anymore, as it's producing the same iterator as you have now in Seq.toIterator:

```
export function toIterator<T>(en: Enumerator<T>): Iterator<T> {
  return {
    next() {
      return en.MoveNext() ? { value: en.Current } : { done: true }
    }
  }
}
```

PR #641, opened 2017-01-09
Fixed netcore build

ncave:


alfonsogarciacaro:
Sorry! I'm always breaking the netcore build 😅 Maybe we should add it to the CI?

ncave:
Nah, it's ok. It will go away anyway in a month or two, I hope.

PR #645, opened 2017-01-11
Added Link to a new use case (Google Script)

easysoft2k15:
I don't know if this may be of interest but I created a library in Google Script (recompiling fable-core.ts with some minor changes) and a simple project in Visual Studio 2015 in order to:
1.  being able to write Google Script in F#
2. compile them with Fable
3. copy and past the generated javascript  in Google App script
This is the link to my github project:
https://github.com/easysoft2k15/GoogleScriptFableTemplate
This is the link of the project (VS Code) with the necessary changes to fable-core.ts:
https://github.com/easysoft2k15/Fable-Core-to-ES5
I inserted the link into the index.html file. I don't know if this is the right to do it. If not please advide.
Thank You
Alessandro

PS: Fable is a  very cool project!

### DIFF
--- a/docs/source/index.html
+++ b/docs/source/index.html
@@ -150,6 +150,7 @@
         <li><a href="https://github.com/Indy9000/noodle-soup">Cook a tasty noodle soup with Fable!</a> - Indy Garcia</li>
         <li><a href="http://kcieslak.io/Getting-Started-with-Fable-and-Webpack">Getting started with Fable and Webpack</a> - Krzysztof Cieslak</li>
         <li><a href="http://kcieslak.io/Creating-VS-Code-plugins-with-F-and-Fable">Creating VS Code plugins with F# and Fable</a> - Krzysztof Cieslak</li>
+        <li><a href="https://github.com/easysoft2k15/GoogleScriptFableTemplate">Writing Google Script in F#</a> - Alessandro Rizzotto</li>
         <li>Written something about Fable? <a href="https://github.com/fable-compiler/Fable/tree/master/docs">Send a pull request</a>!
       </ul>
     </div>


alfonsogarciacaro:
Hi Alessandro! Thanks a lot for your words and your work! Using F# and Fable to write scripts for Google Apps is really interesting. If you don't mind, can I make you a couple of questions to see if I can help you make the integration easier?

- It looks you're using an old version of Fable. I'd recommend using the latest version, as it contains many improvements and bug fixes. The only issue is fable-core 0.7.x is modular (there's no single fable-core.ts anymore) so it's better if we can make the necessary modifications directly in fable-core so you don't need to modify it after every update. Replacing `null` by `this` for `fableGlobal` should be possible, maybe you could send a PR?
- You mention there're some duplicated names in fable-core.ts. Not sure if this is still happening in the latest version but, do you have a specific example?
- To solve at once the problem with the polyfills and the modules, I'd recommend bundling the app and using [Babel Runtime transform](https://babeljs.io/docs/plugins/transform-runtime/) as in [this example](https://github.com/alfonsogarciacaro/fable-pixi-sample1/blob/master/fableconfig.json). This way you get automatically the polyfills you need and a single file with an optimized size.

Please tell me what you think. I can try to send a PR to your project if you want (though I have no experience with Google Script).

easysoft2k15:
Hi Alfonso,
first of all thank You for Your quick replay!
As per Your question:
1. I'll update to the new version and see what needs to be done in order 
to make it work with Google Script. I'll inform You as soon as it's done
2. I prefixed all the changed I've made in fable-core.ts with "GAS_" 
(Google App Script). For example in the FSet class there's a "delete" 
method that get compile to javascript that does not like to Google App 
Script. There are other situations like this one. This project
https://github.com/easysoft2k15/Fable-Core-to-ES5
contain the patched fable-core.ts: if You search for GAS_ You'll find 
all the occurrences of the same problem.
3. I'll definitely take a look at the example that You linked! Again I'm 
not super familiar with javascript and related tools so to me this is an 
opportunity to learn more!
Again Thank You for Your interest!
I'll report back asap.
Ciao
Alessandro

ncave:
@alfonsogarciacaro Can you please increment core/compiler versions? Thanks!

easysoft2k15:
Hi Alonso,
I managed to make Google Script work (almost...) with Fable 7.30 but still have some trouble.
I use a Babel Runtime Transfor as in the example linked below (almost same  fableconfig.js) but in the resulting bundle.js I don't get polyfill for:
* Map
* Symbol
* ArrayBuffer

Did I missed something?
Thank You!

alfonsogarciacaro:
Hmm, are you adding the [`babel-runtime` dependency](https://github.com/alfonsogarciacaro/fable-pixi-sample1/blob/e2dc3c51854e8c5701ed50fa0c5e4a21bf7d2f46/package.json#L7)? (Note this is additional to `babel-plugin-transform-runtime`.) If this doesn't work, you can just try loading the [core-js shim](https://github.com/zloirock/core-js/blob/master/client/shim.min.js) before your script, instead of using the Babel runtime transform. I'm not sure how dependencies are loaded in Google App Script, in HTML it would be just adding a `script` tag.

easysoft2k15:
Yes, I'm using babel-runtime dependency

```
//package.json
{
  "private": true,
  "name": "canvas",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "engines": {
    "fable": "^0.7.24"
  },
  "dependencies": {
    "babel-runtime": "^6.11.6",
    "fable-core": "^0.7.24",
    "fable-powerpack": "latest"
  },
  "devDependencies": {
    "babel-plugin-transform-runtime": "^6.22.0",
    "babel-polyfill": "^6.22.0",
    "fable-compiler": "next",
    "finalhandler": "^0.5.0",
    "open": "0.0.5",
    "rollup-plugin-replace": "^1.1.1",
    "serve-static": "^1.11.1"
  }
}
```

```
//fableconfig.json
{
    //"module": "umd",
    "sourceMaps": true,
    "projFile": "./MyScript.fsx",
	"babelPlugins": ["transform-runtime"],
	"rollup": {
        "dest": "./out/bundle.js"
	}
}
```

Moreover I cannot add a script tag because the way Google Script work is different from usual web development. Basically they are server side. 
What I've done is to create a a library that import Babel-polyfill. The problem is that I must use it with a prefix. So for example I need to write:

```
var myMap=new polyfill.Map();
```

For these reason I manually inserted the following line a the beginning of the generated bundle.js:
```
this.Map = polyfill.Map;
this.Symbol = polyfill.Symbol;
this.ArrayBuffer = polyfill.ArrayBuffer;
```

Some questions:
1. Is there a way to post process the generated bundle.js in order to make the process more smooth?
2. are there other es2015 "thing" like Map, Symbol and ArrayBuffer that I must account for?
3. I use the following babel plugins in the rollup configuration but they don't seems to work:
- babel-plugin-transform-es3-member-expression-literals
- babel-plugin-transform-es3-property-literals
What If done then is pre-process the bundle.js with the command:
```
babel out\bundle.js --plugins=transform-es3-property-literals,transform-es3-member-expression-literals -o out\bundle-2.js
```
This is done by hand and it feels "hacky". Any suggestion to get a way around it?

Thank You!

alfonsogarciacaro:
> Is there a way to post process the generated bundle.js in order to insert the above lines automatically (and in general use my own initialization code)?

Probably the easiest way is to write a node script to insert the necessary lines and run it in a fableconfig `postbuild` script.

> are there other es2015 "thing" like Map, Symbol and ArrayBuffer that I must account for?

About classes, I think only `Set` and `Promise`. And some new methods of existing classes like `Array.from` or `Object.assign` but these should be taken care of by the polyfill, as it monkey patches the prototypes of the already existing objects.

> I use the following babel plugins in the rollup configuration but they don't seems to work:

Can you please try adding them to the `babelPlugins` option in fableconfig.json (you can use a string array)?

Thanks again for your hard work making Fable compatible with Google App Script. This will be a very compelling use case!

easysoft2k15:
Hi Alonso,
thank You for the suggestions!
I updated my repository in order to work with Fable 0.7.30:
https://github.com/easysoft2k15/GoogleScriptFableTemplate
Just a note mainly for my future reference:
I changed a little bit strategy vs the previous solution and now only the poyfill library (from Babel-Polyfill) is used as Google Script Library (more o less same effect of use  <script/>  tag in browser). All Fable-Core functions are part of the main Google Script function together with the compiled FSharp code.
Right now the example in the repo is very contrived. It just make a list of all the files contained in google drive and log it to the Logger.
As soon as I have time I'll try to make a more compelling example.
Alessandro

PR #649, opened 2017-01-13
Problem with deserialized int64

et1975:
The unit test demonstrates the problem: deserialized int64 is not convertible to float, this used to work a couple of releases ago.


alfonsogarciacaro:
Yeah, this is because before `int64` was just being translated to JS number and recently a `Long` class has been introduced to keep the same semantics as in .NET (so you can use the 64 bits to store an integer value). Unfortunately, the `Long` class didn't play well with `ofJson` but this should be fixed in latest `fable-core` release :+1:

PR #662, opened 2017-01-21
Fixed netcore build

ncave:
- Fixed netcore build
- Fixed String.Contains and String.StartsWith with more than one parameter


ncave:
@alfonsogarciacaro Can you please bump/publish as well? Thanks!

alfonsogarciacaro:
Thanks! Version bumped and published :)

PR #663, opened 2017-01-22
Use {} when Window, Global and Self are undefined

easysoft2k15:
If there is not a specific reason to use null in place of {}, I'd would suggest to use the latter (as in Babel-Polyfill).


alfonsogarciacaro:
Thanks! So this is better than using `this`? Actually I introduced the global object to avoid that a Fable library created its own copy of Fable symbols, though this problem is mostly gone with Fable 0.7, so we might as well remove it.

easysoft2k15:
I thought about using ```this``` as well, but then I took a look at the Babel-polyfill and saw they using ```{}```.
I guess i less polluting..
Ciao

PR #668, opened 2017-01-24
F# record equality and comparison checks only declared members (fixes #666)

### LINKED ISSUE #666: F# Record structural equality and comparison include fields added dynamically
### Description

Utils.ts function ``equalsRecords`` gets record object members with ``Object.getOwnPropertyNames`` thus comparing all fields even if they were not declared in F# Record but added dynamically. Same goes for ``compareRecords``. This leads to a confusing behavior sometimes when a UX framework, for example, adds a member to an object for its own purposes.

### Repro code

``` F#
[<Test>]
let ``Record equality ignores dynamic fields``() =
    let r1 = { a = 1; b = 2 }
    r1?c <- 1
    let r2 = { a = 1; b = 2 }
    r2?c <- 2
    equal true (r1 = r2)
```

``equalsRecords`` and ``compareRecords`` uses ``[Symbol.reflection].properties`` to check fields declared in F# only.

### Related information 

* fable-compiler version: 0.7.37
* fable-core version: 0.7.25
* Operating system: Win10x64

valery-vitko:


src/tests/Main/ComparisonTests.fs
```diff
@@ -141,6 +142,24 @@ let ``Record equality works``() =
...
+    equal false (r1 <> r2)
+
+[<Test>]
+let ``Record comparison ignores dynamic fields``() =
+    let r1 = { a = 1; b = 2 }
+    r1?c <- 1
+    let r2 = { a = 1; b = 2 }
+    r2?c <- 2
+    equal 0 (compare r1 r2)
+
```
alfonsogarciacaro: In the CI server test are also run on .NET with NUnit so we need to wrap these with `#if FABLE_COMPILER` to get a green badge :+1:

valery-vitko: Oh, thanks for the tip! Was under impression that ``.\build`` includes ``NUnitTests`` step by default. Fixed.

alfonsogarciacaro: Hmm, probably it should. I'm just running it on the server to save a few seconds on my machine ;)

alfonsogarciacaro:
Awesome, thank you!

PR #672, opened 2017-01-25
[WIP] Allow file types in watch

davidtme:
This allow the watch flag to use paths and paths with file types such as:

```-w ./ ./Shaders/*.glsl```


alfonsogarciacaro:
As Fable 1.0 can watch any type of files through Webpack I think we can close it. Please feel free to reopen if you've any concern.

Issue #571, opened 2016-11-30
Primary constructor of types from signature files (.fsi) are mangled


ncave:
Given the following project:
```
//--- test.fsi ---
module M
type Test =
    new : bool -> Test      // <-- here is what's causing it
    member Status: bool
```
```
//--- test.fs ---
module M
type Test(flag:bool) =
    member x.Status = flag
```

```
//--- project.fsx ---
#load "test.fsi"
#load "test.fs"

[<EntryPoint>]
let main argv =
    let t = M.Test(true)
    printfn "Status: %A" t.Status
    0
```
Here the constructor call will be generated as:
`  const t = _test.Test[".ctor"](true);`

which causes a runtime error. It should be:
`  const t = new _test.Test(true);`

The runtime error does not happen if the constructor definition is not used.

alfonsogarciacaro:
I haven't tested Fable that much with .fsi files, probably this needs some work.

Sometimes it seems F# specs are endless... :/

ncave:
I hear you, unfortunately we can't call the constructor without that constructor definition. The only option in that case is to not use signature files at all, which is a steep limitation for compiling existing code bases.

I was hoping that it's something simple that you can just spring out of your magic hat ;)

alfonsogarciacaro:
Ok, I've investigated about this and it's not simple ;)

The problem happens because when using the .fsi, `meth.IsImplicitConstructor` fails (I guess this is because the method declaration in the .fsi file is taken, where there're no implicit constructors), so [this line](https://github.com/fable-compiler/Fable/blob/21b00ec75d83c3d66e4e38b0f760194a8f259439/src/fable/Fable.Compiler/FSharp2Fable.Util.fs#L226) doesn't evaluate to true.

This is a problem because JS only accepts one single constructor and the other must be compiled as static methods. I have tried a hack by checking that the method is the first of all constructors in the type and this seemed to work but then I had problems with the structs we added recently in #568 because they can have constructors but no implicit constructor (which must be generated automatically).

I haven't found a way to match the constructor reference to the actual implementation either :/

Can you please open an issue in FCS asking that `.IsImplicitConstructor` must evaluate to true also when using signature files?

ncave:
[...]

PR #676, opened 2017-01-27
Fixes issue #571 - ts2fable crashes on module names containing a dot (nested modules).

fsoikin:
ts2fable.js: visitModule now handles the possibility of the module's body being a ModuleDeclaration (aka nested module) rather than ModuleBlock.

### DIFF
--- a/src/ts2fable/ts2fable.js
+++ b/src/ts2fable/ts2fable.js
@@ -735,36 +735,46 @@ function visitModule(node, opts) {
       modules: []
     };
     var modPath = joinPath(mod.path, mod.name);
-    node.body.statements.forEach(function(node) {
-        switch (node.kind) {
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Wonderful, thanks a ton for this! ts2fable needs a big overhaul and I never find the time to do it... And also the code already got quite messy ;)

fsoikin:
There are some more minor issue like this one. I probably won't find time for an overhaul, but I'll push some more PRs shortly, if you don't mind.

alfonsogarciacaro:
I don't mind at all, that would actually be very awesome :tada:

BTW, just to let you know I'm planning to use the [TypeScript Type Checker](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#using-the-type-checker) in the next version.

PR #680, opened 2017-01-27
Don't emit [<AllowNullLiteral>] for enums. Fixes #679.

### LINKED ISSUE #679: ts2fable should not mark enums with [<AllowNullLiteral>]
### Description
When the source `.d.ts` file contains enums, those are translated as DUs. Turns out, it is illegal for DUs to have null literal. F# complains: "_Records, union, abbreviations and struct types cannot have the 'AllowNullLiteral' attribute_".

### Repro code
```
declare module a {
    enum E {
        A = 0,
        B = 1
    }
}
```

**Expected:**
```
module a =
   type E = | A = 0 | B = 1
```

**Actual:**
```
module a =
   type [<AllowNullLiteral>] E = | A = 0 | B = 1
```

fsoikin:


alfonsogarciacaro:
Thank you!

PR #692, opened 2017-01-31
Don't apply [<AllowNullLiteral>] to type aliases. 

### LINKED ISSUE #691: ts2fable shouldn't add [<AllowNullLiteral>] to union types.
### Description
Aliases for TypeScript's non-discriminated union types get an `[<AllowNullLiteral>]` attribute, which is illegal on the `Ux<_>` types that are used to represent such unions.

### Repro code
```
type X = string | number;
```

**Expected:**
```
type X = U2<string, float>
```

**Actual:**
```
type [<AllowNullLiteral>] X = U2<string, float>
```

fsoikin:
Fixes #691.

### DIFF
--- a/src/ts2fable/ts2fable.js
+++ b/src/ts2fable/ts2fable.js
@@ -22,6 +22,10 @@ enum:
 `[TYPE_KEYWORD] [DECORATOR][NAME] =
 `,
 
+alias:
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks!

PR #698, opened 2017-02-05
Replacement for Double.IsNaN and Math.Log(x, b) with tests

Zaid-Ajaj:
This PR adds a replacement for the method `System.Double.IsNaN` and for `Math.Log(x, baseValue)` with their tests and some more tests to make sure the output matches the output in .Net, for example in [these tests](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Math.cs#L217)


alfonsogarciacaro:
This is really nice, thanks a lot!

PR #699, opened 2017-02-06
Updated Fable.Client.Browser project to rc3

ncave:


alfonsogarciacaro:
Thank you! Hmm, for some reason the `dotnet test` command runs too slow and make some tests (MailboxProcessor.post, Async.Parallel) fail most of the times in the CI server. I need to disable them.

BTW, you may have noticed it: the build script is checking now the `DOTNET` environment variable if you want to pass it a different dotnet SDK, so you can keep preview2 to work with FCS. (I finally got how this thing works and now I'm putting a private copy of preview2 in precedence in the PATH to compile FCS for netcore).

PR #707, opened 2017-02-10
Fixed Babel AST json serialization

ncave:
Recent changes in DU serialization seem to have introduced some inconsistent behavior.
This PR fixes the JS Babel AST json serialization (toJson) to match the one produced by Json.NET.
It does not fix the reverse serialization (ofJson) so a few DU serialization tests are still broken.

- Good news: [Demo page](https://github.com/fable-compiler/Fable/tree/narumi/src/dotnet/Fable.Client.Browser/demo) works again!
- Bad news: ofJson DU has to be fixed so back and forth JSON serialization tests can pass.

@alfonsogarciacaro The change in Serialize.ts is just patching the problem after it already happened.
Somewhere there is a root cause of the problem with the duplicate object properties in the Babel AST.

### DIFF
--- a/src/dotnet/Fable.Client.Browser/Fable.Client.fs
+++ b/src/dotnet/Fable.Client.Browser/Fable.Client.fs
@@ -108,7 +108,14 @@ let makeCompiler opts plugins =
 
 let printFile =
     fun (file: AST.Babel.Program) ->
-        printfn "%A" file
+#if DOTNETCORE || DOTNET40
+        // technically there should be a conversion to JSON here
+        // but we're trying to avoid adding the Json.NET reference
+        // as this is just to show it compiles properly on dotnet.
+        printfn "Babel AST: %A" file
+#else
+        printfn "%s" (Fable.Core.JsInterop.toJson file)
+#endif
 
 let printMessages (msgs: seq<CompilerMessage>) =
     msgs
--- a/src/dotnet/Fable.Client.Browser/demo/project.fsx
+++ b/src/dotnet/Fable.Client.Browser/demo/project.fsx
@@ -9,4 +9,6 @@ let compileSource readAllBytes references source =
     let com = makeCompiler opts []
     let fileName = "stdin.fsx"
     let files = compileAst com checker (fileName, source)
-    files |> Array.ofSeq
+    files
+    |> Seq.map (fun file -> Fable.Core.JsInterop.toJson file)
+    |> Array.ofSeq
--- a/src/dotnet/Fable.Client.Browser/demo/repl.js
+++ b/src/dotnet/Fable.Client.Browser/demo/repl.js
@@ -324,21 +324,6 @@
       presets.push('babili');
     }
 
-    function fixInvalidAst(ast) {
-      function isObject(x) {
-        return x !== null && typeof x === "object" && !(x instanceof Number) && !(x instanceof String) && !(x instanceof Boolean);
-      }
-      let fixValue = function (k, v) {
-        return v && v[Symbol.iterator] && !Array.isArray(v) && isObject(v) ? Array.from(v)
-          : v && v.Case && v.Fields && Array.isArray(v.Fields) ?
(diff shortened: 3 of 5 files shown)


alfonsogarciacaro:
Awesome, I'll check the demo page! :clap:

Fields prefixed with `@` are property backing fields automatically generated by the F# compiler. I'll have a look to see what can be done. Thanks for pointing the problem out!

ncave:
@alfonsogarciacaro Thanks!
For the demo page you just need the metadata in the [right place](https://github.com/fable-compiler/Fable/blob/narumi/src/dotnet/Fable.Client.Browser/demo/repl.js#L530) under the demo folder, and then compile with Fable (narumi), but I'm sure you already figured that out.
```
cd Fable_narumi
build
cd ./src/dotnet/Fable.Client.Browser/demo
cp /somewhere/metadata/ ./out/metadata/
node ../../../../build/fable
```
There are still some issues with minifying (it works, you can try it by enabling the babelrc in the demo fableconfig, but it's altering the code in a way so that the bundle does not load fully). But as you can see, minifying is not *that* important when http compression is enabled.

TL;DR. IMO a live playground page would make for a good showcase for Fable on fable.io

alfonsogarciacaro:
Haha, I do appreciate you wrote the build instructions because I'm always really bad at figuring them out. Just a tip

I think I found the problem with the generated AST. The `U2` unions weren't being erased. This is because it was not the type from Fable.Core but another one defined in Fable.AST.Common. This comes from an idea I had at the very beginning (and I discarded later) that the special attributes like `Erase` should be checked only by name and not fullname to remove the Fable.Core dependency in the bindings.

I fixed that so now the AST should match Babel specifications and in theory the JSON serialize/deserialize steps shouldn't be necessary. I removed to them to test it in [this commit](https://github.com/fable-compiler/Fable/commit/4462505e03b3360d328f6a67fcef38535a168285). I also temporarily undid your changes in `Serialize.ts` in the hope the problem you had was indeed the non-erased unions, but we can bring the changes back again if necessary.

So as you can imagine I managed to follow your instructions and build the demo. I use the `ExportMeta` configuration of the fable_poc branch of your FCS fork to generate the metadata but I still had to copy some assemblies manually from .NET Framework 4.0.

Unfortunately I couldn't make the REPL compile a simple `1 + 1` expression. I always get errors [here](https://github.com/ncave/FSharp.Compiler.Service/blob/4f75bd88b5199ae01eef9887caa390fdad9daaae/src/fsharp/FSharp.Compiler.Service/project.json#L272) saying some assemblies couldn't be found. I'm not sure where the requests come from, I started refreshing and debugging the page and adding the references manually, like `System.dll` or `System.Reflections.Primitives.dll`  as you can see in my last commit. But I gave up when the program told me `System.Resources.Manager` was missing.

Do you know why are so many references being asked?

ncave:
@alfonsogarciacaro Yes, I know why ;)

You're trying to export .NET Core binaries instead of .NET 4.0 Full CLR.

I haven't been able to make the .net core binaries work yet, they should work but they don't for some reason (although I haven't spent much time on it). And yes, the dependency list for .net core is [different](https://github.com/ncave/FSharp.Compiler.Service/blob/fable_poc/src/fsharp/Fable.FCS/app.fs#L18-L36). I'll try to make the .net core exported binaries work eventually (because they're so much smaller), perhaps when FCS gets on with the times and the new project build system.

In the mean time, just build/run the [Export.FCS.sln](https://github.com/ncave/FSharp.Compiler.Service/blob/fable_poc/Export.FCS.sln) solution on Windows which will produce classic .net 4.0 binaries, and grab them from there.

P.S. Yes, it was the U2/U3 DU that was breaking the Babel AST, didn't I tell you? No?
Of course I didn't, that would only save you some time and take away the fun of digging stuff out ;)
Sorry about that.

alfonsogarciacaro:
Hehe, no worries. It didn't take very long as there're not many DUs in AST.Babel ;)

Thanks for the tip, the binaries are working now!

PR #716, opened 2017-02-16
FSharpType and FSharpValue tests.

FrankBro:
Some unit tests for FSharpType and FSharpValue that would help making dynamic forms.


alfonsogarciacaro:
Awesome, thank you!

PR #722, opened 2017-02-20
Avoid html elements being blocked by AdBlock in Fable website

codec-abc:
Just changed the class name from `fb-*`to `fable-*` and it does the job. Update the docs.fsx so it still works with the new Suave API


alfonsogarciacaro:
Awesome, thanks! I tried something similar in the past, but for some reason it didn't work. I probably did something wrong :)

codec-abc:
Maybe you forgot to change the publishSite variable in the docs.fsx script. I didn't think to do that at first, and my local pages were still referencing the fable.io's css files.

Issue #697, opened 2017-02-03
Serialization of discriminated union map keys with Json.NET and Fable.JsonConverter


chadunit:
### Description

Following your 0.7 migration guide, I am using Json.NET with the Fable.JsonConverter for server-side serialization. Unfortunately, discriminated unions do not get serialized correctly when used as map keys.

This appears to be an underlying limitation of Json.NET: https://github.com/JamesNK/Newtonsoft.Json/issues/1201

Note that your client-side toJson/ofJson does successfully serialize/deserialize DU map keys. I'm submitting this here in case you might consider addressing this scenario server-side as well in your Fable.JsonConverter.

### Repro code

Server-side (repro in FSI):
```
#r "packages/Newtonsoft.Json/lib/net45/Newtonsoft.Json.dll"
#r "packages/Fable.JsonConverter/lib/net45/Fable.JsonConverter.dll"
open Newtonsoft.Json
let serialize obj = JsonConvert.SerializeObject(obj, Fable.JsonConverter())
type Keys = Key1|Key2|Key3
printfn "DU standalone: %s" (serialize Key3)
printfn "DU map keys: %s" (serialize ([(Key3,333);(Key1,111)] |> Map.ofList))   // INCORRECT
```
Output, note the generic type names for DU map keys:
```
DU standalone: "Key3"
DU map keys: {"FSI_0002+Keys":111,"FSI_0002+Keys":333}
```

Client-side, valid serialization:
```
#r "node_modules/fable-core/Fable.Core.dll"
open Fable.Core.JsInterop
let serialize obj = toJson obj  // Fable client-side
type Keys = Key1|Key2|Key3
printfn "DU standalone: %s" (serialize Key3)
printfn "DU map keys: %s" (serialize ([(Key3,333);(Key1,111)] |> Map.ofList))   // CORRECT
```
Output:
```
DU standalone: "Key3"
DU map keys: {"\"Key1\"":111,"\"Key3\"":333}
```

### Related information 

* Fable.JsonConverter version: 0.1.1
* Newtonsoft.Json version: 9.0.1
* Operating system: Windows 10

alfonsogarciacaro:
Hi @chadunit, thanks for reporting this! Actually, the way we solve that on the client side is precisely by doing the opposite of what James says: putting JSON in the key ;) I guess we can do the same in Fable.JsonConverter, I'll try to do that. You can also give it a go if you want :)

chadunit:
[...]

PR #743, opened 2017-03-05
Enable serialization of union/record map keys #697

chadunit:
Introduces new handling for Maps and Dictionaries with non-string keys in JsonConverter for issue #697. Non-string keys are serialized and the resulting JSON used as the new string key, so union/record keys now work correctly and are uniquely identified by JSON representation instead of the previous generic type names via the default ToString().

I don't know about the DOTNETCORE ReflectionAdapters part as I haven't used .NET Core yet. I think I added what is needed based on docs but couldn't test.


alfonsogarciacaro:
This is great, thanks a lot for this! :clap:

PR #746, opened 2017-03-08
Updated to match FCS-fable latest

ncave:


alfonsogarciacaro:
Thanks! Is this fixing the error reports?

ncave:
@alfonsogarciacaro No, what error reports, can you please elaborate?
This is just preparation for adding the fsac capability, I'll post an example later.

ncave:
@alfonsogarciacaro [Here](https://github.com/ncave/FSharp.Compiler.Service/blob/fable/src/fsharp/fcs-fable/app.fs#L76-L87) is an example for FSAC usage, seems to work at least in this case. Tooltip text shows some composition error in the message, may need further investigation.

alfonsogarciacaro:
@ncave Cool! Thank you, I'll give it a try :+1:

About the errors, sorry, I think I saw something about the exceptions and thought it was related to the REPL sometimes showing `internal error` instead of the proper syntax/type error. But actually I just checked the latest REPL and couldn't reproduce so it seems you already fixed it 😄

ncave:
@alfonsogarciacaro Ok, looks like tooltips work too, I updated the [test](https://github.com/ncave/FSharp.Compiler.Service/blob/fable/src/fsharp/fcs-fable/app.fs#L76-L86).

ncave:
@alfonsogarciacaro Had to rebase FCS-fable to latest FCS 11.0.8, sorry about that, please grab latest.

alfonsogarciacaro:
This is great, @ncave, thank you! Ionide-web running without a backend is much closer now :)

No worries about the rebase, I hadn't grabbed the lates FCS bits yet. I just merged #749 to turn Fable into a Webpack loader, so I need to update the REPL build before anything.

PR #751, opened 2017-03-14
Make TransformOptions not inherit

jgrund:
`TransformOptions` currently inherits from `ReadableOptions` and `WritableOptions`. Because the types are exactly the same, it seems like we can't tell F# which slot to use.

Instead, we can just implement the items directly in the interface, which will resolve this issue.


alfonsogarciacaro:
Good point, thank you!

jgrund:
Thanks!

PR #753, opened 2017-03-15
Socket should inherit from NodeJS.ReadableStream

jgrund:
Socket is an extension of a Duplex stream. We should inherit from NodeJS.ReadableStream. The write methods appear to be implemented on socket already.


jgrund:
Hi @alfonsogarciacaro, Does this change look ok to merge?

alfonsogarciacaro:
It does :) Thank you!

Issue #754, opened 2017-03-18
[narumi] Records signature codegen regression


ncave:
There seems to be a codegen regression for records with signatures, this used to work before:

test.fsi
```
module Test

[<Sealed>]
type TRec =
    static member Create : string * string -> TRec
```
test.fs
```
module Test

type TRec =
    { a: string; b: string }
    static member Create (a, b) = { a=a; b=b }
```
Program.fs
```
module Program
open Test

[<EntryPoint>]
let main argv =
    let t = TRec.Create("1", "2")
    printfn "%A" t
    0
```
project.fsproj
```
<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp1.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <!--<Compile Include="test.fsi" />--> <!-- if you uncomment this it will stop working -->
    <Compile Include="test.fs" />
    <Compile Include="Program.fs" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="FSharp.NET.Sdk" Version="1.0.*" PrivateAssets="All" />
    <PackageReference Include="FSharp.Core" Version="4.1.*" />
  </ItemGroup>
</Project>
```
Here is what gets generated without the signature file (correct):
```
var TRec = function () {
  function TRec(a, b) {
    _classCallCheck(this, TRec);
    this.a = a;
    this.b = b;
  }
```
Here is what gets generated with the signature file (incorrect):
```
var TRec = function () {
  function TRec() {
    _classCallCheck(this, TRec);
  }
```

alfonsogarciacaro:
Hi @ncave. I'm still concerned about the .fsi files, particularly because there is no progress with https://github.com/fsharp/FSharp.Compiler.Service/issues/676. The easiest solution would be if Fable just ignore .fsi files, but you said they're necessary sometimes, right? Do you have an example?

ncave:
@alfonsogarciacaro The thing is, this was working fine fairly recently in narumi.
The .fsi signature files are used to define access (visibility), so removing them would require reworking pretty much every code base that uses signature files. As much as I would like for the [FCS/#676](https://github.com/fsharp/FSharp.Compiler.Service/issues/676) to be fixed, it was more or less the only issue with signature files before this one.

alfonsogarciacaro:
[...]

PR #755, opened 2017-03-18
[WIP] Fix REPL build

ncave:
Outstanding issues:
- Blocked on #754.
- babel standalone is not being generated, needs fixing or replace with cdn version.
- for some reason in testapp cannot import nodejs "fs"
- not an issue, just an observation: using webpack results in 50% bigger bundle size than before.
- not an issue per se, but for some reason with webpack we can no longer use nested .fsx script files as project files, if they use relative paths. The work-around is to use .fsproj and project references.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,9 @@
         "typescript": "^2.2.1",
         "webpack": "^2.2.1",
         "mocha": "^3.2.0",
-        "babel-core": "^6.23.1",
-        "babel-loader": "^6.4.0",
-        "babel-preset-es2015": "^6.22.0"
+        "babel-core": "^6.24.0",
+        "babel-loader": "^6.4.1",
+        "babel-preset-es2015": "^6.24.0",
+        "babel-preset-babili": "^0.0.12"
     }
 }
--- a/src/dotnet/Fable.Client.Browser/Fable.Client.Browser.fsproj
+++ b/src/dotnet/Fable.Client.Browser/Fable.Client.Browser.fsproj
@@ -1,23 +1,36 @@
 ﻿<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
     <Version>1.0.0-alpha</Version>
-    <DefineConstants>$(DefineConstants);DOTNET40</DefineConstants>
-    <DefineConstants>$(DefineConstants);FABLE_COMPILER</DefineConstants>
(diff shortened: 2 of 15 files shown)


alfonsogarciacaro:
Hi @ncave, thanks for this! Sorry, I need to have a look at the REPL after releasing Fable 1.0 beta.

> Blocked on #754.

Replied in the issue

> babel standalone is not being generated, needs fixing or replace with cdn version.

I'm using a custom babel-standalone bundle to include babel-template. It was being generated in a prebuild script in fableconfig.json: `node node_modules/webpack/bin/webpack babel-standalone.js repl/babel-standalone.js --output-library Babel -p`. I need to move this to the FAKE build target instead.

> for some reason in testapp cannot import nodejs "fs"

Probably you need to add [the node target in Webpack config file](https://webpack.js.org/concepts/targets/).

> not an issue, just an observation: using webpack results in 50% bigger bundle size than before.

This is because Webpack tree shaking algorithm is worse than roll-up. However, I did some tests before and after minification the bundle size was very similar for Rollup and Webpack. It seems [Webpack relies much more on uglify dead code elimination](http://2ality.com/2015/12/webpack-tree-shaking.html) for the ES2015 tree shaking feature.

> not an issue per se, but for some reason with webpack we can no longer use nested .fsx script files as project files, if they use relative paths. The work-around is to use .fsproj and project references.

I need to have a look into this, but probably it's related with the fact that, as fable-loaders like all Webpack loaders shouldn't have state, the Fable server needs to [make a guess whether a new request belongs to a new project or a file in the current compilation](https://github.com/fable-compiler/Fable/blob/131819a834df1191add86a4382055e5c4aa8f8eb/src/dotnet/Fable.Tools/State.fs#L174-L192). For .fsproj and .fs files it's very easy: the former define a new project while the latter don't, but for .fsx files it's trickier because they can have two roles.

In any case, I don't think it's a serious issue because using .fsx to list files was just a _hack_. Now that project files are much lighter, I think they should be preferred instead.

ncave:
> This is because Webpack tree shaking algorithm is worse than roll-up. However, I did some tests and after minification the bundle size was very similar for Rollup and Webpack. It seems Webpack relies much more on uglify dead code elimination for the ES2015 tree shaking feature.
Do you have uglify minification on any of the samples? I've tried using the rollup loader, not much of a difference (just a little bit). Tried the babili minifier, works as expected (from 12.5 MB down to 8.5 MB, about the same reduction it was doing before, about 4 MB). The rest of the padding seems to come from webpack itself, even after minification.
Before webpack: approx. 8.5 MB, minified 4.5 MB.
After webpack: approx. 12.5 MB, minified 8.5 MB;

alfonsogarciacaro:
Merging so I can test locally

PR #759, opened 2017-03-23
Int.parse overload for hex parsing

7sharp9:


alfonsogarciacaro:
Awesome, thank you!

ncave:
@7sharp9 If you need other bases, you can use [Convert.ToIntXX](https://github.com/fable-compiler/Fable/blob/master/src/tests/Main/StringTests.fs#L546-L614).
Perhaps those tests need to be moved to ConvertTests.fs

7sharp9:
@ncave I only need to and from hex at the moment, I don't think `int.parse("", NumberStyles)` has octal or other options, I suppose the commas and separator options may be wanted by someone...

PR #760, opened 2017-03-23
Add method to construct a Buffer

jgrund:
Buffer is currently missing a way to create new buffers.

Add `Buffer.from` which can take a string and construct a Buffer from it.


alfonsogarciacaro:
Thanks!

PR #762, opened 2017-03-23
expose deflate function in Fable.Core

johlrich:
Exposes the deflate function to handle a devtool use case needing both deflate / inflate to handle custom serialization.

Should there be additional tests added or would any existing serializations cover this implicitly?


alfonsogarciacaro:
Well, there are a couple of tests for `inflate` like [this one]() but as you say the functionality is a,ready checked in the other serialization tests, and exposing the function is for a very specific case not for the general user. You can add the tests later, but it's not that necessary.

Thanks a lot for the PR!

Issue #761, opened 2017-03-23
Fable.JsonConverter fails de deserialize json to Option types


Zaid-Ajaj:
### Description
Using Fable.JsonConverter to deserialize `Option<T>` that was serialized by Fable's `toJson` does not work:

### Repro code
- Create new C# Library 
- Install Fable.JsonConverter
- Install NUnit 
- Run these tests:

```csharp
    [TestFixture]
    public class FableConverterTests
    {
        static Fable.JsonConverter fableConverter = new Fable.JsonConverter();
        public T Deserialize<T>(string json)
        {
            return (T)JsonConvert.DeserializeObject(json, typeof(T), fableConverter);
        }

        [Test]
        public void DeserializingOptionNoneWorks()
        {
            var json = "null";
            var deserialized = Deserialize<FSharpOption<int>>(json);
            Assert.True(FSharpOption<int>.get_IsNone(deserialized));
        }

        [Test]
        public void DeserializingOptionSomeWorks()
        {
            var json = "5";
            var deserialized = Deserialize<FSharpOption<int>>(json);
            Assert.True(FSharpOption<int>.get_IsSome(deserialized));
            Assert.AreEqual(5, deserialized.Value);
        }
    }
``` 
### Expected 
`toJson (Some 5)` is serialized to just "5", therefore expected `Deserialize<FSharpOption<int>>("5")` (defined above) to be `Some 5` as well

`toJson None` is serialized to "null", therefore expected `Deserialize<FSharpOption<int>>("null")` to be `None : int option`  

### Actual 
Deserialization throws Exception:
![Exception](https://cloud.githubusercontent.com/assets/13316248/24268310/9fc6999a-100d-11e7-863b-4c4725e2691c.png)

### Related information 

* fable-compiler version: ---
* fable-core version: --- 
* Operating system: Windows 10

Zaid-Ajaj:
This seems a bit off, the F# equivalent of these tests do pass: 
```fsharp
let converter = new Fable.JsonConverter()
let deserialize<'a> (json : string) = JsonConvert.DeserializeObject(json, typeof<'a>, converter) :?> 'a
let serialize (value: 'a) = JsonConvert.SerializeObject(value, converter)

[<Test>]
let ``Option<int> deserialization from raw json works``() = 
    match deserialize<Option<int>> "5" with
    | Some input -> Assert.AreEqual(5, input)
    | None -> Assert.Fail("Should not happen")   

[...]

PR #763, opened 2017-03-23
Added tests for Fable.JsonConverter

Zaid-Ajaj:
After opening #761 it seemed that Fable.JsonConverter had no actual issues. It is probaby a problem with my local dev setup with a conflict of which FSharp.Core library I am using across projects. 

Anyways, I wrote a couple of tests to make sure nothing is wrong Fable.JsonConverter (as far as the issue was concerned) which seems to be the case!

Always good to have some tests around as sanity-checks 😄 😄


alfonsogarciacaro:
Awesome! Yes, I love tests too, thank you! :heart: I've to include them in the build. So does it mean that #761 can be closed?

Zaid-Ajaj:
Yes I closed just now, I will have to dig deeper into this. one has to solve his own problems these days, so sad! 😆

PR #764, opened 2017-03-23
[WIP] Fix REPL build

ncave:
This puts us very close to restoring the REPL functionality.

The REPL build is operational, it even works on some samples (tail, fib), but not all - looks like there are some regressions affecting the other samples that need to be investigated further.

@alfonsogarciacaro Please note that he metadata binaries have been renamed from *.txt to *.dll.txt to simplify handling, so please rename yours too.


alfonsogarciacaro:
Thanks! This is weird, the Webpack integration didn't change the code generation, so I don't understand why there have been regressions. Though it's true I had to make some changes to make it possible to compile files in parallel and in arbitrary order. I've to check. We should also probably create an npm package with the Fable/FCS JS version

ncave:
@alfonsogarciacaro There is definitely some regression, try running the testapp compiled with dotnet.
I've heard we can now debug F# in VSCode, is that the case? Does it just work, or are there some prerequisites?

alfonsogarciacaro:
@ncave Should work out of the box using latest FSharp.NET.Sdk, although I'm having some line mismatches. Check https://github.com/ionide/FsAutoComplete/pull/55#issuecomment-288579558

ncave:
@alfonsogarciacaro Ok thanks, just FYI, there is no minimization on the REPL bundle, for easier debugging.

alfonsogarciacaro:
@ncave That's fine, we could also probably use ES2015 directly. Actually the current REPL is ES2015 minus `let/const` because Chrome cannot optimize them yet.

alfonsogarciacaro:
@ncave Actually, now I remember that **there were indeed issues** when using the Babel es2015 preset to compile the REPL, so we should try removing it as it was [done before](https://github.com/fable-compiler/Fable/blob/2f41002fc7c446af38a2ba12427b68ffab0c2ced/src/dotnet/Fable.Client.Browser/demo/fableconfig.json#L4-L9).

PR #767, opened 2017-03-24
Test deserialization for provided type at run-time

Zaid-Ajaj:
Just to make sure this works, it is critical for [Fable,Remoting](https://github.com/Zaid-Ajaj/Fable.Remoting). I couldn't test it there because VS thinks it will just use it's own prefered version of FSharp.Core from my local machine instead of packages folder and mess my references 😑 will go for paket and vs-code I guess


alfonsogarciacaro:
Great, thanks a lot! Please let me know if you need me to implement something on my side :)

PR #769, opened 2017-03-24
Update node foreign interface - WIP

jgrund:
This is a work in progress to update the node foreign interfaces to use only interfaces.

There will be continuing work here, but wanted to push it early.


neoeinstein:
Is progress still being made on this? I'd very much like to see the improved node interfaces merged in once the current build error is resolved.

jgrund:
Hi @neoeinstein, yes I am still updating this. 

The interfaces are much improved but are not yet complete, so I will be fleshing them out over time. 

I'll fix the error and leave to @alfonsogarciacaro if he'd like to merge while not all interfaces are there, or wait until more of them are.

alfonsogarciacaro:
If it's useful to users I think we can merge now :+1: Hopefully the improvements to ts2fable will make it easier to add more interfaces.

Thanks a lot for your work, @jgrund!

et1975:
I think there needs to be  a lot of auto-opened modules, like `stream_types` or `event_types`, etc. otherwise how is one to know to open them? They don't exactly correlate to node's documentation, do they?

jgrund:
I went back and forth over whether to put everything in a single file, as there doesn't seem to be a way to group and re-export modules. Right now, types are tied to their module, ex: `open Fable.Import.Node.Stream` brings in `stream_types`.

jgrund:
In terms of correspondence, there is one in that `stream_types` lives in the `Stream` module. 

That said, I'm happy to re-group everything in a single file if people feel that works better (or if there is some way to re-combine opened modules and re-export).

src/dotnet/Fable.Core/Import/Node/Stream.fs
```diff
@@ -0,0 +1,124 @@
+module rec Fable.Import.Node.Stream
+
+open Fable.Core
+open Fable.Import.JS
+open Fable.Import.Node.Buffer
+open Fable.Import.Node.Events
+
+module stream_types =
```
et1975: What I mean  is, put `[<AutoOpen>]` on it, and then it will open if one opens parent module.

et1975:
I like separate files, I just think auto-openning the type modules will ideal (commented inline).

jgrund:
@et1975 ah, cool :). Didn't know you could do that. I'll add those in

jgrund:
Thinking about this more, not sure there is a difference between having [<AutoOpen>] and removing the nested module. 

The later has the advantage that you can avoid typing `*_types` when not referencing directly / trying to avoid clashes.

alfonsogarciacaro:
The motivation to use the `_types` suffix was to be able to use the module name as a value that corresponded to the `Globals` type (so we could use overloads, optional arguments, etc...). With the new structure it's ok to remove the nested `*_types` module as the types are already wrapped by an outer module (like `Fable.Import.Node.Net`).

But the tricky part is how to invoke the module global functions, as now you have to type `Net.net.createServer` (instead of just `net.createServer` as before) which may not be that much different than `Net.Globals.createServer` what do you think?

jgrund:
I think `Net.Globals` would be preferable to `Net.net`. I could also see something like `Net.Exports` as they correspond to the exports of the module.

PR #770, opened 2017-03-25
[narumi] Add tests checking for FSharp.Core 4.1 Result type support

johlrich:
I noticed projects specify 4.1 and tried to use the Result type, but couldn't yet.

This is currently just some tests that will pass with `dotnet test` but fail through fable.

I could try adding support, but not sure where to start yet; still getting more familiar with the project.


alfonsogarciacaro:
Hi @johlrich, thanks a lot for the tests! I had a quick look at the new Result type but it seems the F# compiler is doing a bit of magic (probably for performance reasons) so supporting it is not straightforward, but it shouldn't be too difficult though)

If you don't mind, I'll merge the PR after supporting Result. For now, there's another Result type in fable-powerpack if you want to use it :)

johlrich:
I don't mind at all, happy to help. Let me know if there are some other tasks towards 1.0 you think others could help with.

alfonsogarciacaro:
That's very neat, thank you! Besides bug fixing, one of the priorities at the moment is to [improve the ts2fable parser](https://github.com/fable-compiler/ts2fable/issues/1), it'd be great if you could help there. Of course, samples, tutorials, posts or templates are always very helpful :)

alfonsogarciacaro:
I'm merging your PR, I'll try to implement the Result type now :+1:

alfonsogarciacaro:
@johlrich I have added support for F# 4.1 Result type :) (cc @et1975) In F# this is a struct type, but it's not possible to pass objects by value in JS, so for Fable is just a normal union type (I also took the chance to optimize union cases with a single field by removing the array allocation). Can you please give it a try? You have to update `dotnet-fable` to 1.0.0-narumi-907 and `fable-core` to 1.0.0-narumi-906. I've also published a new version of fable-powerpack (1.0.0-narumi-3) that removes the customly define Result type, so you can use the Result CE there without conflicts.

alfonsogarciacaro:
BTW, you should be able to do `dotnet restore --no-cache` to update dotnet-fable.

Issue #772, opened 2017-03-25
Reflection: encode input/output types of functions


Zaid-Ajaj:
### Feature Request 🙏 
Please consider encoding type information of functions into the generated code to be available later for retreival.
### Example
Consider the types
```fsharp
type A = A
type B = B

type Record = { 
  func : A -> B
}
```
When fable compiles them, the type information available of `Record` for Reflection become:
```javascript
[_Symbol.reflection]() {
  return {
    type: "Stdin.Record",
    interfaces: ["FSharpRecord"],
    properties: {
      func: "function"
    }
  };
}
```
As you can see, the `func` property has type `function`. 

It would be great if the types of input and output are also available for retrieval, i.e. 

```javascript
properties: {
    func: ["function", [A, B]]
}
```
**Edit: could be worked around using `getType` from `Symbol`. Looking into it ....**

alfonsogarciacaro:
@Zaid-Ajaj No needs for workarounds, I think your proposal is the way to go, I'll try to add it. It's not difficult but it may take some time as fixing bugs in Fable 1.0 has priority.

Could you please PR some tests with the Reflection functions you would like to use (like [this one](https://github.com/fable-compiler/Fable/pull/716/files))? That always helps me a lot :)

Zaid-Ajaj:
@alfonsogarciacaro THANKS A LOT!! 

take your time on this one, I will PR tests soon :smile:

alfonsogarciacaro:
Hi @Zaid-Ajaj. I've finally addressed this. It was actually much easier than I thought 😅 There's one caveat though, for simplification Fable uses the same type for F# functions and delegates (`System.Func`) so `FSharpType.GetFunctionElements` just returns an array with all the argument types (the last one being the return type). The test you PRed works because tuples and arrays are the same for Fable, but I guess it will fail with F# functions of more than one argument. We can find a way later to better match the .NET semantics but for now I hope this will help you do your magic :wink:

Can you please try with the just published dotnet-fable 1.0.0-narumi-912 (remember to use `dotnet restore --no-cache` to upgrade)? You'll have to update the fable-core npm package as well.

Zaid-Ajaj:
@alfonsogarciacaro Thanks a lot, I was testing it just now and it worked!!!! 😄 using the array approach is fine for now. 

Two caveats tho:
 - When using functions of type `'A -> Async<'B>` it compiled to `[A, Async]` and type info of `B` is lost. I think it is an async-specific issue
[...]

PR #774, opened 2017-03-26
Inspecting function range and domain by Reflection

Zaid-Ajaj:
Addresses #772 

Adds failing tests to inspect input (domain) and output (range) types of functions extracted from record fields. 

Missing Replacements:
 - FSharpType.IsFunction
 - FSharpType.GetFunctionElements


alfonsogarciacaro:
Thanks! I'll add support for it as soon as possible :)

alfonsogarciacaro:
Sorry @Zaid-Ajaj, do you need this very soon? I need to prepare my talk for F# eXchange next week and also make some for Fable 1.0 (updating samples, docs...). Would it be OK to wait until the second week of April?

Zaid-Ajaj:
@alfonsogarciacaro That totally fine, don't even think about it, the talk is more important :smile:

PR #780, opened 2017-03-29
Update dotnet-fable.fsproj

jgrund:
Update TargetFramework to point at `netcoreapp1.1`.

I'm trying to use dotnet-fable on a Fedora 24 box and it only supports dotnet core 1.1.

Do you see any issue with bumping the `TargetFramework` here?

### DIFF
--- a/src/dotnet/Fable.Tools/dotnet-fable.fsproj
+++ b/src/dotnet/Fable.Tools/dotnet-fable.fsproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Version>1.0.0-narumi-905</Version>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
     <!-- See https://github.com/dotnet/netcorecli-fsc/issues/98 -->
     <FscToolPath>../../../lib/fsharp.compiler.tools.4.1.0</FscToolPath>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
This should be fine I guess. I was just using netcoreapp1.0 because according to the dotnet site it was more _stablish_ than 1.1 but this is not very clear... as with most netcore things ;)

alfonsogarciacaro:
Unfortunately, after targeting netcoreapp1.1 and republishing the package, I'm getting errors (see capture below). Maybe dotnet SDK can only run netcoreapp1.0 CLI tools at the moment? Pinging @enricosada in case he provide some insight. For now, I'm afraid I'll have to revert this change, sorry :/

capture

jgrund:
@alfonsogarciacaro Thanks for merging and testing out anyway. I've switched to running 1.0 in a docker container and that works well enough.

enricosada:
Sorry for late reply. 
It seems an issue with nuget https://github.com/NuGet/Home/issues/4801
So atm all dotnet cli tools need to be netcoreapp1.0

enricosada:
Maybe this https://github.com/dotnet/cli/issues/4764 is better. 

In practice the tool framework doesn't matter becuase is restored separately for project. 
The netcoreapp1.0 fw is used because that's the LTS (long term support) for .net core. 
1.1 vs 1.0 shouldn't be a big deal for a cli tool
As a note, both runtime (1.0 and 1.1) are bundled with dotnetcli

@alfonsogarciacaro why the need for netcoreapp1.1? Just to understand the requirement

alfonsogarciacaro:
@enricosada Thanks a lot for this and don't worry, actually it was lazy for my part not to look it up myself. @jgrund problems to run netcoreapp1.0 in a Fedora 24 VM, but fortunately this seems to be solved thanks to Docker :)

Issue #789, opened 2017-04-01
Unable to build Fable: Incorrect NUnit plugin path


Zaid-Ajaj:
### Description
A fresh clone won't build (Target "All") correctly becuase of incorrect path to NUnit plugin, I am not sure at which stage the error is happening. It seems that building the NUnit plugin works fine because when I got the error, the plugin was already compiled there. I will look into it...

![screenshot](https://cloud.githubusercontent.com/assets/13316248/24573078/07542326-167f-11e7-9d59-a9a92e7dee11.jpg)

### Expected
To build without errors
### Actual
Not building due to incorrect path to NUnit plugin
### Related information 
* Operating system: Windows 10

PR #790, opened 2017-04-01
Corrected path to NUnit when building

Zaid-Ajaj:
Addresses #789 
When the path is corrected, building with target "All" seems to work fine (on my windows 10).

@alfonsogarciacaro I am guessing you were also compiling fine on your mac without the "/MCD/" part in the path, so could this be os-specific and that the build script should check what os it is running on to determine the correct path?

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -335,7 +335,7 @@ let buildNUnitPlugin () =
     Util.run nunitDir dotnetExePath "restore"
     Util.run nunitDir dotnetExePath "build -c Release"
     CreateDir "build/nunit"
-    FileUtils.cp (nunitDir + "/bin/Release/netstandard1.6/Fable.Plugins.NUnit.dll") "build/nunit"
+    FileUtils.cp (nunitDir + "/bin/MCD/Release/netstandard1.6/Fable.Plugins.NUnit.dll") "build/nunit"
 
 let buildJsonConverter () =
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Wow, where's that `MCD` coming from? Actually I've been using Windows for a couple of months now. Hmm, this is strange. Anyways, I think my solution there of copying the binary is not good, not sure why I did it like that. Maybe you could try by setting the `-o` argument in `dotnet build` as [it's done for the JSON Converter right below](https://github.com/fable-compiler/Fable/blob/426061c92722f75493d9fab439bc93a7ab4a3427/build.fsx#L344-L345)?

Zaid-Ajaj:
Hmm that's weird. Providing output path does the trick and the tests are run. Appveyor doesn't seem to like [async cancellation](https://github.com/fable-compiler/Fable/blob/426061c92722f75493d9fab439bc93a7ab4a3427/src/tests/Main/AsyncTests.fs#L103) but Travis was happy with it. Oh well, tastes differ I guess...

async

alfonsogarciacaro:
Thanks! Yeah, I already fixed that test a couple of times, but it seems that sometimes it runs too slow in Appveyor and doesn't match the timeouts. Interestingly, it fails more since we switched to xunit/netcore.

Zaid-Ajaj:
Sounds like a race condition

PR #791, opened 2017-04-01
make JS `Array.isArray` return `bool`

johlrich:
Minor import signature change to make JS `Array.isArray` return `bool` explicitly.


alfonsogarciacaro:
Thanks!

PR #796, opened 2017-04-03
Update package.json to workaround UseCommonOutputDirectory env variable

mizzle-mo:
This adds the environment variable in a way that should work on all platforms.

This has come up repeatedly in Gitter.

Adds [cross-env](https://github.com/kentcdodds/cross-env) (Dev Dependency) which is a solid lib that allows you to use bash style syntax that will get converted to work on other platforms.


MangelMaxime:
I think the situation is somehow temporary as it will be fixed on Ionide side. (at least I think I read that once).

But could be helpful for new comers. For example, today we had 3-4 people blocked because of this variable setting :)

alfonsogarciacaro:
Thanks @mizzle-mo, this is a very clever fix. My only concern is that `UseCommonOutputDirectory` may be only available when running webpack, not while editing with Ionide which is when the problemn happens. Is this working for you?

Anyway, as @MangelMaxime says, this is a temporary issue in Ionide (actually in FSharp.NET.Sdk) and @enricosada is about to push a fix in the upcoming days :)

mizzle-mo:
@alfonsogarciacaro My bad, I think I might have only read part of the chat.  I thought this was impacting the build/debug experience.

alfonsogarciacaro:
@mizzle-mo NP! OK, if you don't mind we can keep this on hold to see if a fix is coming to Ionide soon. If not, we can apply a solution around these lines :+1:

mizzle-mo:
I am going to close this as the fix has already landed in ionide.  We no longer need the env var.  @alfonsogarciacaro  you might want to update your blog post as well.

PR #799, opened 2017-04-04
add failing test for type converter being composed

johlrich:
adds a test that passes when ran via dotnet but fails in js

### DIFF
--- a/src/tests/Main/ConvertTests.fs
+++ b/src/tests/Main/ConvertTests.fs
@@ -634,6 +634,20 @@ let ``System.Convert.ToString Decimal works``() =
     let x = "101"
     Convert.ToString(101.m) |> equal x
 
+[<Test>]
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks for the test, @johlrich! So it seems to be failing because of the composition then? Hmm, it's probably with the new lambda flattening feature. I'll check :+1:

alfonsogarciacaro:
Sorry for the delay, this should be working now :) Could you please rebase to accept the PR? (I can also just add the test manually if you prefer) Also, it seems the `()` unit argument is missing in the test.

Cheers

johlrich:
:+1: rebased / pushed and tests passed locally on both node/dotnet

alfonsogarciacaro:
Thank you! :heart:

PR #813, opened 2017-04-14
Update cwd for quicktest debugging in vscode

johlrich:
I couldn't launch the node debugger for quicktest without the cwd changed to match the output of the quicktest.js file


alfonsogarciacaro:
Ups, I didn't know I had committed the `.vscode/launch.json` file 😅 Of course this makes much more sense :+1:

PR #825, opened 2017-04-18
Added Convert.ToString

ncave:


src/dotnet/Fable.Compiler/Replacements.fs
```diff
@@ -179,17 +179,26 @@ module Util =
...
         | Fable.Array _ | Fable.Tuple _ | Fable.Function _ | Fable.Enum _ ->
-            GlobalCall ("String", None, false, [arg])
+            GlobalCall ("String", None, false, [args.Head])
+            |> makeCall i.range i.returnType
+        | Fable.Number _ | Fable.ExtendedNumber (Int64 | UInt64) ->
+            let arg =
+                match args.Head.Type, args.Tail with
+                | Fable.Number Int16, [_] -> emit i "$0 < 0 && $1 !== 10 ? 0xFFFF + $0 + 1 : $0" args
+                | Fable.Number Int32, [_] -> emit i "$0 < 0 && $1 !== 10 ? 0xFFFFFFFF + $0 + 1 : $0" args
+                | Fable.ExtendedNumber Int64, [_] -> emit i "$0.isNegative() && $1 !== 10 ? $0.toUnsigned() : $0" args
```
alfonsogarciacaro: We need to be a bit careful when duplicating an argument in an emit expression, as this can cause double evaluation if the argument is a function call. I should probably resolve this automatically somehow as [it is done for inline expressions](https://github.com/fable-compiler/Fable/blob/235245ff2227f12ba81a69e3bd1da31e236ca1f3/src/dotnet/Fable.Compiler/FSharp2Fable.Util.fs#L1255-L1261).

alfonsogarciacaro:
Great, thank you!

src/dotnet/Fable.Compiler/Replacements.fs
ncave: @alfonsogarciacaro Fixed in #829

PR #831, opened 2017-04-19
Fixes #817

### LINKED ISSUE #817: [BUG] TryParse gives an incorrect result
### Repro code

```fsharp
System.Int32.TryParse("9X")
```

### Description

In `fsharpi` the above returns `(false, 0)` but with Fable it [returns `(true, 9)`](http://fable.io/repl.html#?code=printfn%20"%25A"%20<|%20System.Int32.TryParse("9X")&astInput=false&babili=false&evaluate=false&lineWrap=false&presets=es2015%2Creact%2Cstage-2)

In the generated code, it uses `Number.parseInt("9X", 10)`, but the JavaScript `parseInt` function has different behavior from the F# `TryParse` method.

This can be fixed by replacing `Number.parseInt("9X", 10)` with `+"9X"` and adding an integer check, which should give the correct result.

(Also we should add some unit tests for this behavior, e.g. `System.Int32.TryParse("X9TRE34")` and `System.Int32.TryParse("9SayWhat12Huh")`)

If you want, I can probably send a PR to fix this.

### Related information 

* fable-compiler version: 1.0.0-narumi-909
* fable-core version: 1.0.0-narumi-909
* Operating system: Debian GNU/Linux Testing (64-bit)

Pauan:
Also adds unit tests.

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -915,9 +915,9 @@ module AstPass =
         | "toUpperInvariant" -> icall i "toUpperCase" |> Some
         | "toLower" -> icall i "toLocaleLowerCase" |> Some
         | "toLowerInvariant" -> icall i "toLowerCase" |> Some
-        | "chars" -> 
+        | "chars" ->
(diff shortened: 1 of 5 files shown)


alfonsogarciacaro:
This is great, thank you!

PR #834, opened 2017-04-19
Add Ability to choose arbitrary available port

### LINKED ISSUE #833: Add Ability to choose arbitrary available port
When running multiple fable servers in parallel, one has to find an available port for each one.

Instead, what if we provided a way to tell the fable server to choose an available port.

The server could then pass that port as an environment variable when it ran it's sub command.

jgrund:
Fixes #833.

  - Add a function that returns a free port.
  - Add a match for "free" which gets a free port.
  - Pass port into child process as FABLE_SERVER_PORT.

### DIFF
--- a/src/dotnet/Fable.Tools/Main.fs
+++ b/src/dotnet/Fable.Tools/Main.fs
@@ -5,11 +5,12 @@ open System.IO
 open System.Diagnostics
 open System.Reflection
 open System.Runtime.InteropServices
+open System.Net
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
This is very cool, thank you! Now I guess we just need to check FABLE_SERVER_PORT environment variable in the fable-loader :+1:

I was wondering whether we could allow multiple states for multiple projects in the same process, but this is probably a simpler solution :)

PR #835, opened 2017-04-19
add debugger() to Fable.Import.JS

### LINKED ISSUE #814: Add debugger fn to Fable.Import.JS?
Trying to setup debug points with chrome is often flaky for me when doing it from the fs files (breakpoints being set on wrong lines, not being able to set breakpoints on lines I'm expecting) and doing it in the JS is cumbersome so I'll often find myself using a fn like below to force the debugger to hit where I want it to reliably.

```fs
[<Emit("debugger;")>]
let debugger () : unit = jsNative
```

Thoughts on adding this in `Fable.Import.JS`?

johlrich:
calling the Fable.Import.JS.debugger() function will emit the js `debugger` keyword

closes #814


alfonsogarciacaro:
Thanks!

PR #837, opened 2017-04-19
add rollup-pluginutils as dev dependency

johlrich:
When I sync'd latest and ran `build Quicktest` I got the message:

> @ rollup D:\Development\GitHub\Fable\src\tools
> node ../../node_modules/rollup/bin/rollup -c

**�   Cannot find module 'rollup-pluginutils'**

After installing `rollup-pluginutils` I don't see this message anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "fable-utils": "^1.0.0",
     "mocha": "^3.2.0",
     "rollup": "^0.41.6",
+    "rollup-pluginutils": "^2.0.1",
     "typescript": "^2.2.2",
     "webpack": "^2.4.1"
   }


alfonsogarciacaro:
I wonder why I didn't get this :/ Thanks!

jgrund:
@johlrich Were you on latest master when you tried this?

This should have been fixed in https://github.com/fable-compiler/Fable/commit/648200855f16a54226213663fbab7a78a998ea7f

We should be able to revert this one I think.

alfonsogarciacaro:
@jgrund The quick test takes the rollup plugin [directly from source](https://github.com/fable-compiler/Fable/blob/d7428dcf8e6b7f90e1b79a9a0d7c6f3c626a0604/src/tools/rollup.config.js#L1) so I guess that was the problem.

Actually now that I think about it, I'm not sure why the other dependencies (like fable-utils) do work :/ I probably should use the released plugin also from the quick test.

johlrich:
Ahh yeah, having that referenced from source must have been why I saw the issue for quicktest.

@alfonsogarciacaro Another alternative to using the released package: maybe the top level package.json should reference rollup-plugin-fable [via a local path](https://docs.npmjs.com/files/package.json#local-paths) since it's not published?

alfonsogarciacaro:
That's a very sensible proposal @johlrich, thank you! :+1:

PR #839, opened 2017-04-20
Update index.js

jgrund:
FABLE_SERVER_PORT is passed in as a string. Update port to parse as an Int if it exists.


alfonsogarciacaro:
True, I missed that! 😅

PR #843, opened 2017-04-21
Fixed REPL build

ncave:
@alfonsogarciacaro
- All samples work again, finally :)
- Suggestion for future improvement may be to move `Compiler `and `State `up the chain from Fable.Tools to Fable.Compiler as they don't seem to have many dependencies.

### DIFF
--- a/src/dotnet/Fable.Tools/State.fs
+++ b/src/dotnet/Fable.Tools/State.fs
@@ -2,17 +2,28 @@ module Fable.Tools.State
 
 open Fable
 open Fable.AST
-open Newtonsoft.Json
 open System
 open System.Reflection
 open System.Collections.Concurrent
 open System.Collections.Generic
 open Microsoft.FSharp.Compiler
 open Microsoft.FSharp.Compiler.SourceCodeServices
+
+#if !FABLE_COMPILER
+open Newtonsoft.Json
 open ProjectCracker
 open Parser
+#endif
 
-type Command = string * (string -> unit)
+#if FABLE_COMPILER
+type Dictionary<'TKey, 'TValue> with
+    member x.GetOrAdd (key, valueFactory) =
+        match x.TryGetValue key with
+        | true, v -> v
+        | false, _ -> let v = valueFactory(key) in x.[key] <- v; v
+
+type ConcurrentDictionary<'TKey, 'TValue> = Dictionary<'TKey, 'TValue>
+#endif
 
 type State(projectOptions: FSharpProjectOptions, checkedProject: FSharpCheckProjectResults) =
     let entities = ConcurrentDictionary<string, Fable.Entity>()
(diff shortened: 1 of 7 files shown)


src/dotnet/Fable.Tools/State.fs
```diff
@@ -106,20 +129,21 @@ let loadPlugins pluginPaths (loadedPlugins: PluginInfo list) =
...
+//             match er.Severity with
+//             | FSharpErrorSeverity.Warning -> false, "warning"
+//             | FSharpErrorSeverity.Error -> true, "error"
+//         isError, sprintf "%s(L%i,%i) : %s FSHARP: %s"
+//             er.FileName er.StartLineAlternate er.StartColumn
+//             severity er.Message
+//     errors
+//     |> Array.map parseError
+//     |> Array.partition fst
+//     |> fun (ers, wns) -> Array.map snd ers, Array.map snd wns
```
alfonsogarciacaro: Yes, this is not used at the moment as errors don't throw anymore. This is to keep the compilation going on and being able to offer the full error list at the end. See #806.

alfonsogarciacaro:
Awesome, thank you! Now I need to streamline the REPL build process to make it easier to keep the REPL up-to-date, and also push a pure-JS FCS/Fable package.

Agreed about the suggestion! That'll simplify things, maybe you could send a PR?

ncave:
@alfonsogarciacaro I remember you said you got an idea how to work around [FCS issue 676](https://github.com/fsharp/FSharp.Compiler.Service/issues/676), do you mind sharing  the idea?

alfonsogarciacaro:
@ncave Sorry! I just noticed you asked the same in a previous issue and I completely missed it 😓 So basically the problem is that `IsImplicitConstructor` only works if the `FSharpEntity` reference corresponds to the implementation location (not the signature) and, as now [we're making sure Fable.Entity is created using the implementation and not the signature](https://github.com/fable-compiler/Fable/blob/42e2a7e0b7af8da25fb91975bee36e1052325fc9/src/dotnet/Fable.Compiler/FSharp2Fable.fs#L1121-L1122), we can use a similar process to be sure `IsImplicitConstructor` is called on the implementation entity.

I had a quick look and `IsImplicitConstructor` appears in several places so it needs some refactoring and I haven't started yet. Also I wonder which one of these alternatives is better performance-wise:

- Before calling `IsImplicitConstructor` make sure the entity reference comes from the implementation, and look for the implementation entity otherwise. May be slower if there are too many look-ups.
- Before calling `IsImplicitConstructor` translate to Fable.Entity (there's a cache and Fable.Entity is guaranteed to come from the implementation) and compare the arguments of the method reference to seen if it corresponds with the primary constructor.

ncave:
Thanks for sharing.
I was comparing the old and new REPL compile times, looks good, there is even a slight improvement on longer scripts. Hopefully it can be improved further.

PR #850, opened 2017-04-24
Allow compilation of multiple projects in parallel

alfonsogarciacaro:
This PR allows compilation of not only multiple files but also multiple projects in parallel with a single Fable server (projects shouldn't share files). This should be helpful in scenarios were multiple bundles have to be created like Electron apps (one for the main thread and the other for the renderer).

@jgrund I haven't tested this much, but maybe it can help you when compiling different projects in parallel without having to start multiple Fable servers.


jgrund:
@alfonsogarciacaro neat I'll try this out 😄 .

Issue #852, opened 2017-04-25
'dotnet fable webpack' failing for paths with spaces


Kurren123:
### Repro code

For windows: 

```
mkdir "folder with spaces"
cd "folder with spaces"
dotnet new fable-elmish-react
npm install
dotnet restore
rename "folder with spaces.fsproj" folder_with_spaces.fsproj
dotnet fable webpack
```

### Results:

![image](https://cloud.githubusercontent.com/assets/3525980/25390021/323cf65e-29ca-11e7-8848-849e4da22a4d.png)

Text output:
```
cmd /C node C:\Users\Kurren.Nischal\Documents\DevelopmentTests\folder with spaces\node_modules\webpack\bin\webpack.js
module.js:471
    throw err;
    ^

Error: Cannot find module 'C:\Users\Kurren.Nischal\Documents\DevelopmentTests\folder'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:393:7)
    at startup (bootstrap_node.js:150:9)
    at bootstrap_node.js:508:3
Closing server...
```

### Related information 

* fable-compiler version:1.0.0-narumi-*
* fable-core version:1.0.0-narumi-*
* Operating system: windows 10

alfonsogarciacaro:
Related #844. We probably need to quote [the webpack script location](https://github.com/fable-compiler/Fable/blob/afd1f8652cf176781befc6daea397f90c95ad75a/src/dotnet/Fable.Tools/Main.fs#L172). It would be very helpful if you could do it, test it and send a PR if it works :) If not possible, I'll try to do it later.

> To test the fix, make sure you've dotnet SDK installed, pull latest master, implement the fix, run `build.cmd FableTools` (faster than a full build) and from your project folder run `dotnet path/to/Fable/build/fable/dotnet-fable.dll webpack

Kurren123:
@alfonsogarciacaro I would certainly like to give it a go! Will try later today

PR #854, opened 2017-04-26
Quoted the webpack script path (#852)

Kurren123:
For [Issue 852](https://github.com/fable-compiler/Fable/issues/852#issuecomment-297061563)


Kurren123:
I was going to add a test but I don't think we have any for "dotnet fable" yet?

alfonsogarciacaro:
Thanks! Yes, we don't have integration tests atm :/ We should have some of these.

PR #865, opened 2017-05-02
Node Streams should be generic

### LINKED ISSUE #864: Node Streams should be generic
Streams in node can carry `string | Buffer | Object` types.

Make streams generic so we can track the contained types.

jgrund:
Fixes #864.

Streams in node can carry `string | Buffer | Object` types.

Make streams generic so we can track the contained types.

Signed-off-by: Joe Grund <grundjoseph@gmail.com>


alfonsogarciacaro:
Thanks!

PR #866, opened 2017-05-02
Add documentation for the pong-sample (website).

oopbase:
So here is the missing documentation for the pong sample for the website. I hope the markdown is correct, even though I would have liked to test it before creating a pull request :-).
If there is still something missing, please tell me.
Thanks in advance for adding my sample :-).


alfonsogarciacaro:
Awesome, thanks to you and sorry again, I completely forgot to add the sample to the web! 😅

alfonsogarciacaro:
The sample is published :tada: http://fable.io/samples/pong/index.html

The code looks a bit cramped, maybe you could make the lines shorter? Also, it seems the space and arrows interfere with the scroll when the web is too tall, it could be a good idea to change those as well :+1:

oopbase:
Yup, I will shorten the lines and use different keys. Thanks a lot for publishing! 👍

Issue #739, opened 2017-03-02
[discussion] Fable JS performance optimization ideas


ncave:
Currently the JavaScript version of Fable is about 10 times slower than the .NET version.
Please share any performance optimizations ideas that can potentially bridge that gap.

Art:
Fable -> TypeScript -> WebAssembly

ncave:
I've briefly looked at the Firefox and Node profilers results but I don't see any obvious bottlenecks besides a lot of GC (but maybe not enough to explain it all).

alfonsogarciacaro:
This is tricky, I need to check better but I couldn't find any performance bottleneck either. I'm also trying to avoid [performance killers](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers) in JS, so I'm not sure what else can be done to improve performance. I do have some ideas though.

Regarding the profiler:

- Apparently you cannot check the number of calls to the same function with the V8 profiler (or at least I couldn't find how), this would help focus on the parts where optimization could make more impact. Maybe we could use a Babel plugin like [this one](https://github.com/codemix/babel-plugin-trace) to trace function calls.
- There also seem to be ways to check the optimizations and deoptimizations done with the standalone V8 engine, but I haven't checked it yet. [Reference](https://www.html5rocks.com/en/tutorials/speed/v8/)

.NET features missing/different from JS:

- Is FCS using a lot of structs or floats (which in JS are heap values)?
- Maybe the structural equality checks are much more expensive in JS.
- Not exactly a different feature, but checking the profiler I noticed that some _innocent_ properties from FCS objects trigger a very deep stack when called from Fable, maybe some caching could help here.

Collections in fable-core:

- There are several functions in the `Seq` module that could be optimized, for example the different `fold` versions could use specialized loops instead of calling themselves.
- To make a faster release, I left many [functions in List and Array modules unimplemented](https://github.com/fable-compiler/Fable/blob/82f74fcd8b701f29d8749ccd8eb370d36c8e026e/src/fable/Fable.Compiler/Replacements.fs#L1292) and they've remained as such. In these cases, the corresponding function in `Seq` is called and a new List or Array is constructed afterwards. Probably it'll be more performant to have specialized functions instead. But I'm not sure of the impact.

ncave:
Nice links there, thanks @alfonsogarciacaro ! We can look at the deopt flagged functions to see if anything can be improved there: `node --trace-deopt out/testapp/project > deopt.txt`.

[...]

PR #877, opened 2017-05-06
[WIP] Reduced logging pending a verbose mode

mizzle-mo:
to be introduced later

This just reduces the amount of console logging during build.  Ran Fake -> Build -> All and confirmed a successful build.  Please let me know if anything else is required.


mizzle-mo:
@alfonsogarciacaro I added the console.log back, but I made it run Async.  The devdep I added is not being resolved by Travis, but is being resolved when I do Fake Build All.  How do I fix that?

alfonsogarciacaro:
@mizzle-mo Do we really need to make `console.log` asynchronous. According to [this](http://stackoverflow.com/questions/5127532/is-node-js-console-log-asynchronous/5127757), logging with node should not be a performance concern unless lots of data are being output. Sorry for being too picky, but I'm a bit wary of micro-optimizations (particularly at this stage) and adding too many dependencies.

mizzle-mo:
@alfonsogarciacaro I'll be honest, I am the one being too picky.  I just remember reading that you should collect the strings and log them at once, and to avoid unnecessary logging.  I'll take the async part out.  

I think at the end of the day we need to figure out where the code is spending most of the time during build see if we can improve it.  Babel is taking a small amount of time and Fable is taking a very long time even for a small number of files.  This is fine when you have hot-reload, but with Node/VSCode apps, that 20 second build time is brutal.  Any idea where to start looking to speed things up?

jgrund:
I wonder if using pure js F# compilation would speed build times.

mizzle-mo:
@jgrund What do you mean by "pure js F#" compilation?

ncave:
@jgrund Unfortunately no (or not yet), JS compile times are for now 5x-10x slower than dotnet (on Chrome/Node, and even slower on other browsers), see #739.

mizzle-mo:
@alfonsogarciacaro I added back the received logging..  Are there any docs on contributing hiding out there?

alfonsogarciacaro:
@mizzle-mo Unfortunately there're no hiding contributing docs ;) I have to start writing them from now on. For now, just ask me in PRs or issues and I'll try to provide as much guidance as possible.

About compiler performance, first there's like a warm-up time that is always around 4-5 secs for the simple script (I think it's mainly JIT and assembly loading), after that the time [FCS takes to parse the AST](https://github.com/fable-compiler/Fable/blob/7217e20b7d9ea8f6c0ad465343fd021af4722c61/src/dotnet/Fable.Tools/State.fs#L153-L156), and then Fable AST conversions (the Babel part is actually very fast)

@mizzle-mo For VS Code/Node projects, where you don't use the Webpack dev server, probably you can just use Webpack in watch mode (depending on how you start the build, you can do `dotnet fable webpack --args -w`, or just add `-w` in your npm script calling webpack) so you don't need to do a full recompilation every time.

That said, if you want to improve full recompilations, we can try to do some caching as @Pauan suggested. I guess the easiest way to do it for now would be to keep a copy of the [transformed code](https://github.com/fable-compiler/Fable/blob/7217e20b7d9ea8f6c0ad465343fd021af4722c61/src/typescript/fable-loader/index.js#L99-L100) somewhere and then [when Webpack requires compilation of an F#](https://github.com/fable-compiler/Fable/blob/7217e20b7d9ea8f6c0ad465343fd021af4722c61/src/typescript/fable-loader/index.js#L39) check if a cache exists and if the timestamp is more recent than the F# source.

Pauan:
I filed https://github.com/fable-compiler/Fable/issues/882 for caching.

Even though I'm a member, I can't add Labels. :disappointed:

alfonsogarciacaro:
Now that @Pauan has opened the issue (thanks!) we can track the caching there so we don't mix things. Which tag did you want to add?

mizzle-mo:
Sounds good and thank you.

jgrund:
Instead of reducing verbosity, what about a silent flag?

Then the existing workflow stays the same, but there is an avenue to turn off output from the server if needed.

alfonsogarciacaro:
Latest Fable version (1.0.8) already includes a `--verbose` option. This, together with fable-loader 1.0.2 should make the output less verbose (users can still see the name of the compiled files), so we can close this :)

PR #878, opened 2017-05-07
Create Fable transformer for React Native

### LINKED ISSUE #804: Create Fable transformer for React Native
Related (partly) to #800 

We haven't tried yet Fable 1.0 with RN. It should be possible to just add Webpack bundling as a previous step before RN own packaging. However, it seems it's not difficult to create a custom Fable packager which should make RN hot reloading much faster. @jgrund has already made something similar for [Jest](https://github.com/jgrund/jest-fable-preprocessor), maybe he could help here as well?

Links
- [RN transformer](https://github.com/facebook/react-native/blob/master/packager/transformer.js)
- [RN source map transformer](https://github.com/mhzed/react-native-sm-transformer/blob/master/index.js) as reference
- [Fable Webpack loader](https://github.com/fable-compiler/Fable/blob/master/src/typescript/fable-loader/index.js) 

Things to consider: 
- Unlike Webpack loaders, RN packager only accepts one single transformer, so the custom transformer must load RN transformer and delegate to it for normal JS files. See how it's done in the react-native-sm-transformer.
- As we won't have the webpack config file, we must read Babel options from `.babelrc`. We can use the `getBabelRC` function in RN transformer for that. Remember we must also add [Fable custom Babel plugins](https://github.com/fable-compiler/Fable/blob/master/src/typescript/fable-loader/src/babel-plugins.js) when compiling F# files.

If you want to help, please send a minimal WIP PR (please put the transformer in the `src/typescript` folder) so we can continue the discussion there. Thanks a lot for your help in advance!

cc @et1975 @forki

jgrund:
Fixes #804.


Zaid-Ajaj:
@jgrund Can you please resolve the conflicts? If nothing else is blocking, we can merge this :smile:

Issue #856, opened 2017-04-27
[Discussion] Transitive dependencies


et1975:
Continuing the conversation from gitter about transitive dependencies support @alfonsogarciacaro  is adding, started in https://github.com/fable-elmish/react/issues/1:

> Fable dependencies must be `peer` to prevent npm from duplicating them in subdirectories if there're version conflicts. They cannot be `dev` as they must also be installed in the machine of the final user of your lib.
> About non-Fable `peerDependencies`. Fable will just install everything and check if theres an `.fsproj` file in the root folder of the package. In that case it will add a project reference, the others will be ignored.
> I'm just trying to define the process, we could also have the convention that all Fable libraries are prefixed with fable- for example.
> As far as I understand, `devDependencies` are only installed in the developer machine and not in the library consumer. I guess we could use them as a convention for fable libraries but I don't see the advantage of forcing the npm semantics, you can also have devDependencies that are not Fable libraries.

I'm assuming the situation is as follows:
- fable libs distributed as sources, like anything subject to transpilation has been traditionally a `dev` dependency.
- the guidance for fable libs up to this point, has been "don't list fable dependencies as either `peer` or regular, let the user figure it out and install".
- now that we have `dotnet fable add` we get to examine dependencies of the library being added and potentially trigger installation of its (transitive) dependencies.

Question: should we?
Question: if yes, what is the appropriate section `dev` or `peer`?
Question: what is the predicate for triggering a transitive dependency install?
Question: what is the desired behaviour if installed with `yarn` instead of `dotnet fable add`?

alfonsogarciacaro:
My rough guidelines for this would be:

- Something that's easy to do manually, the automation should be a convenience and not a requirement.
- Require a minimal convention, to reduce the possibilities of wrong configurations (we could also add a `dotnet fable validate` command to check the libraries follow the convention, but again this should be simple).

alfonsogarciacaro:
We could also still consider using Nuget to distribute Fable libraries.

Advantages of using npm:

- Makes it easier to include JS files and dependencies in Fable libraries
- Makes it much easier to push a package with the same contents as a project repo
- Conforms to the usual structure of Webpack projects where everything is contained in same directory (although this can be solved by using `resolve` in Webpack config)
- There are packages that must be distributed by npm necessarily: fable-loader and other tools like Babel or Webpack

Advantages of using Nuget:

- Maybe a better distinction between JS and F# dependencies?
- Can use Paket to manage Fable dependencies

[...]

PR #884, opened 2017-05-08
[WIP] Integrate with Paket

alfonsogarciacaro:
Nothing to see here (yet), this is just to inform the rest of the @fable-compiler/code-maintenance team what I'll be working on next: improve integration with Paket so we can use it to distribute Fable libs (see #856 and https://github.com/fsprojects/Paket/issues/2310). I'll use PRs from now on instead of always committing to master ;)

BTW, [I've pushed a couple of brief documents](https://github.com/fable-compiler/Fable/commit/bfc63764c10270ae12694ff6c698e019ef418b99) to make navigation through the code easier. The idea is to have files like this in the main folders and also something similar in the form of a comment on top of the main source files. The READMEs should be self explanatory but not to detailed to prevent they lose sync with the actual code very soon.

Please ask me if you've any comment/suggestion.


Pauan:
How will JS libraries be handled? Yarn? Some sort of integration with Paket?

alfonsogarciacaro:
@Pauan JS libraries will work the same as it's now, but we will recommend yarn as the package manager for JS and we will use it in samples and templates from now on :+1:

Pauan:
@alfonsogarciacaro So I guess the recommended way would be to put in a `postinstall` hook in `package.json` which calls Paket? So that way the user only has to do `yarn install` to have fully working code.

Also, if we're shipping DLL files, doesn't that mean that libraries will have to be recompiled whenever the Fable compiler changes?

alfonsogarciacaro:
@Pauan I guess we can do that. Actually, if there's already a paket.lock file it's not necessary to call Paket directly, just `dotnet restore` works. I prefer no to hide too many things from the user, as they will still need to use Paket directly to add new Fable dependencies, but if yarn postinstall is useful for others we can start using it in the template/samples.

About DLLs, they will be generated by the F# compiler and they are just used for type checking and editor support so they won't be affected by new Fable versions :)

Pauan:
@alfonsogarciacaro Great! That's better than I was expecting.

I haven't used Paket before, so I'll need to figure out some stuff in order to migrate `fable-getting-started` to Paket.

alfonsogarciacaro:
I think this is ready to merge, I'll publish a new version with the changes right away. Crossing fingers so it doesn't destroy everything. I'll try to write a blog post about the new Paket integration tomorrow :+1:

MangelMaxime:
One question through

Using npm we was able to use a local version of the package via npm link to have a simple development process. You make change into the lib git repo and the change are reflected into your projects that use the lib.

Is there an equivalent for nuget ?

To give a better understanding:
I have Hink as the library and Showcase as an application which reference this library for testing it. Actually, I do npm link into Hink and npm link hink into Showcase and that's it.

alfonsogarciacaro:
With the Paket integration we're moving closer to the F#/.NET mindset, at least for pure F# development (before compiling to JS), and I think this is the typical use case where you'll just have a solution (or workspace) with the two projects and just use project references between them.

If you still want to use explicit package references, I'm not sure if Paket can install some packages from local sources and others from remote, maybe @forki can answer to that?

forki:
Sure paket can install from local and remote sources. What's the concrete
question?

Am 16.05.2017 10:01 vorm. schrieb "Alfonso Garcia-Caro" <
notifications@github.com>:

MangelMaxime:
I never created a lib for F# or C#.
 
The question is how can I develop a library and an application (referencing this lib) at the same time.

@forki What you are saying is I can install via paket using local sources but that means I need to generate a new nuget file each time I made a change.

@alfonsogarciacaro , is saying I can use project references so I can access directly the source of the library into the application. Like so, when I make a change into the library, the application will directly have the latest version of the code.

forki:
Ah I misread. Project references are managed without paket.

Am 16.05.2017 1:34 nachm. schrieb "Maxime Mangel" <notifications@github.com

MangelMaxime:
My problem with package reference is it's make you make a change in your application and if you are not carefull you will publish this change into the git repo.

Where using `npm link` make no diff in the project. It's just npm making a symlink to the lib folder source.

forki:
Take a look at paket's paket.local file. Maybe that's what you are looking
for?

Am 16.05.2017 17:12 schrieb "Maxime Mangel" <notifications@github.com>:

MangelMaxime:
Oh looks promising yes :) thanks you.

PR #888, opened 2017-05-09
[Dont Merge] Failing test for `TypeError: _arg1.value is not a function`

et1975:
Stripped down version of elmish url parser to reproduce the problem.

It succeeds when compiled for .NET and fails in JS.

### DIFF
--- a/src/tests/Main/ElmishParserTests.fs
+++ b/src/tests/Main/ElmishParserTests.fs
@@ -1,0 +1,200 @@
+[<Util.Testing.TestFixture>]
+module Fable.Tests.ParserTests
+
+
+[<RequireQualifiedAccess>]
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Option =
+
+    let tuple a b =
+        match (a,b) with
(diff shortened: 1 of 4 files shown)


alfonsogarciacaro:
Thanks a lot, you've saved me a lot of time with the repro code! As expected, this was another side-effect of the lambda uncurrying optimization in Fable 1.0. As `State.value` type was just a generic parameter, Fable couldn't infer the expected arity so the partial application later is not resolved correctly.

I've fixed it by creating a dynamic curried lambda (`CurriedLambda` in fable-core) whenever you pass a lambda to something expecting just a generic parameter (see `AST.Fable.Util.ensureArity` in adc35596f16ae0259f3c26f783907975a77d2815), this will check the number of applied arguments in runtime.

I just hope we don't have many other uncovered edge cases 🙏 

et1975:
Thank you!
is it released yet?

alfonsogarciacaro:
I just pushed the packages, I guess they'll take a few minutes to be available. You should be able to update the dotnet-fable CLI tool with `dotnet restore --no-cache` :)

PR #898, opened 2017-05-12
Reintroduce process.platform property - fixes #897

### LINKED ISSUE #897: Updated Node API is missing process.platform
### Description

On Node's global `process` there is a property called `platform` (see [docs](https://nodejs.org/api/process.html#process_process_platform)). It's missing from the refactor of the Fable Node API

### Repro code

```fsharp
open Fable.Import.Node.Globals

``process``.platform // undefined
```

### Related information 

* fable-core version: Fable.Core 1.0.0-narumi-915

cc @jgrund I believe you worked on this. I don't have time to make a PR at the moment, but will try tomorrow if you don't get to it first.

inosik:


alfonsogarciacaro:
Great, thank you!

PR #900, opened 2017-05-12
Issue #897: Use Platform enum for process.platform

jmmk:
Instead of using a string, use the `Platform` `StringEnum` from `Fable.Import.Node.Base`


alfonsogarciacaro:
Good one, thank you!

PR #901, opened 2017-05-12
Add Arch type to Node.js bindings

inosik:
Values taken from https://nodejs.org/api/os.html#os_os_arch.

/cc @jmmk


alfonsogarciacaro:
Nice one, thanks!

Issue #881, opened 2017-05-07
Update samples to Fable 1.0-beta


Zaid-Ajaj:
Following samples need to be updated to Fable 1.0 beta, the list will help us keep track of what is already done. 
### Browser 
- [x] Canvas [sample](https://github.com/fable-compiler/Fable/tree/master/samples/browser/canvas)
- [x] D3 Map [sample](https://github.com/fable-compiler/Fable/blob/master/samples/browser/d3map/d3map.fsx)
- [x] Funsnake [sample](https://github.com/fable-compiler/Fable/tree/master/samples/browser/funsnake)
- [x] Hokusai (julia fractals) [sample](https://github.com/fable-compiler/Fable/blob/master/samples/browser/hokusai/hokusai.fsx)
- [x] Lindenmayer-Systems (lsystems) [sample](https://github.com/fable-compiler/Fable/tree/master/samples/browser/lsystem)
- [x] Mandelbrot [sample](https://github.com/fable-compiler/Fable/tree/master/samples/browser/mandelbrot)
- [x] Pong sample
- [x] React-TodoMVC (redirect link to the elmish-react version?)
- [x] Mario [sample](https://github.com/fable-compiler/Fable/blob/master/samples/browser/mario)
- [x] Ozmo sample
- [x] Pacman [sample](https://github.com/fable-compiler/Fable/tree/master/samples/browser/pacman)
- [x] Pixi sample
- [x] Samegame
- [x] Vue-TodoMVC 
- [x] ReactRedux-TodoMVC (same as above?)
- [x] WebGL-Terrein
### Node
- [ ] Console app
- [ ] File System 
- [ ] Express

I will work on them starting from next week. 
I suggest to make these samples _independent_ with a README.md for every sample on how to install, run etc. 
If you want to help out, just remember to reference this issue when you PR ;)

alfonsogarciacaro:
The React-TodoMVC and D3 samples are already updated, I marked them in the list above :+1: Not sure why but I had the updated D3 sample in a different branch 😅 

I would keep the React-TodoMVC for users that would like to see how to use _raw_ React without any framework. The Redux version was mainly for the Redux Devtools, but now that fable-elmish also supports it we can replace the sample 😄

Zaid-Ajaj:
@fable-compiler/documentation Replace ReactRedux-TodoMVC with an Elmish sample?

[...]

PR #904, opened 2017-05-14
Updated Vue sample

Zaid-Ajaj:
Alright, for this sample I used 1.x version of Vue. the sample doesn't work with v2.0+ becuase they added some rendering functions to create "vue"s either in code using a `createElement` function much like the one from React or using `.vue` files which require another loader in webpack and obligates you to write javascript with html in that file which I think we don't support (nor we should). 

cc @fable-compiler/documentation please review.

The code all in all looks (to me at least) not really clean because of the overused dynamic programming and I am not sure how someone is supposed to build something else based on this. 

Just redirect to elmish when someone clicks on the vue sample 😆 

#881


alfonsogarciacaro:
This was one of the first samples I wrote when I was interested in the Vue framework before focusing on React. It may be good to have some samples of using dynamic programming to interact with JS libraries, but I don't mind removing the sample to focus on elmish :)

Zaid-Ajaj:
> I don't mind removing the sample to focus on elmish :)

Nah I think it is still nice to see that Fable is able to use such javascript-specific library :smile:

PR #905, opened 2017-05-14
Update WebGL Terrain Sample

Zaid-Ajaj:
For #881 
cc @fable-compiler/documentation please review. 

Didn't change much of code, only added the `externals: { "three": "THREE" }` to webpack config. Other than that only had to change:
```fs
let FirstPersonControls: JsConstructor<Three.Camera * Browser.HTMLElement, IFirstPersonControls> =
    importDefault "./js/FirstPersonControls.js"
```
into
```fs
let FirstPersonControls: JsConstructor<Three.Camera, Browser.HTMLElement, IFirstPersonControls> =
    importDefault "./js/FirstPersonControls.js"
```
Because otherwise the external function `FirstPersonControls` recieved one parameter as an array (tuple in F#) but it needed two parameters `Three.Camera` and `Browser.HTMLElement`

Also I am using local bindings for `Fable.Import.Three.fs`


alfonsogarciacaro:
I think this was still using the first version of JsConstructor which used tuples instead of multiple generic arguments. Thanks for fixing it!

Is it necessary to add the Three.js bindings? Why not use the package?

Zaid-Ajaj:
@alfonsogarciacaro 
> Is it necessary to add the Three.js bindings? Why not use the package?

Not really, I just didn't want to use the old npm dependency to only reference one file since we will be distributing bindings to nuget anyways

PR #906, opened 2017-05-14
Node console app sample

Zaid-Ajaj:
A simple console app sample for Node. 

I couldn't find `console` in a place other than `Fable.Import.Browser` so I chose to use `[<Emit>] ` for `console.log` @fable-compiler/code-maintenance may know a node-native alternative.

cc @fable-compiler/documentation please review 

#881

### DIFF
--- a/samples/node/ConsoleApp/README.md
+++ b/samples/node/ConsoleApp/README.md
@@ -1,0 +1,21 @@
+# Node Console App
+This sample demonstrates a simple console app to be run in node. 
+
+Restore dependencies
+--------------------
+```
+npm install
+dotnet restore
+```
+Build the app
+-------------------
+This will start the Fable server and invoke webpack to build and bundle the app.
+```
+dotnet fable npm-run build
+```
+Run with Node
+-------------
+Run the generated file `app.js` in the build directory
+```
+node build/app.js
+``` 
--- a/samples/node/ConsoleApp/ConsoleApp.fsproj
+++ b/samples/node/ConsoleApp/ConsoleApp.fsproj
@@ -1,0 +1,14 @@
+<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard1.6</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="src/App.fs" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="FSharp.NET.Sdk" Version="1.0.*" PrivateAssets="All" />
+    <PackageReference Include="FSharp.Core" Version="4.1.*" />
+    <PackageReference Include="Fable.Core" Version="1.0.0-narumi-*" />
+    <DotNetCliToolReference Include="dotnet-fable" Version="1.0.0-narumi-*" />
+  </ItemGroup>
+</Project>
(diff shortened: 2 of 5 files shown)


ncave:
@Zaid-Ajaj printfn maps to console.log ...

Zaid-Ajaj:
@ncave Totally forget about it 😅 thanks!

samples/node/ConsoleApp/README.md
```diff
@@ -0,0 +1,21 @@
...
+--------------------
+```
+npm install
+dotnet restore
+```
+Build the app
+-------------------
+This will start the Fable server and invoke webpack to build and bundle the app.
+```
+dotnet fable npm-run build
```
jgrund: Do we want to mention rollup here? I think there may be interest in using it for node apps.

Zaid-Ajaj: Probably not for this sample, trying to keep things (code, mental overhead etc.) as minimal as possible. Could use it for a bigger sample where dead code elimination has more impact

alfonsogarciacaro:
There's already a Node sample with a simple static server. Maybe we could use that one instead of having a single print statement?

Zaid-Ajaj:
@alfonsogarciacaro There will be more of node samples to come but with incremental difficulty. this is for people that wonder: "What would a console app look like with this Fable thing?" and be amazed it is just their normal F# console app without the explicit [<EntryPoint>]. 

However, if you find it too simplistic I will close the PR :smile: 

While we are at it, why not use this for the basic Node template to begin with? no webpack-dev-server, no browser (canvas etc.) stuff with the addition of webpack's option `target: "node"` so that webpack does not import node-native modules (`fs`, `path` etc.) from `node_modules`

samples/node/ConsoleApp/README.md
jgrund: Webpack will also do dead code elimination. 

I'm more wondering why someone would prefer webpack over rollup on a node only app in general? I.E. why not just use rollup as the default for node?

Pauan: Webpack's dead code elimination *suuuuuuuucks*.

Rollup is simpler than Webpack, and the bundle output is **much** smaller. It's just that Webpack has more features than Rollup, and Webpack is more popular than Rollup.

Zaid-Ajaj: I am no expert on webpack nor rollup. The reason I chose webpack here is becuase it seems to be the default way we do stuff in Fable so the choice is arbitrary. However, it would be nice if could send a PR with a similar sample using Rollup

alfonsogarciacaro:
Actually, it's possible to use `[<EntryPoint>]`  with node apps and you'll have access to the arguments. There was an example of such a project compiling both as a .net and node command line app, but it went missing for some reason.

> However, if you find it too simplistic I will close the PR 😄

I don't think we should close the PR, I'd like just to put more meat (as we say in Spanish) to the sample body, so users have a better idea of how a node app with F#/Fable looks like (it's always easy to just delete the code and start from scratch). Just taking the code from the static server sample would work, but we can use something else.

> While we are at it, why not use this for the basic Node template to begin with? 

This is a very good idea! We probably should start a new repo to host the templates. We still need to fix the Node bindings in Fable 1.0, but after that it'd be very nice to create the template.

Zaid-Ajaj:
LOL, closing this for lack of meat 😆 I am already working on another better sample. Just waiting for node bindings to update becuase I don't want to use paket lock files with out-dated Fable.Core versions.

PR #909, opened 2017-05-15
Add SymlinkType string enumeration

inosik:
Values from https://nodejs.org/dist/latest-v6.x/docs/api/fs.html#fs_fs_symlink_target_path_type_callback.

I'm not sure about the type name and if this is the right place to add it, though.

/cc @jgrund


alfonsogarciacaro:
Thanks!

PR #916, opened 2017-05-16
Use Paket + proj file for tamplte

### LINKED ISSUE #915: Update Fable.Template (simple) to use Paket

MangelMaxime:
Fix #915 

@alfonsogarciacaro as you updated Elmish template I return you the favor :D

Also, I switched the .nuspec to .proj file.


MangelMaxime:
@alfonsogarciacaro As the samples are currently broken #921 
Perhaps it could be nice to publish this version of the template.

I know it will have another update as you started for Elmish template. And the user will need to update the configuration when updating to the next version. But I believe the changes are simple. :)

alfonsogarciacaro:
Thanks!

PR #917, opened 2017-05-16
Add Url and Querystring modules

jgrund:
Signed-off-by: Joe Grund <grundjoseph@gmail.com>

### DIFF
--- a/src/dotnet/Fable.Core/Fable.Core.fsproj
+++ b/src/dotnet/Fable.Core/Fable.Core.fsproj
@@ -18,6 +18,8 @@
     <Compile Include="Import/Fable.Import.JS.fs" />
     <Compile Include="Import/Fable.Import.Browser.fs" />
     <Compile Include="Import/Node/Base.fs" />
+    <Compile Include="src/Url.fs" />
+    <Compile Include="src/Querystring.fs" />
     <Compile Include="Import/Node/Events.fs" />
     <Compile Include="Import/Node/Buffer.fs" />
     <Compile Include="Import/Node/Stream.fs" />
--- a/src/dotnet/Fable.Core/Fable.Core.fsproj
+++ b/src/dotnet/Fable.Core/Fable.Core.fsproj
@@ -18,8 +18,8 @@
     <Compile Include="Import/Fable.Import.JS.fs" />
     <Compile Include="Import/Fable.Import.Browser.fs" />
     <Compile Include="Import/Node/Base.fs" />
-    <Compile Include="src/Url.fs" />
-    <Compile Include="src/Querystring.fs" />
+    <Compile Include="Import/Node/Url.fs" />
+    <Compile Include="Import/Node/Querystring.fs" />
     <Compile Include="Import/Node/Events.fs" />
     <Compile Include="Import/Node/Buffer.fs" />
     <Compile Include="Import/Node/Stream.fs" />
--- a/src/dotnet/Fable.Core/Fable.Core.fsproj
+++ b/src/dotnet/Fable.Core/Fable.Core.fsproj
@@ -28,6 +28,7 @@
     <Compile Include="Import/Node/ChildProcess.fs" />
     <Compile Include="Import/Node/Process.fs" />
     <Compile Include="Import/Node/Globals.fs" />
+    <Compile Include="Import/Node/Exports.fs" />
     <Compile Include="Fable.Core.JsInterop.fs" />
     <Compile Include="Fable.Core.Extensions.fs" />
   </ItemGroup>
--- a/src/dotnet/Fable.Core/Import/Node/Querystring.fs
+++ b/src/dotnet/Fable.Core/Import/Node/Querystring.fs
@@ -1,4 +1,3 @@
-[<AutoOpen>]
 module rec Fable.Import.Node.Querystring
 
 open Fable.Core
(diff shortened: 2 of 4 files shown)


src/dotnet/Fable.Core/Import/Node/Url.fs
```diff
@@ -0,0 +1,28 @@
...
+type Globals =
+    abstract parse: urlStr: string -> Url<string>
+    abstract parse: urlStr: string * ?parseQueryString: bool -> Url<obj>
+    abstract parse: urlStr: string * ?parseQueryString: bool * ?slashesDenoteHost: bool -> Url<U2<string, obj>>
+    abstract format: url: Url<string> -> string
+    abstract format: url: Url<obj> -> string
+    abstract resolve: from: string * ``to``: string -> string
+
+[<Import("*","url")>]
+let exports: Globals = jsNative
```
Zaid-Ajaj: This will be really hinder the readability of code because of having to write `Url.exports.someting` or `Querystring.exports.somethingElse` everytime. someone coming from Node expects `url.something` and `queryString.somethingElse` which are a lot more logical given that "exports" is a way to originize js modules not F# modules

jgrund: So, the later example is actually `Url.url.something`. 

I think the issue is really the second dotted name.

Zaid-Ajaj: Not if you open the full module: open Fable.Import.Node.Url 

later in the code just: url.something

you could do even better and have all these submodules auto-open on Fable.Import.Node then have url.something or fs.readFileSync readily available

jgrund: The last option is probably the way to go I suppose. I'll update.

src/dotnet/Fable.Core/Fable.Core.fsproj
```diff
@@ -18,6 +18,8 @@
     <Compile Include="Import/Fable.Import.JS.fs" />
     <Compile Include="Import/Fable.Import.Browser.fs" />
     <Compile Include="Import/Node/Base.fs" />
+    <Compile Include="src/Url.fs" />
+    <Compile Include="src/Querystring.fs" />
```
inosik: These paths are incorrect.

jgrund: oops 😅 , fixing

src/dotnet/Fable.Core/Import/Node/Querystring.fs
```diff
@@ -0,0 +1,20 @@
+[<AutoOpen>]
```
alfonsogarciacaro: As now the types are not nested in a `*_types` submodule, maybe we don't need `AutoOpen` here (also in Url module below). This means that all types from all modules will be globally accessible when opening `Fable.Import.Node` which is probably not desirable and may lead to confusion as some types will shadow each other.

jgrund: Sure. In that case, what should we export for module globals?

I think we have some differing opinions of `url`, `globals` etc...

src/dotnet/Fable.Core/Import/Node/Querystring.fs
```diff
@@ -0,0 +1,20 @@
...
+    abstract decodeURIComponent: (string -> string) option with get, set
+
+type Globals =
+    abstract stringify: obj: 'T * ?sep: string * ?eq: string * ?options: StringifyOptions -> string
+    abstract parse: str: string * ?sep: string * ?eq: string * ?options: ParseOptions -> 'T
+    abstract escape: str: string -> string
+    abstract unescape: str: string -> string
+
+[<Import("*", "querystring")>] 
+let querystring: Globals = jsNative
```
alfonsogarciacaro: As discussed previously, we can rename this to `Exports` (maybe also the type becomes `IExports` instead of `Globals`) then users would do this (if the AutoOpen attribute is removed):

```fsharp
open Fable.Import.Node

QueryString.Exports.escape("my string")
```

We can also try to avoid `Exports` (or `Globals`) as in the old bindings. A solution without having to put everything in a single file, would be to add an extra file at the end with values like:

```fsharp
[<AutoOpen>]
module Fable.Import.Node.Exports

open Fable.Core

let QueryString: QueryString.Globals = jsNative
let Url: Url.Globals = jsNative
```

Then, after opening `Fable.Node.Import`, when users type `QueryString` they'll have both access to the types and the functions. What do you think?

Zaid-Ajaj: I like that last suggestion, having to remember only one namespace to open. Also :+1: for capitalizing the sub-modules

inosik: I like the last suggestion, too. But I wonder if we need `Exports`, if it gets `AutoOpen`ed anyways?

alfonsogarciacaro: @inosik If I use `module Fable.Import.Node` the F# compiler will complain that there's a module and a namespace (e.g. in `Fable.Import.Node.QueryString`) with the same name in the assembly.

jgrund: @alfonsogarciacaro Cool. Having a single export file was something I wanted to do, but couldn't figure out how to do so. 

I'll update these bindings, and the rest now.

jgrund:
Actually, we'd probably want the exports to be lowercase so they don't shadow the types.

alfonsogarciacaro:
This doesn't seem to be a problem: types/modules shadow each other, but apparently you can use both a type/module and a value with the same at the same time. You can test the following code in a F# script:

```fsharp
type List = class end
// Here you cannot access to functions in List module any more

let Seq: System.IDisposable = failwith "foo"
// Here you can still access functions in Seq module
// You can also access value members: Seq.Dispose()
```

ncave:
@alfonsogarciacaro Does it make sense in your opinion to move all the jsNative imports out of Fable.Core into their own assembly, to lower the release churn on Fable.Core?

jgrund:
I'm going to close this, I have an updated PR that includes this and the export fixes at #929.

Let's continue coversation there.

PR #918, opened 2017-05-17
Now throws exception when converting invalid strings

### LINKED ISSUE #725: Missing Exception when casting bad int
## Explanations 
When casting a malformated string into an integer we don't get an exception. [REPL](https://goo.gl/lM8m2c)

When executing the same code in Fsharp interactive I got System.FormatException: Le format de la chaîne d'entrée est incorrect.

## Code to reproduce

```fsharp
printfn "%i" (int "5")
printfn "%i" (int "maxime")
```

@alfonsogarciacaro anwser on gitter:

> @MangelMaxime Yeah, we're just using JS native Number.parseInt, but we do make a isNan check when using System.Int32.TryCheck. It makes sense to follow F# semantics and throw an error if the parsing is not sucessful, can you please open an issue?

Pauan:
Fixes #725

### DIFF
--- a/samples/browser/server/index.js
+++ b/samples/browser/server/index.js
@@ -2767,7 +2767,7 @@ function fresh$2(req, res) {
   if (!modifiedSince && !noneMatch) return false;
 
   // check for no-cache cache request directive
-  if (cc && cc.indexOf('no-cache') !== -1) return false;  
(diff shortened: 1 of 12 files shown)


alfonsogarciacaro:
Awesome, thanks a lot for this! :clap:

PR #919, opened 2017-05-17
Redirect dotnet-fable CLI to Fable.Compiler managed by Paket

alfonsogarciacaro:
After the Paket integration I still wasn't happy with the fact that dotnet-fable CLI tool wasn't included in `paket.dependencies` so the risk of version mismatches between Fable.Compiler and Fable.Core remained high. Also, because nuget puts the DotNetCLIReference tools in a folder outside the workspace, Webpack had troubles in resolving references from the fable-core files (located now in the Nuget package).

The solution I came up with is to make dotnet-fable as thin as possible and simply redirect the commands to the Fable.Compiler package now managed by Paket (it must be included in `paket.dependencies` together with Fable.Core, even if it's not actually referenced by the user project).

Right now I'm running the actual Fable compiler as a subprocess. I explored first to load the assemblies and use reflection to call the main method. It worked but the problem is I had to track all the dependencies and load the assemblies one by one. The problem with missing assemblies happens also when running directly Fable.Compiler.dll within the `lib` folder. To avoid that, I'm also packing the _published_ Fable (that is, including all references) in the `fable` folder. This is probably not the most elegant solution (I get a lot of warnings saying I shouldn't put assemblies outside the `lib` folder) so please tell me if you have any other idea.

Could you please have a look, @forki and @ncave? I guess some adjustments will be also necessary 
for the REPL because of the restructuring.

> I just noticed that running `dotnet fable` takes several seconds at the beginning. I wonder why is this, in my local tests before publishing dotnet-fable commands started faster.


src/dotnet/Fable.Compiler/Utils.fs
```diff
@@ -12,6 +12,10 @@ type ICompilerState =
     abstract GetOrAddEntity: string * (unit->Fable.Entity) -> Fable.Entity
     abstract GetOrAddInlineExpr: string * (unit->InlineExpr) -> InlineExpr
 
+module Constants =
+    let [<Literal>] DEFAULT_PORT = 61225
+    let [<Literal>] VERSION = "1.0.0-narumi-920"
```
ncave: @alfonsogarciacaro Is that version label used somewhere?

ncave:
@alfonsogarciacaro I'm probably not getting what the goal is, can you try explaining it again? Is there a better way of referencing binary dependencies than including them directly in the repo, wouldn't that increase the maintenance cost of keeping them up to date with every build?

src/dotnet/Fable.Compiler/Utils.fs
alfonsogarciacaro: Yes, this is used when you type `dotnet fable --help`

alfonsogarciacaro:
@ncave So basically we have a problem to sync Fable.Core (referenced by the project) and Fable compiler. Fable.Core is now managed by Paket but if the compiler is added in a `DotNetCLIReference` tag, it's not controlled by Paket so it's up to the user to be sure.

Let's say we add a new helper to Fable.Core, but if it's not included in the Replacements module in the compiler, compilation will fail. We could make Fable.Core self contained (so the .dll contains its own replacements) but this will make the REPL more complicated (as JS Fable cannot load assemblies) and we would still have the same problem when updating the AST as the AST in Fable.Core and the compiler may differ.

We could add a runtime Fable.Core version check in the compiler as before, though people seem to not like it very much and it's a pity that users need 3 files for version management (paket.dependencies, package.json and *.fsproj).

forki:
I think one important step for paket is to manage dotnetclireference. Can
you please create an issue in paket and describe what's missing. /cc
@enricosada @cloudroutine

Am 17.05.2017 8:56 vorm. schrieb "Alfonso Garcia-Caro" <
notifications@github.com>:

alfonsogarciacaro:
Ok, it's probably to keep the CLI tool for now and wait until Paket can manage it (I'll open an issue). I'm closing this and merging #923 instead :+1:

PR #929, opened 2017-05-18
Update exports

jgrund:


jgrund:
This is passing with the new changes. 

There are a few nits worth discussing.

1. The module name may be the same as the type name. example `Buffer.Buffer`. Should we prefix types with `I` ? This would fix the ambiguity and remove a bunch of linter errors.

2. Need to figure out how to structure the Base shared types / Globals. This could be done in another patch.

jgrund:
Please review @fable-compiler/code-maintenance

alfonsogarciacaro:
I'm fine with any of the two solutions: either having the same name as types in the TS declaration of adding an `I` prefix to the interfaces 😄

jgrund:
Ok, let's just leave it as is for now then. Can revisit in another patch as needed.

Issue #932, opened 2017-05-19
How to reference local build?


jmmk:
In https://github.com/fable-compiler/Fable/pull/884, @MangelMaxime asked about using a local build of Fable, and @forki mentioned that it might be possible to use [paket.local](https://fsprojects.github.io/Paket/local-file.html) for this.

In the documentation they give an example based on a `NuGet` target in the FAKE script which generates a NuGet package that can then be referenced. Fable has no such target, and the NuGet build seems to be nested inside `pushNuget`, which is part of the `PublishPackages` target.

Now that Fable is using Paket, what's the best way to install a custom build into a project (for testing purposes or otherwise)?

mizzle-mo:
I think it would be helpful to put together some contributing docs and include this with the move to Paket.

alfonsogarciacaro:
I think in #884 @MangelMaxime was talking about referencing other projects locally as if they were packages. But if you want to use a version of the Fable compiler build locally you can take a look at the [QuickTest](https://github.com/fable-compiler/Fable/blob/8253562eedf2a4abb1d498b747f070d085260ecf/src/tools/QuickTest.fsx). The only thing you need to do is [call directly the `dotnet-fable.dll`](https://github.com/fable-compiler/Fable/blob/8253562eedf2a4abb1d498b747f070d085260ecf/build.fsx#L404) and if you also want to a local fable-core, [pass the path to the Fable options](https://github.com/fable-compiler/Fable/blob/8253562eedf2a4abb1d498b747f070d085260ecf/src/tools/rollup.config.js#L17) :+1:

jmmk:
@alfonsogarciacaro I'm not sure I understand. I'll try to clarify exactly what I'm asking:

- I have a `paket.dependencies` that includes `nuget Fable.Core` and a `paket.references` that includes `Fable.Core`

- I have cloned the Fable repository, and I want to make a change to `Fable.Core` and test that change in the context of my project

What's the best way to: 

1. build `Fable.Core` locally

2. Reference this local `Fable.Core`in another project?

alfonsogarciacaro:
Ok, that's more or less the same :+1:

> build Fable.Core locally

After cloning the project, if you touch any of the dotnet files, run `build FableTools`. If you touch any of the fable-core Typescript/JS files, run `build FableCoreJs` (in the latter case remember to pass the fableCore path to the loader options).

> Reference this local Fable.Corein another project?

[...]

PR #933, opened 2017-05-19
Use Console type for console global

jmmk:
Have not verified that this works properly - I'm not sure how to test locally now that Fable is using Paket (See https://github.com/fable-compiler/Fable/issues/932)


alfonsogarciacaro:
This should be fine, so I'll just merge it, thanks! :+1: In any case, you can still build and reference Fable.Core.dll directly [as the tests do](https://github.com/fable-compiler/Fable/blob/0b241759baf51285dd62d15e24b03caf7c3cebc5/src/tests/Main/Fable.Tests.fsproj#L7-L9) without using Paket :)

Issue #800, opened 2017-04-04
Compile JS without bundling


Neftedollar:
Fable 1.0 beta
Subj

Pauan:
Hmm, I wonder if [`rollup-loader`](https://github.com/egoist/rollup-loader) can be used to reduce the amount of boilerplate.

alfonsogarciacaro:
Fable 1.0 is delegating many of the functions implemented (with flaws) in previous versions of Fable like watching, resolving dependencies, etc. However, it shouldn't be difficult for the community to provide alternatives to `fable-loader` that communicate with Fable server independently. Could you give more details of your use case and what would you like the workflow to be, @Neftedollar? If you just want to hide the webpack config file, this would be as easy as creating a JS script that calls webpack under the hood with a default configuration.

Neftedollar:
@alfonsogarciacaro I understand. 
I want to use fable with meteorjs. It has it's own filepreprocessing.(pack/unpack, deploying resolve dependencies etc. without webpack)
Maybe I should go another way and use Fable-Compiler inside meteorjs workflow. 

Or, fireloop has is it's own workflow too.

Neftedollar:
https://guide.meteor.com/build-tool.html

jgrund:
@Neftedollar Fwiw I did something like this with a Jest preprocessor: https://github.com/jgrund/jest-fable-preprocessor

As long as you can take the output from the fable server and push it into Babel, you should be able to get compilation without webpack working.

alfonsogarciacaro:
@Neftedollar In this case, probably the best thing to do would be [to create a custom build plugin for Meteor](https://guide.meteor.com/build-tool.html#writing-build-plugins), but unfortunately the documentation seems to be very thing on this. I also had a look at the Typescript Meteor package and looks quite complicated.

It wouldn't be very difficult to create a standalone version of the fable-loader similar to Fable 0.7: a command to watch and/or compile F# files. The only thing I would like to avoid because it gave many headaches is putting the compiled Js files in a different directory. This is very tricky for files (both F# and JS) outside the project folder as it is not trivial to calculate the most _efficient_ structure and resolve references from other files.

If you don't mind having the compiled JS files side by side with the F# files, I can try to implement something :+1:

> BTW, @jgrund that's very advanced bit ofJS code, it wasn't easy to follow it ;)

Neftedollar:
@alfonsogarciacaro  I've found more useful text 
[...]

PR #935, opened 2017-05-20
Compile JS without bundling

ncave:
[WIP] Initial attempt at #800. It's still very raw, for now it just dumps the compiled .js directly in the outDir.
Please share ideas on what the folder structure under outDir should look like.


src/typescript/fable-splitter/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    options = options || {};
+
+    const babelOptions = options.babel || {};
+    babelOptions.plugins = customPlugins.concat(babelOptions.plugins || []);
+
+    const fableOptions = options.fable || {};
+    fableOptions.path = fableOptions.path || "";
+    fableOptions.define = fableOptions.define || [];
+    fableOptions.plugins = fableOptions.plugins || [];
+    fableOptions.fableCore = fableOptions.fableCore || "../fable-core";
```
alfonsogarciacaro: fable-core JS files are now distributed in the Fable.Core nuget package so we can just use `null` as the default value.

src/typescript/fable-splitter/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    return parsed;
+}
+
+async function transformLoopAsync(path: string, options: FableCompilerOptions) {
+    const result: any = await transformAsync(path, options);
+    if (result && result.ast && result.ast.program) {
+        const dir = Path.dirname(path);
+        const files = result.ast.program.body || [];
```
alfonsogarciacaro: Maybe add `ensureArray(result.ast.program.body).filter(x => x.type === "ImportDeclaration")` to make the intention clearer? I took me a while to understand this line :)

src/typescript/fable-splitter/index.ts
```diff
@@ -0,0 +1,172 @@
...
+export type FableOptions = {
+    path?: string;
+    define?: string[];
+    plugins?: string[];
+    fableCore?: string;
+    declaration?: boolean;
+    typedArrays?: boolean;
+    clampByteArrays?: boolean;
+    extra?: any;
+};
```
alfonsogarciacaro: We need to add an option here to change the extension of the F# files, at the moment I think they remain ".fs" in the imports.

src/typescript/fable-splitter/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (parsed) {
+        console.log(`Fable compiled: ${fullPath}`);
+
+        // if not a .fsproj, save as JS file
+        if (!FSPROJ_EXT.test(path)) {
+            // TODO: create common folder structure
+            const fileName = Path.basename(fullPath).replace(FSHARP_EXT, ".js");
+            const filePath = Path.resolve(Path.join(options.outDir, fileName));
+            const fileDir = Path.dirname(filePath);
+            if (!fs.existsSync(fileDir)) { fs.mkdirSync(fileDir); }
```
alfonsogarciacaro: I think this will fail if the parent dir doesn't exist either. [This function](https://github.com/fable-compiler/Fable/blob/6c8b94e5e1b7e30669a96646859f88f24376d934/src/fable/Fable.Client.Node/ts/lib.ts#L270-L281) should fix the problem.

alfonsogarciacaro:
Awesome work @ncave, thanks a lot for me. Actually, it didn't occur to me that we could just use the imports in the Babel AST to resolve the dependency tree 😅 

About the outDir. That was one of my pain points in Fable 0.7 as I tried to make something that didn't fail (with hashes to avoid conflicts and such) but then many users didn't find the output structure intuitive. My take now would be to make something simple that works for 99% cases. For example:

- Add an option so the user can specify the base directory. If none, assume it's the entry file (usually the .fsproj) directory.
- Assume that all files external to that directory belong to `packages` folder (maybe raise a warning if not so the user can adjust the baseDir?) so if I have a file in `packages/Fable.Elmish/fable/program.fs`, copy it to `[OUT_DIR]/Fable.Elmish/program.fs` (note the `fable` folder has been removed, as it should be common for all Fable packages).

src/typescript/fable-splitter/index.ts
ncave: Sure, although making code understandable is a slippery slope, it will only make the bugs stand out :)

src/typescript/fable-splitter/index.ts
ncave: No need for option IMO.

src/typescript/fable-splitter/index.ts
ncave: Will do.

src/typescript/fable-splitter/index.ts
ncave: x.type === "ImportDeclaration" not enough, went with x.source != null

src/typescript/fable-splitter/index.ts
ncave: Either way should work.

ncave:
@alfonsogarciacaro Thanks for the suggestions. You're right, it's a bit of a pain come up with a folder structure that can handle all the requirements.

I took your advice to heart and went with the easiest thing that works:
- no baseDir, just outDir.
- flat folder structure (one level deep only) for everything.

I know it's a bit controversial, but give it a try and tell me what you think.

More todo:
- handle duplicate files.
- commonjs modules.

src/typescript/fable-splitter/index.ts
```diff
@@ -0,0 +1,205 @@
...
+    const relPath = Path.join(dir, path);
+    const fullPath = Path.resolve(relPath);
+    const newPath = Path.join("..", flatPath(fullPath));
+    return newPath.replace(/\\/g, "/");
+}
+
+function getImportPaths(ast: any): string[] {
+    const decls = ast && ast.program ? ensureArray(ast.program.body) : [];
+    return decls
+        .filter((d) => d.source != null)
```
alfonsogarciacaro: Hmm, I still think that we should check for `d.type === "ImportDeclaration"` here, this is the closest we have to type checking in Babel AST and the `source` field doesn't carry enough semantics (it could be possible that now or in the future another declaration has also a source field).

src/typescript/fable-splitter/index.ts
```diff
@@ -0,0 +1,205 @@
...
+    const newPath = Path.join(fileDir, fileName);
+    return newPath;
+}
+
+function fixPath(dir: string, path: string) {
+    if (!path.startsWith(".")) { return path; } // node package
+    const relPath = Path.join(dir, path);
+    const fullPath = Path.resolve(relPath);
+    const newPath = Path.join("..", flatPath(fullPath));
+    return newPath.replace(/\\/g, "/");
```
alfonsogarciacaro: We can take the chance to fix the .fs(x) extension here.

ncave: @alfonsogarciacaro Fixing the extension is done in flatPath. I'll add another cache there for name dedup, since the alreadyCompiled cache is for checking the source path.

src/typescript/fable-splitter/index.ts
ncave: @alfonsogarciacaro Not possible right now, as the AST for .fsproj only has one declaration with source, which is the "ExportAllDeclaration". Perhaps if that is changed to a "ImportDeclaration" ... Would that break bundlers?

alfonsogarciacaro: Ah, you're right! Didn't notice that 😅 Yes, we can change the autogenerated export all to import + export, or we could make the filter a bit more convoluted:

```
d.type === "ImportDeclaration" || d.type === "ExportAllDeclaration" || (d.type === "ExportNamedDeclaration" && d.source != null)
```

Reference: https://github.com/babel/babylon/blob/master/ast/spec.md#exports

ncave: @alfonsogarciacaro Idk, that seems a bit verbose... There is a reference to "non-null source" in the babel link above, if you know what I mean...

ncave:
@alfonsogarciacaro Added output file deduplication, although it should not be happening too often.
Just tested it on the repl testapp, it works, it should be ready for merging.

alfonsogarciacaro:
Awesome, thanks a lot for this!

PR #927, opened 2017-05-17
Add Path module to Node bindings, documented

Zaid-Ajaj:
@fable-compiler/code-maintenance 
@fable-compiler/documentation Please review :pray:

Not exposing the types just yet until we figure out the best way to do so.

PR #938, opened 2017-05-22
Fix build

inosik:
References #927.

- `open System` in Fable.Import.Node.Path

  Needed for the `ParamArray` attribute.

- Add names to parameters marked with `ParamArray` attribute


alfonsogarciacaro:
Great, thanks!

PR #939, opened 2017-05-22
Updated testapp config

ncave:
@alfonsogarciacaro You can try it yourself:
```
build
tsc -p src/typescript/fable-splitter
cd src/dotnet/Fable.Client.Browser/testapp
npm install
npm run build
npm start
```

### DIFF
--- a/src/dotnet/Fable.Client.Browser/testapp/package.json
+++ b/src/dotnet/Fable.Client.Browser/testapp/package.json
@@ -2,13 +2,15 @@
   "private": true,
   "scripts": {
     "preinstall": "dotnet restore --no-cache",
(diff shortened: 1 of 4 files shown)


ncave:
@alfonsogarciacaro I'm sure a few issues will crop up, for example the source map paths probably need to be fixed, but it works more or less for now.

MangelMaxime:
So splitter is an alternative to webpack and rollup ?

ncave:
@MangelMaxime Yes, but one that produces individual files, not a bundle. Sort of an opposite of a bundler. Perhaps your prototype is doing something else, if so can you please share the concept.

MangelMaxime:
I was working on the same thing than you.

And I stopped my progress on it, by lack of time.

PR #942, opened 2017-05-23
Add types for Http, Https, Crypto and Tls.

jgrund:


jgrund:
Looks like there are some tslint errors that fail the build?

ncave:
@jgrund Not sure why tslint is now trying to lint JS files (it used to work), but you can disable it until resolved [here](https://github.com/fable-compiler/Fable/blob/master/build.fsx#L358).

ncave:
@jgrund Looks like it's the new version of tslint that's doing that, I'll try to fix it.

alfonsogarciacaro:
Thanks! 🥇

Issue #946, opened 2017-05-23
--port free throws exception


Alxandr:
### Description

Supplying `--port free` (as per help output) throws an exception.

```
Unhandled Exception: System.FormatException: Input string was not in a correct format.
   at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
   at System.Number.ParseInt32(String s, NumberStyles style, NumberFormatInfo info)
   at Microsoft.FSharp.Core.LanguagePrimitives.ParseInt32(String s)
   at Fable.Tools.Main.parseArguments(String[] args)
   at Fable.Tools.Main.main(String[] argv)
```

### Related information 

* fable-compiler version: 1.0.4
* fable-core version: 1.0.4
* Operating system: OSX

Zaid-Ajaj:
Alright, I think the problem was [here](:
```fs
let tryFindArgValue key (args: string[]) =
      match args |> Array.tryFindIndex ((=) key) with
      | Some i -> Some args.[i]
      | None -> None
```
which should be (I committed it by mistake, not tested)
```fs
     let tryFindArgValue key (args: string[]) =
         match args |> Array.tryFindIndex ((=) key) with
        // i is the index of the key
        // i + 1 is the index of arg value
        | Some i -> 
            // if args.[i] is the last element
            if args.Length = i + 1 
            then None 
            else Some args.[i + 1] 
         | None -> None
```

alfonsogarciacaro:
Nice catch @Zaid-Ajaj! Could you please send a PR to fix it?

Zaid-Ajaj:
@alfonsogarciacaro Sure thing!

alfonsogarciacaro:
I have published 1.0.5 with the fix, thanks again @Zaid-Ajaj

Alxandr:
Wouldn't it be a good idea to add a unit test to avoid regressions again?

alfonsogarciacaro:
It is! Are you volunteering? :wink:

Alxandr:
While doing random programming things is way funner than writing master thesis, I still should keep on track for the last few weeks :P

Zaid-Ajaj:
Fix was published, I will close this for now, re-open if issue persists

Alxandr:
So, the exception is gone, but now the switch doesn't seem to do anything ^^

```
[...]

PR #948, opened 2017-05-24
Correct parsing arguments in Fable daemon

Zaid-Ajaj:
Addresses #946, 
Fixes reading argument values, also checks if port is parsable as an integer, if thats not the case, log the error to console and use default port

Fable.Tools needs a test project, for now I tested this [here](https://dotnetfiddle.net/EBzpCq)


alfonsogarciacaro:
Awesome, thanks a lot for the quick fix!

Issue #945, opened 2017-05-23
Fable cannot compile project referencing a file from relative directory


Zaid-Ajaj:
### Description
When a project referencing a fsharp file (.fs) from a relative directory, Fable fails to compile that project becuase fable-loader cannot find `paket.dependencies` in that relative directory:

![repro](https://cloud.githubusercontent.com/assets/13316248/26372311/ccf2e48a-3ffd-11e7-9790-bb113ab712b9.png)

# Repro 
Create the following directory structure:
```
root
  |
  | -- App
        | 
        | -- [ here the files of Simple template from "dotnet new fable" ]

  | -- Other
        |
        | -- Code.fs
```

reference file Code.fs from App.fsproj
```xml
<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard1.6</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="../Other/Code.fs" />
    <Compile Include="src/App.fs" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="FSharp.NET.Sdk" Version="1.0.*" PrivateAssets="All" />
    <DotNetCliToolReference Include="dotnet-fable" Version="1.0.4" />
  </ItemGroup>
  <Import Project=".paket\Paket.Restore.targets" />
</Project>
```

Then install the project App.fsproj:
```
cd App
yarn install
dotnet restore 
dotnet fable npm-run build
```

Project from the screen shot
[Project.zip](https://github.com/fable-compiler/Fable/files/1023503/Project.zip)

MangelMaxime:
@Zaid-Ajaj You probably should use `Project References`.

When sharing code in .Net world, you references différentes projects.

Zaid-Ajaj:
@MangelMaxime I could do that but there shouldn't be a requirement to have another project just because  I am referencing a fsharp file from different directory. Fable as able to resolve that file the same way it is resolving the `App.fs` from the `src` directory.

In .NET it is the same, you could link files from anywhere and put them in one project file. 

The fact that fable is looking for `paket.dependencies` in that project is incorrect.

Zaid-Ajaj:
After adding some logging to this function:
```fs
let rec findPaketDependenciesDir dir =
    if File.Exists(Path.Combine(dir, "paket.dependencies")) then
        dir
    else
        match Directory.GetParent(dir) with
[...]

PR #954, opened 2017-05-25
Fix searching for paket dependencies from Fable daemon

Zaid-Ajaj:
Addresses #945 
Fable daemon was looking for paket packages folder whenever it compiled a .fs file which caused problems when there were project files in sibling directories.

@alfonsogarciacaro Proposed the following [here, 5th comment](https://github.com/fable-compiler/Fable/issues/945) which is implemented below and tested (locally, with my project that wasn't working and also by running `fsi build.fsx All` that finishes without any errors)
> We could do it like this:
     - Search (and store) paketDir when creating a Project
     - Right before compiling check if com.Options.fableCore = "fable-core" and in that case create a new               compiler with new options calculating the fable-core location as in here

Also I had to set `paketDir = None` within [here](https://github.com/fable-compiler/Fable/compare/master...Zaid-Ajaj:daemon/paket-deps?expand=1#diff-6dfa1eefb46946c0a176783424e4595fR257) when "cracking" project. 

Also I added some logging when trying to find paket.dependencies and that fails, visited directories are logged

### DIFF
--- a/src/dotnet/Fable.Tools/Parser.fs
+++ b/src/dotnet/Fable.Tools/Parser.fs
@@ -91,9 +91,7 @@ let parse (msg: string) =
     let opts =
         { fableCore =
             match parseString "fable-core" "fableCore" json with
-            | "fable-core" ->
-                let paketDir =  ProjectCracker.findPaketDependenciesDir (Path.GetDirectoryName(path)) []
-                IO.Path.Combine(paketDir, "packages", "Fable.Core", "fable-core") |> makePathRelative
+            | "fable-core" -> "fable-core"
             | path -> (makePathRelative path).TrimEnd('/')
         ; declaration = parseBoolean false "declaration" json
         ; typedArrays = parseBoolean true "typedArrays" json
--- a/src/dotnet/Fable.Tools/Parser.fs
+++ b/src/dotnet/Fable.Tools/Parser.fs
@@ -92,7 +92,7 @@ let parse (msg: string) =
         { fableCore =
             match parseString "fable-core" "fableCore" json with
             | "fable-core" ->
-                let paketDir = Path.GetDirectoryName(path) |> ProjectCracker.findPaketDependenciesDir
+                let paketDir =  ProjectCracker.findPaketDependenciesDir (Path.GetDirectoryName(path)) []
                 IO.Path.Combine(paketDir, "packages", "Fable.Core", "fable-core") |> makePathRelative
             | path -> (makePathRelative path).TrimEnd('/')
         ; declaration = parseBoolean false "declaration" json
--- a/src/dotnet/Fable.Tools/ProjectCracker.fs
+++ b/src/dotnet/Fable.Tools/ProjectCracker.fs
@@ -208,11 +208,7 @@ let rec findPaketDependenciesDir dir searchedDirs =
 let tryFindPaketDirFromProject projFile =
     let projDir = Path.GetDirectoryName(projFile)
     if File.Exists(Path.Combine(projDir, "paket.references"))
-    then 
-        try 
-            findPaketDependenciesDir projDir [] |> Some
-        with
-        | _ -> None
+    then findPaketDependenciesDir projDir [] |> Some
     else None
 
 let checkFableCoreVersion paketDir =
@@ -253,7 +249,7 @@ let getPaketProjRefs paketDir projFile =
             []
 
 let getProjectOptionsFromFsproj projFile =
-    let paketDir = tryFindPaketDirFromProject projFile
+    let paketDir = findPaketDependenciesDir projFile [] |> Some
     paketDir |> Option.iter checkFableCoreVersion
     let rec crackProjects (acc: CrackedFsproj list) extraProjRefs projFile =
         acc |> List.tryFind (fun x ->
--- a/src/dotnet/Fable.Tools/ProjectCracker.fs
+++ b/src/dotnet/Fable.Tools/ProjectCracker.fs
@@ -191,27 +191,28 @@ let crackFsproj (projFile: string) =
         |> List.map (fun x -> Path.Combine(projDir, Path.normalizePath x) |> Path.GetFullPath) }
 
 let rec findPaketDependenciesDir dir searchedDirs = 
-    printfn "findPaketDependenciesDir %s %A" dir searchedDirs
     let path = Path.Combine(dir, "paket.dependencies")
     if File.Exists(path) then
-        printfn "Found %s inside %s" path dir
+        Log.print false "Found %s inside %s" path dir
         dir
     else
         match Directory.GetParent(dir) with
         | null -> 
-            let errorMessage = 
-                let dirs = String.concat "\n" searchedDirs
-                sprintf "Couldn't find paket.dependencies directory, searched in: \n%s" dirs
-            failwith errorMessage
+            searchedDirs 
+            |> String.concat "\n" 
+            |> failwithf "Couldn't find paket.dependencies directory, searched in: \n%s"
         | parent -> 
-            let searched = searchedDirs |> List.append [dir]
+            let searched = dir :: searchedDirs
             findPaketDependenciesDir parent.FullName searched
 
 let tryFindPaketDirFromProject projFile =
-    printfn "tryFindPaketDirFromProject %s" projFile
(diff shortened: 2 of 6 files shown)


src/dotnet/Fable.Tools/State.fs
```diff
@@ -260,9 +261,22 @@ let compile (com: Compiler) (project: Project) (fileName: string) =
...
+                let prevOptions = (com :> ICompiler).Options
+                let fableCoreLocation = 
+                    let paketDirectory = project.PaketDirectory
+                    IO.Path.Combine(paketDirectory, "packages", "Fable.Core", "fable-core") 
+                    |> Parser.makePathRelative
+                let nextOptions = { prevOptions with fableCore = fableCoreLocation }
+                Compiler(nextOptions, prevPlugins)
+        FSharp2Fable.Compiler.transformFile nextCompiler project project.CheckedProject fileName
+        |> Fable2Babel.Compiler.transformFile nextCompiler project
+        |> addLogs nextCompiler
```
alfonsogarciacaro: If we move the indentation here it means we are not applying `addLogs` to the other `if` branches any more.

src/dotnet/Fable.Tools/State.fs
```diff
@@ -260,9 +261,22 @@ let compile (com: Compiler) (project: Project) (fileName: string) =
...
+            | false -> com
+            | true -> 
+                let prevPlugins = (com :> ICompiler).Plugins
+                let prevOptions = (com :> ICompiler).Options
+                let fableCoreLocation = 
+                    let paketDirectory = project.PaketDirectory
+                    IO.Path.Combine(paketDirectory, "packages", "Fable.Core", "fable-core") 
+                    |> Parser.makePathRelative
+                let nextOptions = { prevOptions with fableCore = fableCoreLocation }
+                Compiler(nextOptions, prevPlugins)
```
alfonsogarciacaro: This is a bit nitpicking, but when a piece code is long it seems to take more visual importance. The `compile` function is small so now it looks as if `fable-core` resolution is the core of it. I'd try to make the code more compact and add a comment to make the meaning of it clearer to future readers:

```fsharp
        let com =
            // Resolve the fable-core location if not defined by user
            if (com :> ICompiler).Options.fableCore = "fable-core" then
                let fableCoreLocation =
                    IO.Path.Combine(project.PaketDirectory, "packages", "Fable.Core", "fable-core")
                    |> Parser.makePathRelative
                Compiler({ (com :> ICompiler).Options with fableCore = fableCoreLocation }, (com :> ICompiler).Plugins)
            else com
```

> We can also add the `Options` and `Plugins` members directly to the `Compiler` type so it's not necessary to cast it to `ICompiler` all the time.

src/dotnet/Fable.Tools/Parser.fs
```diff
@@ -92,7 +92,7 @@ let parse (msg: string) =
         { fableCore =
             match parseString "fable-core" "fableCore" json with
             | "fable-core" ->
-                let paketDir = Path.GetDirectoryName(path) |> ProjectCracker.findPaketDependenciesDir
+                let paketDir =  ProjectCracker.findPaketDependenciesDir (Path.GetDirectoryName(path)) []
                 IO.Path.Combine(paketDir, "packages", "Fable.Core", "fable-core") |> makePathRelative
```
Zaid-Ajaj: Yeah I was just refactoring code without fully understanding what it was doing at which point 😅 this one is fixed

src/dotnet/Fable.Tools/ProjectCracker.fs
```diff
@@ -190,18 +190,28 @@ let crackFsproj (projFile: string) =
         |> List.filter (fun x -> not(x.EndsWith("Fable.Core.dll")))
         |> List.map (fun x -> Path.Combine(projDir, Path.normalizePath x) |> Path.GetFullPath) }
 
-let rec findPaketDependenciesDir dir =
-    if File.Exists(Path.Combine(dir, "paket.dependencies")) then
+let rec findPaketDependenciesDir dir searchedDirs = 
+    printfn "findPaketDependenciesDir %s %A" dir searchedDirs
```
Zaid-Ajaj: Yeah these were for me when building a local project. I wil just remove them

src/dotnet/Fable.Tools/ProjectCracker.fs
```diff
@@ -190,18 +190,28 @@ let crackFsproj (projFile: string) =
...
+    let path = Path.Combine(dir, "paket.dependencies")
+    if File.Exists(path) then
+        printfn "Found %s inside %s" path dir
```
Zaid-Ajaj: done

src/dotnet/Fable.Tools/ProjectCracker.fs
```diff
@@ -190,18 +190,28 @@ let crackFsproj (projFile: string) =
...
-        | null -> failwith "Couldn't find paket.dependencies directory"
-        | parent -> findPaketDependenciesDir parent.FullName
+        | null -> 
+            let errorMessage = 
+                let dirs = String.concat "\n" searchedDirs
+                sprintf "Couldn't find paket.dependencies directory, searched in: \n%s" dirs
+            failwith errorMessage
+        | parent -> 
+            let searched = searchedDirs |> List.append [dir]
+            findPaketDependenciesDir parent.FullName searched
```
Zaid-Ajaj: Yeah the cons `::` operator is much faster (`O(1)`). Because this function will only run a list of 5 - 10 elements, I didn't put much thought in it :smile: fixed!

src/dotnet/Fable.Tools/ProjectCracker.fs
```diff
@@ -242,8 +252,9 @@ let getPaketProjRefs paketDir projFile =
             []
 
 let getProjectOptionsFromFsproj projFile =
-    let paketDir = tryFindPaketDirFromProject projFile
-    paketDir |> Option.iter checkFableCoreVersion
+    //let paketDir = tryFindPaketDirFromProject projFile
+    //paketDir |> Option.iter checkFableCoreVersion
+    let paketDir = None
```
Zaid-Ajaj: Actually if I put this back, it causes the errors I am getting, I will put these lines back:
```fs
    let paketDir = tryFindPaketDirFromProject projFile
    paketDir |> Option.iter checkFableCoreVersion
```
and have `tryFindPaketDirFromProject` actually return None when an exception is thrown, i.e.:
```fs
let tryFindPaketDirFromProject projFile =
    let projDir = Path.GetDirectoryName(projFile)
    if File.Exists(Path.Combine(projDir, "paket.references"))
    then 
        try 
            findPaketDependenciesDir projDir [] |> Some
        with
        | _ -> None
    else None
```

alfonsogarciacaro: Why are you getting errors? Is there a project with `paket.references` but no `paket.dependencies` in a parent project? It that's the case I think it makes sense to throw an error.

Zaid-Ajaj:
@alfonsogarciacaro Comments reviewed and code edited accordingly, I also added a `Log` module that we could use later when there is a `--verbose` argument passed but that's for another occasion :smile:

src/dotnet/Fable.Tools/ProjectCracker.fs
```diff
@@ -190,18 +190,29 @@ let crackFsproj (projFile: string) =
...
 
 let tryFindPaketDirFromProject projFile =
     let projDir = Path.GetDirectoryName(projFile)
     if File.Exists(Path.Combine(projDir, "paket.references"))
-    then findPaketDependenciesDir projDir |> Some
+    then 
+        try 
+            findPaketDependenciesDir projDir [] |> Some
+        with
+        | _ -> None
```
Zaid-Ajaj: I had a structure similar to that explained in #945 where there is a fsharp file alone in a directory, sibling to the directory containing paket.references, .fsproj etc. and that gave me errors. 

I will check it again tomorrow to make sure this was cause of the error

Zaid-Ajaj: @alfonsogarciacaro Confirmed that the try-catch is not needed. My project compiled fine

alfonsogarciacaro:
Thanks!

PR #966, opened 2017-05-26
Added prepack support in fable-splitter

ncave:
It's trivial and optional, so why not.


ncave:
@alfonsogarciacaro Thanks, published.

PR #974, opened 2017-05-31
Add cache

### LINKED ISSUE #882: Cache output of Fable compiler
Right now, whenever doing a build of a Fable project, it will compile everything from scratch, including your code and library code.

Webpack's watch mode helps out a lot, but every time you restart watch mode it has to compile everything from scratch again.

This is especially a problem if you have many libraries, because even though the library code hasn't changed at all, it still has to be recompiled, which makes the initial startup time for watch mode very long.

Therefore, I propose to cache the JavaScript output of Fable, so that way if a file hasn't changed, it can simply use the cache. This should speed up build times significantly.

This caching is done in the application, not in the library, so the behavior should be backwards compatible, and it doesn't put additional pressure on library authors.

alfonsogarciacaro:
Fix #882 

This PR tries to add cache capabilities as mentioned in #882. I'm not entirely happy with the result, as it requires two caches: one for the .NET side and another for the JS side. I did it this way because it's necessary to know the compilation order to validate the cache (to avoid problems with inline functions).

I guess we can try to move the cache entirely to JS side but this means the loader needs to keep state, which is not recommended for Webpack loaders (besides having to worry that Fable deamon and fable-loader states are in sync) and may create problems when compiling projects in parallel. Also, the results are not particularly impressive (in a big project of mine first compilation went from around 1 minute to around 40 secs) so I'm not sure it's worth to spend much time in this feature.

At the moment I just added it as an opt-in from fable-loader (add `extra: { useCache: true }` to the options) if somebody wants to build locally and give it a try.

@ncave I've tried to prevent this breaks the Fable.Client.Browser project (I added a couple of files to the .fsproj) but I haven't tested it yet. Please have a look.


Pauan:
A 33% speed increase sounds fantastic to me! :heart:

ncave:
@alfonsogarciacaro I have fixed the build in #975. Works fine if new caching is not enabled.
Enabling the cache unfortunately seems to break the Babel AST, we have to look at what's wrong there.

ncave:
@alfonsogarciacaro Here is what comes back from the fable server instead of Babel AST:
```
{
  cache: "C:\Users\username\AppData\Local\Temp\fable\3C4817350BFAF16DD6C936F4DAC30311",
  fileName: "C:/Projects/Fable/src/dotnet/Fable.Client.Browser/testapp/app.fs"
}
```

ncave:
@alfonsogarciacaro Ok, the caching issue is fixed in #975, it works now.

PR #975, opened 2017-05-31
Fixed REPL build

ncave:
- Resolved REPL build issue.
- Added fallback fable-core path resolution.
- Resolved cache AST issue.


ncave:
@alfonsogarciacaro Rebased to master if that makes it easier.

ncave:
@alfonsogarciacaro Congratulations on the round number of commits ([1000](https://github.com/fable-compiler/Fable/graphs/contributors))!

ncave:
Rebased to latest.

alfonsogarciacaro:
Thanks! Now I only need 1000 commits more to get another round number :wink:

Unfortunately the change below will invalidate the cache on JS side and also, in my tests, I've noticed the cost of reading from disk, serializing to bytes and parsing the JSON on client side mostly consumes the benefits of the cache.

```diff
         match cachePath with
          | Some cachePath ->
              Log.logVerbose("From cache: " + fileName)
 -            ["cache", cachePath; "fileName", fileName] |> dict |> toJson
 +            let json = IO.File.ReadAllText(cachePath)
 +            json
          | None ->
              let com =
                  // Resolve the fable-core location if not defined by user
```

This convinces me that it's not a good idea to have a feature that must work in coordination between the .NET and JS sides, as we already have several JS clients and we should avoid tight dependencies. I'm already going to revert the cache (in the .NET app) in #977, so for now I'll merge this and make the necessary changes later :+1:

Issue #781, opened 2017-03-29
Error: An implementation of file has already been given


et1975:
### Description

Same file name *across different project references* causes [the error](/Microsoft/visualfsharp/issues/2679):

```
Fable loader sent: fable-elmish/samples/react/navigation/app.fsproj
F# PROJECT: fable-elmish/samples/react/navigation/app.fsproj
   fable-elmish/samples/react/node_modules/fable-powerpack/src/Result.fs
   fable-elmish/samples/react/node_modules/fable-powerpack/src/Promise.fs
   fable-elmish/samples/react/node_modules/fable-powerpack/src/PromiseSeq/Type.fs
   fable-elmish/samples/react/node_modules/fable-powerpack/src/PromiseSeq/Module.fs
   fable-elmish/samples/react/node_modules/fable-powerpack/src/PromiseSeq/Extensions.fs
   fable-elmish/samples/react/node_modules/fable-powerpack/src/Fetch.fs
   fable-elmish/samples/react/node_modules/fable-powerpack/src/IndexedDB.fs
   fable-elmish/samples/react/node_modules/fable-react/Fable.Import.React.fs
   fable-elmish/samples/react/node_modules/fable-react/Fable.Helpers.React.fs
   fable-elmish/samples/react/node_modules/fable-elmish/prelude.fs
   fable-elmish/samples/react/node_modules/fable-elmish/cmd.fs
   fable-elmish/samples/react/node_modules/fable-elmish/program.fs
   fable-elmish/samples/react/node_modules/fable-elmish-react/common.fs
   fable-elmish/samples/react/node_modules/fable-elmish-react/react.fs
   fable-elmish/samples/react/node_modules/fable-elmish-react/react-native.fs
   fable-elmish/samples/react/node_modules/fable-elmish-browser/prelude.fs
   fable-elmish/samples/react/node_modules/fable-elmish-browser/parser.fs
   fable-elmish/samples/react/node_modules/fable-elmish-browser/navigation.fs
   fable-elmish/samples/react/node_modules/fable-elmish-debugger/Fable.Import.RemoteDev.fs
   fable-elmish/samples/react/node_modules/fable-elmish-debugger/debugger.fs
   fable-elmish/samples/react/navigation/app.fs
Parsing F# project...
F# project contains errors:
fable-elmish/samples/react/node_modules/fable-elmish-browser/prelude.fs(L4,0) : error FSHARP: An implementation of the file or module 'Prelude' has already been given
```

### Repro code

[...]

PR #978, opened 2017-06-02
Resolved file name clashing

ncave:
Resolves the following error when building Fable.Client.Browser.fsproj:

`error FS0239: An implementation of the file or module 'Util' has already been given`


alfonsogarciacaro:
Ah, so that was the reason you changed the name! I thought it was a mistake 😅 

I'd rather not use `Log` because that file now contains several modules, we can use `ToolsUtil` or similar. But anyways this is weird because that error should have been already fixed with #781. The Elmish templates, for example, contains files with the same name.

ncave:
@alfonsogarciacaro Thanks, I renamed the file.

ncave:
@alfonsogarciacaro If you don't like the name change, it's optional ATM as that file is not required in the project (but it may be again going forward, so not duplicating names may be a good idea). This is passing as of right now.

alfonsogarciacaro:
Ah, ok. Actually afaik the fix should have been propagated to the F# compiler, but maybe we have to wait for F# 4.2 for that.

Thanks! 🎻

Issue #976, opened 2017-06-01
Wrong JS generated


Alxandr:
### Description

I have the following piece of F#:

```fsharp
    let followsLensLaws lens outer inner dummy f = prop {
      if inner = outer then
        do! Assert.pass
        do! Assert.pass
        do! Assert.pass
        do! Assert.pass
      else
        do! getSetIdentity lens outer
        do! setGetSymmetry lens outer inner
        do! setSetOrderDependence lens outer inner dummy
        do! getSetMapCorrespondence lens f outer
    }
```

which when I generate JS from it does something wrong. I'm not entirely sure what exactly it does wrong yet as the output is fairly unreadable, but what I do know is that when `getSetMapCorrespondence` is called, `f` is `undefined`, whereas when it was passed into the `followsLensLaws` it was not. The output JS looks like this:

```js
const followsLensLaws = __exports.followsLensLaws = function (lens_0, lens_1, outer, inner, dummy, f) {
      const lens = [lens_0, lens_1];
      return function (builder_) {
        return (0, _Util.equals)(inner, outer) ? builder_.Bind_1(_assert.Assert.pass, function () {
          return builder_.Bind_1(_assert.Assert.pass, function () {
            return builder_.Bind_1(_assert.Assert.pass, function () {
              return builder_.Bind_1(_assert.Assert.pass, function () {
                return builder_.Return();
              });
            });
          });
        }) : builder_.Bind_0(function (tupledArg) {
          const lens_2 = [tupledArg[0], tupledArg[1]];

          const f_1 = function (lensGet, lensSet, outer_1) {
            return getSetIdentityWith(lensGet, lensSet, outer_1);
          };

          const lens_3 = [lens_2[0], lens_2[1]];
          return (($var1, $var2) => $var3 => f_1($var1, $var2, $var3))(get(lens_3[0], lens_3[1]), set(lens_3[0], lens_3[1]));
        }(lens)(outer), function () {
          return builder_.Bind_0(function (tupledArg_1) {
            const lens_4 = [tupledArg_1[0], tupledArg_1[1]];

            const f_2 = function (lensGet_1, lensSet_1, outer_2, inner_1) {
              return setGetSymmetryWith(lensGet_1, lensSet_1, outer_2, inner_1);
            };

            const lens_5 = [lens_4[0], lens_4[1]];
[...]

PR #980, opened 2017-06-04
Currying: failing tests

Zaid-Ajaj:
Thanks to @Alxandr and @ed-ilyin  we were able to boil down #976 from a convoluted code base to a managable piece of code. This PR adds a test for basic currying. Again, the bug seems to be the closure-optimization "feature".

### First attempt: the following actually passes (try in the repl)
which surprised me because we checked that it didn't work correctly
```fs
let equal x y = 
 printfn "%A = %A was %A" x y (x = y)
 
let ``Basic currying works``() =
  
 let curry (fn: 'a -> 'b -> 'c) =
   let first = fun (a: 'a) ->
    let second = fun (b: 'b) ->
      let result = fn a b
      result
    second
   first
   
 let plus = curry (+)
 let result = plus 2 3
 equal 5 result            // "5 = 5 was true"
 equal 5 (plus 2 3)        // "5 = 5 was true"
 equal 5 ((curry (+)) 2 3) // "5 = 5 was true"
 
``Basic currying works``()
```
Until you pull `curry` outside of the test method, which then fails. Try it in the repl
```fs
let equal x y = 
    printfn "%A = %A was %A" x y (x = y)

let curry (fn: 'a -> 'b -> 'c) =
  let first = fun (a: 'a) ->
    let second = fun (b: 'b) ->
      let result = fn a b
      result
    second
  first
  
let ``Basic currying works``() =
    let plus = curry (+)
    let result = plus 2 3
    equal 5 result            // "5 = 0 was false"
    equal 5 (plus 2 3)        // "5 = undefined was false"
    equal 5 ((curry (+)) 2 3) // "5 = undefined was false"
    
``Basic currying works``()
```

Makes me wonder what other tests will fail if you just pull the definitions out of the test methods...

### DIFF
--- a/src/tests/Main/ApplicativeTests.fs
+++ b/src/tests/Main/ApplicativeTests.fs
@@ -442,3 +442,19 @@ let ``Unit expression arguments are not removed``() =
         x <- i
     doNothing <| foo 5
     equal 5 x
+
+[<Test>]
+let ``Basic currying works``() =
+    let curry (fn: 'a -> 'b -> 'c) =
+      let first = fun (a: 'a) ->
+        let second = fun (b: 'b) ->
+          let result = fn a b
+          result
+        second
+      first
+
+    let plus = curry (+)
+    let result = plus 2 3
+    equal 5 result
+    equal 5 (plus 2 3)
+    equal 5 ((curry (+)) 2 3)
--- a/src/tests/Main/ApplicativeTests.fs
+++ b/src/tests/Main/ApplicativeTests.fs
@@ -443,16 +443,17 @@ let ``Unit expression arguments are not removed``() =
     doNothing <| foo 5
     equal 5 x
 
+
+let curry (fn: 'a -> 'b -> 'c) =
+  let first = fun (a: 'a) ->
+    let second = fun (b: 'b) ->
+      let result = fn a b
+      result
+    second
+  first
+
 [<Test>]
 let ``Basic currying works``() =
-    let curry (fn: 'a -> 'b -> 'c) =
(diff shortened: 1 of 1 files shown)


Alxandr:
It might actually be an idea to go down the route that F# did with `OptimizedClosures` in JS as well. That way you wind up with functions that can be called with a single parameter, or call `.fastInvoke(...args)`.

alfonsogarciacaro:
Thanks for the test and the detailed explanation @Zaid-Ajaj! This has saved an incredible amount of time. Team work FTW! ✋ 

I was already concerned about functions returning curried lambdas but I wasn't sure how to fix it. I did it now just by wrapping the method body with a `CurriedLambda` call. Not sure if this the best solution but it seems to work (though I had to add a `.curried` marker to functions returned by `CurriedLambda` to avoid stack overflows.

I'd rather not do something like `OptimizedClosures` because the main purpose of the uncurrying feature was to make F# lambdas compatible with JS functions more than the optimization by itself.

Alxandr:
@alfonsogarciacaro that should be easy enough though. Just do a implicit conversion from `a -> b -> ... -> n -> returnValue` to `(a, b, ..., n) -> returnValue`.

Alxandr:
Also, while I'm already rambling, given that F# does not have varadic functions, and JS does, doing inline auto-currying is obviously an option.

You could do the following transform:

```fsharp
let add3 a b c = a + b + c
```

```js
import partialApply from 'fable-core';

const add = (a, b, c) => {
  if (arguments.length < 3) return partial(add, [ ...arguments ]);
  return a + b + c;
}
```

Issue #926, opened 2017-05-17
Running `dotnet fable npm-run` without `start` or `build` causes an exception


FrankBro:
### Description

Running `dotnet fable npm-run` without `start` or `build` causes an exception

`Unhandled Exception: System.IndexOutOfRangeException: Index was outside the bounds of the array. at Fable.Tools.Main.main(String[] argv)`

### Repro code

`dotnet fable npm-run`

### Expected and actual results

Provide a helpful recommendation on what to run.

### Related information 

* fable-compiler version: Latest
* fable-core version: Latest
* Operating system: Win10

alfonsogarciacaro:
You're right, we should add an array length check for [the npm-run command](https://github.com/fable-compiler/Fable/blob/7ac95ae2adf4236073f9e62790f995a3032c4114/src/dotnet/Fable.Tools/Main.fs#L166) (and also for node-run and shell-run) and print a more helpful error message.

Could you please PR it?

jeremyabbott:
@alfonsogarciacaro if it can wait until this weekend I'd be interested in giving it a try.

mizzle-mo:
@alfonsogarciacaro Would it be possible to replace npm-run with just build and start to simplify things?  npm-run is not only extra characters, but I believe it is a bit misleading since there is extra magic that happens prior to npm run.

MangelMaxime:
@mizzle-mo `npm-run` allow you to the commands from `package.json`

It's an alias to `npm run start`. Only when using `dotnet fable` to run, this will automatically start the fable compilation server.

You could archieve the same result using two consoles with this commands: `dotnet fable start` && `npm run start`.

mizzle-mo:
@MangelMaxime Could we not get the same result with ```dotnet fable start``` and ```dotnet fable build```  The npm-run seems redundant and misleading.  I initially thought it just ran ```npm run``` since that is the name of the command.  I figured I could do that with less typing by just using ```yarn start``` which obviously didn't work as expected.  Just wondering why we have extra sugar on the command if not necessary.  Why is the npm-run command required in the cli?

MangelMaxime:
Because fable isn't dedicated to be used with npm only.

You could start the Fable compilation server and pass it the files you want to compiler via FAKE, bash etc.

`dotnet fable npm-run` is an helper which start for you the compilation server before executing the command from `package.json`

Perhaps the help can give you more info:

```fs
[...]

PR #984, opened 2017-06-05
Check for missing arguments when using npm-run

Zaid-Ajaj:
Addresses #926 

When running `dotnet fable npm-run` without any more arguments, an exception is thrown which is not very helpful.

With this PR, when you run `dotnet fable npm-run` you would get:
```
Missing argument(s) after npm-run, expected at least one more argument corresponding with the name of an npm-script.

Examples: 

  `dotnet fable npm-run start`
  `dotnet fable npm-run build`

Where 'start' and 'build' are the names of two npm-scripts located at package.json:

"scripts" :{
    "start": "webpack-dev-server"
    "build": "webpack" 
}
```


alfonsogarciacaro:
Nice one, thank you! 🆘 😼

PR #985, opened 2017-06-06
Added package lock file

ncave:


alfonsogarciacaro:
Thanks! So now yarn is deprecated? ;)

Pauan:
@alfonsogarciacaro It's a bit complicated.

Just like yarn, npm 5 now has automatic support for lock files (it uses `package-lock.json`). Of course this won't work with earlier versions of npm.

The intention is for `package-lock.json` to work with any tool. In other words, it's likely that yarn will gain support for `package-lock.json`, and therefore `package-lock.json` can be used as a one-size-fits-all lock file.

So rather than "deprecating" yarn, instead it's more like "use whatever tool you prefer, yarn or npm5 or whatever, the same lock file will work with anything"

PR #989, opened 2017-06-08
Adding parse to allow use of test on a non-english culture thread

arthis:
Added parse to enable execution of tests and use Convert.ToSingle in dotnet test while current thread point to fr-FR.


arthis:
Appveyor falure is about :

Failed   Fable.Tests.Async.Async cancellation works
Error Message:
 Assert.Equal() Failure
Expected: False
Actual:   True

It seems like quite a recurrent problem BTW.

alfonsogarciacaro:
Thanks! Yes, it seems some of the asynchronous tests fail in slow machines. Probably it could be fixed by making the timeouts longer. I already did it once, but apparently it was not enough 😅

Issue #536, opened 2016-11-08
PrintfFormat.Value doens't return a string


davidtme:
### Description

`PrintfFormat.Value` doens't return a string i.e. what the compiler expects

### Repro steps

```f#
let pathScanTest (pf:PrintfFormat<_,_,_,_,'t>) s =
    let formatStr  = pf.Value
    printfn "pf.Value = %s" formatStr

pathScanTest "foo %s" "foo bar"
```

### Expected behavior

```
pf.Value = 'foo %s'
```

### Actual behavior

```
pf.Value = 'function (cont) {
            _cont = cont;
            return fsFormatRegExp.test(str) ? makeFn(str) : _cont(str);
        }'
```

### Related information 

* Fable Version: 0.7.5-alpha.22
* .NET Runtime
* Windows

davidtme:
@alfonsogarciacaro did anything ever happen with this, I was hoping to use it to create a suave.io style pathScan for elmish

alfonsogarciacaro:
Ups, it seems this issue got buried by others and forgot about it, sorry! 😅 I need to check it... unless you want to give it a try, of course :wink:

davidtme:
I'm not even sure what PrintfFormat does or how it works really just hoping I can copy and paste my way to victory.

alfonsogarciacaro:
Me neither ;) Anyway I'll have a look. Ping me here again next week if I happen to forget!

alfonsogarciacaro:
Anyone from @fable-compiler/code-maintenance interested in having a look? 😉

MangelMaxime:
Should be close ?

alfonsogarciacaro:
It'd be nice if @davidtme can confirm whether it's working for him. But the test and fix added by @Zaid-Ajaj are working so I guess we can close this.

PR #990, opened 2017-06-08
Make print format keep track of the input format

Zaid-Ajaj:
Addresses #536

When using 
```fs
sprintf "something %s"
```
instead of translating to 
```js
String.fsFormat("something  %s")(function(x) { return  x; })
```
It will translate to:
```js
{  
  formatFn: String.fsFormat("something  %s")(function(x) { return  x; }),
  input: "something  %s" 
}
```
Making it possible to retrieve the input string later:
```fs
[<Test>]
let ``input of print format can be retrieved``() =
      let pathScan (pf:PrintfFormat<_,_,_,_,'t>) =
          let formatStr = pf.Value
          formatStr
          
      equal "/hello/%s" (pathScan "/hello/%s")
```

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -708,10 +708,14 @@ module AstPass =
     let fsFormat com (i: Fable.ApplyInfo) =
         let emit macro =
             let emit = Fable.Emit macro |> Fable.Value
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Awesome, thank you! I never found time to tackle this :clap: :clap: I hope this can help @davidtme :)

Zaid-Ajaj:
thanks, it was a nice exercise to get to know the compiler api's better :smile:

Issue #991, opened 2017-06-09
Fable.JsonConverter prerelease only available for dotnet core


coolya:
### Description

Fable.JsonConverter prerelease only available for dotnet core

### Repro code

Have project that is not targeting dotnetcore but frame 4.6.2 try to use latest Fable.JsonConverter in order to work with fable 1.0.

### Expected and actual results

Fable.JsonConverter available for none dotnet core projects.

For various reasons I can't migrate my backend project that sends json data to the fable front end to dotnet core. Mostly because not all dependencies aren't available for dotnet core yet. 

### Related information 

* fable-compiler version: 1.0.7
* fable-core version: 1.0.7
* Operating system: OSX

MangelMaxime:
Duplicate with https://github.com/fable-compiler/Fable/issues/847

In this issue, the solution proposed is to use Paket to reference a Github dependency. Would this work for you ?

coolya:
That is what I already did.  I just found it very confusing that the nuget package didn't work. How about making the Paket file reference the "officially" and documented way of using it?

alfonsogarciacaro:
Yeah, we should just publish a version for net45. It's just the couple of times I tried for one reason or the other (nuget problems, macOS) it didn't work so I gave up for the moment 😕

alfonsogarciacaro:
I published a version (1.0.0) with both netstandard1.6 and net45 support. If you have moment, could you please if check the latest Nuget package works for your net45 project? Cheers!

coolya:
Works like charm, thanks a lot @alfonsogarciacaro

PR #992, opened 2017-06-10
update documentation about JSON serialisation

coolya:
updated the docs section to use Fable.JsonConverter and document how to use it in
in projects other than dotnet core. To prevent confusions that I reported in #991 

http://fable.io/blog/Introducing-0-7.html#JSON-Serialization


alfonsogarciacaro:
Great, thanks! 📝

Issue #947, opened 2017-05-24
error FABLE: Cannot find replacement for System.Environment.get_NewLine


forki:
### Description

error FABLE: Cannot find replacement for System.Environment.get_NewLine

### Repro code

    printfn "%s" System.Environment.NewLine

alfonsogarciacaro:
This should be an easy one. Anyone in @fable-compiler/code-maintenance team interested in taking it? I can point out where it should be added in `Replacements` so it's a good way to get in touch with that module.

Zaid-Ajaj:
For reference: https://stackoverflow.com/questions/10864486/node-js-constant-for-platform-specific-new-line

alfonsogarciacaro:
I would just use `\n` as most of the times the code will be run in the web.

forki:
+1 for \n

ncave:
In general, +1 for matching the BCL functionality as close to documented as possible to minimize surprises.
This may or may not be one of the exceptions, but IMO it's not, just use the string literal if you need to.
But I can see this very rapidly evolving into spaces-vs-tabs-like argument, so I'll stop here, please ignore...

forki:
Ok. Let me correct myself. I'm personally fine with simple line breaks but
yes doing it properly will give bonus points.

Am 25.05.2017 5:46 nachm. schrieb "ncave" <notifications@github.com>:

PR #993, opened 2017-06-10
Add replacement for System.Environment.NewLine

coolya:
As I wasn't sure how to add a more generic replacement that would
work as suggested in #947 I took a simple approach and replaced it with `\n`

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -655,10 +655,10 @@ module AstPass =
                 then "awaitPromise" else "startAsPromise"
             CoreLibCall("Async", Some meth, false, i.args)
             |> makeCall i.range i.returnType |> Some
-        | "ofJsonAsType" ->            
+        | "ofJsonAsType" ->
             match i.args with
             | [arg; typ] -> fableCoreLib com { i with methodName = "ofJson"; args = [arg; makeJsObject None ["T", typ]] }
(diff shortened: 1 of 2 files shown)


coolya:
Any idea why the test fails on windows?

Zaid-Ajaj:
Nice one!
> I wasn't sure how to add a more generic replacement 

I want full control, you can write a function in one of fable-core modules (in typescript) and from the replacements you just call that function passing in the necessary arguments, if any. See an example from the replacements like [this one here](https://github.com/fable-compiler/Fable/blob/master/src/dotnet/Fable.Compiler/Replacements.fs#L321) 

On a side-note, I don't think you need to check the the `methodKind` to match with Fable.Getter because `System.Enviroment.NewLine` is read-only

> Any idea why the test fails on windows?

Some async test timeout is usually the cause, we need to fix this :smile: cc @fable-compiler/code-maintenance

alfonsogarciacaro:
Great, @coolya, thanks for this! :clap:

I think the test fails on Windows because when the tests run on .NET `Environment.NewLine` returns `\r\n`. I guess the easiest way to fix this problem for now is to make the test available only on JS by wrapping it with the `#if FABLE_COMPILER` directive :)

coolya:
I think it's ready to get merged now. Though now the tests on windows fail because of the async stuff 😄

alfonsogarciacaro:
Awesome, thank you! Yes, the Async tests fail from time to time in AppVeyor (in the .NET run not in JS), I need to make the timeouts a bit longer...

Issue #988, opened 2017-06-08
Show relative path during compile


forki:
create the fable react template and start fable.

![image](https://user-images.githubusercontent.com/57396/26923763-6ea48d6e-4c43-11e7-897c-1f0d4bb220ce.png)

It would be nice if fable would show `Counter\Types.fs" and so on

MangelMaxime:
I believe it's been fixed now ?

![image](https://user-images.githubusercontent.com/4760796/28250278-b253b910-6a64-11e7-8ff6-61a56702b797.png)

alfonsogarciacaro:
Yeah, it is. Thanks!

PR #994, opened 2017-06-10
print relative path of compiled files

coolya:
In project with several files that have the same name but different paths
it's not clear form the message which file was compiled. #988

I changed the log outout to print the relative path of file.

Previously the output looked like:

```
fable: Compiled ConvertTests.fs
fable: Compiled ResizeArrayTests.fs
fable: Compiled ResultTests.fs
fable: Compiled RegexTests.fs
```

not it looks like this:

```
fable: Compiled src/tests/Main/ConvertTests.fs
fable: Compiled src/tests/Main/ResizeArrayTests.fs
fable: Compiled src/tests/Main/ResultTests.fs
fable: Compiled src/tests/Main/RegexTests.fs
```

### DIFF
--- a/src/typescript/fable-loader/index.js
+++ b/src/typescript/fable-loader/index.js
@@ -153,7 +153,7 @@ module.exports = function(buffer) {
                     })
                 }
                 var babelParsed = transformBabelAst(data, babelOptions, data.fileName, buffer);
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Superb, thanks a ton! 🐎

PR #1000, opened 2017-06-15
Exposing Console among JS import globals

et1975:
As discussed, since console is available in most (all?) JS runtimes, would make sense to make it available via Import.JS globals.

Notes:
- Copied the impl that lives in Import.Browser and removed methods referencing `Element` type - it doesn't appear in MDN/spec, but maybe it makes sense to keep that version of Console type in `Browser` package "as is".
- Added `table` method, which is defined in MDN/spec and implemented by most browsers.


alfonsogarciacaro:
Thank you! Will be released as Fable.Core 1.1.1

alfonsogarciacaro:
@et1975 From now on you can always say with proud yours was the #1000 issue/PR in Fable :wink:

Issue #1005, opened 2017-06-17
Option.defaultValue from FSharp.Core missing


haf:
### Description

nuget Fable.Core 1.0.8
    FSharp.Core (4.2.1) - redirects: force

Create a new Elmish app.

Use `Option.defaultValue`. Crashes in JS.

```
28:15-27 "export 'defaultValue' was not found in '../../packages/Fable.Core/fable-core/Seq'
webpack: Compiled with warnings.
```

In sample view code:

```
      dl
        [ ]
        [
          yield dd
            []
            [ str "Total taxes"]
          yield dt
            []
            [ input [] ]
          if Option.defaultValue 0m model.totalTaxes > 0m then
            yield rateSubcomponents model dispatch
          else
            ()
        ] ]
```

alfonsogarciacaro:
Fixed by #1009

PR #1009, opened 2017-06-17
Add support for Option.defaultValue

coolya:
A propper replacement for `Option.defaultValue` was missing #1005. I added a new replacement for it to issue a call into `Utils.defaultArg`.


alfonsogarciacaro:
👐  Awesome, thanks! 👐

Issue #987, opened 2017-06-08
Fable should warn on the use of nested Option


generalelectrix:
### Description

Since Fable compiles Option down into a nullable object, using nested options will produce unexpected behavior as `Some(None)` compiles to `null`, so `Some(None) = None` evaluates to true in Fable yet false in regular F#.  Pattern matching of course has the same issue.

### Expected and actual results

This can lead to pretty heinous bugs if you are using existing F# code that expects nested options to work.  It is possible to work around by essentially defining a fresh version of Option with a different name, thus lacking Fable's nullable representation, but it would probably be good to issue a compiler warning on the use of nested options.

alfonsogarciacaro:
Hi @generalelectrix! This is a nice suggestion and shouldn't be too difficult to implement, though it probably won't make it into the first Fable 1.0 stable version. We'll try to implement it in future versions or, if you want to give it a try, we can provide you some guidance on how to do it.

Cheers!

generalelectrix:
If you think it wouldn't be too hard, I could dive in with some guidance presuming the coding to be done is on the F# side (I'm here because I'm a JS ignoramus).

alfonsogarciacaro:
Relevant: https://github.com/fsharp/fslang-suggestions/issues/546

Closing as this is fixed thanks to @coolya's PR! :clap: :clap:

Please note, this has been made a warning and not an error at the end as it affected Fable self-compilation (to create the REPL).

PR #1010, opened 2017-06-17
WIP: Show error when nested options are used

coolya:
As mentioned in #987 falbe should point out when nested options are used. I made it an error because the code will definitely not behave correctly at runtime. 

```
ERROR in ./src/tests/Main/MiscTests.fs
/Users/kdummann/source/fsharp/Fable/src/tests/Main/MiscTests.fs(831,12): (831,18) error FABLE: Nested option in option won't work at runtime
```

Should it be an error?

How do you test compiler error messages? I wasn't able to find anything in the tests folders where something like that was done.

### DIFF
--- a/src/dotnet/Fable.Compiler/FSharp2Fable.fs
+++ b/src/dotnet/Fable.Compiler/FSharp2Fable.fs
@@ -113,35 +113,40 @@ let private transformNewList com ctx (fsExpr: FSharpExpr) fsType argExprs =
 
 let private transformNonListNewUnionCase com ctx (fsExpr: FSharpExpr) fsType unionCase argExprs =
     let unionType, range = makeType com ctx.typeArgs fsType, makeRange fsExpr.Range
-    match fsType with
-    | OptionUnion ->
+    let genericArg = fsType.GenericArguments |> List.ofSeq |> List.tryHead
+
+    match (fsType, genericArg) with
+    | (OptionUnion, Some OptionUnion) ->
+        "Nested option in option won't work at runtime"
+        |> addErrorAndReturnNull com ctx.fileName (Some range)
+    | (OptionUnion, _) ->
         match argExprs: Fable.Expr list with
         // Represent `Some ()` with an empty object, see #478
         | expr::_ when expr.Type = Fable.Unit ->
             Fable.Wrapped(Fable.ObjExpr([], [], None, Some range), unionType)
         | expr::_ -> Fable.Wrapped(expr, unionType)
         | _ -> Fable.Wrapped(Fable.Value Fable.Null, unionType)
-    | ErasedUnion ->
+    | (ErasedUnion, _) ->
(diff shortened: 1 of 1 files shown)


MangelMaxime:
Just to help people think about the problem [here](https://goo.gl/piUw1w) is a repro code.

```fs
open Fable.Import
open Fable.Core.JsInterop


type Value1 = string option

type Value2 = Value1 option

let printValue2 = 
  function 
  | Some value1 ->
    match value1 with
    | Some v -> Browser.console.log ("Value found at Value1 level: " + v)
    | None -> Browser.console.log "No value found at Value1 level"
  | None -> Browser.console.log "No value found at Value2 level"
  
printValue2 (Some (Some "Maxime"))
printValue2 (Some (None))
printValue2 (None)
```

We probably should print an error if we want to make sure no runtime error is garanted.

alfonsogarciacaro:
Thanks for this @coolya! :clap: Unfortunately, we don't have an automated way at the moment to test compiler errors 😕 But the tests are passing with this change and it looks good to me (I'm also OK with throwing an error to force the user refactor the code) so we can merge it :+1:

Thanks also for the repro code @MangelMaxime!

coolya:
@alfonsogarciacaro  would a PR that allows testing for error messages be welcome? If So I would be interested in looking into it. Is there a way to sent a FSharp AST to Fable and get the result back? Maybe from a quotation rather than a file and then check if errors were emitted.

alfonsogarciacaro:
@coolya That'd be a very nice idea! In general, we're lacking integration tests. The current tests are good because they are run both in JS and .NET so we can be sure the same semantics is kept in both platforms. However, it'd be also nice to use something like Canopy with the Fable samples (now in the `samples` repo) and make sure the samples are still working with each Fable version.

About sending an F# AST, unfortunately this is not possible because the FCS AST is not exactly the same as the AST from quotations. Also it seems that FCS needs the whole F# project (an .fsx script at a minimum) for the parsing. If you see the tests and samples in the FCS repo (like [this one](https://github.com/fsharp/FSharp.Compiler.Service/blob/c0054d7b715d162de37046dc32de6cd3725b3d9a/docs/content/typedtree.fsx#L40-L46)) you'll see they just write a string with code to a temporary file and send the path to the compiler.

About using the Fable compiler more directly for testing/debugging, I have [a hidden `debug` option](https://github.com/fable-compiler/Fable/blob/2ed466d8a94d2d2409cd076f377fe5077c749e51/src/dotnet/Fable.Tools/Main.fs#L121-L131) to just parse a file and print the generated json (that is usually sent to JS clients like fable-loader) on screen. You can use it as reference to compile a file programatically. Then you can just take the logs from the `Compiler` instance (`com.Logs`).

coolya:
Thanks for all the input. I will try to build something this weekend and then PR it so we can discuss details. :)

Issue #1008, opened 2017-06-17
Docs: usage of npm vs yarn is confusing


haf:
### Description

When getting started, npm and yarn are both "used" in the sense that `dotnet fable npm-run` uses npm but the templates are installing yarn.

Either alias `yarn-run` to `npm-run` or make it explicit (for those who don't know what npm/yarn is), that we're "executing javascript" via `js-run` or similar.

A suggestion. Feel free to close if you don't agree.

MangelMaxime:
We had something similar with elmish template.

Finally we used the command: `dotnet fable shell-run "yarn start"`

Do you find it less confusing ?

alfonsogarciacaro:
We can also add a `yarn-run` command, that's an easy one, so @MangelMaxime suggestion becomes `dotnet fable yarn-run start`.

JS cannot be run by itself so I'd rather mention the tool explicitly. We also have a `dotnet fable node-run` command to run a JS file directly.

MangelMaxime:
Problem with adding `yarn-run` is we would end up adding more and more alias command. I already heard of 3-4 npm like commands (welcome in JS worlds :trollface: )

haf:
I'm for `yarn-run` since it follows the principle of least surprise and mentions the tool it'll call thereby making it explicit.

mike-morr:
In this case ```yarn run``` may not be accurate, as what you are actually running is an npm script.  Even if you use yarn to execute, you are still running an NPM script under the hood.

alfonsogarciacaro:
Latest fable version (1.1.1) adds a `yarn-run` command. Even if it's a new command, it shouldn't be difficult to remember as it follows the same pattern as for the others: npm-run, node-run, shell-run.

I don't know exactly how `yarn run` executes, I'm not sure either if scripts in package.json are exclusively considered npm scripts. Yarn has an interesting features as it can run local modules like `yarn run webpack`.

Also, for both npm and yarn, Fable 1.1.1 allows a shortcut following the pattern `dotnet fable yarn-run start` ->  `dotnet fable yarn-start`.

alfonsogarciacaro:
I tried to make it so the use of `yarn` and `npm` keywords are exchangeable in the commands to give the choice to the user. This is explained in [a note here](http://fable.io/docs/getting-started.html#tl-dr) but please feel free to send a PR if that's not clear enough :+1:

Issue #1007, opened 2017-06-17
Docs: make d/l & usage of `dotnet` clear


haf:
### Description

As we started testing, it wasn't that clear from the docs that .Net Core had to be installed and where to/how to install it.

This is a suggestion to make that clearer.

alfonsogarciacaro:
The dotnet SDK has been added to the requirements :+1: http://fable.io/docs/getting-started.html#requirements

PR #1011, opened 2017-06-17
Templates update

### LINKED ISSUE #1004: Auto compilation not working on the default installed example
I followed the instructions given at "getting started" for Fable 1.0
(https://github.com/fable-compiler/Fable/blob/master/docs/pages/getting-started.md)
I started the Webpack developement server.
But, making changes to the App.fs and saving make no effect in the browser page.
I ran "dotnet fable npm-run build" in another terminal successfully. But no changes in the browser page.
The only way to browser page for changes of the App.fs is to stop the Webpack development server and restart it.

### Related information 

* fable-compiler version: don't know where to find this info
* fable-core version: don't know where to find this info

MangelMaxime:
This PR update the template to Fable 1.1.0.

It's also try to provide a better user experience as some people discuss about it on Twitter and even the different tchats.

I added support of fake to abstract the build process to the user. We can imagine this template being the first step of people into Fable world. They just want the thing to work and easy to start with.

Now they can start simply by doing: `build.cmd` or even `build.cmd Watch`.

One things I am unsure of is checking the dotnet version installed as it can take sometime to download the right version if not present. And also if they installed the template they should have dotnet core on their machine.

Pinging @alfonsogarciacaro @haf as this can be related to #1008 #1007 

cc @fable-compiler/documentation 

This version of the template should also fix #1004


MangelMaxime:
I just put the PR under [WIP] because when using a cmd I can't do `Ctrl-c` at the moment. I need to investigate the reason always fun playing with Windows cmd.

MangelMaxime:
Ok so the problem about `Ctrl-c` was a in all the cmd of my windows. I restarted and it's ok now.

So this version of the template is ready for review and merge if everyone is ok.

alfonsogarciacaro:
Thanks a lot for this @MangelMaxime! Though I've to say I've some mixed feelings about adding so many files to the simple template. Probably it's only me but I like to know what each file committed to my repo is exactly doing and I had a hard time when using the F# project scaffold back in the days to clean the files and the build script and get only what I needed. Some other considerations:

- We should try to avoid [the explosion of packages folder](https://github.com/fable-compiler/fable-suave-scaffold/issues/116). This is mainly due to netcore System.XXX packages (hopefully the situation will improve with netcore 2) but things like Paket groups also contribute to that.

- We still need .NET Framework/Mono for tooling like Paket or Ionide, but in my vision we should aim for just having netcore + node as system requirements.
- At the moment, we are dealing with node.js and dotnet SDK very differently. Node.js must be a system requirement and we don't check the version, while for dotnet SDK we check the version and can be locally installed. I guess it's better to treat both dependencies in a similar way.
- Downloading and installing dotnet SDK takes a while, I'd like to avoid it just because the template asks for dotnet SDK 1.0.4 and the user has 1.0.5 (supposedly, users already need dotnet SDK to download the template).
- Now that npm also supports lock files, it'll be nice to offer both the possibility of using `npm` too, 

> In latest Fable, it'll be possible to run `yarn-run start` and `yarn-start` as shorcut. Same thing for `npm`, so using either tool should be just a matter of changing `npm` with `yarn` or vice versa.

Maybe we could keep the simple template as slim as possible and create another one with a more full-fledged project (using FAKE, etc)? What do you think?

MangelMaxime:
- If we can avoid the explosion of packages folder, I am totally agree that we should :).

- I already removed the downloading and installer dotnet SDK locally. We only use the global one. Because the templates should be installed via `dotnet new ...` so user already have a dotnet version installed :)

- Using yarn, allow us to have a really quick `yarn install` compare to `npm install` where even having nothing to install take a lot of times. We could solve this by adding a choici in the template asking the user which manager he want to use.

I am ok to make the template slim. But the user should not have to run a lot of commands to start working with it.

We could provide `build.cmd` and `watch.cmd` (with their equivalent for `*.sh`)
We need to imagine that this is the first time the user will play with Fable so we need to make his experience very clean and easy :)

mike-morr:
I like the new experience using this template.  I don't have to worry about "did I remember all the commands", this version of the template simplifies things a lot.  That being said, I also agree with @alfonsogarciacaro.  We should have a minimal template with just dotnet and node as well, however I think this template offers the best getting started experience.  Yarn/dotnet cli are abstracted away so that  with one command I can get this up and running.  I like this template a lot.

mike-morr:
@alfonsogarciacaro Can you check the build script on Mac, I am running into a lot of issues with it on Mac.  It ran fine on Windows.

alfonsogarciacaro:
I think some of the confusion also came because Fable was still buggy. I just released a new template for Fable 1.1.1 with an [updated README](https://github.com/fable-compiler/Fable/blob/99ab837723aaeec6d036394b346b2b5a54395baf/src/templates/simple/Content/README.md). Can you please check if everything is clearer now? Later we may create a new repo for templates and add a new one with a more complex project and using Fake or even better, we could turn the fable-suave-scaffold into a template.

MangelMaxime:
I updated so the template so the user can choose to use Fake or no.

```
C:\Users\mange\Downloads>dotnet new fable -h
Template Instantiation Commands for .NET Core CLI

Usage: new [options]

Options:
  -h, --help          Displays help for this command.
  -l, --list          Lists templates containing the specified name. If no name is specified, lists all templates.
  -n, --name          The name for the output being created. If no name is specified, the name of the current directory is used.
  -o, --output        Location to place the generated output.
  -i, --install       Installs a source or a template pack.
  -u, --uninstall     Uninstalls a source or a template pack.
  --type              Filters templates based on available types. Predefined values are "project", "item" or "other".
  --force             Forces content to be generated even if it would change existing files.
  -lang, --language   Specifies the language of the template to create.


Simple Fable App (F#)
Author: Alfonso Garcia-Caro; Maxime Mangel
Options:
  -f|--fake  If set then, the template will include Fake as build system.
             bool - Optional
             Default: false
```

I needed to customize the README.md and paket.lock file. 
README.md is not a problem however pakcet.lock can be [see the lines starting by `//` [see this line](https://github.com/fable-compiler/Fable/pull/1011/files#diff-ec67061d394802c0be69afc1f8a051abR568)

I believe if we do `paket.exe update` then they will be overwritten and we will need to add them again. Also, modifying a `paket.lock` isn't recommanded ^^ @forki Do you have an idea to solve the situations ?


Also, @alfonsogarciacaro I see that you make the template follow the Fable version. Is it intended ?

I tested the template in both mode with donet `2.0.0-preview1-005977` with success.

mike-morr:
@MangelMaxime Is there anyway to have fake surface the error message from ```dotnet fable yarn-start```?  Right now it just throws a cryptic stack trace for something simple like the Fable port already being in use.  I did get it to build on macOS though!

MangelMaxime:
@mike-morr Not sure it's worth having it in the templates. Fake is already consider complex, if we had custom integration in it users could get lost.

But yes you probably could add it, by parsing the error code. Please, note it's not the only error we can get. Fable port already used, etc.

mike-morr:
@MangelMaxime I agree, and great work on this!

forki:
What is the concrete issue with the lock file?

Am 02.07.2017 11:51 nachm. schrieb "Mike Morrison" <notifications@github.com

MangelMaxime:
The issue is I don't have the same dependencies if I use the template with Fake or without it. So the `paket.dependencies` and `paket.lock` are not the same depending if we want to use Fake or no.

Paket.dependencies:
```
source https://nuget.org/api/v2

nuget FSharp.Core

# Keep this version synced with dotnet-fable in .fsproj
nuget Fable.Core 1.1.5

# Add here other Fable libraries like Fable.PowerPack
nuget Fable.Import.Browser

//#if (fake)
group Build
framework: net46

    source https://nuget.org/api/v2
    nuget FSharp.Core  redirects:force, content:none
    nuget FAKE
//#endif
```

When the user install the template, `dotnet new` will remove the `Build` group if user don't specify to use Fake. That's why I am adding the comment: `//#if (fake) [....] //#endif`

It's not a bug in the lock file. I am just asking if we have another solution than manually editing the lock file to add the comments. Or if we do `paket.exe update` will this preserve the comments ? Or rewrite the whole file ?

forki:
no sorry we don't have comments in lock file.

MangelMaxime:
Ok thanks for the answer.

@alfonsogarciacaro What's your opinion with the situation ? To test the template we need to install it can't install the paket deps directly from the template folder, as we get this error:

```
Paket failed with:
-> Error during parsing of 'C:\Users\mange\Workspace\Github\fable-compiler\Fable\src\templates\simple\Content\paket.lock'.
-> Can't parse "fake".
   Le format de la chaîne d'entrée est incorrect.
```

IMO, we have more or less two options here:

1. Split the template in two. One `fable` (which use no Fake) and one `fable-fake` which use fake. IMO, it's less user friendly and give move confusion than needed.

2. Use one template with optional parameters fake (like done in this PR). We could edit the lock file at each update (which we will probably forget sometime^^) or let the lock file have the build group. This would download only two extra package: `FAKE` and `FSharp.Core` into the build folder. Without much more effect.

Next time the user update his deps via: `paket.exe update` it will remove the build group as it would no more be defined in the `paket.dependencies`

I am personally in favour of the solution n°2.

mike-morr:
@MangelMaxime Check this out https://github.com/xyncro/freya-template.  Not sure if it will help, but they have two versions of the template one with hopac and one with async.

MangelMaxime:
@mike-morr Yes but they don't use paket :) So they have no problem with the lock file.

forki:
I consider that a bug in Freya templates. ;-) 

But yes I already discussed with @enricosada about that. What happens if you have multiple templates that come with own paket dependencies. Dotnet cli is not made for this. And even if you use nuget that will lead to horrible issues. 

For current situation I think Best way would be to call paket install after template restore. Then it would just garbage collect the fake dependency

mike-morr:
These seem to be the lines we need to mirror in Paket.      
```xml
<PackageReference Include="Freya" Version="4.0.0-alpha-*" Condition="'$(freyaCore)' == 'async'"/>
<PackageReference Include="Freya.Hopac" Version="4.0.0-alpha-*" Condition="'$(freyaCore)' == 'hopac'"/>
```

forki:
@mike-morr yes but that's fundamentally different to what paket does. (and also wrong ;-))

MangelMaxime:
If they don't use Fake :) They aren't to one more command 👿 😄

forki:
Why do you want to scaffold without fake at all?

MangelMaxime:
Not me, I believe people want a version without Fake because it's harder to understand. If I understand well.

But IMO, if the people start with Fable it's better to just do `build.cmd` only.

forki:
So where are does the request come from?

MangelMaxime:
Hum hum... Don't find them, it's been a while since I started this work.

Or perhaps, I was trying to address a solution to @alfonsogarciacaro [message](https://github.com/fable-compiler/Fable/pull/1011#issuecomment-309403219)
And we should just debate, if we want user to use Fake or no with the simple template. Personally, I believe this is the way to go. But we probably need to add a disclaimer explaining why then need all this setup. 

I mean JS developers get used to all the boilerplate. But C# users at least the one I know, aren't because they are used to VS delegate everything to it. Nuget, transpilation via plugins, etc.

alfonsogarciacaro:
Awesome work @MangelMaxime, thanks for that! :clap: :clap:

Yes, I still would prefer to keep a minimal `dotnet new fable` template so it's easier for users to understand what are the exact steps to run a Fable project. Though I'm ok to add an option to include FAKE for users who already know it. Also, the README already includes a note about other templates, so more advanced users can search for more complete templates like Fable.Elmish.React or the fable-suave-scaffold.

About the comments in the lock file, I think this is a bug of the current dotnet templating and it should be fixed with dotnet CLI 2.0 (see [this](https://github.com/dotnet/templating/issues/432)). `postActions` are also coming so it should be possible to install dependencies automatically after restoring the template.

MangelMaxime:
The comment in the lock is "normal" :) And is not generated by dotnet CLI 2.0. :)
It's me adding it to make the template work in both mode (with and without fake).

postActions could be a solution to trigger `paket.exe install` after template install.

MangelMaxime:
@alfonsogarciacaro We put on hold waiting for the `postActions` and so dotnet 2.0 ?

It's mostly to explain why we don't merge yet :)

alfonsogarciacaro:
@MangelMaxime Sorry I haven't checked this in a while. What's the current situation? Is FAKE installed by default or do you need to set an option for that?

MangelMaxime:
Fake is only installed if we do `dotnet new fable -n MyProject -fake`

The latest concern of this PR, was the manual edition of **paket.lock** to add this two lines:
- https://github.com/fable-compiler/Fable/pull/1011/files#diff-ec67061d394802c0be69afc1f8a051abR568
- https://github.com/fable-compiler/Fable/pull/1011/files#diff-ec67061d394802c0be69afc1f8a051abR575

This avoid the user who don't use Fake to download it.

We could remove this lines, and so first time user will have Fake installed by paket (whether they use it or not). But the next time he do `.paket\paket.exe install` (or update) it will be removed as the [paket.dependencies](https://github.com/fable-compiler/Fable/pull/1011/files#diff-7ab7e49eaf6b3990e86023e87be44cd8) file will not list them.
We could automate that thanks to `postActions` but it's not available for dotnet 1.x

And of course, I will have to resolve the conflicts of this PR. :)

alfonsogarciacaro:
Ok, thanks for the explanation. If you don't mind, I'd prefer to wait a bit until dotnet SDK 2 (or at least 2 preview3) is out and the template engine is more stable. At the moment there are so many moving parts the chances we have to have to change things later is high, so I'd rather keep this template simple for the time being :smile:

MangelMaxime:
No problem :)

It's was just so we understand why the PR in on Hold since a long time :)

MangelMaxime:
I close this PR as the templates moved into another repo.

PR #1012, opened 2017-06-19
Fixed REPL decimals and build

ncave:
- Fixed REPL decimal constants (needs FCS_fable latest).
- Moved Decimal to ExtendedNumber so it can be easier to migrate to a different implementation.
- Unfortunately the nested option error was breaking the REPL build, so changed the error to a warning.


alfonsogarciacaro:
Great, thanks! 🔁

PR #1013, opened 2017-06-19
Updated Fable-kids-blog-post

whitetigle:
Minor corrections for blog post with: image update and text corrections


whitetigle:
OOps, will update PR from current state.

PR #1015, opened 2017-06-20
Update documentation regarding the JsonConvert plugin

coolya:
Previously it suggested to use a file reference on projects that are not dotnetcore.
Since the library is now also available for net45 from nuget that part is obsolet.

### DIFF
--- a/docs/source/docs/interacting.md
+++ b/docs/source/docs/interacting.md
@@ -513,7 +513,7 @@ Caveats:
 
 Fable.JsonConverter:
 
-You can install directly via [nuget](https://www.nuget.org/packages/Fable.Core/) by simply adding it as a dependency to you `paket.dependencies` file:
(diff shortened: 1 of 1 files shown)


docs/source/docs/interacting.md
```diff
@@ -513,26 +513,10 @@ Caveats:
 
 Fable.JsonConverter:
 
-You can install directly via [nuget](https://www.nuget.org/packages/Fable.JsonConverter/1.0.0-narumi-908) if your project is a dotnet core project by simply adding it as a dependency to you `paket.dependencies` file:
+You can install directly via [nuget](https://www.nuget.org/packages/Fable.Core/) by simply adding it as a dependency to you `paket.dependencies` file:
```
alfonsogarciacaro: Should be `Fable.JsonConverter`, right?

coolya: 🤦‍♂️ you are right.

alfonsogarciacaro:
Awesome, thank you!

PR #1016, opened 2017-06-20
Version check as a warning

ncave:
@alfonsogarciacaro Converted missing .nuspec build error to a warning.
Instead of adding --no-version-check flag everywhere, this avoids build crashing when referencing Fable.Core that does not have a .nuspec. Again, this is IMO a slightly better behavior than flags, but if you don't like it let me know and I'll close it.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -429,12 +429,12 @@ let runFableServer args f =
 
 let runTestsJs () =
     Npm.install __SOURCE_DIRECTORY__ []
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
You're right, it's probably too strict to fail when the .nuspec file is missing for testing scenarios where you usually make a direct reference to Fable.Core.dll. Thanks!

PR #1018, opened 2017-06-22
Add DU comparion test

forki:


forki:
@alfonsogarciacaro looks like this behaving correctly now. It is broken for me on fable 0.7

forki:
merging as regression test.

alfonsogarciacaro:
Thanks! I'm not sure how DU comparison is supposed to work but Fable 0.7 uses case names as tags (so probably they do alphabetical comparisons) while Fable 1.x uses indices. That should be the reason of the difference you spotted.

forki:
Alphabetical is definitely wrong ;-) 
That's what this test is about.

PR #1029, opened 2017-06-25
Prefix generated lambda expression arguments

coolya:
In situationes where the user defined variables like x and y it could lead
to naming conflicts with generated lambda expressions because these use the
same names. I deconflicted that by using __ as a prefix. __ is used in many
other languages (e.h C) to indicate a compiled namespace.

See #1028

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -1257,7 +1257,7 @@ module AstPass =
             CoreLibCall("Util", Some  "defaultArg", false, [expr; defValue; f])
             |> makeCall r Fable.Any
         let toArray r arg =
-            let ident = makeIdent "x"
+            let ident = makeIdent (com.GetUniqueVar())
             makeLambdaExpr [ident] (makeArray Fable.Any [Fable.IdentValue ident |> Fable.Value])
             |> runIfSome r arg (makeArray Fable.Any [])
         let getCallee() = match i.callee with Some c -> c | None -> i.args.Head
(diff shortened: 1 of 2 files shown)


coolya:
Any idea why the tests fail with this strange exception?

alfonsogarciacaro:
Thanks for this @coolya! Prefixing the variables with `__` should do the trick, but for consistency it may be better to use the `GetUniqueVar()` method [as in here](https://github.com/fable-compiler/Fable/blob/3e3249577a934b140e105e260bf137e45b366c38/src/dotnet/Fable.Compiler/Replacements.fs#L832). Could you please do the change? It'll be also nice to take the chance to fix the other places with hard-coded temp variable names in the Replacements module: [here](https://github.com/fable-compiler/Fable/blob/3e3249577a934b140e105e260bf137e45b366c38/src/dotnet/Fable.Compiler/Replacements.fs#L1260) and [here](https://github.com/fable-compiler/Fable/blob/3e3249577a934b140e105e260bf137e45b366c38/src/dotnet/Fable.Compiler/Replacements.fs#L1585).

About the tests failure, sometimes xunit fails in Travis for apparently random reasons. The way to fix is usually just to rebuild the commit ;)

coolya:
How about [theses](https://github.com/fable-compiler/Fable/blob/3e3249577a934b140e105e260bf137e45b366c38/src/dotnet/Fable.Compiler/Replacements.fs#L193)? Are they also candidates for name conflicts?

alfonsogarciacaro:
I guess those are fine because there's no reference to any variable outside in the body of lambda :)

coolya:
Are theses test failures related to my changes? It looks like they fail in the .net part and not JS tests.

alfonsogarciacaro:
Yeah, for some reason the .NET part of the tests fails a lot in Travis :grin: But don't worry, if the JS tests are passing that's all good 😃

PR #1033, opened 2017-06-29
Test DateTime in pattern match

forki:


forki:
@alfonsogarciacaro  did some change here. with fable 0.5 this is a runtime error.

alfonsogarciacaro:
I think almost nothing has changed for DateTime in Fable 1.x, but there were indeed some fixes about pattern matching and scoping, though I'm not sure if that was the problem here :/

forki:
it said d.getTime is not a function

alfonsogarciacaro:
Then it can be that something in DateTime has actually been fixed, I don't remember well 😅

PR #1034, opened 2017-06-29
There is no yarn-run command, but npm-run.

maestrow:


MangelMaxime:
Actually we added `yarn-run` in Fable CLI

PR #1035, opened 2017-06-30
Set exit codes

forki:
@alfonsogarciacaro would that help for https://github.com/fable-compiler/fable-react_native-demo/pull/20#issuecomment-312003952 ?

### DIFF
--- a/src/dotnet/Fable.Tools/Server.fs
+++ b/src/dotnet/Fable.Tools/Server.fs
@@ -41,8 +41,7 @@ let rec private loop timeout (server: TcpListener) (buffer: byte[]) (onMessage: 
             if data = SIGTERM
             then
                 Log.logAlways("Closing Fable daemon...")
-                Environment.Exit 1
-                return ()
+                return 1
             else
                 onMessage(data, fun (reply: string) ->
                     let msg = Encoding.UTF8.GetBytes(reply)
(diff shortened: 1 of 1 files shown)


src/dotnet/Fable.Tools/Server.fs
```diff
@@ -41,6 +41,7 @@ let rec private loop timeout (server: TcpListener) (buffer: byte[]) (onMessage:
             if data = SIGTERM
             then
                 Log.logAlways("Closing Fable daemon...")
+                Environment.Exit 1
```
enricosada: maybe instead of that,

```
Environment.Exit 1
return ()
```

to force exit, just do

```
return 1
```

i seen the flow, and error code is returned for all (from `p.ExitCode`), minus this part

alfonsogarciacaro:
Hmm, I'm not sure this will work. First, there're several situations where the server is stopped with SIGTERM but it's not an error (e.g., when user presses Ctrl+C). The way to propagate an error from Webpack, etc, is checking the ExitCode of the subprocess ([as in here](https://github.com/fable-compiler/Fable/blob/ffbfc2cd66b99839595b98bf0151e6443d63ba3c/src/dotnet/Fable.Tools/Main.fs#L164)). 
I will write more explanations at: https://github.com/fable-compiler/fable-react_native-demo/pull/20#issuecomment-312003952

PR #1036, opened 2017-06-30
Add paket 5 targets

forki:


alfonsogarciacaro:
Thanks for this! A couple of notes/questions:

- Please note this only affects Fable self-compilation, not Fable projects. You need to edit the templates for that.
- Isn't Paket updating the targets automatically when running `paket install`?
- The paket.exe binary has been modified. I was using magic mode, so I would expect that file doesn't need to change.

forki:
it's the newer bootstrapper

Issue #1043, opened 2017-07-03
System.Uri Type not found


coolya:
### Description

I tried to add replacements for some of the `System.Uri` static methods. The `UnescapeDataString` in particular. But when I use the `Uri` Type somewhere in the problem Fable throws an error even before replacing when it tried to load the FSharp AST.
### Repro code

Please provide the F# code to reproduce the problem.
```
System.Uri.UnescapeDataString "Kevin+van+Zonneveld%21" |> equal "Kevin van Zonneveld!"

/Users/kdummann/source/fsharp/Fable/src/tests/Main/MiscTests.fs(830,11): (830,14) error FSHARP: The value, constructor, namespace or type 'Uri' is not defined.
```

### Expected and actual results

At least the FSharp AST should be build and Fable should complain afterwards that there is no replacement. Any Ideas why parsing the AST already fails in this case?

alfonsogarciacaro:
Hmm, I guess a reference is missing somewhere. Does the code work if you compile it as a netcoreapp? @ncave Maybe we're missing something [here](https://github.com/fable-compiler/Fable/blob/51ed85132225491a2fe472051db98b08b4301c54/src/dotnet/Fable.Tools/ProjectCracker.fs#L210-L230)?

coolya:
Works fine if I compile the code `netcoreapp1.1`

alfonsogarciacaro:
Ok, thanks for the confirmation. Hopefully we can move soon to netcore 2.0 and leave behind all these holes in the API and crazy references 😄

coolya:
I think I found the problem ... and its a really weird one. 

`dotnet build` resolves `System.Runtime` to a nuget package:

```
         -r:/Users/kdummann/.nuget/packages/system.runtime/4.3.0/ref/netstandard1.5/System.Runtime.dll
```

But we resolve it to:

```
/usr/local/share/dotnet/shared/Microsoft.NETCore.App/1.0.5/System.Runtime.dll
```

Which is strange assembly only containing some basic subset of the original. The actual uri types are located in 

```
/usr/local/share/dotnet/shared/Microsoft.NETCore.App/1.0.5/System.Private.Uri.dll
```

I was able to make the code compile by adding:

```
        yield "-r:" + resolve "System.Private.Uri"
```

to the `ProjectCracker` but I don't know if this is the right way to fix it.

alfonsogarciacaro:
[...]

PR #1048, opened 2017-07-04
Replace syste.uri

coolya:
Added replacements for the static methods of the System.Uri class:

- `UnescapeDataString`: To unescape a URI/Data string
- `EscapeDataString`: Escapes a string to a repesentation safe to put in URI, this includes characters that are safe in URI but should be escaped.
- `EscapeUriString`: Escapes a URI, all characters valid in a URI will remain.

In order to make it work I had to add System.Private.Uri to the compiler path. See #1043 for details.

### DIFF
--- a/src/typescript/fable-core/Util.ts
+++ b/src/typescript/fable-core/Util.ts
@@ -532,3 +532,18 @@ export function parse<A>(v: string | null, initial: A, parser: RegExp, fn: (s: s
     throw new Error("Input string was not in a correct format.");
   }
 }
+
+export function urldecode(s: string): string {
+  // https://stackoverflow.com/a/4458580/524236
+  return decodeURIComponent((s).replace(/\+/g, "%20"));
+}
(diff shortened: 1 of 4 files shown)


src/typescript/fable-core/Util.ts
```diff
@@ -532,3 +532,18 @@ export function parse<A>(v: string | null, initial: A, parser: RegExp, fn: (s: s
     throw new Error("Input string was not in a correct format.");
   }
 }
+
+export function urldecode(s: string): string {
```
coolya: I will change the name to match the one from the Sytem.Uri class

alfonsogarciacaro: Should we wait then until the name is changed before merging?

coolya: done

alfonsogarciacaro:
Awesome, thanks!

PR #1049, opened 2017-07-04
Fixed Repl fsproj

ncave:
- fixed Repl fsproj
- building fable-splitter in situ

### DIFF
--- a/src/dotnet/Fable.Client.Browser/Fable.Client.Browser.fsproj
+++ b/src/dotnet/Fable.Client.Browser/Fable.Client.Browser.fsproj
@@ -20,8 +20,6 @@
     <Compile Include="../Fable.Core/AST/AST.Babel.fs" />
     <Compile Include="../Fable.Core/Plugins.fs" />
     <Compile Include="../Fable.Core/Import/Fable.Import.JS.fs" />
-    <!--<Compile Include="../Fable.Core/Import/Fable.Import.Browser.fs" />-->
-    <!--<Compile Include="../Fable.Core/Import/Node/*.fs" />-->
     <Compile Include="../Fable.Core/Fable.Core.JsInterop.fs" />
     <Compile Include="../Fable.Core/Fable.Core.Extensions.fs" />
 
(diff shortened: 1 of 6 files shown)


src/dotnet/Fable.Client.Browser/Fable.Client.Browser.fsproj
```diff
@@ -30,8 +28,7 @@
     <Compile Include="../Fable.Compiler/FSharp2Fable.Util.fs"/>
     <Compile Include="../Fable.Compiler/FSharp2Fable.fs"/>
     <Compile Include="../Fable.Compiler/Fable2Babel.fs"/>
-
-    <Compile Include="../Fable.Tools/State.fs"/>
+    <Compile Include="../Fable.Compiler/State.fs"/>
```
alfonsogarciacaro: I may have asked you already but, can't we just use ProjectReferences here for Fable.Core and Fable.Compiler, they're compatible with Fable too.

ncave: @alfonsogarciacaro Unfortunately no, as we need to compile them with the FABLE_COMPILER flag. This could just be my ignorance though, let me know if there is an easy way to do that.

ncave: Obviously we can have another Fable.Compiler/Fable.Compiler_fable.fsproj project there that uses the flag, and perhaps that's better as they will be together, so it's more likely to update both, but I wonder if there is a way to do this with one project.

alfonsogarciacaro: Fable merges all the files of current and referenced projects into a single project and then [adds the basic compiler args](https://github.com/fable-compiler/Fable/blob/61d44e078e6f6068365ffde11bb460364509441b/src/dotnet/Fable.Tools/ProjectCracker.fs#L403) which always contains the `FABLE_COMPILER` flag so this shouldn't be a problem.

ncave: @alfonsogarciacaro Yes, it does, I actually mean compile with dotnet. It helps a lot to be able to compile the same source with dotnet, for sanity check.

alfonsogarciacaro: I think there's a way to set the configuration of the referenced project and then use that configuration as a condition to set the FABLE_COMPILER flag. At least in the old project format, not sure if it's the same (or maybe there's an easier way) in the new project. Summoning @enricosada the wizard :wink:

ncave: Yeah, perhaps referencing a specific target with Targets="sometarget" can work, but I have not tried it.

alfonsogarciacaro: Ok, thanks a lot for the clarification. We can check that later and merge this now :+1:

PR #1050, opened 2017-07-05
Adding test case for IsNullOrWhiteSpace

forki:


forki:
It looks like the issue is that s.Length is undefined

alfonsogarciacaro:
The test is passing :confused: AppVeyor only fails because of the Async.Cancellation test as usual, I'll have to kill that for dotnet testing at the end.

forki:
yes I have to types of strings apparently. And for one s.Length is undefined.

alfonsogarciacaro:
Sorry, not sure I understand what the problem is. Do you have a failing sample?

forki:
working on it.

forki:
ok found the issue. Json parser restores the field as datetime (but property type is string) and therefor it looks like a string but isn't

alfonsogarciacaro:
Yeah, that's a problem because there's no DateTime type for JSON serialization so we're just using [a regex to recognize DateTime](https://github.com/fable-compiler/Fable/blob/8c6758be9b0e1d0e8428fa93da62d481dfe94a68/src/js/fable-core/Serialize.ts#L353) (in ISO format).

PR #1052, opened 2017-07-05
Support exception equality

i-p:
F# exceptions compiled to JS are missing methods for equality comparison. As a result, equal F# exceptions are not equal in JS.


alfonsogarciacaro:
Nice one, thanks! :clap: :clap:

PR #1053, opened 2017-07-06
Array inside keyValueList is lost forever

2sComplement:
Not sure if this is expected behaviour, but an Array inside a list of DU's being passed into `keyValueList` is being transpiled as a plain object.

```
  1233 passing (5s)
  1 failing

  1) Arrays Array inside keyValueList is preserved:

      Error: Expected: [object Object] - Actual: [object Object]
      + expected - actual

       {
      -  "names": {
      -    "Name": "name"
      -  }
      +  "names": [
      +    {
      +      "Name": "name"
      +    }
      +  ]
       }
```


alfonsogarciacaro:
This was indeed an error of the compile-time resolution of `keyValueList`. Thanks a lot for pointing it out!

PR #1058, opened 2017-07-10
Support String.Join with objects

i-p:
Implementation of "join" function in JS is missing a call to "toString" for every value and doesn't process multiple non-string arguments.

As a result, following tests return ``[object Object]--[object Object]`` and an empty string when compiled to JS:
```
     System.String.Join("--", [|3I; 5I|])
      |> equal "3--5"
      System.String.Join("--", 3I, 5I)
      |> equal "3--5"
```


alfonsogarciacaro:
Thanks!

PR #1060, opened 2017-07-11
Fixed CI issue

ncave:


alfonsogarciacaro:
Great, thanks. Apparently updating npm worked (so the package-lock.json) is respected, but this will be needed anyways when updating tslint :+1:

PR #1061, opened 2017-07-11
Updated lock file

ncave:
regenerated to prevent changes on build.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -119,13 +119,13 @@
       "integrity": "sha1-jzuCf5Vai9ZpaX5KQlasPOrjVs8=",
       "dev": true,
       "requires": {
-        "arr-flatten": "1.0.3"
+        "arr-flatten": "1.1.0"
       }
     },
     "arr-flatten": {
-      "version": "1.0.3",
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Hmm, package-lock.json should have been regenerated in 1e188c7a4aacd1334aaf614283debf3e7385032a

Which npm version did you use to create this lock file? (I used 5.2.0) Looks like most of the changes are about the format of the lock file itself, not the versions of the packages.

BTW, it's true that npm 5 doesn't update the lock file by default when package.json is modified and `npm install` is run. The lock file has to be manually deleted first. I hope they address that quickly.

ncave:
5.0.3 is what comes with node 8.1.3, but I see you're on the cutting edge with 5.2.0. I regenerated it with 5.2.0.

alfonsogarciacaro:
Thanks! I still wonder why your lock file is different from the one I generated after updating package.json, but I guess that's fine :)

alfonsogarciacaro:
Your PR is the first one to pass Travis in a long time so that's a good sign ;)

ncave:
> Thanks! I still wonder why your lock file is different from the one I generated after updating package.json, but I guess that's fine :)

“No man ever steps in the same river twice, for it’s not the same river and he’s not the same man”.
Heraclitus

PR #1065, opened 2017-07-12
Repl target fix

ncave:
- Repl target fix
- downgraded reraise level from error to warning

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -620,7 +620,7 @@ Target "All" (fun () ->
 // - In `FSharp.Compiler.Service_fable` run `build CodeGen.NetCore -d:FABLE_COMPILER`
 // > Attention: the generation of libraries metadata is not included in this target
 Target "REPL" (fun () ->
-    let replDir = "src/dotnet/Fable.Client.Browser/demo"
+    let replDir = "src/dotnet/Fable.JS/demo"
     Npm.install replDir []
 
     // Compile fable-core
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -894,7 +894,7 @@ module AstPass =
             Fable.Throw (args.Head, typ, r) |> Some
         | "reraise" ->
             "`reraise` is not yet supported, please use `raise` instead"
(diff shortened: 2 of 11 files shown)


alfonsogarciacaro:
Great @ncave, thank you!

I think we may start creating a JS-only FCS/Fable package. I think I've asked you like a million times but the responses get lost in the issue labyrinth: so can we add the FCS files from your fork as a github dependency? Is the Fslex step still needed? Where should we put the needed dlls?

As you suggested before, maybe we should create a new repo for the REPL/JS package and make self-buildable without external dependencies on the system to make it easier for starting contributors?

ncave:
@alfonsogarciacaro Sure, you can add a paket github dependency to FCS-fable, it's going to bloat Fable somewhat but I guess that's ok. Unfortunately the fslexyacc step is still needed as there is no netcore tooling for that and I see no plans for it (even though it should be fairly easy to port it now with netstandard 2.0). We already have the dlls in the github pages, so they can be paket referenced from there, or somewhere else.

On the other hand, if we flip this around, all this does not need to happen in Fable, it can (and should) be in a separate repo. I was suggesting starting a new repo that can have all that and also paket-reference Fable Core and Fable Compiler for the source. I have a proto repo called [fable-repl](https://github.com/ncave/fable-repl), for now it just has the monaco bindings. If you agree, I can transfer its ownership to the fable-compiler org and we continue the work there. The old repl can stay in Fable until the new one is operational, to minimize throw-away effort.

Let me know what you think is the best way forward.

alfonsogarciacaro:
Absolutely, I think that's a great idea! Can you please transfer the ownership of the repo and we can continue work there? Thanks a lot!

ncave:
@alfonsogarciacaro Done. To to clarify, I still think you're right to push for a JS-only FCS/Fable package, so it can be used not only bu the fable-repl but other repls too. Which repo that belong into, this is the question - fable, fable-repl or a new one, say fable-js.

MangelMaxime:
I think it's should not belongs to fable-repl as it's not explicit that user could use JS-only FCS/ from it.

Personally, I tend in favor of fable-js

PR #1068, opened 2017-07-14
the fsproj is now in src

forki:


forki:
Yes this was only quick fix

Am 14.07.2017 5:01 nachm. schrieb "Alfonso Garcia-Caro" <
notifications@github.com>:

PR #1069, opened 2017-07-14
Prevent dotnet template engine to parse Paket.Restore.targets

MangelMaxime:
@alfonsogarciacaro Careful when you run `paket.exe update/install` perhaps this will override this changes.

@forki could probably confirm/infirm the behaviour of paket on this file. :)

### DIFF
--- a/src/templates/simple/Content/.paket/Paket.Restore.targets
+++ b/src/templates/simple/Content/.paket/Paket.Restore.targets
@@ -1,4 +1,5 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--/-:cnd:noEmit-->
   <PropertyGroup>
     <!-- Mark that this target file has been loaded.  -->
     <IsPaketRestoreTargetsFileLoaded>true</IsPaketRestoreTargetsFileLoaded>
@@ -127,5 +128,5 @@
               NuspecBasePath="$(NuspecBasePath)"
               NuspecProperties="$(NuspecProperties)"/>
   </Target>
-
+  <!--/+:cnd:noEmit-->
 </Project>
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Sorry, I don't understand exactly why this change is needed, could you explain more? I usually use the template as a normal project to test everything works before publishing a new version, so I'm afraid I or another contributor at one point will run `paket update` and forget about this. Also, if this is something necessary for Paket, shouldn't it be added directly to Paket targets as it will affect all templates using Paket?

forki:
This is a workaround against dotnet template engine bug. Nothing to do with
paket

Am 15.07.2017 12:07 nachm. schrieb "Alfonso Garcia-Caro" <
notifications@github.com>:

alfonsogarciacaro:
@forki I see, but I think my points above are still valid. Sorry, but I'm not crazy supporting a non-stable version of dotnet SDK by adding patches and maintenance chores... specially not in Summer :wink: 😎 🌞 🌻 

I'd rather change the README to say dotnet SDK 2.0 is not supported yet or add the recommendation to run `paket install` after running the template.

forki:
Just take a look at recent fable gitter. Everyone runs into this. Everyone. So we need to do something...

alfonsogarciacaro:
These are the scenarios I foresee:

- A couple of weeks later I make a fix to Fable and want to update the template. I run `paket update` to test the template works with latest Fable, this overwrites the change from this PR, and the chances that I forget to undo that while I'm in a hurry and focusing on another thing are really high.
- Even if I remember to undo the change for Paket.restore.targets, what if Paket is actually changing something to fix another issue? Do I have to check the diffs every time and move the `<!--/-:cnd:noEmit-->` comments manually?

Because of this, I would prefer an alternative solution:

- State in the README that Fable explicitly requires dotnet SDK 1.0.4; or
- Add the recommendation to run `paket install` if you're using dotnet SDK 2 (actually I think with dotnet SDK 2 you can run operations automatically).
- If this affects all templates, why not add it to Paket targets directly? It's just a comment and shouldn't affect normal Paket users.

forki:
Yes adding it to paket itself is probably best.

PR #1075, opened 2017-07-16
Added failing test for missing Option.flatten.

irium:


alfonsogarciacaro:
Ups, I didn't even know this existed 😮 But Fable now warns you against nested options, right?

Anyways, this shouldn't be difficult to implement, I'll do it.

irium:
> Ups, I didn't even know this existed 😮 But Fable now warns you against nested options, right?

Not always :) Such result maybe returned from functions like `<*>` (`apply`) mentioned [here](https://github.com/fable-compiler/Fable/blob/master/src/tests/Main/ApplicativeTests.fs#L530) in recent PR #1071 when `f` itself returns `Option`.

BTW, all `Option` public APIs could be seen [here](https://github.com/fsharp/fsharp/blob/master/src/fsharp/FSharp.Core/option.fsi).

PR #1082, opened 2017-07-18
Save the daylight!

forki:
'isDaylightSavingTime' is not exported by packages\Fable.Core\fable-core\Date.js

### DIFF
--- a/src/tests/Main/DateTimeTests.fs
+++ b/src/tests/Main/DateTimeTests.fs
@@ -130,6 +130,14 @@ let ``DateTime.IsLeapYear works``() =
     DateTime.IsLeapYear(2016) |> equal true
 
 [<Test>]
+let ``DateTime.IsDaylightSavingTime works``() =
+    let d1 = DateTime(2017, 7, 18, 2, 0, 0, DateTimeKind.Utc)
+    let d2 = DateTime(2017, 12, 18, 2, 0, 0, DateTimeKind.Utc)
+    d1.IsDaylightSavingTime() |> equal true
+    d2.IsDaylightSavingTime() |> equal false
+    
+
+[<Test>]
 let ``DateTime.DaysInMonth works``() =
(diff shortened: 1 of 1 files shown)


inosik:
Should be possible with this: https://stackoverflow.com/a/30280636.

Your test will always fail, because `DateTime`s with `DateTimeKind.Utc` are never DST. They either have to be `Local` or `Unspecified`.

Also, I don't know if there is an appropriate test for this at all, because some countries in the world don't have a DST or "summer time" anymore. So, testing if a certain date is DST or not will always fail for a contributor from Russia, for example.

forki:
can we create a test for german daylight settings?

inosik:
`DateTime.IsDaylightSavingTime()` gets its return value from [`TimeZoneInfo.Local`][src], I don't know if we can change the time zone in .NET.

  [src]: http://referencesource.microsoft.com/#mscorlib/system/datetime.cs,703

forki:
I think you can't. but we catually only want to do it in the js

inosik:
What about this:

``` javascript
// t is the date object to check, returns true if daylight saving time is in effect.
function isDaylightSavingTime(t) {
    var jan = new Date(t.getFullYear(), 0, 1);
    var jun = new Date(t.getFullYear(), 6, 1);
    return isDST(jan.getTimezoneOffset(), jun.getTimezoneOffset(), t.getTimezoneOffset());
}

// The real logic
function isDST(janOffset, junOffset, tOffset) {
    return Math.min(janOffset, junOffset) == tOffset;  
}
```

And we only test `isDST` with known values?

alfonsogarciacaro:
@inosik Looks good! Can you PR it?

inosik:
I can start with the change to `fable-core/Date.ts`, but you'd need to guide me from there on :sweat_smile:

alfonsogarciacaro:
Sure! Just send the PR to `Date.ts` and I'll take care of editing the `Replacements` module :+1:

PR #1083, opened 2017-07-18
Implement isDaylightSavingTime

inosik:
References #1082.

Because JS doesn't have a native function for this, we need to check if the `Date`s UTC offset is equal to either the offset in January or in June.

If a users time zone doesn't have DST, this function should always return `false`.

@alfonsogarciacaro where can I add tests for the `isDST` function?

### DIFF
--- a/src/js/fable-core/Date.ts
+++ b/src/js/fable-core/Date.ts
@@ -265,3 +265,13 @@ export function op_Addition(x: Date, y: number) {
 export function op_Subtraction(x: Date, y: number | Date) {
   return subtract(x, y);
 }
+
+export function isDaylightSavingTime(x : Date) {
+  var jan = new Date(x.getFullYear(), 0, 1);
+  var jun = new Date(x.getFullYear(), 6, 1);
+  return isDST(jan.getTimezoneOffset(), jun.getTimezoneOffset(), x.getTimezoneOffset());
+}
+
+function isDST(janOffset : number, junOffset : number, tOffset : number) {
+  return Math.min(janOffset, junOffset) == tOffset;
+}


alfonsogarciacaro:
Thanks a lot for this! @forki's test should be enough to check `isDaylightSavingTime`, do you want to test `isDST` separately?

inosik:
I actually wanted to test it with some predefined values, but maybe it's just me being paranoid. :smile: 

In this case, there's probably no reason to make it a separate function.

forki:
Don't worry. If it's releasedocs then I will write UI tests for our app
starten and test it indirectly ;-)

Am 18.07.2017 12:54 schrieb "Ilja Nosik" <notifications@github.com>:

I actually wanted to test it with some predefined values, but maybe it's
just me being paranoid. 😄

In this case, there's probably no reason to make it a separate function.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/fable-compiler/Fable/pull/1083#issuecomment-316028841>,
or mute the thread

forki:
@inosik thanks for implementing it.
you win a music video: https://www.youtube.com/watch?v=7aJYK3PnWqU

inosik:
Sweet! Now I need to win some speakers for my workstation :smile:

alfonsogarciacaro:
This is what I see most of the times when I try to share a video between Germany and Spain ;)

screen shot 2017-07-18 at 1 42 07 pm

forki:
sorry guys you are missing out.

inosik:
Unfortunately, the fix [breaks Travis CI][failure]. AppVeyor seems to be completely broken right now.

  [failure]: https://travis-ci.org/fable-compiler/Fable/builds/254814708#L4613

alfonsogarciacaro:
Luckily my advanced hacker skills prevail over Youtube algorithm... by typing the video title in the Search Bar ;)

alfonsogarciacaro:
I hate Travis, it's been failing continuously for a while now for not apparent reason 😬

inosik:
AppVeyor doesn't seem to run at all at the moment, but Travis fails because of the test from #1082.

inosik:
``` javascript
// index.js
function isDaylightSavingTime(x) {
    var jan = new Date(x.getTimezoneOffset(), 0, 1);
    var jul = new Date(x.getTimezoneOffset(), 6, 1);
    return isDST(jan.getTimezoneOffset(), jul.getTimezoneOffset(), x.getTimezoneOffset());
}

function isDST(janOffset, julOffset, tOffset) {
    return Math.min(janOffset, julOffset) === tOffset;
}

var d1 = new Date(2017, 6, 18, 2, 0, 0);
var d2 = new Date(2017, 11, 18, 2, 0, 0);

console.log(d1.toString() + " is DST: " + isDaylightSavingTime(d1).toString());
console.log(d2.toString() + " is DST: " + isDaylightSavingTime(d2).toString());
```

```
$ node index.js
Tue Jul 18 2017 00:00:00 GMT+0200 (Mitteleuropäische Sommerzeit) is DST: true
Mon Dec 18 2017 00:00:00 GMT+0100 (Mitteleuropäische Zeit) is DST: false
```

Works for me. Must be the time zone of the Travis instance.

alfonsogarciacaro:
Let's see what Appveyor says and, if necessary, disable that test in the CI servers.

inosik:
As I wrote in [#1082](https://github.com/fable-compiler/Fable/pull/1082#issuecomment-316005395), this test will fail for everybody who runs it in a country without DST. That's why I wanted to test the `isDST` function instead of `DateTime.IsDaylightSavingTime()`.

Looks like we can [control the time zone in Travis CI][travis] with the `TZ` environment variable and [in AppVeyor with `tzutil`][appveyor], which is a Windows program. But I don't think that it's a good idea to rely on this.

  [travis]: https://stackoverflow.com/a/27052708
  [appveyor]: https://stackoverflow.com/a/33617323

forki:
can ask for the current zone? we might then just enable US, Germany and that's it

PR #1091, opened 2017-07-24
Check that distinctBy doesn't stackoverflow

forki:


forki:
I can confirm that this is fixed by @alfonsogarciacaro in lastest Fable.

wohoo

alfonsogarciacaro:
Related commit: https://github.com/fable-compiler/Fable/commit/8366cd7788f7df9e2ce81e88280167a27e535462

PR #1099, opened 2017-07-30
Improve fable-splitter and add watch mode

alfonsogarciacaro:
This is actually a present for @forki 💋 as fable-splitter is probably a much better fit for React Native instead of Rollup, which we're using only for the watch mode (RN has its own bundler). This should improve compilation speed and enable Live Reload.

@ncave These are the changes I've introduce to the splitter:

- Add CLI
- Put files in same directory as entry file in outDir's root (otherwise it's probably counterintuitive)
- Add watch mode (when passing `-w` to the CLI)

> BTW, I've disabled tslint for now as it's telling me all the time my code sucks :wink: Just kidding, but it's a bit annoying to fix the tslints errors with the messages from the build. @ncave, do you use any tool to see the errors in the editor?

I'll try to update the README later, publish and use it with the RN demo to check if it works :+1:

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -366,7 +366,7 @@ let buildCoreJS () =
 let buildSplitter () =
     let buildDir = __SOURCE_DIRECTORY__ </> "src/js/fable-splitter"
     Npm.install __SOURCE_DIRECTORY__ []
-    Npm.script __SOURCE_DIRECTORY__ "tslint" [sprintf "--project %s" buildDir]
+    // Npm.script __SOURCE_DIRECTORY__ "tslint" [sprintf "--project %s" buildDir]
     Npm.script __SOURCE_DIRECTORY__ "tsc" [sprintf "--project %s" buildDir]
 
 let buildCore isRelease () =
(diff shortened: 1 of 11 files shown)


ncave:
@alfonsogarciacaro Those improvements sound great, thanks. Yes, there is a TSLint plugin for VSCode that help a lot.

forki:
I love presents!

Am 31.07.2017 05:36 schrieb "ncave" <notifications@github.com>:

build.fsx
```diff
@@ -366,7 +366,7 @@ let buildCoreJS () =
 let buildSplitter () =
     let buildDir = __SOURCE_DIRECTORY__ </> "src/js/fable-splitter"
     Npm.install __SOURCE_DIRECTORY__ []
-    Npm.script __SOURCE_DIRECTORY__ "tslint" [sprintf "--project %s" buildDir]
+    // Npm.script __SOURCE_DIRECTORY__ "tslint" [sprintf "--project %s" buildDir]
```
forki: Remove comment

PR #1105, opened 2017-08-05
Template update

MangelMaxime:
- Update the deps to latest versions
- Use paket to handle FSharp.NET.Sdk

I propose to manage FSharp.NET.Sdk via paket because there seems to have issues when opening a clean template into VS2017. And like that alls the deps are manage by paket or npm/yarn.

@alfonsogarciacaro I already updated the version in the `.proj` file


alfonsogarciacaro:
True, better to manage FSharp.NET.Sdk with Paket. But it's still not possible to open the new .fsproj format with VS 2017 yet, right?

MangelMaxime:
I believe it's opening now. I don't use VS anymore myself. 


cc @Zaid-Ajaj

Zaid-Ajaj:
@alfonsogarciacaro Nope, VS2017 on Windows seems to be looking for a globally installed FSharp.NET.Sdk on the machine that you don't get even you install the "F# Language Support". Installing the SDK  locally with the project dependencies, VS2017 still complains and can't load the project. VS2017 really not worth it, VS Code "just works"

PR #1110, opened 2017-08-16
JS line endings

### LINKED ISSUE #1109: Need to republish fable-splitter
### Description

On Linux Mint (and probably others unix), if we try to use fable-splitter I got this error: 

`[1] /usr/bin/env: ‘node\r’: No such file or directory`

This is caused because the end of line used in the `cli.js` file is windows specific. (From my research).
If I convert the file using `dos2unix` then I can use fable-splitter

      
> The Dos2unix package includes utilities "dos2unix" and "unix2dos" to
>        convert plain text files in DOS or Mac format to Unix format and vice
>        versa.
> 
> In DOS/Windows text files a line break, also known as newline, is a
>        combination of two characters: a Carriage Return (CR) followed by a
>        Line Feed (LF). In Unix text files a line break is a single character:
>        the Line Feed (LF). In Mac text files, prior to Mac OS X, a line break
>        was single Carriage Return (CR) character. Nowadays Mac OS uses Unix
>        style (LF) line breaks.

I checked the version of the file in the Github repo and it's looks like to have the right end of line. I suppose it's a problem when one of you published the fable-splitter version on npm.
 @ncave @alfonsogarciacaro

ncave:
Closes #1109

### DIFF
--- a/src/js/.gitattributes
+++ b/src/js/.gitattributes
@@ -1,0 +1,63 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
+###############################################################################
+* text=auto eol=lf
+
+###############################################################################
+# Set default behavior for command prompt diff.
+#
+# This is need for earlier builds of msysgit that does not have it on by
+# default for csharp files.
+# Note: This is only used by command line
+###############################################################################
+#*.cs     diff=csharp
+
+###############################################################################
+# Set the merge driver for project and solution files
+#
+# Merging from the command prompt will add diff markers to the files if there
+# are conflicts (Merging from VS is not affected by the settings below, in VS
+# the diff markers are never inserted). Diff markers may cause the following 
+# file extensions to fail to load in VS. An alternative would be to treat
+# these files as binary and thus will always conflict and require user
+# intervention with every merge. To do so, just uncomment the entries below
+###############################################################################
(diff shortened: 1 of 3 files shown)


ncave:
@alfonsogarciacaro It may be ok to switch all text files to LF endings in the root .gitattributes instead of just src/js/*, but I'll leave that decision up to you.

ncave:
@alfonsogarciacaro Just fyi, this is already published to npm.

alfonsogarciacaro:
Great, thanks! Does .gitattributes affect also the local files? I think I already have "commit UNIX endings" in global git settings, but the problem is npm packages are pushed from the local versions.

ncave:
@alfonsogarciacaro yes it does, it forces checkouts to be LF only.

alfonsogarciacaro:
Ah, cool! But maybe there's still the risk that I (or someone else) make a change to the file with the shebang in Windows (so the `\r` character is introduced) and publish right away. Probably I should do like I did for the old fable-compiler package and make the file with a shebang just call the cli.js file so it doesn't need to be touched very often. Or at least putting a comment making sure the line endings must be correct before publishing.

ncave:
@alfonsogarciacaro Some editors warn you that you have mixed line endings, and some don't.

PR #1119, opened 2017-08-29
Adding a test for negative number for TryParse

FrankBro:
Should pass but fails right now.


alfonsogarciacaro:
Strange, tests pass on my machine and at least on Travis after merging the PR 🤔

MangelMaxime:
Last time I checked on the repl it was failing but i couldn't understand the bug looking at fable core...

PR #1125, opened 2017-09-01
updated REPL to FCS latest

ncave:
- Because of the new [error](https://github.com/fable-compiler/Fable/blob/master/src/dotnet/dotnet-fable/ProjectCracker.fs#L227) I had to add Fable.Core references in all of the REPL projects, but that's fine.
- Webpack creates some problems for the FableMap constructor that takes comparer, so I had to switch to Rollup which works fine.
- Please make sure you are using the latest [FCS-fable](https://github.com/ncave/FSharp.Compiler.Service/tree/fable).


ncave:
@alfonsogarciacaro Let me know if you want me to republish the REPL github pages, perhaps after Fable 1.2.0 is finalized.

alfonsogarciacaro:
Thanks as always!

> Because of the new error I had to add Fable.Core references in all of the REPL projects, but that's fine.

Argh! I should have imagined this would affect the REPL build. Anyway, the error should probably be removed if the `fableCore` option is set, and even if it's not, we may replace it with a warning.

> Webpack creates some problems for the FableMap constructor that takes comparer, so I had to switch to Rollup which works fine.

Interesting. Do you know which kind of problems? May this affect other Fable projects?

> Let me know if you want me to republish the REPL github pages, perhaps after Fable 1.2.0 is finalized.

Yes, we can update the REPL when Fable 1.2.0 is out (which has to be soon because... FableConf is coming). I'm also updating the web to make updates and contributions easier :+1:

ncave:
> Interesting. Do you know which kind of problems? May this affect other Fable projects?

It's possible. It has something to do with the fable-core/Map constructor that takes comparer:
```
Uncaught TypeError: Constructor Map requires 'new'
    at e.Map (native)
    at new e (http://localhost:8080/repl/bundle.min.js:1:634995)
    at Function.value (http://localhost:8080/repl/bundle.min.js:1:635437)
    at Object.<anonymous> (http://localhost:8080/repl/bundle.min.js:1:818973)
...
```
Again, it works fine with rollup and splitter so I did not investigate further.

alfonsogarciacaro:
That rings a bell... I think it happened with the first versions of Webpack 3.x. It may be a good idea to test again with latest Webpack version to check if this has been fixed.

ncave:
@alfonsogarciacaro The repl uses webpack 3.3.0 from the root package.json, so perhaps now is a good time to update the npm dependencies to latest, before fable 1.2.0 is out.

PR #1127, opened 2017-09-02
Fix 2 digits years datetime

MangelMaxime:
In JavaScript date have this behavior:

> If a year between 0 and 99 is specified, the method converts the year to a year in the 20th century (1900 + year); for example, if you specify 95, the year 1995 is used.

In F#, if you set a datetime via `Datetime(1, 1, 1)` you want the year 1 and not 1901.

### DIFF
--- a/src/js/fable-core/Date.ts
+++ b/src/js/fable-core/Date.ts
@@ -60,6 +60,11 @@ export function create(
   } else {
     date = new Date(Date.UTC(year, month - 1, day, h, m, s, ms));
   }
+
+  if (year <= 99) {
+    date.setFullYear(year);
+  }
+
   if (isNaN(date.getTime())) {
     throw new Error("The parameters describe an unrepresentable Date.");
   }
--- a/src/tests/Main/DateTimeTests.fs
+++ b/src/tests/Main/DateTimeTests.fs
@@ -49,6 +49,20 @@ let ``DateTime can be JSON serialized forth and back``() =
     utc.Kind = DateTimeKind.Utc |> equal true
     utc.ToString("HH:mm") |> equal "17:30"
 
+[<Test>]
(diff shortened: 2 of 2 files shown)


alfonsogarciacaro:
Good catch, thanks! 📅

ncave:
The second test is always failing for me locally for some reason, idk why is it passing in CI.
```
  1268 passing (5s)
  1 failing

  1) DateTime DateTime UTC from Year 1 to 99 works:

      Error: Expected: 1 - Actual: 2
```

MangelMaxime:
That's really strange... Because it's the year and we use UTC so there no timezone taken into account here...

What is the result you get when running the same code in your fsi ?

ncave:
@MangelMaxime The dotnet version prints 1, but the javascript version (i.e. fable-core/Date/year) consistently prints 2.

ncave:
@MangelMaxime If that helps, if I change the DateTimeKind to Local, it prints 1 for the year. Or if I change the month or the day to 2 or more, it prints 1 for the year. Confusing, I know.

MangelMaxime:
@ncave So this have something to do with the edge case.

The date used for the tests is `0001-01-01` (1 January of year 0001).


Can you please tell me the result of each line of this code on your machine ? (It's JavaScript code) I am trying to break down the code step by step to understand the situation.

```js
var d = new Date(Date.UTC(1, 0, 1, 0, 0, 0))
d.toString()
d.setFullYear(1)
d.toString()
d.getYear()
d.getUTCFullYear()
```

ncave:
@MangelMaxime Fixed in #1139.

MangelMaxime:
Ah thanks :)

I love how `setFullYear` takes month and day too... as argument.

PR #1128, opened 2017-09-05
Fix tests targets and string format over local dates

MangelMaxime:


alfonsogarciacaro:
Awesome, thanks a lot for this! Hmm, the CI servers are complaining with the following:

```
> @ mocha /home/travis/build/fable-compiler/Fable
> mocha "./build/tests/bundle.js"
/home/travis/build/fable-compiler/Fable/build/tests/bundle.js:11652
        throw new AssertionError(msg || "Expected: " + expected + " - Actual: " + actual, actual, expected);
        ^
Error: Expected: 2014-09-11T16:37:02.000+02:00 - Actual: 2014-09-11T16:37:02.000+00:00
    at AssertionError.Error (native)
    at new AssertionError (/home/travis/build/fable-compiler/Fable/build/tests/bundle.js:11641:122)
    at equal (/home/travis/build/fable-compiler/Fable/build/tests/bundle.js:11652:15)
```

Maybe we could separate the test paths fix and the change in the format function to focus on the problem? Also, it's a pity the package-lock.json is always changing so much every time the project is built with a different OS, would it be possible to remove it from the PR?

MangelMaxime:
Ah yes sorry, I forgot to no push the package-lock.json.

Can't we go back to yarn :) ? It would also make the build faster as Yarn is faster to check that nothing need to be done :D

You are right, I will split the PR as it's failing on CI.

alfonsogarciacaro:
Yeah, I guess using yarn makes sense to avoid these problems with package-lock.json. I don't remember exactly, but I think I removed yarn because it was giving many problems with Travis. I don't know if it's better now. Do you know how to set it up?

MangelMaxime:
I think I know yes. I will test it and make another PR to add it so you can accept this one.

alfonsogarciacaro:
Great, thank you!

PR #1129, opened 2017-09-05
Fix/format date

MangelMaxime:
To merge after #1128.

@alfonsogarciacaro  I didn't re-add the Date tests because it's was the time zone which make the tests fails.

When using a local date and doing `mydate.ToString("O")` the output depends on your current time zone. The problem, is I don't have the same time zone than the server, or Eugene if he build Fable etc.

I am not sure how to make a reliable test for the local date format. One way would probably be to build dynamically the reference string but I am not sure of it. And feel like reporting the time zone problem on to another point etc.


alfonsogarciacaro:
Ah, true! Can you please at least add the tests commented out and with a note saying they only work in a specific time zone? We can try to fix the issue later, maybe detecting the time zone in advance or using a regex to check the result, etc.

alfonsogarciacaro:
You may also need to rebase with master, now that #1128 is merged :+1:

MangelMaxime:
Done waiting for the CI to be green as I activated the utc date test. (not dependant on the machine time zone)

alfonsogarciacaro:
Thanks!

PR #1130, opened 2017-09-05
Move to yarn

MangelMaxime:
I tested the CI with my fork and they passed:

- Travis: https://travis-ci.org/MangelMaxime/Fable/jobs/272035001
- AppVeyor: https://ci.appveyor.com/project/MangelMaxime/fable


alfonsogarciacaro:
Super fast, thank you! 🏎 🏎 🏎

ncave:
Should the yarn package be added as paket dependency? The build is currently failing on Windows with:
```
System.Exception: Start of process C:\Projects\Fable\packages\Yarnpkg.js\tools\yarn.cmd failed. The system cannot
find the file specified
   at Fake.ProcessHelper.ExecProcessWithLambdas@103-16.Invoke(String message) in D:\code\fake\src\app\FakeLib\Proc
```

MangelMaxime:
yarn is recommanded to be installed globally on your system. So it's a manual action. 

After yarn is also distributed via npm so we could cheat but i know no one going this way.

MangelMaxime:
I think I so something about installing yarn via fake I will take a look

ncave:
@MangelMaxime yarn is already installed globally, but that's not what the error is about, it seems to be missing a package in the packages folder.

MangelMaxime:
I don't understand is why you have yarn under your packages folder. 

Because for me your error is it's doesn't find the file `C:\Projects\Fable\packages\Yarnpkg.js\tools\yarn.cmd`. Could you try to clean your packages folder just to check ? The CI are green so it's seems to have something to do with your computer.

ncave:
@MangelMaxime That's the first thing I did, it doesn't help. I also tried on two different machines. Looks like Fake is looking for \Fable\packages\Yarnpkg.js\tools\yarn.cmd, is that package there in CI? I see you do some additional pre-install on TravisCI with `curl -o- -L https://yarnpkg.com/install.sh`, but I don't see anything similar in the appveyor.yml

alfonsogarciacaro:
My guess is the FAKE helper first checks if Yarn is installed globally and then tries the `packages` folder. Yarn is already installed in AppVeyor and this PR also installs it for the Travis image, so that's probably why CI passes.

It'd be good to have Yarn downloaded locally so it's a global requirement less, but I'm not sure how to do it. Apparently this [FAKE PR](https://github.com/fsharp/FAKE/pull/1494) just tries to locate yarn as if it were the Nuget npm package but that doesn't look quite right.

alfonsogarciacaro:
@ncave The FAKE PR checks the path first, I guess this is how it's located in AppVeyor? Do you have `yarn` in your path?

https://github.com/fsharp/FAKE/pull/1494/files#diff-b7934a24ef87f312fdb430e3e0355861R12

ncave:
@alfonsogarciacaro No, I had it installed globally through npm. I'll install it with the windows installer, thanks for the help.

ncave:
Works now. Thanks for the help, @MangelMaxime and @alfonsogarciacaro, the problem was located between the keyboard and the chair.

MangelMaxime:
Nice if you got it working :)

For AppVeyor, if it detect a yarn.lock at the root of the project it will install yarn.
However, it's still not the case of Travis so we need to install it our self.

PR #1133, opened 2017-09-05
Add support for String.filter

### LINKED ISSUE #1132: Implement Microsoft.Fsharp.Core.String.filter
I was trying to filter characters out of a string and realized that `String.filter` is not implemented.

It should be able to be done in JS as something like:

```
export function filter(pred:(i: string) => string, x: string) {
  return x.split('').filter(pred).join('');
}
```

jgrund:
Fixes #1132.


alfonsogarciacaro:
Awesome, thank you! 😍

PR #1134, opened 2017-09-05
Update docs to v1.2 (finally!)

alfonsogarciacaro:
From now this repo will contain only markdown files for the documentation. The web will be generated in [dev branch of fable-compiler.github.io repo](https://github.com/fable-compiler/fable-compiler.github.io/tree/dev).


inosik:
As I see, you removed the plugin documentation. Are plugins going to be deprecated or do we just need a newer version? I think that's actually a really useful feature.

alfonsogarciacaro:
Don't worry, plugins won't be removed :) It's just the feature didn't have the response I expected, so for now I want to focus on the other documents for the new web. Time allowing, in the future I'd like to add an API to make it easier to emit Fable AST and then rewrite the documentation about the plugins :+1:

PR #1138, opened 2017-09-08
Rollup config update

ncave:


src/dotnet/Fable.JS/demo/rollup.config.js
```diff
@@ -5,14 +5,15 @@ function resolve(filePath) {
...
-//   "presets": [
-//     ["es2015", {"modules": false}]
-//   ]
-// };
+var babelOptions = {
+  "presets": [
+    // ["es2015", { "modules": false }],
+    // ["minify", { "mangle": false }],
+  ]
+};
```
alfonsogarciacaro: Is the Fable plugin causing problems if you don't pass Babel Options with an empty `presets` array? If that's the case, I guess we need to fix it?

alfonsogarciacaro:
Awesome, thank you!

src/dotnet/Fable.JS/demo/rollup.config.js
ncave: @alfonsogarciacaro I forgot to comment the babel options, but no, it doesn't seem to cause problems.

Issue #1141, opened 2017-09-09
Rollup plugin doesn't generate source maps


inosik:
### Description

When using the Rollup plugin, Rollup generates a warning about broken source maps. The [condition][condition] in the plugin seems to be incorrect, when I change it to `if (true)`, the warning described below disappears. The [Rollup wiki][wiki] says that plugins that transform source code should generate source maps automatically, so I wonder if we should remove the condition altogether.

  [condition]: https://github.com/fable-compiler/Fable/blob/542a34c379eb41577b53df72d1dc557703a50bdb/src/js/rollup-plugin-fable/index.js#L100 
  [wiki]: https://github.com/rollup/rollup/wiki/Plugins#conventions

### Repro code

Clone my [VS Code extension sample repo][repo], restore the packages and build it with `dotnet fable yarn-run build`.

  [repo]: https://github.com/inosik/fable-vscode-rollup-sample

### Expected and actual results

```
(!) Broken sourcemap
https://github.com/rollup/rollup/wiki/Troubleshooting#sourcemap-is-likely-to-be-incorrect
Plugins that transform code (such as 'fable') should generate accompanying sourcemaps
```

### Related information

* Fable version (`dotnet fable --version`): 1.2.0
* Rollup plugin version: 1.1.0

alfonsogarciacaro:
You're right, probably it's a good idea to remove the condition for the Rollup plugin, could you please send a PR? Thanks a lot in advance!

inosik:
See #1142.

alfonsogarciacaro:
rollup-plugin-fable 1.1.1 published :+1:

inosik:
Yes, I already updated my packages and the warning is gone. Thanks!

PR #1142, opened 2017-09-09
Always generate source maps for Rollup

inosik:
References #1141.

What about the `babelOpts` object? Should we define a local copy of it in this `then` call?

### DIFF
--- a/src/js/rollup-plugin-fable/index.js
+++ b/src/js/rollup-plugin-fable/index.js
@@ -96,12 +96,14 @@ module.exports = (
             });
           });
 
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Yeah, this is probably safer, maybe something like this?

```js
const newBabelOpts = Object.assign({}, babelOpts, {
   sourceMaps = true,
   sourceFileName = path.relative(process.cwd(), data.fileName.replace(/\\/g, '/'))
});
```

inosik:
Done.

alfonsogarciacaro:
Thanks!

PR #1153, opened 2017-09-19
Show compile times in fable splitter

forki:
Not sure if correct


alfonsogarciacaro:
I think the preferred way to measure time performance in node is to use [process.hrtime](https://nodejs.org/docs/v0.8.0/api/all.html#all_process_hrtime)

forki:
like this? 1fc7cd4

alfonsogarciacaro:
Thanks!

alfonsogarciacaro:
Published as fable-splitter 0.1.17

PR #1157, opened 2017-09-20
Added subsecond time in splitter

ncave:
@alfonsogarciacaro Added subsecond time in splitter, if not needed, disregard.


alfonsogarciacaro:
I think this is very useful, thank you! I'll publish a new version right away :)

PR #1179, opened 2017-10-05
repl testapp - netcore metadata support

ncave:
- not all samples work on netcore binaries (e.g. sudoku and raytracer), something to do with comparers.
- but the other 4 samples work, so it's almost there.


alfonsogarciacaro:
Awesome, thank you!

PR #1189, opened 2017-10-18
Enable nested options

alfonsogarciacaro:
This PR adds a new JS runtime type `Some`, to represent `Some` options that are sometimes interpreted as `None` by Fable, mainly nested options but also unit options or options from unknown types.

With this we have now a mixed of erased options and cases where options have a runtime representation (only for the `Some` case), let's see how it works :wink:


Zaid-Ajaj:
How does this affect code generation and interop when using options? :smile:

alfonsogarciacaro:
@Zaid-Ajaj Hopefully it shouldn't, for most cases, options continue to be erased so they can still be used to interop with JS in the same way they've been used so far. Only special cases like nested or generic options would be wrapped, but these shouldn't be used to interact with JS... again, hopefully :wink:

PR #1202, opened 2017-10-24
fixes #1201

### LINKED ISSUE #1201: error FABLE: Cannot find replacement for System.Char::ToUpper
I'm working on https://github.com/fable-compiler/ts2fable/pull/19 and was surprised that `Char.ToUpper`. Good news is that string.ToUpper is, so easy work-a-round. May be close as won't fix.

``` fs
let upperFirstLetter (s: string): string =
    // sprintf "%c%s" (Char.ToUpper s.[0]) (s.Substring 1)
    sprintf "%s%s" (s.Substring(0,1).ToUpper()) (s.Substring 1)
```

ncave:


alfonsogarciacaro:
Fast as lightning, thank you! ⚡️

alfonsogarciacaro:
And with this, we already have more than 1300 tests! :tada:

PR #1209, opened 2017-10-31
Add Char.isLetter replacements

MangelMaxime:


src/js/fable-core/Char.ts
```diff
@@ -0,0 +1,12 @@
+export function isLetter(input: string) {
+    if (typeof input !== "string") {
+        return false;
+    }
+    if (input.length !== 1) {
+        return false;
+    }
+    if (input.toLowerCase() !== input.toUpperCase()) {
```
ncave: @MangelMaxime Just out of curiosity, how accurate is that definition?

MangelMaxime: This support ACSII, Cyrillic letters (probably more too).

In fact, this repose on the browser support. From what I understood, the browser keep a list associated the lower/upper of the support letters.

@ncave At first, I was needing this replacement in order to build the fable-repl with the latest version of Fable etc. but seems like I need at least one more now. 

> [!] (fable plugin) Error: C:/Projects/FSharp.Compiler.Service_fable/src/fsharp/tast.fs(3210,148): (3210,161) error FABLE: Cannot find replacement for Microsoft.FSharp.Core.PrintfModule::PrintFormatToStringThen

Perhaps, this one need to be a change in `FSharp.Compiler.Service_fable` project. I am still discovering the whole thing.

ncave: @MangelMaxime This is already supported, not sure why you get that.
https://github.com/fable-compiler/Fable/blob/master/src/dotnet/Fable.Compiler/Replacements.fs#L761

MangelMaxime: Ok, forget my last comment about a change needed into `FSharp.Compiler.Service_fable` there is already a replacement implemented in Fable.

Just looks like Fable don't catch it. Will try to understand why

MangelMaxime: Oups my tabs wasn't refresh :)

And about the regex, I can't make it work on this website: https://regex101.com/

Is it supported by JavaScript ?

ncave: idk, something like that perhaps?
function isLetter(str) {
  return str.length === 1 && str.match(/[^\W\d_]/);
}

MangelMaxime: This don't work for `β` char.

Zaid-Ajaj: Just for reference, this is how dotnet implements [Char.IsLetter](https://referencesource.microsoft.com/#mscorlib/system/char.cs,231)

ncave: @MangelMaxime Yes, I forgot JS char classes only work for latin1:
> Word boundary doesn’t work for non-English alphabets

That's fine, your solution is good enough for now, the repl doesn't yet have full unicode support anyway.

MangelMaxime: @Zaid-Ajaj Thanks for the implementation.

I think current implementation is good enough in the PR. I don't know how to implement the dotnet implemenation in JavaScript ^^

alfonsogarciacaro:
Great, thanks!

Issue #1210, opened 2017-11-01
Map with None value throws key not found


zaaack:
### Description

Map with None value throws key not found

### Repro code

http://fable.io/repl/#?code=%0Aopen%20Fable.Core%0Aopen%20Fable.Core.JsInterop%0A%0Alet%20a%3A%20Map%3Cstring%2C%20string%20option%3E%20%3D%20Map%20(%5Byield%20%22aa%22%2C%20None%5D)%0A%0Aa.ContainsKey%20%22aa%22%20%7C%3E%20printfn%20%22ContainsKey%3A%20%25A%22%0Aa.%5B%22aa%22%5D%20%7C%3E%20printfn%20%22aa%3A%20%25A%22%0A%0A

### Expected and actual results

Expected: Returns None

Actual: Throws key not found Error

### Related information

* Fable version (`dotnet fable --version`): 1.2.4
* Operating system: macos

PR #1212, opened 2017-11-01
Fix Map.find None value throws key error (Fix 1210)

zaaack:
Fix: https://github.com/fable-compiler/Fable/issues/1210

I ran into nested option issue several times these days, would it be better to compile option to a more complicate type?

### DIFF
--- a/tests/Main/MapTests.fs
+++ b/tests/Main/MapTests.fs
@@ -120,6 +120,28 @@ let ``Map.find works``() =
     xs |> Map.find 2
     |> equal 4.
 
+
+[<Test>]
+let ``Map.find option works``() =
+    let xs = Map [1,Some 1.; 2,None]
+    xs |> Map.find 2
+    |> Option.isNone
+    |> equal true
+    xs |> Map.find 1 |> (fun x -> x.Value)
+    |> equal 1.
+
+[<Test>]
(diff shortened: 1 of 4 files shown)


alfonsogarciacaro:
Hmm, latest Fable 1.3.0-beta does compile nested options [to a special type](https://github.com/fable-compiler/Fable/blob/4a55e602779cd8e2c06bec5900fd2ff9eb262046/src/js/fable-core/Option.ts#L3-L42) but it seems this wasn't enough to fix your use case. Thanks a lot for the fix!

tests/Main/MapTests.fs
```diff
@@ -120,6 +120,28 @@ let ``Map.find works``() =
...
+    |> equal 1.
+
+[<Test>]
+let ``Map.tryFind option works``() =
+    let xs = Map [1,Some 1.; 2,None]
+    // nested option cannot work in js
+    // xs |> Map.tryFind 2 |> (fun x -> x.Value)
+    // |> Option.isNone
+    // |> equal true
+    xs |> Map.tryFind 1 |> (fun x -> x.Value.Value)
```
zaaack: Actually here is still some cases not cover, like (Some None).Value , so map's TryFind still not work well...

PR #1224, opened 2017-11-08
Show that working with DateTimeOffset doesn't work

forki:


forki:
yes I work around by unboxing to string and parsing again... but this would help me a lot

PR #1232, opened 2017-11-14
Add support for System.DateTimeOffset

alfonsogarciacaro:
@forki Buf, I thought support for `DateTimeOffset` could be added in a cheap and dirty way but it took much longer than expected. Hopefully this also helps to clean the code for Date handling a bit. I also made it so the offset is kept when serializing to JSON (JS Date always serializes to UTC) so clients in different timezones can share the offset with the server.

I tried to make `DateTime` and `DateTimeOffset` ES6 classes extending JS Date but this didn't work when compiling to ES5 (not sure how long will this be necessary) so they are still JS Date in disguise with an additional `kind/offset` property.

Most of the tests are just copied from DateTests.fs. It will be nice to have more to check the offset with different conversion. Maybe we can import some tests from the netcore repo?


forki:
:hugs:

forki:
begging for release ;-)

alfonsogarciacaro:
Released as 1.3.0-beta-008 :)

forki:
Cool. Thanks

Am 14.11.2017 21:13 schrieb "Alfonso Garcia-Caro" <notifications@github.com

alfonsogarciacaro:
I've released another beta with minor fixes. @forki could you check and confirm everything works fine for you? If it does, I'll release 1.3 stable :+1:

> Please note that fable-splitter doesn't support dependencies yet, so if you're using it and want to force the recompilation of a file you have to make a change (e.g. adding a space) and resave the file.

forki:
Currently Bad timing, because I need to port back end from suave to
giraffe. Because reasons.

So app is not building right now

Am 15.11.2017 12:09 schrieb "Alfonso Garcia-Caro" <notifications@github.com

alfonsogarciacaro:
I'll take that as a yes, release 1.3 stable and pass the responsibility of testing it to users... because that's how we do OSS! ;)

forki:
+1

PR #1233, opened 2017-11-15
Fixed DateTime test for other time zones

ncave:


alfonsogarciacaro:
Awesome, thanks as always!

PR #1239, opened 2017-11-20
Fix #1185: Importing functions via importMember

### LINKED ISSUE #1185: Fable 1.2.4 broke importMember function
### Description

I updated my [VSCode extension sample][sample] to Fable 1.2.4 from 1.2.3 and this broke the `importMember` function. If I change my code to use the `import` function or the `Import`-Attribute it works.

  [sample]: https://github.com/inosik/fable-vscode-rollup-sample

### Repro code

Clone the repository from above and checkout the `fable-broken-import-member` branch. Then follow the instructions from the readme.

This is what I get:

``` shell
./src/wordcount.fsproj → ./out/extension.js...
fable: Compiled src\wordcount.fsproj
[!] (fable plugin) Error: D:/projects/experiments/fable/vscode-rollup-sample/src/extension.fs(1,1): error FABLE: `importMember` must be assigned to a variable
 Make sure Fable server is running on port 61225
src\extension.fs
Error: D:/projects/experiments/fable/vscode-rollup-sample/src/extension.fs(1,1): error FABLE: `importMember` must be assigned to a variable
 Make sure Fable server is running on port 61225
    at error (D:\projects\experiments\fable\vscode-rollup-sample\node_modules\rollup\dist\rollup.js:185:14)
    at Promise.resolve.then.catch.err (D:\projects\experiments\fable\vscode-rollup-sample\node_modules\rollup\dist\rollup.js:9182:6)
```

### Related information

* Fable version (`dotnet fable --version`): 1.2.4
* Operating system: Windows 7 x64

inosik:
References #1185.

It works now, but I'm not sure if `CurriedLambda` should be used in this case.

BTW, @alfonsogarciacaro, how do you debug/step through the .NET side of Fable? I can't set breakpoints with VS Code in the Fable.Compiler project, not on my Mac and not on my Windows workstation at work. I already tried setting the `DebugType` property to `portable` and building with the `Debug` configuration, but that didn't work.

### DIFF
--- a/src/dotnet/Fable.Compiler/FSharp2Fable.fs
+++ b/src/dotnet/Fable.Compiler/FSharp2Fable.fs
@@ -1108,7 +1108,11 @@ let private addMethodToDeclInfo com ctx (declInfo: DeclInfo) range import meth a
                 // When a member returns a function, there are issues with the uncurrying
                 // optimization when calling & applying at once (See #1041, #1154)
                 if isBodyMultiArgFunction then
-                    let body = transformExpr com ctx body
+                    let body =
+                        match transformExpr com ctx body with
+                        | Fable.Value (Fable.ImportRef (Naming.placeholder, path, importKind)) ->
+                            Fable.Value (Fable.ImportRef (meth.DisplayName, path, importKind))
+                        | body -> body
                     let body = makeDynamicCurriedLambda body.Range body.Type body
                     getMemberKind meth, args, extraArgs, body
                 else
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Awesome, thank you! 😻 

About breakpoints... hehe, I cannot use them since we moved to netcore (neither in VS Code or VS for Mac). They work in Main.fs for example, but as soon as the program moves to FSharp2Fable.fs, the debugger and the actual code lines don't match. I've tried several things, like unifying all the files in the same project, but the only way to properly debug the program was to use an old .fsproj file, compile Fable for .NET Framework and debug on Windows with VS. I already removed the old .fsproj version though, maybe we can try again now that VS accepts the new .fsproj format.

Usually I just iterate all the time using the `QuickFableCompilerTest` FAKE target: put the code I want to test in `src/tools/QuickTest.fs`, change the code in `src/dotnet` (adding `printfn` statements where necessary to get information) run the target, check if it works and if it doesn't repeat. It's a bit of a pain because recompiling dotnet-fable (`FableCLI` target) takes a bit but I haven't found a better method yet.

inosik:
That's what I did too, resorting to good old `printf` debugging :smile:

PR #1249, opened 2017-11-24
Fixes #838 Return empty group values when group is not matched

### LINKED ISSUE #838: [BUG] System.Text.RegularExpressions.Group.Value is incorrect
### Repro code

```fsharp

for m in Regex.Matches("\n\n\n", @"(?:([^\n\r]+)|\r\n|\n\r|\n|\r)") |> Seq.cast<Match> do
    printfn "%A" m.Groups.[1].Value
```

Expected:

```js
import { matches } from "fable-core/RegExp";
import { fsFormat } from "fable-core/String";
const inputSequence = matches("\n\n\n", "(?:([^\\n\\r]+)|\\r\\n|\\n\\r|\\n|\\r)");

for (let m of inputSequence) {
  fsFormat("%A")(x => {
    console.log(x);
  })(m[1] || "");
}
```
```
""
""
""
```

[Actual](http://fable.io/repl.html#?code=open%20System.Text.RegularExpressionsfor%20m%20in%20Regex.Matches("\n\n\n"%2C%20%40"(%3F%3A([^\n\r]%2B)|\r\n|\n\r|\n|\r)")%20|>%20Seq.cast<Match>%20do%20%20%20%20printfn%20"%25A"%20m.Groups.[1].Value&astInput=false&babili=false&evaluate=false&lineWrap=false&presets=es2015%2Creact%2Cstage-2):

```js
import { matches } from "fable-core/RegExp";
import { fsFormat } from "fable-core/String";
const inputSequence = matches("\n\n\n", "(?:([^\\n\\r]+)|\\r\\n|\\n\\r|\\n|\\r)");

for (let m of inputSequence) {
  fsFormat("%A")(x => {
    console.log(x);
  })(m[1]);
}
```
```
"undefined"
"undefined"
"undefined"
```

### Description

The [`System.Text.RegularExpressions.Group.Value`](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.capture.value(v=vs.110).aspx) property is always a string. If it doesn't match, then it will be the empty string `""`

However, Fable compiles it to `m[1]`, which is sometimes `undefined` (because regexp matches in JS *sometimes* return `""` and *sometimes* return `undefined`). This can be fixed by using `m[1] || ""` instead.

I can probably send a PR to fix this.

Zaid-Ajaj:
According to #838, group values can be undefined which is not correct because they should always be strings. This PR adds a guard for such values and returns `(value || "")` instead of just `value`.

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -1259,7 +1259,11 @@ module AstPass =
                  |> addErrorAndReturnNull com i.fileName i.range |> Some
         | "value" ->
             if isGroup
-            then i.callee.Value |> wrap i.returnType |> Some
+            then 
+                i.callee.Value 
+                |> wrap i.returnType
+                |> fun value -> emit i "($0 || \"\")" [value] 
+                |> Some
             else propInt 0 i.callee.Value |> Some
         | "length" ->
             if isGroup
--- a/tests/Main/RegexTests.fs
+++ b/tests/Main/RegexTests.fs
@@ -261,3 +261,12 @@ let ``Regex.Replace with evaluator, limit and offset works``() =
 let ``Replacing with $0 works``() = // See #1155
     let s = Regex.Replace("1234567890", ".{2}", "$0-")
     equal "12-34-56-78-90-" s
+
+// See #838
+[<Test>]
+let ``Group values are correct and empty when not being matched``() = 
+    Regex.Matches("\n\n\n", @"(?:([^\n\r]+)|\r\n|\n\r|\n|\r)") 
+    |> Seq.cast<Match> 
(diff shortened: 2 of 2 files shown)


alfonsogarciacaro:
I forgot about that issue 😅 Awesome, thanks!

alfonsogarciacaro:
FYI, I just [replaced the JS emit with Fable AST](https://github.com/fable-compiler/Fable/commit/38df179e3bec8fcf60c51bd8df39c6e679506950) as I'm trying to reduce the number of plain JS strings in the replacements module, I hope that's OK with you :smile:

Zaid-Ajaj:
@alfonsogarciacaro Absolutely! It's just that the `emit` is the first thing that comes to my mind, it looks better now :+1:

PR #1250, opened 2017-11-25
Add Char.IsWhiteSpace with tests

Zaid-Ajaj:
Because why not :smile:


alfonsogarciacaro:
Ups, seems the replacement was half-cooked 👨‍🍳 Thanks!

PR #1252, opened 2017-11-25
Fix optional tuple in record raise errors for jsonConverter

zaaack:
I got `invalid token` error when trying to deserialize optional tuple in records.

e.g.
```json
{"name":null,"id":null,"range":null}
``` 
to 
```F#
type LocationQuery = {
    name: string option
    id: int64 option
    // (minLongitude, minLatitude), (maxLongitude, maxLatitude)
    range: ((double * double) * (double * double)) option
}
```
Then I found this related PR: https://github.com/fable-compiler/Fable/pull/506 , looks it didn't fix optional tuple case.


alfonsogarciacaro:
Nice catch ⚾️ Thanks!

alfonsogarciacaro:
Fable.JsonConverter 1.0.3 with the fix has been published :+1:

PR #1263, opened 2017-11-28
added opt-in no regex

ncave:
Not a large perf difference, but it can be a little bit faster on big projects. Opt-in with compilation define.

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -3,7 +3,6 @@ open Fable
 open Fable.Core
 open Fable.AST
 open Fable.AST.Fable.Util
-open System.Text.RegularExpressions
 
 module Util =
     let [<Literal>] system = "System."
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
Performance improvements are always welcome! I'm OK with just replacing the Regex code to avoid the need of using the compilation define, what do you think? Or do you prefer to keep the old code in cases there are scenarios the new code cannot handle?

ncave:
@alfonsogarciacaro No, I was just trying to avoid controversy. I'll take it out entirely.

alfonsogarciacaro:
Gimme my Regex back! ...just kidding, you know in F# controversies only happen around the `fun` keyword ...well, not only :wink:

Thanks!

src/dotnet/Fable.Compiler/Replacements.fs
```diff
@@ -2537,8 +2536,9 @@ let checkLiteral com fileName range (value: obj) (typ: Fable.Type) =
     match typ with
     | Fable.Enum("System.Text.RegularExpressions.RegexOptions") ->
         match value with
-        | :? int as i when i = int RegexOptions.IgnoreCase
-                        || i = int RegexOptions.Multiline
-                        || i = int RegexOptions.ECMAScript -> ()
+        | :? int as i when i = 0x0001 // "i" - RegexOptions.IgnoreCase
+                        || i = 0x0002 // "m" - RegexOptions.Multiline
+                        || i = 0x0100 // "e" - RegexOptions.ECMAScript
```
inosik: This is not necessary. The IL code is exactly the same.

ncave: @inosik The point is to not use anything from the regex namespace (in this case, RegexOptions) to lower the API surface.

PR #1272, opened 2017-12-01
Money literals are not converted

forki:

PR #1274, opened 2017-12-01
Money parsing

forki:


alfonsogarciacaro:
The fix for this and #1272 has been published with version 1.3.5. A caveat though, atm decimals are just converted to JS numbers so they may not have the precision you expect.

PR #1271, opened 2017-11-30
Build REPL in CI server

alfonsogarciacaro:
I'm trying to use Paket Github dependencies to get the files from `fable` branch of @ncave FCS fork, however I'm getting the following error in AppVeyor (Path is too long):

screen shot 2017-12-01 at 12 03 23 am

I can see that Paket install the files and then creates a duplicated folder with the commit hash as suffix. It's possible this folder is creating the issue. Would there be some way to prevent the creation of this extra folder, @forki?

screen shot 2017-12-01 at 12 03 50 am

PR #1275, opened 2017-12-02
added GetOptimizedAssemblyContents for the REPL

ncave:
just the repl (for now)


alfonsogarciacaro:
Thanks! In #1271 I removed the files for the REPL UI to avoid duplication (they're also in fable-compiler.github.io repo and @MangelMaxime is working in the new version in fable-repl). Could you please send also the PR to fable-compiler.github.io? Or maybe we can just forget about these files for now and focus on the new version :)

Issue #1266, opened 2017-11-30
Implement replacements for optimization switch


alfonsogarciacaro:
Follow-up to https://github.com/fable-compiler/Fable/pull/1197#issuecomment-348034660

Replacements needed so the optimization switch can pass the test suite:
- [x] Microsoft.FSharp.Collections.FSharpList::CompareTo
- [x] Microsoft.FSharp.Collections.FSharpList::Equals
- [x] Microsoft.FSharp.Collections.FSharpList::GetHashCode
- [x] Microsoft.FSharp.Core.LanguagePrimitives.ErrorStrings::InputArrayEmptyString
- [x] Microsoft.FSharp.Core.LanguagePrimitives.ErrorStrings::InputMustBeNonNegativeString
- [x] Microsoft.FSharp.Core.LanguagePrimitives.ErrorStrings::InputSequenceEmptyString
- [x] Microsoft.FSharp.Core.LanguagePrimitives.HashCompare::GenericComparisonIntrinsic
- [x] Microsoft.FSharp.Core.LanguagePrimitives.HashCompare::GenericEqualityIntrinsic
- [x] Microsoft.FSharp.Core.LanguagePrimitives.HashCompare::GenericGreaterOrEqualIntrinsic
- [x] Microsoft.FSharp.Core.LanguagePrimitives.HashCompare::GenericGreaterThanIntrinsic
- [x] Microsoft.FSharp.Core.LanguagePrimitives.HashCompare::GenericLessThanIntrinsic
- [x] Microsoft.FSharp.Core.LanguagePrimitives.HashCompare::PhysicalEqualityIntrinsic
- [x] Microsoft.FSharp.Core.LanguagePrimitives::GenericEqualityComparer
- [x] Microsoft.FSharp.Core.LanguagePrimitives::GenericEqualityERComparer
- [x] Microsoft.FSharp.Core.LanguagePrimitives::ParseInt32
- [x] Microsoft.FSharp.Core.LanguagePrimitives::ParseInt64
- [x] Microsoft.FSharp.Core.LanguagePrimitives::ParseUInt32
- [x] Microsoft.FSharp.Core.LanguagePrimitives::ParseUInt64
- [x] Microsoft.FSharp.Core.Operators.OperatorIntrinsics::PowDouble
- [x] Microsoft.FSharp.Core.Operators.OperatorIntrinsics::RangeChar
- [x] Microsoft.FSharp.Core.Operators.OperatorIntrinsics::RangeDouble
[...]

PR #1277, opened 2017-12-02
Added ErrorStrings

Zaid-Ajaj:
Addresses #1266 
Again, no tests, I get compile errors when I try to access `LanguagePrimitives.ErrorStrings`

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -1289,6 +1289,15 @@ module AstPass =
             propStr "length" i.callee.Value |> Some
         | _ -> None
 
+    // Compile static strings to their constant values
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
@alfonsogarciacaro Should these be `get_InputArrayEmptyString` instead of `inputArrayEmptyString`?

alfonsogarciacaro:
Awesome, thanks @Zaid-Ajaj! 💪 

TBH, I had to check it myself. Dealing with method names in the Replacements module is a bit messy (for example, originally I was thinking to compile every method to lower case and that's why everything starts with lower case), but in any case [this line](https://github.com/Zaid-Ajaj/Fable/blob/cc908c7442b77fb3ec4cf9060d3309303364ff39/src/dotnet/Fable.Compiler/Replacements.fs#L2502) removes the `get_` prefix so it should be fine as it is 👍

PR #1197, opened 2017-10-20
WIP: Try to optimize in FCS

forki:
Corresponding VF# PR: https://github.com/Microsoft/visualfsharp/pull/3784

/cc @dsyme @ncave

PR #1278, opened 2017-12-02
Make Double.Parse and Single.Parse compile with IFormatProvider

Zaid-Ajaj:
Addresses #1266 

According to @ncave's comment [here](https://github.com/fable-compiler/Fable/pull/1197#issuecomment-348034660) it seems that `System.Double.Parse` and `System.Single.Parse` cannot be compiled and because there is no replacement is to be found for `CultureInfo.InvariantCulture` I *assumed* that the underlying problem has to do with the `Parse` functions being called somewhere like this:

```
let result = Double.Parse(input, CultureInfo.InvariantCulture)
// or
let result = Single.Parse(input, CultureInfo.InvariantCulture)
```
Which is rather very common when parsing numbers

@ncave Can you check this to be sure?

This PR makes such code work. Fable does not yet support `CultureInfo` so the value `CultureInfo.InvariantCulture` is compiled to an empty object literal and it is ignored on call-site. 

I know this is kind of a hack so I added warnings where appropriate :smile:

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -1149,8 +1149,6 @@ module AstPass =
             // just ignore the second args (IFormatProvider) and compile 
             // to System.Double.Parse(string)
             | "parse", [str; _ ] ->
-                "The value of IFormatProvider passed to the Parse function will be ignored"
-                |> addWarning com i.fileName i.range
                 CoreLibCall (numberModule, Some "parse", false,
                     [str; (if isFloat then makeNumConst 10.0 else makeIntConst 10)])
                 |> makeCall i.range i.returnType |> Some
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
My wild guess here is that `Double.Parse(input)` just calls `Double.Parse(input, CultureInfo.InvariantCulture)` so the optimizer just replaces the code to remove the indirection. I'm not sure about the warning because it the optimizer makes this change automatically users will see the warning even if they're only using one argument. Maybe we can just check the second argument and raise the warning only if it's not `CultureInfo.InvariantCulture`?

Zaid-Ajaj:
Hmm converting `Double.Parse(input)` to `Double.Parse(input, CultureInfo.InvariantCulture)` seems rather counter-intuitive because AFAIK `Double.Parse` will use the default culture of the system (I had many problems with that one :sweat:). 

I don't think we need another warning and removing the current one should be enough because the warning was intended to let the user know that his usage of `CultureInfo` won't have any effect.

alfonsogarciacaro:
Ah, you're right! Makes total sense, thanks again! 🥇

PR #1280, opened 2017-12-02
Added repl bench app

ncave:


src/dotnet/Fable.JS/Interfaces.fs
```diff
@@ -41,8 +41,9 @@ type IParseResults =
...
-    abstract CreateChecker: references:string[] * readAllBytes:(string->byte[])->IChecker
+    abstract CreateChecker: references:string[] * readAllBytes:(string->byte[]) -> IChecker
     abstract CreateCompiler: fableCorDir:string * ?replacements: seq<string * string> -> IFableCompiler
     abstract ParseFSharpProject: checker:IChecker * fileName:string * source:string->IParseResults
-    abstract GetCompletionsAtLocation: parseResults:IParseResults * line:int * col:int * lineText:string->Async<Completion[]>
-    abstract CompileToBabelJsonAst: com: IFableCompiler * parseResults:IParseResults * fileName:string * ?optimized:bool->string
\ No newline at end of file
+    abstract GetCompletionsAtLocation: parseResults:IParseResults * line:int * col:int * lineText:string -> Async<Completion[]>
+    abstract CompileToBabelAst: com: IFableCompiler * parseResults:IParseResults * fileName:string * optimized:bool -> Fable.AST.Babel.Program
+    abstract CompileToBabelJsonAst: com: IFableCompiler * parseResults:IParseResults * fileName:string * ?optimized:bool -> string
```
alfonsogarciacaro: Maybe make `optimized` optional for `CompileToBabelAst` too? We can also remove `CompileToBabelJsonAst`, when called from F# we can apply directly `JsInterop.toJson` and to call it from JS we probably need another layer to include Babel compilation anyways.

alfonsogarciacaro:
Thanks! And sorry I shouldn't have removed the benchmarks without asking 😅 

Did I do that

src/dotnet/Fable.JS/Interfaces.fs
ncave: @alfonsogarciacaro Alright, although explicit parameters force you to express the intent better than optional parameters.

alfonsogarciacaro: It's mainly for consistency, I'm also OK with the other way around: making `optimized` mandatory in `CompileToBabelJsonAst` (in case we keep it) :smile:

alfonsogarciacaro: Anyways, I'll merge this and we can do the amendment later :+1:

PR #1302, opened 2017-12-20
Fix `dotnet new fable` documentation

worldbeater:
Apparently, 
```
dotnet new fable -n FableApp
```
command does not do anything when using dotnet 2.1.2 cli tools. I tried my best to figure out why it can't create a new project and finally succeed. It needs -lang parameter. So the command should be:
```
dotnet new fable -n FableApp -lang F#
```
Now it works and I finally managed to create a Fable sample app.


alfonsogarciacaro:
Thanks! This is actually a bug in latest dotnet CLI, but it's good to put it in the docs until they fix it :)

https://github.com/dotnet/templating/issues/1373

Issue #715, opened 2017-02-15
COSMETICS: Don't introduce lambda if it's not needed


forki:
I'm compiling the Sudokusolver https://gist.github.com/alfonsogarciacaro/e9cfd5fbf18937cd132b4781ffc85ece with Fable v1.0.0-narumi, FCS v10.0.1, Babel v6.22.1 in the browser

![image](https://cloud.githubusercontent.com/assets/57396/22983242/d1197a22-f3a1-11e6-93d7-58936ccbb935.png)

I think it could be 

    return map(Int32Array.from, list);

alfonsogarciacaro:
The F# compiler always wraps in a lambda method calls used as values. In theory, these are not needed in JS but I decided to keep them because sometimes they're actually needed. For example:

- If you pass an object method this way, JS doesn't capture the `this` context. That's the reason, for example, why they have to use `bind` [here](https://github.com/tastejs/todomvc/blob/feb8465b4e7f2631195c64ded82f45aa7a6cb982/examples/react/js/app.jsx#L29-L33). It works if you use an ES2015 arrow function which does capture `this`.
- Some JS APIs don't work like this. Actually that very sample above won't work. You can try the following code in Chrome's console:

```js
var ar = [[1,2],[3,4],[5,6]]
ar.map(Int32Array.from) // This throws an exception
ar.map(x => Int32Array.from(x)) // This works
```

PR #1307, opened 2017-12-25
Remove empty intermediate functions

vbfox:
What
------

When a function or method is passed as parameter Fable was generating
a wrapper function around it.

```fsharp
let x = map foo bar
let y = map abs bar
```
became
```javascript
var x = map(function(x){return foo(x);},bar);
var y = map(function(x){return Math.abs(x);},bar);
```
It will now be:
```javascript
var x = map(foo,bar);
var y = map(Math.abs.bind(Math),bar);
```

Why ?
-------

My original motivation was to investigate why functional react components were appearing as "Unknown" in React extension. The reason was that `createComponent` received the wrapper function (that has no name) instead of the real one.

Testing a little the difference it also seem to be a little better on performance on both edge and Firefox (It's the same on node/v8 where it seem that the engine recognize this pattern and already optimize it away)

Implementation questions / details
--------------------------

My first instinct was do the transformation on the F# AST side as we have more information there to make this choice and could even avoid the `bind` call in most cases. But it was problematic with the rest of the code as the transformations currently don't expect to have any of the methods they transform (`Math.abs` for exemple) to be passed as parameters instead of being called and so the change would have been bigger here.

In the end I preferred to not do the `bind` removal and to do the transformation in the babel tree, but it's 100% open to discussion 😉 

I also didn't pass locations (for source maps) as I don't fully understand them yet. I'd like to see if my approach is sound and get some feedback from maintainers first but i'll do it.


alfonsogarciacaro:
This is great, @vbfox! Thanks a lot for this :clap: :clap: We had some discussions about this in the past (see #715) but I was a bit reluctant to add this optimization because of the `this` binding issue and also because some JS APIs (like `Int32Array.from`) don't work this way. But I agree it's a very good addition as it improves performance, readability and compatibility with React tools.

It'd be nice to find a way to identify the failing APIs. Passing the source location shouldn't be difficult. I'll have a deeper look at the code later :+1:

It's true that optimizations right now are a bit messy, some of them occur in the FSharp2Fable pass and others (tail-call) in the Fable2Babel pass. For Fable 1.4 I'd like to add [an extra optimization pass](https://github.com/fable-compiler/Fable/commit/6c2af36178cc8dfcc4dc3acd5e24612c491d6cac) and move most of the optimizations there. In this case, as you need to apply `bind` it may make sense to keep it in the Babel pass anyways (I still have hopes that Fable would be useful to compile F# to other languages, though it's true it's mainly designed for JS right now).

alfonsogarciacaro:
For reference I just read this comment in the following post:

> However, we would typically not use bind because is demonstrably slower than using a closure.

[GET READY: A NEW V8 IS COMING, NODE.JS PERFORMANCE IS CHANGING.](https://medium.com/the-node-js-collection/get-ready-a-new-v8-is-coming-node-js-performance-is-changing-46a63d6da4de)

jgrund:
@alfonsogarciacaro worth noting that the comment refers to older versions of v8. Later in the article it goes on to state that bind is the fastest on newer v8s (by a tiny margin).

alfonsogarciacaro:
@jgrund Hehe, you were fast. I just read that part and was going to write the same ;)

vbfox:
Ah yes sorry I need to update the PR with the source location, completely forgot :)

For the bind my tests show that v8 is essentially the same between both on the current version (node & chrome ones).

`Int32Array.from` is one of the case that break without the bind yes :) But any JS-imported method could break so even if one day the bind removal is done it need to be disabled for all imported methods.

vbfox:
For me this version is ready for review :)

alfonsogarciacaro:
Thanks @vbfox, I'm currently working in a [extensive cleanup of the code](https://github.com/fable-compiler/Fable/commits/cleanup) for Fable 2.0. If you don't mind, we can retarget the PR to `dev2.0` branch when it's finished so we are sure your contribution fits nicely in the next Fable version :)

forki:
Is this something that the F# optimizer would do?

vbfox:
@forki Nope.

Example: https://sharplab.io/#v2:DYLgZgzgPgtg9gEwK7AKYAIAqqIBcLoC8AsAFDoXpwAOqAdugMoCeeqMZZl6au6YcOEXQAKMEgYAPdAFoAfOmkBqdAEYAlF0paKvdAGNEqEeqI7uPVHwBGAQwBOwgNqqA3ACZXAZgC6ncha6VorCADIAlngAdDC21PyC6Hb25tx6zGGRuDFx6LbWBMmp2gGBJkA= 

The thing is that in .Net functions aren't first class objects (contrary to javascript) and F# must create an instance of `FSharpFunc` to pass so it must add this level of indirection.

alfonsogarciacaro:
AFAIK the F# compiler treats differently methods and lambdas (aka functions as values), methods are _called_ while lambdas are _applied_, and every time a method is passed as a value, the F# compiler wraps it in a lambda.

@vbfox Fable 2.0 is going to take a bit longer and you already need this so let's merge it :+1: Sorry for taking so long and thanks a lot for your contribution!

PR #1310, opened 2017-12-31
Support TimeSpan correctly in Fable.JsonConverter

FrankBro:
Not sure I could put tests for this since by itself, Fable handles TimeSpan correctly. The problem was using Fable.JsonConverter is not enough for TimeSpan.

### DIFF
--- a/src/dotnet/Fable.JsonConverter/Fable.JsonConverter.fs
+++ b/src/dotnet/Fable.JsonConverter/Fable.JsonConverter.fs
@@ -38,6 +38,7 @@ type Kind =
     | MapOrDictWithNonStringKey = 7
     | Long = 8
     | BigInt = 9
+    | TimeSpan = 10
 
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks! Yes, Fable.JsonConverter tests may be confusing, there are some in the test suite that are executed in the dotnet run pass and there're some others [here](https://github.com/fable-compiler/Fable/blob/7229be70aa0a5acef9413c720cc7708a593bce34/src/dotnet/Fable.JsonConverter/tests/Fable.JsonConverter.Tests.fs).

alfonsogarciacaro:
I've added [a test for Fable.JsonConverter TimeSpan](https://github.com/fable-compiler/Fable/commit/fba8e6cef0bbfd717eaa6422dc241bdfbc5d943d), please let me know if it's OK or PR any additional test that you found failing before.

alfonsogarciacaro:
Published as 1.0.5 :+1:

PR #1320, opened 2018-01-10
use directly i link, so i can click and see the demo

enricosada:
following the doc, was nice to just click and see it :D


alfonsogarciacaro:
Oh, good one 👍 Thanks!

Issue #1323, opened 2018-01-15
Complete fable-core/Array module


alfonsogarciacaro:
When I started writing Fable, I replaced many of the [Array and List module functions with the Seq module](https://github.com/fable-compiler/Fable/blob/736d1e10bafe7f9065abef025ca053fb366e3271/src/dotnet/Fable.Compiler/Replacements.fs#L1960-L1977) in order to release faster and reduce the size of fable-core (before it was distributed as independent modules). However, this has remained like this until today and we should change it for Fable 2.0: having specialized functions for arrays and lists should provide performance opportunities.

This is also a good timing to start writing part of fable-core in F# (something I should have done long ago). I've already started [adding the `src/js/fable-core/Array2.fs` file](https://github.com/fable-compiler/Fable/commit/44426ec004df860df54028ce74cda9541b6c6c5c) in `dev2.0` branch, if you want to help with this issue (much appreciated!) you can start from there. Things to consider:

- At the beginning I ~~stole~~ _got inspiration_ from FunScript.Core code, so you can also use [this file](https://github.com/ZachBray/FunScript/blob/master/src/main/FunScript/Core/Array.fs) as reference, but check the following point.
- Fable translates arrays as Typed Arrays (for numbers) or dynamic JS Arrays. In order to tell each case apart, _Fable will send the array constructor as the last argument_ that you can use to build the new array to be returned (check the current `map` function for reference).

**Please send a WIP PR to `dev2.0` branch as soon as you start to know you're working on it and to help you with any question.**

### Writing F# sources for fable-core

As of now, please use `src/js/fable-core/Array2.fs`. We'll change the name of the file when it's complete to replace the current Array.ts. Before starting, make sure to run `dotnet restore` on repo root so the IDE doesn't give errors.

### Building fable-core

Run `sh build.sh FableCoreJS` on repo root (or `.\build.cmd FableCoreJS` on Windows). You will see the resulting files in `build/fable-core`. After the first time you can run `sh build.sh FableCoreJSFast`, which doesn't build the Fable CLI.

alfonsogarciacaro:
These are the functions that need to be implemented.

### Functions returning an array

- [ ] append
- [ ] choose
- [ ] collect
- [ ] concat
- [ ] countBy
- [ ] distinct
- [ ] distinctBy
- [ ] except
- [ ] filter
- [ ] where
- [ ] groupBy
- [ ] initialize
- [ ] map2
- [ ] mapIndexed2
[...]

PR #1324, opened 2018-01-15
WIP: Complete fable-core/Array module

valery-vitko:
Related issue: #1323

### DIFF
--- a/src/js/fable-core/Array2.fs
+++ b/src/js/fable-core/Array2.fs
@@ -8,9 +8,25 @@ type ArrayCons =
     [<Emit("$0.from($1)")>]
     abstract From: 'T seq -> 'T array
 
-let map (f: 'T->'U) (source: 'T[]) (cons: ArrayCons) =
+let mapIndexed (f: int -> 'T -> 'U) (source: 'T[]) (cons: ArrayCons) =
+    let len = source.Length
+    let target = cons.Create(len)
+    for i = 0 to (len - 1) do
+        target.[i] <- f i source.[i]
+    target
+
+let map (f: 'T -> 'U) (source: 'T[]) (cons: ArrayCons) =
+    // TODO: Should we involve currying and re-use mapIndexed here?
+    // Don't quite know yet how is it optimized so can't decide on performance vs code reuse.
+    // mapIndexed (fun _ x -> f x) source
+
     let len = source.Length
     let target = cons.Create(len)
     for i = 0 to (len - 1) do
         target.[i] <- f source.[i]
     target
+
+let indexed (source: 'T[]) (cons: ArrayCons) =
+    // TODO: How do we get concrete instance of ArrayCons for dynamic JS Arrays creation?
+    // Passing as a parameter for now.
+    source |> mapIndexed (fun i x -> i, x) <| cons;


src/js/fable-core/Array2.fs
```diff
@@ -8,9 +8,25 @@ type ArrayCons =
...
+    let len = source.Length
+    let target = cons.Create(len)
+    for i = 0 to (len - 1) do
+        target.[i] <- f i source.[i]
+    target
+
+let map (f: 'T -> 'U) (source: 'T[]) (cons: ArrayCons) =
+    // TODO: Should we involve currying and re-use mapIndexed here?
+    // Don't quite know yet how is it optimized so can't decide on performance vs code reuse.
+    // mapIndexed (fun _ x -> f x) source
```
alfonsogarciacaro: You can run `sh build.sh FableCoreJS` to see the generated code, [or just use the REPL](http://fable.io/repl/#?code=open%20Fable.Core%0A%0Atype%20ArrayCons%20%3D%0A%20%20%20%20%5B%3CEmit(%22new%20%240(%241)%22)%3E%5D%0A%20%20%20%20abstract%20Create%3A%20capacity%3A%20int%20-%3E%20'T%20array%0A%20%20%20%20%5B%3CEmit(%22%240.from(%241)%22)%3E%5D%0A%20%20%20%20abstract%20From%3A%20'T%20seq%20-%3E%20'T%20array%0A%0Alet%20mapIndexed%20(f%3A%20int%20-%3E%20'T%20-%3E%20'U)%20(source%3A%20'T%5B%5D)%20(cons%3A%20ArrayCons)%20%3D%0A%20%20%20%20let%20len%20%3D%20source.Length%0A%20%20%20%20let%20target%20%3D%20cons.Create(len)%0A%20%20%20%20for%20i%20%3D%200%20to%20(len%20-%201)%20do%0A%20%20%20%20%20%20%20%20target.%5Bi%5D%20%3C-%20f%20i%20source.%5Bi%5D%0A%20%20%20%20target%0A%0Alet%20map%20(f%3A%20'T%20-%3E%20'U)%20(source%3A%20'T%5B%5D)%20(cons%3A%20ArrayCons)%20%3D%0A%20%20%20%20mapIndexed%20(fun%20i%20x%20-%3E%20f%20x)%20source%20cons%0A&html=%3Chtml%3E%3Cbody%3E%3C%2Fbody%3E%3C%2Fhtml%3E%0A). As you can see, in this case it should be fine as the generated is very simple. But we need to be careful with curried function in fable-core as they can get slow if the compiler wraps them with `CurriedLambda`.

src/js/fable-core/Array2.fs
```diff
@@ -8,9 +8,25 @@ type ArrayCons =
...
+
     let len = source.Length
     let target = cons.Create(len)
     for i = 0 to (len - 1) do
         target.[i] <- f source.[i]
     target
+
+let indexed (source: 'T[]) (cons: ArrayCons) =
+    // TODO: How do we get concrete instance of ArrayCons for dynamic JS Arrays creation?
+    // Passing as a parameter for now.
```
alfonsogarciacaro: Fable knows this will return `int*'T[]` so `cons` should already be a dynamic JS Array constructor.

src/js/fable-core/Array2.fs
valery-vitko: Thanks for the tip about the REPL (impressive stuff, btw). Yeah, I saw `CurriedLambda` in the output, but thought of possibility of it being optimized-away in the babel/rollup/webpack/whatever minification pipeline. I'll avoid it in this particular casein (and minimize usage in fable-core).

src/js/fable-core/Array2.fs
valery-vitko: OK, thx, got it. What about functions like `partition` returning `'T[] * 'T[]`? It's in the list of "functions not returning an array", by the way. I believe it should be handled by do passing `cons`, or am I still missing something here?

alfonsogarciacaro: It was in the list of functions already implemented in Array.ts, but you're right, it needs the constructor to be passed. I moved it to the first group, thanks!

valery-vitko:
@alfonsogarciacaro, could you please review my latest changes? Sorry for being a bit noisy with TODO comments, but I haven't been following Fable development lately, so not quite sure of proper directions to follow sometimes :)

alfonsogarciacaro:
@valery-vitko That's absolutely fine, no worries. I'll review them :+1:

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+
+[<Emit("$0.slice($1)")>]
+let private sliceFrom (array: 'T[]) (``begin``: int): 'T[] = jsNative
+
+[<Emit("$0.splice($1, $2)")>]
+let private splice (array: 'T[]) (start: int) (deleteCount: int): 'T[] = jsNative
+
+// TODO: Should we reuse declarations from Import.JS instead of raw emits?
+let inline private spliceN<'T> (array: 'T[]) (start: int) (deleteCount: int) =
+    (box array :?> JS.Array<_>).splice(float(start), float(deleteCount)) |> box :?> 'T[]
```
alfonsogarciacaro: Maybe using Import.JS can be a bit safer than just using JS code in a plain JS (and help us find problems with the bindings), but we should have the tests already so I don't really have a strong preference. If you prefer the emits because you can control the function signature that's fine with me :)

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+
+[<Emit("$1.findIndex($0)")>]
+let inline private findIndexImpl predicate (array: 'T[]): int = jsNative
+
+// TODO: Should we tend to use native js counterparts as much as possible
+// (and how do Replacements nativeArrayFunctions relate?), i.e. chase performance,
+// or should we do the opposite and implement all the function in F# (with small exceptions when typed arrays chime in),
+// go for correcntess for starters.
+[<Emit("$0.concat($1)")>]
+let append (array1: 'T[]) (array2: 'T[]): 'T[] = jsNative
```
alfonsogarciacaro: If the semantics are the same, I tend to think native JS methods should produce less code and yield better performance ([though this is not always true](https://twitter.com/FableCompiler/status/890488515926122496)). Yes, the Replacements module [use native JS functions for arrays when possible](https://github.com/fable-compiler/Fable/blob/a844ca0621548c3dfb20229f9c7c04a030856207/src/dotnet/Fable.Compiler/Replacements.fs#L2030-L2037).

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+    | len ->
+        let mutable acc = state
+        let res = cons.Create len
+        for i = 0 to array.Length-1 do
+            let h',s' = mapping acc array.[i]
+            res.[i] <- h'
+            acc <- s'
+        res, acc
+
+// TODO: Move function to the list of "Functions returning an array"
```
alfonsogarciacaro: Done, thanks!

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+   result
+
+let map3 f (source1: 'T[]) (source2: 'U[]) (source3: 'U[]) (cons: ArrayCons) =
+   if source1.Length <> source2.Length || source2.Length <> source3.Length then failwith "Arrays had different lengths"
+   let result = cons.Create(source1.Length)
+   for i = 0 to source1.Length - 1 do
+      result.[i] <- f source1.[i] source2.[i] source3.[i]
+   result
+
+// TODO: Move function to the list of "Functions returning an array"
```
alfonsogarciacaro: Done!

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+    let mutable i = 0
+    for group in dict do
+        result.[i] <- group.Key, group.Value
+        i <- i + 1
+
+    result
+
+let ofArray (arr: 'T[]) = arr
+
+let empty<'T> : 'T[] = [||] // TODO: Shoud it be a ctor function with cons.Create 0?
```
alfonsogarciacaro: As this is an empty array, I guess it doesn't matter if it's empty or not. You can also use `Array.empty`.

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+
+let inline singleton value (cons: ArrayCons) = cons.From [|value|]
+
+// TODO: CompiledName: Initialize
+let init count initializer (cons: ArrayCons) =
```
alfonsogarciacaro: Fable is going to use the compiled names (with the first letter in lower case) to locate the methods in the core library.

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+    if count < 0 then invalidArg "count" LanguagePrimitives.ErrorStrings.InputMustBeNonNegativeString
+    let result = cons.Create count
+    for i = 0 to count - 1 do
+        result.[i] <- initializer i
+    result
+
+[<CompiledName("Pairwise")>]
+let pairwise (array: 'T[]) (cons: ArrayCons) =
```
alfonsogarciacaro: Although it uses the compiled names (e.g. `iterate` instead of `iter`) Fable lowers the first letter of core methods. I did it like that because the core library was originally written in JS and that's the language standards. So in this case we shouldn't need the attribute.

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+
+let replicate count initial (cons: ArrayCons) =
+    // Shorthand version: = initialize count (fun _ -> initial)
+    if count < 0 then invalidArg "count" LanguagePrimitives.ErrorStrings.InputMustBeNonNegativeString
+    let result : 'T array = cons.Create count
+    for i = 0 to result.Length-1 do
+        result.[i] <- initial
+    result
+
+// TODO: CompiledName: Reverse
```
alfonsogarciacaro: Same as above: `reverse`

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
 module Array
 
+// This value is for use by compiled F# code and should not be used directly.
+// We can use LanguagePrimitives.ErrorStrings.
+// TODO: Approve please if we really need to use ErrorStrings here
+#nowarn "1204"
```
alfonsogarciacaro: This is fine. Yes, the "value is unused" warning is annoying when using emits :wink: Another option is to prefix the arguments with an underscore, but the `#nowarn` directive is absolutely ok :+1:

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+let tryItem index (array:'T[]) =
+    if index < 0 || index >= array.Length then None
+    else Some array.[index]
+
+let toList array = 
+    List.ofArray array
+
+
+// TODO: Consider optimizing functions below via manual inlining
+// -or- re-implementing via native js arrays functions (i.e. reduce(...))
```
alfonsogarciacaro: Sometimes a `for` loop is faster than native JS array functions so your current approach should be fine. About inlining, because this code will be distributed in JS form we cannot use `inline` here, inlining would have to be done in the `Replacements` module if necessary.

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,732 @@
...
+let rec existsOffset2 predicate array1 (array2:_ []) index =
+   if index = length array1 then false
+   else predicate array1.[index] array2.[index] || existsOffset2 predicate array1 array2 (index+1)
+
+let rec exists2 predicate array1 array2 =
+   if length array1 <> length array2 then failwith "Arrays had different lengths"
+   existsOffset2 predicate array1 array2 0
+
+let inline sum (array: 'T []) : 'T =
+   fold (+) LanguagePrimitives.GenericZero array
```
alfonsogarciacaro: Unfortunately we cannot use `inline` in public methods here because Fable would erase them. Currently fable-core is distributed in JS form and Fable won't have access to the F# source to read the inlined expression.

valery-vitko:
thanks for the review, and a status update jfyi - planning to resolve these this weekend

src/js/fable-core/Array2.fs
valery-vitko: Oh, indeed. Got it!

src/js/fable-core/Array2.fs
valery-vitko: Thanks for the tip with underscores! Will do it, `Emit`s are going to be a less common citizen in the module.

As for the FS1204, do we need `LanguagePrimitives.ErrorStrings` as in the following example:
```f#
    if array.Length = 0 then invalidArg "array" LanguagePrimitives.ErrorStrings.InputArrayEmptyString
```
(The code is borrowed from FSharp.Core sources). Or is Fable compiler message localization going to be a whole different story and we should stick with hardcoded strings for now?

src/js/fable-core/Array2.fs
valery-vitko: I'll consider re-using `Fable.Import.JS` bindings then unless it hurts generated code just for the sake
 of avoiding duplicated definitions in different places 😃

src/js/fable-core/Array2.fs
valery-vitko: Oh, that feeling when you thought you knew that js<->native transitions do hurt performance, but underestimate consequences by a magnitude... I think it's better to leave things as is right now and consider optimizing anything only after benchmark results chime in. Do you agree?

src/js/fable-core/Array2.fs
valery-vitko: Got it!

src/js/fable-core/Array2.fs
valery-vitko: Ok, got it - will replace all function names with compiled names with the first letter in lower case (e.g. `iteri` -> `iterateIndexed`).

src/js/fable-core/Array2.fs
valery-vitko: Agree! And as for "manual inlining" I meant that stuff when you unwrap/inline function by hand, and not asking the compiler via `inline` keyword. E.g. when you have `sum` implemented via 'fold` first, but then replaces `fold` with a loop/recursion code yourself.

Anyway, as I've said already in a comment above, it smells like a premature optimization to me :)

src/js/fable-core/Array2.fs
valery-vitko: So we don't mind if it's `[]` or, say, `new Int8Array()` here? Sounds confusing to me. Am I missing something?

valery-vitko:
@alfonsogarciacaro, I think I resolved all the issues/questions I had, except for the [one](https://github.com/fable-compiler/Fable/pull/1324#discussion_r163172292), [two](https://github.com/fable-compiler/Fable/pull/1324#discussion_r164270735) and [three](https://github.com/fable-compiler/Fable/pull/1324#discussion_r164271641) (sorry, got buried in "outdated" spoilers). So, I'm thinking, what should I do next?

src/dotnet/Fable.Compiler/Replacements.fs
```diff
@@ -1342,6 +1342,8 @@ module AstPass =
         | "physicalHash", _ ->
             CoreLibCall("Util", Some "getHashCode", false, i.args)
             |> makeCall i.range i.returnType |> Some
+        | "genericZero", _ -> makeEmit i.range i.returnType [] "0" |> Some
+        | "genericOne", _ -> makeEmit i.range i.returnType [] "1" |> Some
```
alfonsogarciacaro: I try to avoid `emit` in the Replacements module when possible as it prevents optimizations. We also need to check the type, because some classes can have a custom `Zero` static member. I can deal with that, could you please add a comment like `// TODO: Emit Fable AST and check for types with custom zero/one (strings?)`.

src/js/fable-core/Array2.fs
alfonsogarciacaro: Sorry, I noticed a typo in my comment above, I just fixed it. In theory, using `[]` or `new Int8Array(0)` should be the same, because Fable is prepared to accept both dynamic and typed arrays. But it's true this can affect JS optimizations expecting monomorphic functions so it may be better to use `cons.Create(0)` instead.

src/js/fable-core/Array2.fs
alfonsogarciacaro: Oh, I didn't know we could use `LanguagePrimitives.ErrorStrings` 😮 I've using hardcoded strings in JS but it makes sense to use the error strings here in F#, thanks for the tip!

src/js/fable-core/Array2.fs
alfonsogarciacaro: Sounds good! Yes, let's not worry too much about performance at this point, but leave a comment where in doubt so we can benchmark late :smile:

alfonsogarciacaro:
Thanks a lot for checking @valery-vitko! I've replied to your comments, please tell me if you've any other question.

I really appreciate your help! ❤️ I have the following in mind for fable-core at the moment, please have a look and tell me what you think:

- Do more or less the same as for Array with the List module.
- The Map, Set, and BigInt modules where converted from F# and adapted (the first two were taken from FunScript and BigInt from FSharp.Core). It'd be nice to keep the original F# (with necessary modification) instead.
- The Long.ts module was "borrowed" from another repo but it has two problems: uses class methods instead of module functions so tree shaking doesn't work well, and the operators have different as the F# compiled names, so we need [a special treatment in Replacements module](https://github.com/fable-compiler/Fable/blob/a844ca0621548c3dfb20229f9c7c04a030856207/src/dotnet/Fable.Compiler/Replacements.fs#L405-L422) and other places. It'd be great if we could change the code to use module members with the expected operator names.

If you'd like to take any of these tasks, please tell me and I'll write an issue with more information. I also need to make sure I didn't break the build script for `dev2.0` 😅

src/dotnet/Fable.Compiler/Replacements.fs
valery-vitko: Got it!

Though I replaced GenericZero/GenericOne and FSharp built-in generic ops with 
```f#
    [<Emit("$0 + $1")>]
    let inline nativeAddition (_x: 'T) (_y: 'U): 'V = jsNative

    [<Emit("$0 / $1")>]
    let inline nativeDivision (_x: 'T) (_y: 'U): 'V = jsNative

    [<Emit("0")>]
    let inline nativeZero<'T> (): 'T = jsNative
```
inside of Array2.fs. ~~Please don't hit me but~~That's a quick workaround to provide non-inlined version of sum's and average's. Not quite sure though to combine both non-inlined and generic zero/ops in Fable. But maybe at least this generic stuff should be moved somewhere else (e.g. to `LanguagePrimitives.Fable` module).

src/js/fable-core/Array2.fs
```diff
@@ -1,16 +1,750 @@
...
+        result.[i] <- group.Key, cons.From group.Value
+        i <- i + 1
+
+    result
+
+let ofArray (arr: 'T[]) = arr
+
+let inline private emptyImpl (cons: ArrayCons) = cons.Create(0)
+
+let empty = emptyImpl
```
valery-vitko: So, `empty` expects `cons` now, and `skip`, `skipWhile`, `take` and `takeWhile` also do as depending on `empty`.

valery-vitko:
@alfonsogarciacaro, no problem, my pleasure 😄 

As for the tasks for fable-core, they all sound fun. I could go with List module (if we feel confident about Array module port that there are no hidden drawbacks I'll repeat in List). But I'd be glad to take any of these.

As for the build scripts, there are indeed some `dotnet-fable` leftovers here and there. Clean them up locally and fixed Fable.Plugins.NUnit. Many tests still fail, though, I believe due to codebase being in transition. Anyway, I can send a PR if you'd like.

src/dotnet/Fable.Compiler/Replacements.fs
alfonsogarciacaro: Cool! If necessary Fable can inject zero and one as arguments the same way it does with the array constructor.

alfonsogarciacaro:
Yes, I got carried away and I'm breaking a lot of things... always with the illusory hope that this time things will be _right_ 😅 Don't worry too much about the tests now (I'm also planning to remove the NUnit plugin). If you want we can merge this now and I'll try to repair the tests to see which are failing, while you move on to the List module :)

valery-vitko:
I don't think that breaking some stuff in a dev branch at the beginning makes much trouble 😃 At least it didn't for me this time.

I don't mind if we go ahead with the merge and move on to the next thing 👍

src/dotnet/Fable.Compiler/Replacements.fs
valery-vitko: Yes, indeed! Injecting as arguments! `ArrayCons` being injected already should've given me a hint... If it's more feasible than `LanguagePrimitives` with a special treatment in replacements, then I'll go for it in `Lists` module and also make changes in `Array2.fs` also.

alfonsogarciacaro:
Sorry for taking so long to merge this, @valery-vitko. `dev2.0` is finally in a state where I can start integrating this, I'll try to restore the array tests as soon as possible. Thanks again for the awesome work!

PR #1325, opened 2018-01-15
Removed duplicated word "like"

LiteracyFanatic:


alfonsogarciacaro:
That's intended: we _like_ VS Code very much... sorry, I had to make the bad pun :wink:

Nice catch, thank you!

Issue #797, opened 2017-04-03
Pass current instance as first argument to imported extension methods


Zaid-Ajaj:
### Description
When writing a binding (function import) as an extension method, the current instance (`self`) should be *created* and *passed* as first argument to the imported function. 

for example, here I am just importing an existing function `toUniversalTime` from `fable-core/Date`:
```fsharp
open System
open Fable
open Fable.Core
open Fable.Core.JsInterop

[<AutoOpen>]
module Extensions = 
  type System.DateTime with
    [<Import("toUniversalTime", "fable-core/Date")>]
    member self.ToVerySpecialUniversalTime() : DateTime = jsNative

let x = DateTime(2015, 1, 1).ToVerySpecialUniversalTime()
```

See [sample here](http://fable.io/repl#?code=open%20System%0Aopen%20Fable%0Aopen%20Fable.Core%0Aopen%20Fable.Core.JsInterop%0A%0A%5B%3CAutoOpen%3E%5D%0Amodule%20Extensions%20%3D%20%0A%20%20type%20System.DateTime%20with%0A%20%20%20%20%5B%3CImport(%22toUniversalTime%22%2C%20%22fable-core%2FDate%22)%3E%5D%0A%20%20%20%20member%20self.ToVerySpecialUniversalTime()%20%3A%20DateTime%20%3D%20jsNative%0A%0Alet%20x%20%3D%20DateTime(2015%2C%201%2C%201).ToVerySpecialUniversalTime()%0A) 

### Why
Extension methods can be overloaded, perfect for importing foreign functions with different parameters instead of using U1, U2 ....etc.

I am not sure how the compiled code should look like, in the compilation in seems that Babel has detected that `DateTime_ToVerySpecialUniversalTime` is just the same as `toUniversalTime` and reduced the call to `toUniversalTime` (see code below). That doesn't matter as long as the instance of of the `DateTime` is created and passes as first argument which is not the case:

### Actual
```javascript
import { toUniversalTime } from "fable-core/Date";

export const Extensions = function (__exports) {
  const DateTime_ToVerySpecialUniversalTime = __exports["DateTime.ToVerySpecialUniversalTime"] = toUniversalTime;
  return __exports;
}({});

// Missing definition of DateTime(2015, 1, 1)
export const x = toUniversalTime();
```

### Expected
```javascript
import { toUniversalTime } from "fable-core/Date";
import { create } from "fable-core/Date";

export const Extensions = function (__exports) {
[...]

PR #1327, opened 2018-01-15
WIP: Pass current instance as first argument to imported extension methods #797

valery-vitko:
As discussed in #797, Fable 2.0 should fix the issue. So added a failing test to revisit the issue when the time comes.


alfonsogarciacaro:
Good idea, thank you!

PR #1329, opened 2018-01-17
Display the status of the correct build in Readme

vbfox:
I checked the readme and fable isn't in a good state: All builds are failing, windows linux :-(

Turns out it's false! It's only the configuration of the build badges that was taking the latest branch or PR and displaying that instead of master that was beautifully green...

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ### The compiler that emits JavaScript you can be proud of!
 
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Ups! I usually checked the green ticks next to the commits so I didn't notice. Thanks a lot for fixing it :tada:

PR #1331, opened 2018-01-18
Implement replacements for eprintf and friends

### LINKED ISSUE #1328: eprintfn not implemented
When trying to use `eprintfn` I am getting:

```
error FABLE: Cannot find replacement for Microsoft.FSharp.Core.ExtraTopLevelOperators::PrintFormatLineToError
```

I would expect this to compile to `console.error`

inosik:
Closes #1328.

I had to do this twice, because we have `eprintf` and `Printf.eprintf`.

Some questions:

1. How to test this?

    I tested it with `FableCompilerQuickTest` and it worked, but I guess actually printing to stderr would make the CI build fail. Maybe add a method which uses `eprintf` but omit the `Test` attribute?

2. Should we omit the lambda expression in the printing functions? In other words, do this change:

    ``` patch
     export function toConsole(arg: IPrintfFormat) {
    -   return arg.cont((x) => { console.log(x); });
    +   return arg.cont(console.log);
     }
    ```

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -779,6 +779,9 @@ module AstPass =
         | "printFormatLineToTextWriter" ->
             // addWarning com i.fileName i.range "fprintfn will behave as printfn"
             ccall i "String" "toConsole" i.args.Tail |> Some
+        | "printFormatToError"
+        | "printFormatLineToError" ->
+            ccall i "String" "toStdErr" i.args |> Some
         | "printFormat" ->
             // addWarning com i.fileName i.range "printf will behave as printfn"
             ccall i "String" "toConsole" i.args |> Some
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
That was super fast, thanks @inosik!

This is a good concern. TBH, we don't have a good solution for testing anything beyond assertions 😕 In fact, `printfn` is not being tested at the moment, only `sprintf`, but given they only differ in the continuation I just hope everything will be fine 🙏 :wink: If you have any idea about how we could test these things it'll be very welcome!

About using `arg.cont(console.log)`, yes, that makes sense 😄 It's in line of #1307 Could you please implement it in your PR?

inosik:
> I just hope everything will be fine :pray: :wink:

My favorite type of testing :smile:

As I see it, we have two parts which we might want to test: if we have a replacement in the first place and if it does the right thing. The first one is easy: just try to compile a snippet which contains the code. The second one is difficult without actually executing the code.

Issue #1341, opened 2018-02-13
List.tail [] yields undefined instead throwing an error


robkuz:
### Description

    let x = List.tail []

should throw an error like in .NET. instead yields an `undefined` which yields a RT error at a later in unrelated code.

### Expected and actual results

Throw the same error as .NET

### Related information

* Fable version (`dotnet fable --version`): 1.3.8
* Operating system: Windows 10

alfonsogarciacaro:
@inosik Fixed this :+1:

PR #1342, opened 2018-02-16
Throw errors on List.Head and List.Tail with empty lists

inosik:
I hope this works 😄

### DIFF
--- a/src/dotnet/Fable.Compiler/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Replacements.fs
@@ -1795,7 +1795,21 @@ module AstPass =
                 // A static 'length' method causes problems in JavaScript -- https://github.com/Microsoft/TypeScript/issues/442
                 let seqMeth = if meth = "length" then "count" else meth
                 ccall "Seq" seqMeth (staticArgs c args)
-            | List -> let c, _ = instanceArgs c args in prop meth c
+            | List ->
+                match meth with
+                | "head" | "tail" ->
+                    match c, args with
+                    | Some list, _ ->
+                        // xs.Head or xs.Tail
+                        ccall "List" meth [list]
+                    | _, [_] ->
+                        // List.head or List.tail
+                        ccall "List" meth args
+                    | _ ->
+                        addErrorAndReturnNull com i.fileName i.range "WTF?"
+                | _ ->
+                    let c, _ = instanceArgs c args
+                    prop meth c
             | Array ->
                 let c, _ = instanceArgs c args
                 if meth = "head" then makeGet i.range i.returnType c (makeIntConst 0)
--- a/src/js/fable-core/List.ts
+++ b/src/js/fable-core/List.ts
@@ -128,3 +128,19 @@ export function splitAt<T>(index: number, xs: List<T>): [List<T>, List<T>] {
   }
   return [ofArray(first), last];
 }
+
+export function head<T>(xs: List<T>): T {
+  if (xs.head !== undefined) {
+    return xs.head;
+  } else {
+    throw new Error("The input list was empty.");
+  }
+}
+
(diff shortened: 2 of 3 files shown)


src/js/fable-core/ListClass.ts
```diff
@@ -16,12 +16,26 @@ export function ofArray<T>(args: T[], base?: List<T>) {
...
 
   constructor(head?: T, tail?: List<T>) {
-    this.head = head;
-    this.tail = tail;
+    this._head = head;
+    this._tail = tail;
+  }
+
+  get head(): T {
+    if (this._head != null)
```
davidtme: Will this work? what if I had a list of null?

inosik: Good catch, `[null].Head` fails.

inosik:
Wow, that makes 36 tests fail.

inosik:
Okay, the only things left are pattern matching over lists and serialization. I'm not sure how to fix this.

src/js/fable-core/ListClass.ts
alfonsogarciacaro: Indeed this can be an issue, even `undefined` can be a problem if we change the representation of `()` and you have a `unit list`. In theory, you should be able to check `this._tail` here as well. Empty list will always have null tails.

alfonsogarciacaro:
Thanks a lot for this @inosik! This should a good solution to #1341 for Fable 1.x (the CI failure seems unrelated, I need to check). Unfortunately it probably won't work for Fable 2 as I'm considering representing lists with arrays for performance and readability (as Bucklescript does). Another option is to [replace calls `tail` or `head`](https://github.com/fable-compiler/Fable/blob/a8d4213e7c3edba82685da95680545c6c363338c/src/dotnet/Fable.Compiler/Replacements.fs#L1798) with a special method that checks for empty lists, [as in the case of Option/getValue](https://github.com/fable-compiler/Fable/blob/a8d4213e7c3edba82685da95680545c6c363338c/src/js/fable-core/Option.ts#L46).

inosik:
@alfonsogarciacaro I think they are, all of the failing tests either involve pattern matching over lists or serialize lists.

> Another option is to replace calls to tail or head with a special method that checks for empty lists

I tried something like this yesterday, but my problem was that we don't distinguish between explicit calls to `Head` or `Tail` and the implicit calls that are generated when pattern matching over lists. That made the two new tests I added fail.

alfonsogarciacaro:
[This is the current code for accessing head and tail in pattern matching](https://github.com/fable-compiler/Fable/blob/a8d4213e7c3edba82685da95680545c6c363338c/src/dotnet/Fable.Compiler/FSharp2Fable.fs#L770). In Fable 2, [there will be specific cases in the AST](https://github.com/fable-compiler/Fable/blob/2065c4b0419b3cf7bbd17eb3114b4b4d80dfa3b4/src/dotnet/Fable.Compiler/AST/AST.Fable.fs#L133-L134) so it will probably be easier then to discriminate how we are accessing list members.

I'm currently a bit focused on Fable 2 development. We can try to fix this in Fable 1 but, as it's not critical, we can wait a bit until Fable 2 is alpha and revisit this then (the solution will necessarily be different). What do you think?

inosik:
Cool, thanks! I'll try to make this work in Fable 1 today or tomorrow in this PR, later we can cherry pick the new tests and then fix it in Fable 2.

inosik:
@alfonsogarciacaro The code you pointed out above isn't used in pattern matching, but I managed to get around this. Every call from F# will use the functions in the `List` module and the rest of the TS/JS code can use the public properties of the `List<T>` class.

alfonsogarciacaro:
Awesome, thank you!

PR #1348, opened 2018-02-26
Fix expecto args order

MangelMaxime:
Expect.equal signature is:

```fs
val equal: 
   actual  : 'a     ->
   expected: 'a     ->
   message : string 
          -> unit
```


alfonsogarciacaro:
Thanks!

Issue #1374, opened 2018-03-22
toJson serializes Date in a DU as an object


valery-vitko:
### Description

`toJson` serializes Date in a DU field as an object. It gets  "{"kind":1}" instead of a round-trip datetime string.

### Repro code

```fsharp
DateTime(2016, 8, 4, 17, 30, 0, DateTimeKind.Utc) |> UnionWithDateTime.CaseWithDateTime  |> Fable.Core.JsInterop.toJson
```

### Expected and actual results

Expected: `{"CaseWithDateTime":"2018-03-21T23:00:00.000Z"}`
Actual: `{"CaseWithDateTime":{"kind":1}}`

### Related information

* Fable version (`dotnet fable --version`): 1.3.11
* Operating system: Windows 10

PR #1375, opened 2018-03-22
Fix client toJson serialization of DateTime in a DU field #1374

valery-vitko:


alfonsogarciacaro:
CI failure seems unrelated (I need to look into that), tests are passing so I'm merging this. Thanks a lot! I'll release a new version shortly :+1:

alfonsogarciacaro:
Released as 1.3.12 :+1:

PR #1379, opened 2018-03-27
Fix non-ValueType option deserialization

eugene-g:
Encountered a weird deserialization bug today. See tests for repro, hope they're self-explanatory.

@draganjovanovic1 suggested a hack with double deserialization, which solves the problem, but doesn't feel right. Any ideas?


eugene-g:
Sorry for noisy PR. The relevant commit is "Fix non-ValueType option deserialization".

Note to self: https://stackoverflow.com/questions/39048673/github-pull-request-shows-too-many-changes-commits

draganjovanovic1:
Let's try to explain what the problem is and how we found it.

We made our types in such way that invalid state is unrepresentable. Let's take the one from @eugene-g's test as an example and add validation for sake of clarity:

```fsharp
module NonValueType =
    type ValueTypeJsonConverter<'a> (toJson: 'a -> string, fromJson: string -> 'a) =
        inherit JsonConverter ()
        override __.CanConvert objType = objType = typeof<'a>
        override __.ReadJson (reader, _, _, _) =
            let rawJson = Linq.JToken.Load (reader) |> string
            (fromJson rawJson) :> obj
        override __.WriteJson (writer, value, _) =
            let value = value :?> 'a
            let json = toJson (value)
            writer.WriteRawValue (json)

    let private isValid = String.IsNullOrWhiteSpace >> not

    [<Newtonsoft.Json.JsonConverter(typeof<FoobarJsonConverter>)>]
    type T = private Foobar of string
        with
            static member CanCreate = isValid
            static member TryCreate (value) =
                if T.CanCreate value
                then Ok <| Foobar value
                else Error "Value can't be null or empty"
            static member Create (value) =
                match T.TryCreate value with
                | Ok value -> value
                | Error msg ->
                    invalidArg "value" msg
            static member ToJson (Foobar value) = JsonConvert.SerializeObject (value, Fable.JsonConverter())
            static member FromJson = T.Create
            static member Value (Foobar s) = s
            override x.ToString () =
                let (Foobar value) = x
                value
    and FoobarJsonConverter () =
        inherit ValueTypeJsonConverter<T> (T.ToJson, T.FromJson)

    let canCreate = T.CanCreate
    let tryCreate = T.TryCreate
    let create = T.Create
    let value = T.Value
```

Type is single case DU with private constructor so we force use of the `create`/`tryCreate` function or `T.Create`/`T.TryCreate` static member. Because of the private constructor and to prevent creation of invalid instance via deserialization we implemented custom JsonConverter as well.

If we have an instance of `NonValueType.T` we are able to serialize and deserialize it w/o any issues. When we try to work with `NonValueType.T option`, serialization works but deserialization breaks if serialized value was `None` (serializes as `null`).

Let's analyze the deserialization source code before PR:

```fsharp
    override x.ReadJson(reader, t, existingValue, serializer) =
        ...
        | true, Kind.Option ->
            let innerType = t.GetGenericArguments().[0]
            let innerType =
                if innerType.IsValueType
                then (typedefof<Nullable<_>>).MakeGenericType([|innerType|])
                else innerType
            // here we know that innerType = typeof<NonValueType.T>
            // and try to deserialize and check for null
            // our T.FromJson is being called
            // and we get our exception (see invalidArg "value" msg above)
            let value = serializer.Deserialize(reader, innerType) 
            let cases = FSharpType.GetUnionCases(t)
            if isNull value
            then FSharpValue.MakeUnion(cases.[0], [||])
            else FSharpValue.MakeUnion(cases.[1], [|value|])
        ...
```

If we take a look in the comment I added in the source code above we may understand why deserialization failed, it's because we don't allow instance of `NonValueType.T` to be created with invalid value (`null` in our case).

The fix:
```fsharp
    override x.ReadJson(reader, t, existingValue, serializer) =
        ...
        | true, Kind.Option ->
            // here we check for null before we try deserialization and return None immediately
            match reader.Value with
            | null -> upcast None
            | _ ->
                let innerType = t.GetGenericArguments().[0]
                let innerType =
                    if innerType.IsValueType
                    then (typedefof<Nullable<_>>).MakeGenericType([|innerType|])
                    else innerType
                let value = serializer.Deserialize(reader, innerType)
                let cases = FSharpType.GetUnionCases(t)
                if isNull value
                then FSharpValue.MakeUnion(cases.[0], [||])
                else FSharpValue.MakeUnion(cases.[1], [|value|])
        ...
```

Here we added null check w/o trying to deserialize value which makes deserialization of `null` to `NonValueType.T option` possible.

Interesting point is that the current implementation used to work even for F# record types since `null` to record deserialization does not throw exception even it deserializes to null reference and can lead to nasty `NullReferenceException` if not checked manually with e.g. `obj.ReferenceEquals(x, null)`.

draganjovanovic1:
Ouch, this PR was pretty naive fix. Please do not merge, new amend with explanation and tests coming soon :)

eugene-g:
Should be fine now, ready to merge.

draganjovanovic1:
I promised explanation as well but one look to newly added tests should be self explanatory 😄

alfonsogarciacaro:
Thanks a lot for the fix, the explanation and the tests! Sorry, I had a cold yesterday and couldn't review this. I just merged it and published it as Fable.JsonConverter 1.0.8 :+1: Thanks again!

draganjovanovic1:
No problem. We temporarily switched from NuGet to GitHub reference since it is one file only to not wait. Now we can switch back to NuGet reference :-) Thanks for the new version.

PR #1385, opened 2018-04-12
Add some replacements

zaaack:
@alfonsogarciacaro Here are the rest replacements except for collections, I didn't write many tests but I think it's time to send a PR to avoid more conflicts.

I passed `argTypes` to all `Helper.CoreCall` that I write, but I'm curious is there any differences between `args |> List.map (fun a -> a.Type)` and `i.ArgTypes`, because they look really similar to me.


src/dotnet/Fable.Compiler/Transforms/Transforms.Util.fs
```diff
@@ -122,6 +122,8 @@ module AST =
     let makeNumConst (x: float) = NumberConstant (float x, Float64) |> Value
     let makeDecConst (x: decimal) = NumberConstant (float x, Float64) |> Value
 
+    let argTypes (args: Expr list) = args |> List.map (fun a -> a.Type)
```
alfonsogarciacaro: This is actually not called, right? Maybe we should remove it to avoid confusion with `info.ArgTypes` as I explain in the comment to the PR.

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -1401,6 +1406,498 @@ let unchecked (com: ICompiler) r t (i: CallInfo) (_: Expr option) (args: Expr li
...
+let log (com: ICompiler) r t (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
+    let v =
+        match args with
+        | [] -> []
+        | [v] -> [v]
+        | (Fable.Value (Fable.StringConstant _))::_ ->
+            Helper.CoreCall("String", "format", t, args, i.ArgTypes)
+            |> List.singleton
+        | _ -> [args.Head]
+    Helper.GlobalCall("console", t, args, ?argTypes=None, ?thisArg=None, memb="log", ?loc=r)
```
alfonsogarciacaro: It's not necessary here to write `?argTypes=None`. Optional arguments only need the name when they don't appear in order (as `memb="log"` here) or when we pass an option value, prefixed with `?` (as in `?loc=r`), but we can omit them if the value is `None`.

alfonsogarciacaro:
This is awesome @zaaack, thanks a lot for this! :clap: :clap: :clap: I left a couple of comments, but I can make the adjustments myself so let's merge this :)

About `info.ArgTypes`, it's true it's confusing. If you do `args |> List.map (fun a -> a.Type)` you get `the types of the argument expressions` while `info.ArgTypes` gives `the types of the arguments as defined in the method signatures`. These will match most of the times, but there's a case where they don't exactly and it's important for the uncurrying optimization: _function arguments returning a generic value_.

Consider the following code:

```fsharp
let apply2 (f:'a->'b->'c) x y = f x y

let test() =
    let f: int->int->int->int = fun x y z -> x + y + z
    let f2 = apply2 f 1 2
    f2 3
```

In `apply2` signature `f` has 2-arity. However, as it returns a generic value, the signature can also accept functions returning functions, or _curried functions with a higher arity_. In `test` we pass a function with 3-arity to `apply2`. If we just uncurry it with 3 arguments, it'll be a problem, because the body of `apply2` will assume `f` only has two arguments. That's why the uncurry optimization pass needs to know the argument types as defined in the signature, not the actual types of the argument expressions.

Currently Fable 2 compiles this code as follows:

```js
function apply2(f, x, y) {
  return f(x, y);
}

function test() {
  return apply2(function f(x, y) {
    return function (z) {
      return x + y + z;
    };
  }, 1, 2)(3);
}
```

> As you can see, the lambda is uncurried with 2-arity (corresponding to the signature of `apply2`) instead of using 3-arity, as in the actual type of the expression.

We should probably change the name of the info field to `SignatureArgTypes` or similar to be more explicit about its meaning, what do you think?

I hope this helps! Please let me know if you have more questions and thanks again for your help.

alfonsogarciacaro:
Ah, and don't worry about the tests. @EdouardM is taking care of them 😉 So when they're ready we can run all of them and see what's broken :+1:

zaaack:
Thanks for your fast and detailed explanation! Just wake up and looking on my phone. I think changing the name or add some comments on the field definition would help me lot.

Here actually has other questions, I see you removed Expr.Wrap, and wondering whether it's Ok to remove the wrap type in guids constructor replacement.https://github.com/fable-compiler/Fable/pull/1385/files#L1795 ,  because I think I removed the right typeinfo of thisArg.

alfonsogarciacaro:
Good question. I removed `Expr.Wrap` because I realized I it was a sloppy way to bypass type checking. For other cases where the type in Fable AST and the final expression in JS may differ (like options) I have added new elements to the AST (e.g. `ValueKind.NewOption`) but I'm not sure here. There's [another thread](https://github.com/fable-compiler/repl/issues/39#issuecomment-378829275) suggesting we use a proper type for Guid instead of just strings. Either we do that, or we add Guids to Fable AST 🤔

zaaack:
At first actually I created a `create` function in Guid.ts and using CoreCall to add type. Then I removed it because Im not sure it's needed. Looks it's still needed I guess.

alfonsogarciacaro:
You're right, that's another option to add the type. Maybe the easiest for now.

PR #1390, opened 2018-04-18
Implement Set.fs

MangelMaxime:
@alfonsogarciacaro 

This PR only implement Set.fs in fable-core and also activate the tests for it.

I still wasn't able to solve the last two test.

PS: I didn't want to squash all the comits because we already included some commits from dev2 in this branch


alfonsogarciacaro:
This is awesome @MangelMaxime, thanks a lot for this! :clap: 🥇:clap: Fable 2 release is much closer now 🏎

PR #1392, opened 2018-04-21
Fix map construction

alfonsogarciacaro:
Hey @MangelMaxime! This should fix the map construction, however I've noticed most of the static members in fable-core Set.fs and Map.fs are not actually exposed in F# code, so we can improve the code in Replacements. I've to stop now, but if you want, please wait until I make also the changes for Set and merge them here so you can keep working on your branch with cleaner code.

> Apparently I also broke dev2.0 branch with latest commit, so I also need to check that.


MangelMaxime:
Thank you :)

I will wait your next changes and go before working on the Map again then.

alfonsogarciacaro:
I've merged with `dev2.0` and cleaned the code. Please feel free to merge this and uncomment the rest of map tests (except the JSON ones). Hopefully most of them should pass now :)

PR #1393, opened 2018-04-22
Feature/work on map

MangelMaxime:
We are down to 3 failing tests I added them to `QuickTest.fs`

## Failing test: Map.map works

```fs
testCase "Map.map works" <| fun () ->
    let xs = Map [1,1.; 2,4.; 3,9.; 4,16.]
    let ys = xs |> Map.map (fun k v -> v * 2.)
    ys.[3] |> equal 18.
```

## Reason

Here is the def of `MapTree`
```fs
type MapTree<'Key,'Value when 'Key : comparison > =
    | MapEmpty
    | MapOne of 'Key * 'Value
    | MapNode of 'Key * 'Value * MapTree<'Key,'Value> *  MapTree<'Key,'Value> * int
```

The problem is that `MapEmpty` is translated to null and so when doing `m[0]` we got an exception `Cannot read property '0' of undefined`

```js
function MapTreeModule_mapi(f, m) {
  if (m[0] === "MapOne") {
    return ["MapOne", m[1], f(m[1], m[2])];
  } else if (m[0] === "MapNode") {
    return ["MapNode", m[1], f(m[1], m[2]), MapTreeModule_mapi(f, m[3]), MapTreeModule_mapi(f, m[4]), m[5]];
  } else {
    return MapTreeModule_empty();
  }
}
```

----

## Last 2 tests

The last 2 failing tests:

- `"Map.tryFindKey works"`
- `"Map.toArray works"`

seems related to an iterator problems:

Error: `xs[Symbol.iterator] is not a function`

----

⚠️ **Important:** I also added a basic support for `Option.isSome` and `Option.isNone` ⚠️ 

The code is [here](https://github.com/fable-compiler/Fable/pull/1393/files#diff-a34e57b30a53951ff9b99ce4076feaa4R1271). I just added what I needed so I think we are not covering the case where we call `instance.IsSome` and `instance.IsNone`.


alfonsogarciacaro:
Awesome, thanks a lot for this! I'll merge and try to fix the failing tests :+1:

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -1248,19 +1261,23 @@ let options (com: ICompiler) r (t: Type) (i: CallInfo) (thisArg: Expr option) (a
...
-    //     // For some reason, F# compiler expects them to be applicable
-    //     | _ -> makeLambda [] comp |> Some
+    | "IsSome", _, [c]
+    | "IsNone", _, [c] ->
+        let op =
+            if i.CompiledName = "IsSome"
+            then BinaryUnequal
+            else BinaryEqual
+
+        makeEqOp r c (Fable.Value (ValueKind.Null (firstGenArg com r i.GenericArgs))) op |> Some
```
alfonsogarciacaro: Now the AST contains `OptionTest`. It's better to use it because it makes it easier to change the option representation if necessary. I'll do that :+1:

MangelMaxime: Oh ok, I didn't know.
Thank you :)

PR #1401, opened 2018-05-09
Fix #1400: Fix Deserializing into int64 from json

### LINKED ISSUE #1400: ofJson unable to parse int64 when not represented as strings
### Description

ofJson unable to parse int64 when not represented as strings. It will however parse a string representation of the same number

### Repro code

```
module Program


open Fable.Core
open Fable.Core.JsInterop
open Fable.Import.Browser

type Account =
    {
        Id: int64
    }
let val1: Account = {Id = 123L }


// works
let val2 = """ {"Id": "123"} """ 
let val3 = ofJson<Account> val2 //
let deserialized = ofJson<Account> val2

//doesnot work
let apiResponse = """ {"Id": 123} """ // .net Api returns int64 like this without quotes around 123
let deserializedError = ofJson<Account> apiResponse
```

### Related information

* Fable version (`dotnet fable --version`): 1.3.12
* Operating system: Windows

vikranthc:


alfonsogarciacaro:
Awesome, thanks a lot for this! :clap: :clap: I'll release a new version with the fix right away :+1:

alfonsogarciacaro:
Released as 1.3.17. Thanks again!

Issue #1398, opened 2018-05-06
Use the WarningsAsErrors-Property of MSBuild


inosik:
### Description

The `WarningsAsErrors`-Property is currently ignored by Fable. @nojaf asked about that on Gitter yesterday. He wanted to make the compilation fail on incomplete pattern matches.

### Repro code

``` xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <!--
    Warning FS0025: Incomplete pattern matches on this expression.
    -->
    <WarningsAsErrors>FS0025</WarningsAsErrors>
  </PropertyGroup>
</Project>
```

### Expected and actual results

Fable should fail on a project with this setting, just as `dotnet build` does. But we currently generate a warning instead of an error.

### Related information

* Fable version (`dotnet fable --version`): 1.3.16

alfonsogarciacaro:
Thanks for the detailed report @inosik. It shouldn't be difficult to do this. I'd prefer to do it in `dev2.0` branch because it has already diverged a lot with master and probably I will have to merge the changes manually checking the commits one by one.

Maybe you or someone else could send a PR. I think the necessary steps are:

1 - Read `WarningsAsErrors` [from MSBuild properties](https://github.com/fable-compiler/Fable/blob/467c56dafee68df47c35f63bddfcb773c4a04dfc/src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs#L188-L189) and add it as an option to FableCompiler object.
2 - When adding a warning and the tag is `FSHARP` (not `FABLE`), [replace it with an error](https://github.com/fable-compiler/Fable/blob/467c56dafee68df47c35f63bddfcb773c4a04dfc/src/dotnet/Fable.Compiler/Transforms/State.fs#L104-L109) if the option is set.

inosik:
> I'd prefer to do it in dev2.0 branch

Sure.

> Maybe you or someone else could send a PR?

I can look into that 👍 

1. We should already get all the properties from Dotnet.ProjInfo, but most of them seem to get lost somewhere. I'll look closer at the Project Cracker.

2. We don't have to do anything special. If we pass the `--warnaserror:FS0025` option to the FCS, it will turn the warning to an error and Fable will report it correctly. I already tried that yesterday real quick by simply appending the option to the `OtherOptions` array here:

[...]

PR #1402, opened 2018-05-10
Pass more compiler options to FCS

inosik:
See #1398.

As it is now, this would pass *all* options we previously ignored to FCS, which means stuff like `--optimize` would be passed too. Here's a list of options which are set by default for `QuickTest.fsproj`:

``` text
"-o:obj/Debug/netcoreapp2.0/QuickTest.dll"; "-g"; "--debug:portable";
  "--noframework"; "--define:TRACE"; "--define:DEBUG"; "--define:NETCOREAPP2_0";
  "--optimize-"; "--target:exe"; "--warn:3"; "--warnaserror:76";
  "--fullpaths"; "--flaterrors"; "--highentropyva-"; "--targetprofile:netcore";
  "--simpleresolution"; "--nocopyfsharpcore"
```

I think we need a filter function, which will filter the options that are useful for us, like `--warnaserror`, or `--nowarn`. What other options should we accept?

### DIFF
--- a/src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs
+++ b/src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs
@@ -178,6 +178,13 @@ let private getFsprojName (fsprojFullPath: string) =
     let i = fsprojFullPath.LastIndexOf('/')
     fsprojFullPath.[(i + 1) .. (fsprojFullPath.Length - 8)]
 
+let private isUsefulOption (opt : string) =
+    [ "--define"; "-d"
+      "--nowarn"
+      "--warnon"
+      "--warnaserror" ]
+    |> List.exists opt.StartsWith
+
 /// Use Dotnet.ProjInfo (through ProjectCoreCracker) to invoke MSBuild
 /// and get F# compiler args from an .fsproj file. As we'll merge this
 /// later with other projects we'll only take the sources and the references,
@@ -200,8 +207,10 @@ let fullCrack (projFile: string): CrackedFsproj =
                 let dllName = getDllName line
                 dllRefs.Add(dllName, line)
                 src, otherOpts
-            elif line.StartsWith("-") then
+            elif isUsefulOption line then
                 src, line::otherOpts
+            elif line.StartsWith("-") then
+                src, otherOpts
             else
                 (Path.normalizeFullPath line)::src, otherOpts)
     let projRefs =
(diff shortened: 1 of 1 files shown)


inosik:
Based on the [documentation of the F# compiler options](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/compiler-options), I think we should accept these options:

| Option | Description |
|---|---|
| `--define:<string>`, `-d:<string>` | Defines a symbol for use in conditional compilation. |
| `--nowarn:<int-list>` | Disables specific warnings listed by number. Separate each warning number by a comma. You can discover the warning number for any warning from the compilation output. |
| `--warnon:<int-list>` | Enable specific warnings that might be off by default or disabled by another command line option. In F# 3.0, only the 1182 (unused variables) warning is off by default. |
| `--warnaserror[+\|-] [<int-list>]` | Enables or disables the option to report warnings as errors. You can provide specific warning numbers to be disabled or enabled. Options later in the command line override options earlier in the command line. For example, to specify the warnings that you don't want reported as errors, specify `--warnaserror+ --warnaserror-:<int-list>`. |

alfonsogarciacaro:
Thanks for this, @inosik! I agree we should send over all the "warn" related options (Fable passes "--warn:3" by default). But I have some doubts about `--define`. So far, Fable is ignoring the define symbols from the .fsproj/MSBuild and uses the ones coming from the JS client configuration (besides adding `FABLE_COMPILER`). This is because it's a pain to add symbols in the .fsproj under some condition, and I find it much easier to do it in the JS config file. Also, in some cases (like when compiling Fable itself to JS), we use the `NETCOREAPP2_0` symbol to identify if the app is being compiled for netcore.

One option that I've sometimes considered is passing **both** symbols from MSBuild and the JS config, but I've the feeling this can be confusing for the users. What do you think?

inosik:
I'd prefer to do as much configuration in the project file as possible (if it makes sense), so that editor tooling can work with it as well. But I don't mind ignoring `--define` for now. We can still reconsider that and add it back later.

alfonsogarciacaro:
Perfect! Let's merge this then :+1:

inosik:
We should document that these options are now supported in the project file. The MSBuild properties are `DisabledWarnings` (`--nowarn`), `TreatWarningsAsErrors` (`--warnaserror+`) and `WarningsAsErrors` (`--warnaserror:<numbers>`). Where should we add that?

alfonsogarciacaro:
Hmm, good question. There's actually no documentation about how the .fsproj is being parsed at the moment. Maybe a section in the [.NET compatibility document](http://fable.io/docs/compatibility.html)? Or we can also create a different page for this, and also explain how to create Fable libraries (with instructions on what should be written in the .fsproj to include the sources in the package).

PR #1407, opened 2018-05-24
Fix list comprehensions returning None

inosik:
On a side note: It got a little bit harder to just recompile and run tests. You have to remember to remove the `.fable/fable-core` directory if you changed something in `fable-core`.

Should I create the same PR for `master` or can you just cherry pick the change to `Seq.ts`, @alfonsogarciacaro?

### DIFF
--- a/src/js/fable-core/Seq.ts
+++ b/src/js/fable-core/Seq.ts
@@ -605,7 +605,7 @@ export function scanBack<T, ST>(f: (x: T, st: ST) => ST, xs: Iterable<T>, seed: 
 }
 
 export function singleton<T>(y: T) {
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Ah, you're right, I'm deleting the folder in [the script I usually use for the tests]() but I forgot to do it in build.fsx. Thanks for pointing it out!

About master, `dev2.0` has diverged so much that I'll probably just replace the branch (if I manage to learn how to do that in git). I'll have to check manually which things in master we need to port to dev2.0. I'm trying to release a Fable 2 alpha version, so let's focus on that for now and not worry about master :)

Thanks a lot for the fix!

PR #1409, opened 2018-05-27
benchmark update

ncave:
- Fable benchmark running on dotnet works now.
- Fable benchmark running on JS doesn't compile yet:
  - bunch of property getter errors: FABLE: Cannot resolve X.Y.Z.get_SomePropertyName
  - FABLE: Object member has kind ObjectMethod false but value is not a function


src/dotnet/Fable.Compiler/CLI/Agent.fs
```diff
@@ -207,7 +207,8 @@ let startAgent () = MailboxProcessor<Command>.Start(fun agent ->
                 let msg = Parser.parse msg
                 // lazy sprintf "Received message %A" msg |> Log.logVerbose
                 let isUpdated, state, activeProject = updateState checker state msg
-                let com = Compiler(msg.path, activeProject, Parser.toCompilerOptions msg)
+                let fableCore = (Path.getRelativePath msg.path activeProject.FableCore).TrimEnd('/')
+                let com = Compiler(fableCore, msg.path, activeProject, Parser.toCompilerOptions msg)
```
alfonsogarciacaro: Hmm, Fable 2 now uses [the CLI to pass an alternative `fable-core` location](https://github.com/fable-compiler/Fable/blob/37bb279213b21a80a00fe125e5859984bafcbf0e/src/dotnet/Fable.Compiler/CLI/Main.fs#L112-L117). This is because it was annoying in Fable 1 to edit the webpack.config.js every time I wanted to test a project with Fable development version. Maybe it's better to use it for the benchmark [as in here](https://github.com/fable-compiler/Fable/blob/37bb279213b21a80a00fe125e5859984bafcbf0e/src/tools/quicktest.sh#L41) and deprecate the option in the JS client (I think nobody else uses it anyway)?

src/dotnet/Fable.Compiler/Fable.Compiler.fsproj
```diff
@@ -40,7 +40,7 @@
       <PackagePath>prefercliruntime</PackagePath>
       <Pack>true</Pack>
     </Content>
-    <Content Include="..\..\..\build\fable-core\" PackagePath="fable-core\" />
+    <Content Include="../../../build/fable-core" PackagePath="fable-core" />
```
alfonsogarciacaro: I remember in previous dotnet SDK version there was a problem where MSBuild wouldn't include subdirectories if you didn't add `\` at the end of the path: https://github.com/NuGet/Home/issues/4795

Apparently it's fixed, but we need to test that :)

src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs
```diff
@@ -314,7 +314,7 @@ let createFableDir rootDir =
     fableDir
 
 let copyDirIfDoesNotExist (source: string) (target: string) =
-    if isDirectoryEmpty target then
+    if Directory.Exists source && isDirectoryEmpty target then
```
alfonsogarciacaro: Why did you have to add this, because non-existent `fable-core`? Maybe we can take two birds with one stone and fix this (and the comment above) by passing `--fable-core` through the CLI?

src/dotnet/Fable.Compiler/Transforms/State.fs
```diff
@@ -81,10 +81,9 @@ type Project(projectOptions: FSharpProjectOptions, implFiles: Map<string, FSharp
 
 /// Type with utilities for compiling F# files to JS
 /// No thread-safe, an instance must be created per file
-type Compiler(currentFile, project: Project, options) =
+type Compiler(fableCore, currentFile, project: Project, options) =
     let mutable id = 0
     let logs = Dictionary<string, string list>()
-    let fableCore = (Path.getRelativePath currentFile project.FableCore).TrimEnd('/')
```
alfonsogarciacaro: Again, if we deprecate passing fableCore from the JS client and use only the CLI we don't need these changes.

src/dotnet/Fable.JS/Fable.JS.fsproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../../../paket-files/ncave/FCS/fcs/fcs-fable/fcs-fable.fsproj" />
+    <ProjectReference Include="../../../../FSharp.Compiler.Service_fable/fcs/fcs-fable/fcs-fable.fsproj" />
```
alfonsogarciacaro: Hmm, going back to using repos in parallel directories is convenient for local testing, but will prevent building the repl as an Appveyor artifact. Maybe leave a comment at least so we can use Paket dependencies again when we make everything work with Fable 2?

src/dotnet/Fable.JS/Main.fs
```diff
@@ -195,14 +206,15 @@ let defaultManager =
         member __.GetCompletionsAtLocation(parseResults:IParseResults, line:int, col:int, lineText:string) =
             let res = parseResults :?> ParseResults
             getCompletionsAtLocation res line col lineText
-        member __.CompileToBabelAst(com: IFableCompiler, parseResults:IParseResults, fileName:string, optimized: bool) =
-            let com = com :?> CompilerImpl
+        member __.CompileToBabelAst(fableCore:string, parseResults:IParseResults, fileName:string, optimized: bool) =
```
alfonsogarciacaro: Ah! I realize now that you actually wanted to pass fableCore from the API not from the JS config file, sorry! Well, this makes sense, but then it may be confusing to have two ways to manage the fable-core path (using a global variable or passing it as a local parameter) probably we should remove the global variable for simplicity....

src/dotnet/Fable.JS/bench/package.json
```diff
@@ -1,12 +1,11 @@
...
   "private": true,
   "scripts": {
-    "prebuild": "dotnet restore --no-cache",
-    "fable": "dotnet ../../../../build/fable/dotnet-fable.dll start",
+    "fable": "dotnet ../../../../build/fable/dotnet-fable.dll start --port 61225",
     "build": "dotnet ../../../../build/fable/dotnet-fable.dll npm-run splitter",
     "webpack": "node ../../../../node_modules/webpack/bin/webpack -p",
     "rollup": "node ../../../../node_modules/rollup/bin/rollup -c",
-    "splitter": "node splitter.config.js",
+    "splitter": "node ../../../../src/js/fable-splitter/dist/cli -c splitter.config.js",
```
alfonsogarciacaro: Just an idea, probably we should check for the `splitter.config.js` (or `fable-splitter.config.js`) file by default like Webpack does.

src/dotnet/Fable.Compiler/Fable.Compiler.fsproj
ncave: This one actually was motivated by not being able to open the project in VS2017, it was complaining about the slash at the end.

src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs
ncave: Yeah, it doesn't hurt to check if source exists, it's just a check.

src/dotnet/Fable.Compiler/CLI/Agent.fs
ncave: - Perhaps, but OTOH it's a config option so it makes sense for it to be in the config object that is being passed to the splitter, and everywhere else. You can have an override in the CLI line as you already do, it's a bit strange with the mutable global and such, but whatever works.

src/dotnet/Fable.Compiler/Transforms/State.fs
ncave: This line was not changed, just moved it from within the constructor to a parameter so we can easily construct a Compiler with specific fableCore path, this way the Compiler is a bit more flexible.

src/dotnet/Fable.JS/Main.fs
ncave: I agree.

src/dotnet/Fable.JS/bench/package.json
ncave: That would be awesome, or default the config name to that.

src/dotnet/Fable.JS/Fable.JS.fsproj
ncave: Fixed.

ncave:
@alfonsogarciacaro I can't really say if we're closer, we'll see more when we get past the property getter errors.

alfonsogarciacaro:
Ok, thanks a lot for the explanation. Let's merge this for now and we'll keep working on this, are the errors about trait calls? Do you know where they happen?

Thx!

src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs
alfonsogarciacaro: Hmm, it may be good to get the error if the source doesn't exist, I can't think right now in which situations missing dependency's sources (either fable-core or another Nuget package) are not an issue 🤔

src/dotnet/Fable.Compiler/Fable.Compiler.fsproj
alfonsogarciacaro: I had to [add back the slash at the end](https://github.com/fable-compiler/Fable/commit/2c26046df98e7cb06c5132a4b52f5127ad15028b#diff-1715577b9f0a4893c7db757910fc247aR43) for the `Include` attribute, without it I was getting the following error when creating the package:

```
  Fable.Compiler -> /Users/alfonsogarciacaronunez/dev/fable/src/dotnet/Fable.Compiler/bin/Release/netcoreapp2.0/dotnet-fable.dll
/usr/local/share/dotnet/sdk/2.1.200/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(198,5): error NU5019: File not found: '/Users/alfonsogarciacaronunez/dev/fable/build/fable-core'. [/Users/alfonsogarciacaronunez/dev/fable/src/dotnet/Fable.Compiler/Fable.Compiler.fsproj]
```

src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs
ncave: If we want a "fable-core is missing" error, perhaps an explicit check and friendly error message would be better than just throwing.

src/dotnet/Fable.Compiler/Fable.Compiler.fsproj
ncave: I guess there is some disconnect between VS2017 and MSBuild here, perhaps you need to reopen your issue above, if it's still an issue.

PR #1412, opened 2018-05-31
benchmark update

ncave:
benchmark update to FCS 23.0.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "devDependencies": {
     "@types/babel-core": "^6.25.3",
-    "@types/node": "^10.1.2",
+    "@types/node": "^10.1.4",
     "@babel/core": "^7.0.0-beta.49",
     "@babel/preset-env": "^7.0.0-beta.49",
     "babel-core": "^6.26.3",
@@ -19,15 +19,15 @@
     "fable-splitter": "file:./src/js/fable-splitter",
     "fable-utils": "file:./src/js/fable-utils",
     "mocha": "^5.2.0",
-    "rollup": "^0.59.3",
+    "rollup": "^0.59.4",
     "rollup-plugin-babel": "^3.0.4",
     "rollup-plugin-commonjs": "^9.1.3",
     "rollup-plugin-fable": "file:./src/js/rollup-plugin-fable",
     "rollup-pluginutils": "^2.3.0",
     "rollup-watch": "^4.3.1",
     "tslint": "^5.10.0",
-    "typescript": "^2.8.3",
+    "typescript": "^2.8.4",
     "uglify-es": "^3.3.9",
-    "webpack": "^4.0.1"
+    "webpack": "^4.10.2"
   }
 }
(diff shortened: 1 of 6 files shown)


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "2.1.200"
+    "version": "2.1.201"
```
alfonsogarciacaro: Netcore 2.1 has just been released (sdk 2.1.300, MSFT versioning rules). Should we use that?

package.json
```diff
@@ -19,15 +19,15 @@
     "fable-splitter": "file:./src/js/fable-splitter",
     "fable-utils": "file:./src/js/fable-utils",
     "mocha": "^5.2.0",
-    "rollup": "^0.59.3",
+    "rollup": "^0.59.4",
```
alfonsogarciacaro: Are we still using Rollup? I haven't updated the plugin for Fable 2. I need to try if we can use Webpack 4 and uglify now when compiling FCS + Fable. Fable 2 uses less weird ES2015 features so I hope we've more luck now.

global.json
ncave: I know, but this just aligns with FCS 23.0.1.
Yes, we should use it, but the change to dotnet core 2.1 is bigger and it should be in it's own PR, as the API surface may have changed.

package.json
ncave: I don't know, I still like rollup for it's clean output, but I don't follow who's the winner of the top bundler competition, so you probably know better.

global.json
ncave: I see you already made that PR to switch to 2.1, so I've updated FCS-Fable too.

package.json
alfonsogarciacaro: The winner is clearly Webpack, there've been many improvements in v4 and it enjoys big momentum now. Rollup still produces a much cleaner output but it's probably not worthy to maintain both plugins at this point. When we want to check the generated JS we can still use fable-splitter. Anyways, don't worry. When we start updating the REPL to Fable 2 I'll remove the dependency :+1:

src/dotnet/Fable.JS/bench/splitter.config.js
```diff
@@ -1,12 +1,14 @@
-const babelOptions = {
+const fableUtils = require("fable-utils");
+
+const babelOptions = fableUtils.resolveBabelOptions({
```
alfonsogarciacaro: It's not necessary any more to resolve the paths because Fable copies fable-core and sources from Nuget packages into the current directory (in the hidden `.fable` folder) to avoid problems :)

alfonsogarciacaro:
I see you already rebased after upgrading to netcoreapp2.1 (it was easier than I expected), thanks!

src/dotnet/Fable.JS/bench/splitter.config.js
ncave: but I'll take it out from here, as it works either way.

ncave: a minute too late, already merged :) but you can take it out if not needed.

PR #1416, opened 2018-06-05
Add compiler version to defined directives

Zaid-Ajaj:
This will make it possible to have libraries be backwards-compatible, especially ones that rely on specific features or compilation results from Fable 1.x and need custom implementations for Fable 2+

### DIFF
--- a/src/dotnet/dotnet-fable/Parser.fs
+++ b/src/dotnet/dotnet-fable/Parser.fs
@@ -93,7 +93,7 @@ let parse (msg: string) =
     { path = path
       define =
         parseStringArray [||] "define" json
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
No need to add another directive to Fable2X for next versions because I think I could simply detect it like this: 
```fs
type internal InternalUnionForTestingCompilerVersion = One | Two of int 
let isUsingFableTwo() = Fable.Import.JS.Array.isArray(Two 20) 
```

alfonsogarciacaro:
Good idea, thanks!

Zaid-Ajaj:
Can you please publish a version with this directive included? I would like to try out some compatibility hacking :wink:

PR #1417, opened 2018-06-05
Fix range check in getCharAtIndex

marcpiechura:


alfonsogarciacaro:
Nice catch! Thank you!

PR #1418, opened 2018-06-05
Replacement fixes

ncave:
- fixed incorrect replacement logic (to match what it was before).
- fixed language intrinsic replacements and tests.
- splitter now not stopping on babel errors, so we can see all errors (we might need to track build status and output at the end if build has failed).


alfonsogarciacaro:
This is great, thank you!

PR #1419, opened 2018-06-07
Add failing test for Array.concat with strings

Zaid-Ajaj:
the following test fails at [simple-json-repro](https://github.com/Zaid-Ajaj/simple-json-repro) with error `TypeError: a.set is not a function`:
```fs
testCase "Array.concat works" <| fun test -> 
    [| [| "One" |]; [| "Two" |] |]
    |> Array.concat 
    |> List.ofArray
    |> test.areEqual [ "One"; "Two" ]
```

### DIFF
--- a/tests/Main/ArrayTests.fs
+++ b/tests/Main/ArrayTests.fs
@@ -277,6 +277,12 @@ let tests =
         ys.[0] + ys.[1]
         |> equal 3.
 
+    testCase "Array.concat works with strings" <| fun test -> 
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Ups, it seems the implementation [only works with typed arrays](https://github.com/fable-compiler/Fable/blob/13517dc948a673896cc60f4478f832c37ea20591/src/js/fable-core/Array.fs#L181). I'll change that, thanks!

Issue #1415, opened 2018-06-04
dev2.0 branch - Babel error: "You must pass a scope and parentPath unless traversing a Program/File."


ncave:
See #700, same thing basically (dev2.0 branch only).

ncave:
Just for reference, a few more errors in the build log: [build log](https://gist.github.com/ncave/941fe05e5d0d1de122cc28e735efc5de)

alfonsogarciacaro:
Thanks @ncave! This error usually happens when we declare two `let` or `const` variables with same name within a block scope in Babel AST. It's unfortunate that Babel doesn't give you the location in the error message, but I guess we can add an extra check to locate the problem.

Hmm, looking at the log it seems this is going to hit hard: `error FABLE: Current version cannot implement ToString for records and unions`. This happens because in Fable 2 methods are static functions except for overrides, which are attached to the prototype. But records and unions are compiled as plain JS objects or arrays so they don't have their own prototype... I need to find a way to solve this.

ncave:
@alfonsogarciacaro Here is an updated build log (with #1418 applied): [build_log](https://gist.github.com/ncave/941fe05e5d0d1de122cc28e735efc5de)

ncave:
@alfonsogarciacaro Thanks for adding the extra logging, it's so much easier to locate the problem now.
Here is a small repro:
```
module App

let test arg =
    match arg with
    | 0 | 1 ->
        let test x = x
        test
    | _ -> id

[<EntryPoint>]
let main argv =
    (test 1) 0
```

causing this Fable error: TypeError: unknown: Duplicate declaration "test$1"
The problem seems to be related to ho many cases are in that match, i.e.:
- match arg with  | 0 | 1 -> fails with above error
- match | 1 -> works

alfonsogarciacaro:
You mean the range in identifiers? That was supposed to be one of the improvements in Fable 2, but I just realized I wasn't copying the location info from the Fable to the Babel AST 😅 Happy to see it has improved the error info for this case!

It seems things that were solved in Fable 1 are hitting me in Fable 2 because I forgot to move the patches in the refactoring. In this case it was about wrapping the `case` statements in a switch with braces to give its own block scope to each. I also took the chance to remove target duplication when possible. Can you please try with latest commit?

ncave:
@alfonsogarciacaro No, I was talking about the commit after tis issue was opened that changed the error to `TypeError: unknown: Duplicate declaration`, which helped a lot.

[...]

PR #1420, opened 2018-06-09
Added BigInt

ncave:
- added some constants
- added BigInt and tests
- fixed ResizeArray tests

Note:  I've spent a lot of time trying to make the FSharp BigInt files compile into proper shape for ingesting, but I guess the dev2.0 branch is just not that polished yet as the master branch, as far as generated code quality goes. So for now we're better off sticking with the already generated JS version, slightly altered to fit. It's fine the way it is, since it doesn't change often. Perhaps in the future we can try again, when the Fable 2.0 code generation matures, but there are other more important things to do first, I'm sure.


tests_external/bar.js
```diff
@@ -1 +1 @@
-exports.bar = 5
\ No newline at end of file
+export let bar = 5;
```
alfonsogarciacaro: Did you get any linting fail because of this? I think originally the idea was to check import attribute/expressions would work also with commonjs modules directly.

tests/Main/js/1foo.js
```diff
@@ -1,42 +1,46 @@
...
+export const fooOptional = {
+  Foo1() {
+    return arguments.length;
+  },
+  Foo2() {
+    return arguments.length;
+  },
+  Foo3() {
+    return arguments.length;
+  }
```
alfonsogarciacaro: Please see the comment below about commonjs modules. If possible, I'd prefer to keep them to check importing from commonjs modules also works. There are already other JS files in the test project with ES2015 modules.

src/js/fable-core/ListClass.js
```diff
@@ -15,13 +15,13 @@ function compareList(self, other) {
...
 
-export class List {
+export default class List {
   constructor(head, tail) {
     this.head = head;
     this.tail = tail;
   }
 
-  toString() {
+  ToString() {
```
alfonsogarciacaro: Maybe we need to change it but right now Fable 2 is [compiling `ToString` declaration with the first letter in lower case](https://github.com/fable-compiler/Fable/blob/c3a30aef66eebca2495706cbf19429dfc6021c16/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs#L573-L575). The reasoning is this way JS debuggers will call `toString` if the prototype of an object declares it (though I haven't really checked that yet). In any case, thanks for fixing the custom `toString` implementation check in `Util/toString`, I forgot to do that 😅

alfonsogarciacaro:
This is great, @ncave, thanks a lot! I just left some minor comments, could you please check?

I'm sorry that you had to spend a lot of time on this. Using the JS files should be fine for now, although code generation by Fable 2 should be quite good at this point. And in fact, compiling some fable-core modules from F# (like Map and Set) has helped to improve how the functions were being exposed. But it's true some knowledge of Fable 2 internals is needed. Specially about [the name mangling of functions in internal modules](https://github.com/fable-compiler/Fable/blob/c3a30aef66eebca2495706cbf19429dfc6021c16/src/dotnet/Fable.Compiler/Transforms/Replacements.fs#L1406), to call them correctly in `Replacements`.

ncave:
@alfonsogarciacaro Yes, and also mangling of namespaces, classes, structs etc. But it's fine, the longer names would probably go away during minification.

tests_external/bar.js
ncave: No, the reason I changed it was to be able to generate and bundle all tests directly to/From ES6 modules. I'll change it back, but it's forcing all the tests to be compiled to commonjs just for this test.

src/js/fable-core/ListClass.js
ncave: Ok, if JS Debuggers are the reason, I'll change it to camelCase.

ncave: I just added a camelCase toString, this way you can do both if you decide to change.

ncave:
@alfonsogarciacaro The fixes are in, let me know if you really dislike having both toString and ToString in the ListClass / Long / BigInteger prototypes, and I'll take it out.

src/js/fable-core/Long.js
```diff
@@ -48,17 +48,22 @@ export default function Long(low, high, unsigned) {
...
+Long.prototype.add = function (x) { return op_Addition(this, x); }
+Long.prototype.sub = function (x) { return op_Subtraction(this, x); }
+Long.prototype.mul = function (x) { return op_Multiply(this, x); }
+Long.prototype.div = function (x) { return op_Division(this, x); }
+Long.prototype.mod = function (x) { return op_Modulus(this, x); }
+Long.prototype.shl = function (x) { return op_LeftShift(this, x); }
+Long.prototype.shr = function (x) { return op_RightShift(this, x); }
+Long.prototype.and = function (x) { return op_BitwiseAnd(this, x); }
+Long.prototype.or = function (x) { return op_BitwiseOr(this, x); }
+Long.prototype.xor = function (x) { return op_ExclusiveOr(this, x); }
```
alfonsogarciacaro: Sorry I didn't notice this in my previous review. I'd prefer to keep most of the methods in `Long` as module functions instead of attaching them to the prototype. I did this intentionally  for Fable 2, also following the new structure for F# classes (most of the members are now module functions instead of instance members, which btw is what happens in F# AST). The reason is to make tree shaking more effective. Any module function that's not used can be eliminated from the final bundle, but if everything is attached to the prototype the whole `Long.js` file will be included as soon as you make any reference to it.

alfonsogarciacaro:
Hi @ncave. Yes, the idea is to compile most of the things as module functions (avoiding nested modules) and use mangling to prevent conflicts. It makes the code a bit less readable sometimes but it makes it work better with tree shaking and, as you say, the long names will be removed during minification (unlike prototype members, the names of module functions can easily be minified in the bundle).

I left another comment that I miss in my previous review. I guess you're attaching the functions to the prototype to make it work with the JS BigInt files. That's ok for now, I'll merge and I'll try to make BigInt work directly from F# so this is not needed.

ncave:
@alfonsogarciacaro That's fine, I can fix the BigInt to not use Long prototypes. Your time is probably better spent with other, more important things.

ncave:
@alfonsogarciacaro I would really appreciate your help with #1415, as it is a showstopper for the repl.

alfonsogarciacaro:
No worries :) I'm trying [to do it in this branch](https://github.com/fable-compiler/Fable/tree/dev2.0-bigint), compiling BigInt modules from F# works but tests are failing. I'll investigate why later.

ncave:
@alfonsogarciacaro See PR #1424.

PR #1421, opened 2018-06-09
Weird failing type injection test: type names affect runtime metadata

Zaid-Ajaj:
Ok so I stumbled upon this very weird bug in alpha-017, the following code works fine: 

```fs
type Maybe<'t> = 
    | Just of 't
    | Nothing 

type GenericTestRecord<'t> = { Other: 't; Value : Maybe<string> }

// helper type with function getNameOf to extract the name of the injected type
type Types() = 
    static member getNameOf<'t> ([<Inject>] ?resolver: ITypeResolver<'t>) : string = 
        let resolvedType = resolver.Value.ResolveType()
        resolvedType.Name

let name = Types.getNameOf<Maybe<list<GenericTestRecord<string>>>>()
```
So far so good, now rename `GenericTestRecord` to `RecWithGenDU` and it fails with error `$type$$22 is not defined`

> typeof(T).Name seems to work here as well, but this was not my problem, I only simplified my code down to this snippet

It is only this combination of types: `Maybe<list<RecWithGenDU<string>>>` just fails but if you make slight (seemingly irrelevant) changes, it will just work again 
 - removing the field `Value : Maybe<int>`  
 - if you make the resolved type `list<RecWithGenDU<string>>` 
 - Other variations I tried but can't remember :joy: 

Let's see if CI can confirm this

### DIFF
--- a/tests/Main/ReflectionTests.fs
+++ b/tests/Main/ReflectionTests.fs
@@ -383,6 +383,21 @@ type Helper =
     static member Make(values: obj[], [<Inject>] ?res: ITypeResolver<'U>) =
         let t = res.Value.ResolveType()
         FSharpValue.MakeRecord(t, values) :?> 'U
+        
+type Maybe<'t> = 
+    | Just of 't
+    | Nothing 
+
+// The !!!NAME!!! of this type fails with somethings, I don't know
+type RecWithGenDU<'t> = { Other: 't; Value : Maybe<string> }
+// But this one works, very weird
+type GenericTestRecord<'t> = { Other: 't; Value : Maybe<string> }
+
+// helper class to extract the name of an injected type
+type Types() = 
+    static member getNameOf<'t> ([<Inject>] ?resolver: ITypeResolver<'t>) : string = 
+        let resolvedType = resolver.Value.ResolveType()
+        resolvedType.Name
 
 let injectTests = [
     testCase "ITypeResolver can be injected" <| fun () ->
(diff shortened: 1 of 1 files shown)


Nhowka:
Weird! CI confirmed that...

If `RecWithGenDU` comes second the behaviour is the same? I'm curious about what is causing that, because right now feels like sorcery! 🧙

alfonsogarciacaro:
Thanks for catching this @Zaid-Ajaj! This is only my fault 😅 As we talked, Fable 2 is generating now reflection info in place. But I had some trouble with types with cross references, so in the code generated now types are defined now and fields/cases are lazily evaluated. Example:

```js
testCase("Name can be extracted from RecWithGenDU", function () {
  const name = Types$$$getNameOf({
    ResolveType() {
      return function () {
        const $type$$5 = (0, _Reflection.union)("QuickTest.Maybe`1", () => [(0, _Reflection.list)($type$$6)], () => [["Just", [(0, _Reflection.list)($type$$6)]], ["Nothing", []]]);
        const $type$$6 = (0, _Reflection.record)("QuickTest.RecWithGenDU`1", () => [_Reflection.string], () => [["Other", _Reflection.string], ["Value", $type$$5]]);
        return $type$$5;
      }();
    }

  });
  equal(false, name === "");
});
```

But for some reason I thought this was not necessary for generics 🤦‍♂️ In the example above, if we don't wrap the generics with the lambda `$type$$6` won't be defined when passed to the TypeInfo. Not sure why it worked for `GenericTestRecord` but I guess the names were ordered alphabetically.

Zaid-Ajaj:
@alfonsogarciacaro I am glad you could identify the issue this quick, very interesting indeed! Thats why it worked when the `Maybe<string>` field was removed or when the type was just `list<RecWithGenDU<string>>`: the cross references disappeared

PR #1427, opened 2018-06-13
ToString for records and unions

ncave:


alfonsogarciacaro:
Awesome @ncave, thanks a lot for this! I think it's a great solution for now, then we can decide if we should implement records/unions agains with JS classes/function constructors :+1:

Zaid-Ajaj:
> then we can decide if we should implement records/unions agains with JS classes/function constructors

Do you expect to start using js classes for the 2.0 release? I am already building stuff that assume object literals and arrays for records and unions 😅

alfonsogarciacaro:
@Zaid-Ajaj Yeah, right now I'm thinking that we'll have to use JS classes/function constructors at the end, because using plain objects/arrays creates problems in many cases: overriden methods, custom comparison, type testing, etc. I'm finally realizing that we shouldn't rely too much on Fable internals because this can break in future versions, I will try from now to reduce the magic tricks and only use reflection if necessary like in _standard_ F#.

However the good news is I'm also planning to add a `toJSON` method by default to records and unions ([as for lists](https://github.com/fable-compiler/Fable/blob/7d65193dabd47449b5806b519c9f92f6b3526a84/src/js/fable-core/ListClass.js#L28-L30)). This method [will be called automatically by `JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) so when you're serializing records and unions will be converted to plain object/arrays.

I hope this is compatible with your current work. Do you have any example where you rely on a specific JS representation for records/unions?

Nhowka:
In the case of overridden _virtual_ methods. Couldn't we have a slot for each method so that when each subclass that overrides the method is instantiated, it replaces the slot with the current implementation on construction? Would be like a poor dynamic dispatch, so when called from the base class or from the child, it will always call the child implementation.

Zaid-Ajaj:
> I will try from now to reduce the magic tricks and only use reflection if necessary like in standard F#.

:heart: 

> I hope this is compatible with your current work. Do you have any example where you rely on a specific JS representation for records/unions?

I was building a JSON converter that doesn't rely on reflection (at least, kind of optional) so when deserializing the data, I am constructing my records and unions manually, but I guess I won't escape reflection and I would have to use `FSharpValue.MakeRecord` and `FSharpValue.MakeUnion`. 

here is a sample from my code that constructs records from data and type information: 
```fs
// Records
| JObject dict, TypeInfo.Record fields -> 
    // Match the JSON object literal keys with their types
    let valuesWithTypes =   
        let values = Map.toList dict
        values
        |> List.map (fun (key, value) -> 
            fields 
            |> Array.tryFind (fun (fieldName, _) -> fieldName = key)
            |> function 
                | Some (_, fieldType) -> key, value, fieldType
                | None -> failwithf "Could not find key '%s' in the type information" key)
    // create an empty object literal 
    let record = obj() 
    // add each key and value to the object literal
    for (key, value, fieldType) in valuesWithTypes do 
        setProp record key (unbox (fromJsonAs value fieldType))
    // lookup optional fields in the type info and if they don't exist 
    // in the JSON object, then add them manually as None 
    for (fieldName, fieldType) in fields do 
        match fieldType with 
        | TypeInfo.Option _ -> 
            match Map.tryFind fieldName dict with 
            | None -> setProp record fieldName (unbox None)
            | Some _ -> () 
        | _ -> ()

    unbox record
```

alfonsogarciacaro:
@Nhowka Hmm, that's an interesting idea. Though it's probably very similar to what we are already doing. Fable doesn't really change the object when upcasting it, so if you call a virtual method that's been implemented by attaching the member to the prototype, the implementation in the derived class will be called indeed. In JS it's more performant to attach methods to the prototype than to a field in the object itself. The point here is either to attach methods to the object/prototype or keep them as static module functions (which work better with tree shaking/minification, and can also be better optimized by the JS engine).

@Zaid-Ajaj Thanks a lot for sharing the code. Very good, reminds me of type shape :+1: I guess you need to get the reflection info at one point, so the only thing that you need is to keep a reference to the Type somewhere and then, as you said, call `FSharpValue.MakeRecord` [like Thoth will do](https://github.com/alfonsogarciacaro/Thot/blob/03d73567b5b2c51cf792466daf1d9b1ca7fdca2e/src/Thoth.Json/Decode.fs#L559).

BTW, I already started implementing base classes for unions and records in the `dev2.0-types` branch.

Zaid-Ajaj:
@alfonsogarciacaro The full working project is now published at [SimpleJson](https://github.com/Zaid-Ajaj/Fable.SimpleJson), would be nice to have if you wanted to test some Fable project that uses `ofJson<'t>`  and it will act as a test suite for reflection support as for now it doesn't rely on internal Fable implementations. 

> reminds me of type shape

Indeed, I don't use the types directly, but rather I first convert them to a [type info](https://github.com/Zaid-Ajaj/Fable.SimpleJson/blob/master/src/TypeInfo.fs) representation that is easier to work with and pattern-match against (it should be it's own library tbh).

alfonsogarciacaro:
I've just pushed dotnet-fable 2.0.0-alpha-020 with the new types. Can you please try with that?

ncave:
@alfonsogarciacaro Preliminary bundle sizes for FCS+Fable:

- dev2.0-barebones:
```
ES2015 modules:         13,589,077 bytes
rollup (not minified):  11,284,788 bytes
webpack (minified):      3,664,782 bytes
```
- dev2.0 (i.e. dev2.0-types):
```
ES2015 modules:         13,606,582 bytes
rollup (not minified):  11,282,221 bytes
webpack (minified):      3,692,028 bytes
```
So I guess the bundle size is more or less the same.

alfonsogarciacaro:
Thanks a lot for the numbers @ncave, I also compiled another project (much smaller) and got similar results. I guess what we add on one side (function declarations for every record/union) we compensate on the other (not repeating field names for every record constructor, etc). It'd be also nice to see if there's any big difference in performance, though it may be difficult to keep the two branches up-to-date with all the bug fixes.

In any case, it's great minification is working with Fable 2. IIRC it didn't work in Fable 1 for the REPL. Maybe we can try later if Prepack also works now.

Zaid-Ajaj:
@alfonsogarciacaro Just updated and ran my tests, everythings looks good! it only broke maps deserialization that was relying on internal structure: maps need a dedicated `toJSON` function, otherwise they produce these strings:
```fs 
[ ]
|> Map.ofList
|> Json.stringify // alias to native JSON.stringify
|> log // => {"comparer":{},"tree":"MapEmpty"}

[ "one", "two"; ]
|> Map.ofList
|> Json.stringify
|> log // => {"comparer":{},"tree":["MapOne","one","two"]}

[ 1, "one"; ]
|> Map.ofList
|> Json.stringify
|> log // => {"comparer":{},"tree":["MapOne",1,"one"]}

[ "one", "two"; "three", "four" ]
|> Map.ofList
|> Json.stringify 
|> log // => {"comparer":{},"tree":["MapNode","three","four",["MapOne","one","two"],"MapEmpty",2]}
```
I was able to workaround it anyway with a bit of reverse engineering,  but I would rather avoid doing things this way
```fs
    type InternalMap = 
        | MapEmpty 
        | MapOne of string * Json 
        | MapNode of string * Json * InternalMap * InternalMap

    let (|MapEmpty|_|) json = 
        match json with  
        | JString "MapEmpty" -> Some json
        | _ -> None 

    let (|MapKey|_|) = function
        | JNumber number -> Some (string number)
        | JString key -> Some key 
        | _ -> None

    let (|MapOne|_|) = function 
        | JArray [ JString "MapOne"; MapKey key; value ] -> Some (key, value)
        | _ -> None 
        
    let (|MapNode|_|) = function 
        | JArray [ JString "MapNode"; MapKey key; value; left; right; JNumber _  ] -> 
            Some (key, value, left, right)
        | _ -> None 

    let rec generateMap json = 
        match json with
        | MapEmpty _ -> Some InternalMap.MapEmpty
        | MapOne (key, value) -> Some (InternalMap.MapOne (key, value))
        | MapNode (key, value, left, right) -> 
            match generateMap left, generateMap right with 
            | Some leftMap, Some rightMap -> 
                Some (InternalMap.MapNode(key, value, leftMap, rightMap))
            | _ -> None
        | _ -> None 

    let rec flattenMap = function 
        | MapEmpty -> [ ] 
        | MapOne (key, value) -> [ key, value ] 
        | MapNode (key, value, left, right) -> 
            [ yield! flattenMap left 
              yield! flattenMap right 
              yield  (key, value) ] 

// then the conversion happens here 
| JObject map, TypeInfo.Map (keyType, valueType) -> 
    // check whether the map is serialized to it's internal representation
    // and convert that to back to a normal map from the data
    match Map.tryFind "comparer" map, Map.tryFind "tree" map with 
    | Some (JObject comparer), Some (JArray tree) when Map.isEmpty comparer ->
        match generateMap (JArray tree) with 
        | Some internalMap -> 
            flattenMap internalMap
            |> List.map (fun (key, value) -> 
                let nextKey = unbox (fromJsonAs (JString key) keyType)
                let nextValue = unbox (fromJsonAs value valueType)
                unbox<string> nextKey, nextValue)
            |> Map.ofList 
            |> unbox   
        | None -> 
            let inputJson = SimpleJson.toString (JArray tree)
            failwithf "Could not generate map from JSON\n %s" inputJson       
```

alfonsogarciacaro:
I've added `toJSON` to most of JS types in fable-core. However, `Map` and `Set` are now implemented in F# and it's not so easy. I was thinking (and probably will do) to include a `Fable.Core.IJsonSerializable` interface that Fable can recognize and attach the members to the prototype (currently this happens for `System.IComparable`). This will also be useful for people writing their custom types.

However, I wouldn't recommend it for this case. It will fail for example with `Dictionary` because Fable implements it as JS `Map` (depending on the key type) and we cannot touch its prototype (at least we shouldn't). My recommendation is to test against `System.Collections.Enumerable` (note this is the non-generic version) and then convert to array [as it's done here](https://github.com/MangelMaxime/Thoth/pull/30/files#diff-758afc1951a5313d55cfe2c63282de2fR155). This will automatically expand to any collection so it's probably a more standard solution.

Zaid-Ajaj:
@alfonsogarciacaro Thanks for the detailed explanation, I will try it out and see where it goes :smile:

PR #1430, opened 2018-06-15
Fixes #1428

### LINKED ISSUE #1428: (dev2.0) Cannot type test any exception
```
module App

let length xs =
    try xs |> Array.length
    with :? System.NullReferenceException -> 0

[<EntryPoint>]
let main argv =
    length [| 1; 2; 3 |]
```
error FABLE: Cannot type test: System.NullReferenceException

ncave:
- fixed exceptions replacements.
- fixed HashSet replacements.

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
@@ -827,6 +827,8 @@ module Util =
                 jsInstanceof (makeCoreRef Fable.Any "default" "Long") expr
             | Some Types.bigint ->
                 jsInstanceof (makeCoreRef Fable.Any "default" "BigInt") expr
+            | Some name when name.EndsWith "Exception" ->
+                jsInstanceof (makeIdentExpr "Error") expr
             | _ ->
                 if ent.IsFSharpExceptionDeclaration then
                     let expr = com.TransformAsExpr(ctx, expr)
--- a/src/dotnet/Fable.Compiler/Transforms/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Replacements.fs
@@ -1744,13 +1744,13 @@ let hashSets (com: ICompiler) (_: Context) r t (i: CallInfo) (thisArg: Expr opti
         Helper.InstanceCall(c, methName, t, args, i.SignatureArgTypes, ?loc=r) |> Some
     | "Add", Some c, [arg] ->
         Helper.CoreCall("Util", "addToSet", t, [arg; c], ?loc=r) |> Some
-    | ("isProperSubsetOf" | "isProperSupersetOf"
-    |  "unionWith" | "intersectWith" | "exceptWith"
-    |  "isSubsetOf" | "isSupersetOf" | "copyTo" as methName), Some c, args ->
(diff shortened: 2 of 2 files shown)


alfonsogarciacaro:
The `AutoReset` event is failing on .NET side sometimes, not an issue so let's merge this :+1:

HashSet reminded me that I just [merged the `hash` and `getHashCode` functions in fable-core/Util.ts](https://github.com/fable-compiler/Fable/commit/ad08f8d97d5c39d63ffe9e6fd179670b17ce13a9#diff-2603074b9d8223997863fcce63c88c61R185) because I wasn't sure why there were two different ones. Do you see any problem with this?

ncave:
@alfonsogarciacaro Yes, they are different, it would be better if you can reverse that and keep them separate.

alfonsogarciacaro:
Thanks @ncave! I'm trying to remove the string argument and be sure when the `.GetHashCode` (instance member) should be called, as well as make the semantics clearer. I found the differences to be:

- `hash`, `Unchecked.hash` first check if `GetHashCode` is implemented and then default to structural hash.
- `.GetHashCode` called directly defaults to identity hash (for reference types except string) if not implemented.
- `LanguagePrimitive.PhysicalHash` creates an identity hash no matter whether `GetHashCode` is implemented or not.

Does this look right? Would you mind to quickly review [this commit](https://github.com/fable-compiler/Fable/commit/de485ba7a6243b6bd240e9f24aedb335f409baea) to check I'm not making any obvious mistake?

ncave:
@alfonsogarciacaro Sounds about right, but I don't have the full picture of what you're trying to do with the `dev2.0-types` branch.

I would just rename `Utils.getHashCode` to `Utils.hash`, and perhaps revert the change to `Replacements.hash` to just call `Utils.hash`, which already checks if `GetHashCode` is there.

Also I'm not sure if the name `Replacements.hasStructuralComparison` is an accurate description, as some of those might have `CustomComparison`, and also may not be a good idea to drive `ToString` with the it, it may need more specific approach.

alfonsogarciacaro:
Thanks @ncave, I'll apply the suggestions :+1:

The `dev2.0-types` is about having a base type (actually a JS function constructors) for unions, records and F# exceptions (structs are assimilated into records) implementing the basic methods (ToString, Equals, CompareTo) so we can be sure to have the proper behavior at runtime (I tried to use Fable 2.0.0-alpha-019 to compile another project with multiple records and unions implementing `ToString` and found many situations where the static member couldn't be found at compile time).  About `hash` I just stumbled on it and thought it was a good chance to try to formalize it a bit (that's probably not what you should do in a feature branch but...).

You're right about `Replacements.hasStructuralComparison`, the name is confusing. I will rename it to something like `hasBaseImplemeningBasicMethods`. This means it's safe to call the instance member for these types, because either the base method or the custom implementation in the derived type will be activated.

BTW,  I implemented `GetHashCode` for [Union and Record bases](https://github.com/fable-compiler/Fable/commit/27f277d85cc9cbac5d1f21c5df93c4c2f36816f9#diff-d82dcbae1b4ccb3f458452ef2ccc5703R117). Could you please also have a quick look and see if they make sense? Thanks!

ncave:
@alfonsogarciacaro Looks good.

PR #1431, opened 2018-06-17
Dev2.0 types

alfonsogarciacaro:
One of the main drivers of Fable 2 was to use lightweight types for unions and records to reduce bundle size, however this has been proven to be complicated because there are many situations where we want to have more control of the behavior of unions and records at compile time (converting to string, serializing to JSON, comparison, etc). Because of this, this PR creates again actual types for records and unions but hopefully we can still reduce the bundle size thanks to:

- Now there are base types for records and unions that contains the common methods (ToString, Equals, CompareTo...) so we don't have to repeat them in every declaration (types can still include custom implementations).
- Type declarations don't contain reflection info (that takes a lot of space for unions with many cases). Reflection info is only generated ad-hoc where necessary.

@ncave, I've saved the branch `dev2.0` before this merge in the `dev2.0-barebones` branch so we can make the benchmark comparison when we manage to compile FCS with Fable 2. Hopefully there's not a big difference ;)

Note that I'm using JS function constructors instead of ES2015 classes for the base types. There're two main reasons for this:

- Function constructors are enough in this case and, as most of the projects still compile to ES5 (_old JS_) using functions instead of classes should save space in the bundle as we don't nee to comply with the full semantics of classes.
- For F# classes, I actually need to use function constructors to work around the fact that JS classes can only derive from the base class basic constructor. This is not necessary for unions and records (as they cannot inherit) but using functions here as well makes the approach more consistent.

In any case, in the future we can consider moving to classes if they become more generalized and JS engines are more capable of optimizing them.

> I don't know why this is failing in Travis, but tests are passing in Appveyor and locally so there should be no problem to merge.


inosik:
> I don't know why this is failing in Travis, but tests are passing in Appveyor and locally so there should be no problem to merge.

That's because the .NET SDK download fails. For some reason all of the known URLs return 404s. Even the official Docker images fail.

alfonsogarciacaro:
Ah, ok. Thanks a lot for the info @inosik! :+1:

inosik:
FYI, see dotnet/dotnet-docker#599 and all the linked issues.

PR #1432, opened 2018-06-18
Replacement updates

ncave:
- fixes duplicate constructors
- fixes object expression compiler (by converting it to type), will open an issue for it eventually


alfonsogarciacaro:
Great, thanks! :clap: :clap:

PR #1440, opened 2018-06-22
Replacement fixes

ncave:


src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
```diff
@@ -804,9 +804,8 @@ let private transformDeclarations (com: FableCompiler) fsDecls =
                         EntityName = entityName
                         IsPublic = isPublicEntity ent }
                     [Fable.UnionConstructor info |> Fable.ConstructorDeclaration]
-                | None when ent.IsFSharpRecord
-                        || ent.IsFSharpExceptionDeclaration
-                        || ((ent.IsClass || ent.IsValueType) && not ent.IsMeasure && not(hasImplicitConstructor ent)) ->
+                | None when (ent.IsFSharpRecord || ent.IsValueType || ent.IsFSharpExceptionDeclaration)
+                        || (ent.IsClass && not ent.IsMeasure && not (hasImplicitConstructor ent)) ->
```
alfonsogarciacaro: Is this necessary? Value types can also have implicit constructors [as in here](https://github.com/fable-compiler/Fable/blob/91cf4ea0b5cf7e0ae7a2dc4bad5e751db18cee85/tests/Main/TypeTests.fs#L181) and if we remove this check, there will be duplicated code (the auto-generated constructor and the user-defined constructors, besides the `inherits` instructions).

src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
```diff
@@ -1302,7 +1302,7 @@ module Util =
                     |> declareModuleMember info.IsPublic info.EntityName false
             if info.Entity.IsFSharpExceptionDeclaration
             then yield coreValue com ctx "Types" "FSharpException" |> inherits com ctx consIdent
-            elif info.Entity.IsFSharpRecord
+            elif info.Entity.IsFSharpRecord || info.Entity.IsValueType
```
alfonsogarciacaro: I removed this because I noticed value types don't have structural equality by default in .NET F#, do you need it?

alfonsogarciacaro: Hmm, I'm checking the equality checks in the tests below and they eval to false in my computer's FSI. But the tests pass in Appveyor. Is it because I have an old FSI?

Microsoft (R) F# Interactive version 4.1

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
ncave: I think we need it, cause value type records and value types compile to the same AST, the only difference seems to be the constructor is marked as implicit, and it's body is a bit different, but that's beside the point. They both still need to inherit from Types.Record.
If there is any differentiation needed it can happen on the generation side (in FableToBabel.fs), but I think its fine the way it is.

alfonsogarciacaro: Ok, thanks for the clarification. I think I'll revert this change and add the inheritance in Fable2Babel when transforming the implicit constructor for value types, as it was a couple of commits ago.

src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
ncave: See above. I thought all things have structural equality/comparison by default in F# (especially structs, it's in the name), unless they have declared CustomEquality/CustomComparison. From [here](https://blogs.msdn.microsoft.com/dsyme/2009/11/08/equality-and-comparison-constraints-in-f/):
> Tuple types support "structural" equality. The same applies to lists, options, arrays and user-defined record, union and struct types whose constituent field types permit structural equality, hashing and comparison.

PR #1446, opened 2018-06-26
Fixed List.distinctBy

ncave:


src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -651,7 +651,6 @@ let makeEqualityComparer (com: ICompiler) typArg =
     // TODO: Use proper IEqualityComparer<'T> type instead of Any
     ObjectExpr
         ([ObjectMember(makeStrConst "Equals", f, ObjectValue)
-          ObjectMember(makeStrConst "Compare", makeComparerFunction com typArg, ObjectValue)
```
alfonsogarciacaro: I added this to speed up [comparerFromEqualityComparer](https://github.com/fable-compiler/Fable/blob/7a1dbeaa3613d28a0713a706a37ac63785453801/src/js/fable-core/Util.ts#L49) but you're right it's probably better to remove it as I tend to regret this kind of shortcuts more often than not.

alfonsogarciacaro:
Thanks! I'll push a new release with this and the other fixes. Hopefully we're closer now. Really excited to see those benchmark results! 🤞

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
ncave: Yes, it actually breaks stuff, that's why I took it out.
If you really need to add it back, you can replace it with CompareTo, but it's clearer this way.

PR #1450, opened 2018-06-28
Updated distinctBy

ncave:
Note: Second tests in Seq.distinctBy are temporarily disabled, can't figure out what's wrong. It's like the iterator doesn't rewind after the first test.


alfonsogarciacaro:
I'll have a look, thanks!

src/js/fable-core/Set.fs
```diff
@@ -603,80 +603,81 @@ let iterate f (s : Set<'T>)  = s.Iterate(f)
...
+let maxElement (s: Set<'T>) = s.MaximumElement
 
-// let create (l : seq<'T>) = Set<_>.Create(l)
+// let create (l: seq<'T>) = Set<_>.Create(l)
 
 type IMutableSet<'T> =
     inherit IEnumerable<'T>
     abstract size: int
-    abstract add: 'T -> IMutableSet<'T>
+    abstract add: 'T -> bool
```
alfonsogarciacaro: @ncave Do you need this? We'll probably have to revert this change because the `IMutableSet` interface is intended to match [JS Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add) (I should have left a comment about this, JS Set is used to emulate .NET HashSet, however when the key has custom comparison we call `createMutable` to emulate the JS Set interface with an F# Set). Returning bool is actually [done by `Util/addToSet`](https://github.com/fable-compiler/Fable/blob/43768e73f3cd9957d312e3788d3fcece98f6df04/src/dotnet/Fable.Compiler/Transforms/Replacements.fs#L1739-L1740).

src/js/fable-core/Set.fs
```diff
@@ -695,11 +696,19 @@ let createMutable (source: 'T seq) (comparer: IComparer<'T>) =
...
 
-let distinct (xs: 'T seq) (comparer: IComparer<'T>): 'T seq =
+let distinct (xs: seq<'T>) ([<Inject>] comparer: IComparer<'T>) =
     createMutable xs comparer :> _ seq
 
-let distinctBy projection (xs: 'T seq) (comparer: IComparer<'T>): 'T seq =
-    createMutable (xs |> Seq.map projection) comparer :> _ seq
+let distinctBy (projection: 'T -> 'Key) (xs: seq<'T>) ([<Inject>] comparer: IComparer<'Key>) =
+    let hashSet = createMutable Seq.empty comparer
+    xs |> Seq.filter (projection >> hashSet.add)
```
alfonsogarciacaro: @ncave I found the issue: the `hashSet` reference was kept every time `Seq.filter` run away. So in the next iterations it discarded most of the times (the set already contained them). Pushing a fix right away.

PR #1451, opened 2018-06-28
Fixed Seq.groupBy test

ncave:


alfonsogarciacaro:
Thanks! Yes, that test was failing in latest Appveyor builds too. It's a pity we cannot run it on CI, but hopefully we should notice if there's a problem when running tests locally :)

PR #1456, opened 2018-06-30
REPL benchmark update

ncave:
Please note the benchmark uses the [fable2](https://github.com/ncave/FSharp.Compiler.Service/tree/fable2) FCS branch.

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs
@@ -540,10 +540,18 @@ let optimizeExpr (com: ICompiler) e =
     List.fold (fun e f -> f com e) e optimizations
 
 let rec optimizeDeclaration (com: ICompiler) = function
+
+#if FABLE_COMPILER
+    // optimization for actions and values does not result in a valid expr
+    // TODO: enable optimization when it's fixed
+    | ActionDeclaration expr -> ActionDeclaration expr
+    | ValueDeclaration(value, info) -> ValueDeclaration(value, info)
+#else
     | ActionDeclaration expr ->
         ActionDeclaration(optimizeExpr com expr)
     | ValueDeclaration(value, info) ->
         ValueDeclaration(optimizeExpr com value, info)
+#endif
     | ConstructorDeclaration kind as consDecl ->
         match kind with
         | ClassImplicitConstructor info ->
--- a/src/dotnet/Fable.JS/bench/app.fs
+++ b/src/dotnet/Fable.JS/bench/app.fs
@@ -13,13 +13,16 @@ let metadataPath =
 let readFileSync: System.Func<string, byte[]> = failwith "JS only"
 [<Fable.Core.Import("readFileSync", "fs")>]
 let readTextSync: System.Func<string, string, string> = failwith "JS only"
+[<Fable.Core.Import("writeFileSync", "fs")>]
+let writeTextSync: System.Action<string, string> = failwith "JS only"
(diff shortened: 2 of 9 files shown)


src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs
```diff
@@ -540,10 +540,18 @@ let optimizeExpr (com: ICompiler) e =
     List.fold (fun e f -> f com e) e optimizations
 
 let rec optimizeDeclaration (com: ICompiler) = function
+
+#if FABLE_COMPILER
+    // optimization for actions and values does not result in a valid expr
+    // TODO: enable optimization when it's fixed
+    | ActionDeclaration expr -> ActionDeclaration expr
+    | ValueDeclaration(value, info) -> ValueDeclaration(value, info)
+#else
```
alfonsogarciacaro: It'd be good to find why this is failing as ValueDeclaration includes most of the functions and optimizations do a lot of inlining. At a minimum we should try to apply the uncurrying optimizations because they're necessary to interact with fable-core files (in Fable 1 the uncurrying was done in Replacements module, but Fable 2 does it here).

src/dotnet/Fable.JS/bench/app.fs
```diff
@@ -58,12 +64,17 @@ let main argv =
...
+            let ms1, fsAst = measureTime parseFSharp ()
+            let errors = fable.GetParseErrors fsAst
+            errors |> Array.iter (printfn "Error: %A")
+            if errors.Length > 0 then failwith "Too many errors."
+            let ms2, babelAst = measureTime parseFable fsAst
+            if i = 1 then
+                // fable.PrintFSharpAst (fsAst, false, printfn "%s")
+                // printfn "Babel AST: %s" (toJson babelAst)
+                // writeAllText fsAstFile (toJson fsAst)
+                writeAllText babelAstFile (toJson babelAst)
```
alfonsogarciacaro: Is `toJson` working? JSON de/serialization has been moved out of Fable.Core for Fable 2, and `toJson` is not defaulting to `JSON.stringify` but I don't think this will work with types in Babel AST.

MangelMaxime: I am not sure it's a good idea to let `toJson` in Fable 2.

Because, people could think it's still working as before when in fact it isn't.

src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs
ncave: @alfonsogarciacaro For whatever reason, applying any optimization for action and value resuts in invalid expr.

alfonsogarciacaro: Is it happening in any case? At what point do you see it's an invalid expression? Do you have any example?

Anyway, I'll try to debug later :)

src/dotnet/Fable.JS/bench/app.fs
alfonsogarciacaro: @MangelMaxime I had to leave `toJson` to make Fable 2 alpha work because some libs still use (for example, Elmish to print values to console). It's marked as obsolete, but hopefully the libs will be updated and we can remove it for the stable release.

@ncave I've mode [some changes in Babel AST types to prevent mangling of field names](https://github.com/fable-compiler/Fable/commit/f672dc24aa365724d4c8a65156e8228857e23275) and other (use arrays instead of list, immediately convert unions into strings...). In theory, it should be possible to send those types to Babel now without any JSON serialization at all.

PR #1467, opened 2018-07-08
Remove some allocations

ncave:


src/dotnet/Fable.Compiler/Global/Prelude.fs
```diff
@@ -53,10 +53,10 @@ module List =
         List.ofArray xs
 
     let mapToArray f (xs: 'a list) =
-        List.toArray xs |> Array.map f
+        xs |> List.map f |> List.toArray
 
     let mapiToArray f (xs: 'a list) =
-        List.toArray xs |> Array.mapi f
+        xs |> List.mapi f |> List.toArray
```
alfonsogarciacaro: Thanks for this! I had the feeling that `Array.map` would be more performant but probably I was wrong. I was also considering the following, would it avoid allocating the intermediate list?

```fsharp
    let mapiToArray f (xs: 'a list) =
        let ar: 'a[] = List.length xs |> Array.zeroCreate 
        xs |> List.iteri (fun i x -> ar.[i] <- f x)
        ar
```

ncave: @alfonsogarciacaro Yes, I was thinking of that too, definitely better, anything helps.

Issue #1457, opened 2018-06-30
(v2.0) REPL benchmark stats


ncave:
Some preliminary baseline stats:
```
------------------------------------------------
Fable 2.0.0 alpha, FCS 24.0.1
------------------------------------------------

Medium test file (10 kB):

------------------------------------------------
REPL benchmark - DotNet Core 2.1.300
------------------------------------------------
InteractiveChecker created in 1315 ms
iteration 1, FCS time: 2418 ms, Fable time: 598 ms
iteration 2, FCS time: 78 ms, Fable time: 42 ms
iteration 3, FCS time: 67 ms, Fable time: 41 ms
iteration 4, FCS time: 74 ms, Fable time: 41 ms
iteration 5, FCS time: 67 ms, Fable time: 44 ms
iteration 6, FCS time: 66 ms, Fable time: 62 ms
iteration 7, FCS time: 71 ms, Fable time: 104 ms
iteration 8, FCS time: 74 ms, Fable time: 52 ms
iteration 9, FCS time: 67 ms, Fable time: 43 ms
iteration 10, FCS time: 88 ms, Fable time: 41 ms

------------------------------------------------
REPL benchmark - Fable 1.3.7, Node.js 10.5.0
------------------------------------------------
InteractiveChecker created in 5232 ms
iteration 1, FCS time: 1866 ms, Fable time: 736 ms
iteration 2, FCS time: 1077 ms, Fable time: 402 ms
iteration 3, FCS time: 969 ms, Fable time: 348 ms
iteration 4, FCS time: 846 ms, Fable time: 328 ms
iteration 5, FCS time: 834 ms, Fable time: 322 ms
iteration 6, FCS time: 764 ms, Fable time: 304 ms
iteration 7, FCS time: 745 ms, Fable time: 293 ms
iteration 8, FCS time: 744 ms, Fable time: 286 ms
iteration 9, FCS time: 729 ms, Fable time: 289 ms
iteration 10, FCS time: 728 ms, Fable time: 287 ms

------------------------------------------------
REPL benchmark - Fable 2.0.0 alpha, Node.js 10.5.0
------------------------------------------------
InteractiveChecker created in 2840 ms
iteration 1, FCS time: 1018 ms, Fable time: 235 ms
iteration 2, FCS time: 681 ms, Fable time: 131 ms
iteration 3, FCS time: 559 ms, Fable time: 69 ms
iteration 4, FCS time: 524 ms, Fable time: 145 ms
[...]

PR #1473, opened 2018-07-10
[WIP] Compile chars as numbers

alfonsogarciacaro:
This PR tried to compile chars as numbers in the hope this would give a performance boost when parsing but apparently there's not much gain. [Some discussion here](https://github.com/fable-compiler/Fable/issues/1457#issuecomment-402765942).

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
@@ -198,16 +198,19 @@ module Util =
             com.TransformAsExpr(ctx, entRef)
 
     let makeTypedArray (com: IBabelCompiler) ctx typ (arrayKind: Fable.NewArrayKind) =
-        match typ with
-        | Fable.Number kind when com.Options.typedArrays ->
-            let cons = getTypedArrayName com kind |> Identifier
+        let makeJsTypedArray jsName =
             let args =
                 match arrayKind with
                 | Fable.ArrayValues args ->
                     [| List.mapToArray (fun e -> com.TransformAsExpr(ctx, e)) args
                        |> ArrayExpression :> Expression |]
                 | Fable.ArrayAlloc(TransformExpr com ctx size) -> [|size|]
-            NewExpression(cons, args) :> Expression
+            NewExpression(Identifier jsName, args) :> Expression
+        match typ with
+        | Fable.Char when com.Options.typedArrays ->
+            makeJsTypedArray "Uint16Array"
+        | Fable.Number kind when com.Options.typedArrays ->
+            getTypedArrayName com kind |> makeJsTypedArray
         | _ ->
             match arrayKind with
             | Fable.ArrayValues args ->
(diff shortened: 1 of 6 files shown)


ncave:
@alfonsogarciacaro
For completeness, there was a pretty big gain in FCS after we switched chars to uint16, and not much more after from this PR.
So one argument for this PR would be that FCS-like char-buffer processing applications might be able to perform better out of the box without alteration.
I don't know if that's a common case or not.

alfonsogarciacaro:
I think some people like @chadunit were writing a parser with Fable. Could be beneficial in that case?

ncave:
@alfonsogarciacaro Generating better performing code is beneficial in itself. If you have time, please list any counter-arguments to this PR from your point of view.

alfonsogarciacaro:
So basically the counter-arguments would be:

1. In JS, when indexing or enumerating a string, we get each character as a string as well not a number. This means if we decide to represent chars as numbers we need to make the conversion every time users try to index or enumerate the string. The PR is actually doing this, but it may open the opportunity to edge cases not covered by the tests that fail at runtime.

2. My last PRs are causing breaking changes so I'm afraid of accepting this one ;) Maybe we can wait until the REPL works fine, and then try again with this code to see if it still works fine.

ncave:
@alfonsogarciacaro Just FYI, I already tried this PR with the REPL bench and it was fine.

chadunit:
For what it's worth, I had already changed my parser to convert string input to int sequence up front for unrelated reasons.

But here is Scott Wlaschin's sample parser combinator library (which is char-based) and some naive timing code. I replaced src/tools/QuickTest.fs with this: https://gist.github.com/chadunit/e4e30c483b1575b2f06b92a84ac8b509

I didn't see much difference between dev2.0 and dev2.0-chars. Here is an example run in which I replaced the test JSON in the file with a 100KB sample from https://www.json-generator.com.

dev2.0:
```
[1] Parsed 100 objects in 1139 ms
[2] Parsed 100 objects in 985 ms
[3] Parsed 100 objects in 971 ms
[4] Parsed 100 objects in 990 ms
[5] Parsed 100 objects in 982 ms
[6] Parsed 100 objects in 966 ms
[7] Parsed 100 objects in 971 ms
[8] Parsed 100 objects in 1001 ms
[9] Parsed 100 objects in 1002 ms
[10] Parsed 100 objects in 989 ms
```

dev2.0-chars:
```
[1] Parsed 100 objects in 1165 ms
[2] Parsed 100 objects in 945 ms
[3] Parsed 100 objects in 998 ms
[4] Parsed 100 objects in 971 ms
[5] Parsed 100 objects in 985 ms
[6] Parsed 100 objects in 985 ms
[7] Parsed 100 objects in 1000 ms
[8] Parsed 100 objects in 1011 ms
[9] Parsed 100 objects in 994 ms
[10] Parsed 100 objects in 983 ms
```

alfonsogarciacaro:
Thanks a lot for benchmarking @chadunit! So in principle the performance boost is coming mainly from using typed arrays which can be done just by using `uint16[]` (if the user knows about this).

This PR still uses regular expressions for `IsDigit` and `IsWhitespace`. I wonder if this has a big influence in the benchmark. I'll try to convert them to use codes instead and run the Scott Wlaschin parser again. Thanks again for providing the gist @chadunit!

alfonsogarciacaro:
REPL is compiling fine with this but when trying to use it to compile a script containing char literals it fails. First it was giving the error `Unexpected type Char for literal 123`, but after [this commit](https://github.com/fable-compiler/Fable/pull/1473/commits/583e9a819b08325a3a63332abfef32b2bda5416a) now the UI freezes (even with the web worker) so I guess it just fell into an endless loop.

alfonsogarciacaro:
Ups, no, my bad. Actually this is working, I might have made a mistake when downloading the Appveyor artifact (again, I shouldn't program after midnight). Stand back, I'm merging this!

Issue #1472, opened 2018-07-10
(dev2.0) Object constructor and member overloads


ncave:
Currently it's a bit hard to implement BCL objects with overloading constructors/members in Fable.Core, since there is no way to know during replacement what the mangled name will be (as that happens after).
(F# modules/functions are fine as there are no overloads or interfaces to worry about.)

Would it be easier if the overloads get disambiguated (numbered with a predictable suffix) *before* they get to replacement, and then the suffix can be stripped during replacement matching, but is still there for generating the right call to the correct overload?

Just thinking aloud, please ignore if it does not make sense.

alfonsogarciacaro:
You're right. I've never liked the way to suffix overloads that requires traversing all the members every time and is easily breakable (as with the problem we had with different members in signature files). I was also considering ways to create a predictable suffix for overloads, and I'm definitely open to do it if we find a good solution.

Funscript just attached the types of the arguments to the method name. It's an option but I didn't like it very much because the member names would become much longer. It's not a big problem because we are already mangling member names in Fable 2, and this shouldn't affect minified code. But still can make the results of the REPL somewhat less nice to see.

I'm not an expert, but do you know a way to create a unique and predictable (and not too long) hash out of the type arguments that we could attach to the member name?

alfonsogarciacaro:
What [about this](https://github.com/fable-compiler/Fable/commit/3c6afbcec0af0202985d7ebe7518c30b2455fef1) (branch `dev2.0-overload-suffix`)? It creates an overload suffix combining the string hashes of the argument type names. Some tests still need to be fixed (those that call F# Map or Set), but do you think it could work?

ncave:
Sure, creating a name from the parameters could work too. My suggestion above was to add a ReplaceCallInfo.MagledName for the overloads (or have a way to compute it), simple enumeration would do (e.g. Member$1, Member$2, etc.), the point is to have it available during replacement so the full mangled name can be composed, but your solution is pretty much the same thing.

alfonsogarciacaro:
Nah, using the hash for the overloads in fable-core F# types doesn't work, because the replaced types (`Microsoft.FSharp.Collections.FSharpSet` for example) have a different names than the types in fable-core (`Set.FSharpSet`).

alfonsogarciacaro:
[...]

PR #1475, opened 2018-07-11
Use hashes as overload suffix (but indices for Replacements)

alfonsogarciacaro:
Discussion here: https://github.com/fable-compiler/Fable/issues/1472

I've made it so types in fable-core still use enumeration. So, as long as the overloads keep the same order as the original types in the BCL, it should work. I tried it with the StringBuilder constructors (check System.Text.fs) and it seems to work.

@ncave If possible, please have a look and tell me what you think. As commented in the issue, I like the hash approach (if chances of collision are low) because it's predictable, but we still need to keep the enumeration approach for Replacements. So this may be introducing too much complication in the code base.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -112,10 +112,13 @@ let buildCoreJsTypescriptFiles () =
 
 let buildCoreJsFsharpFiles () =
     nugetRestore coreJsSrcDir
-    sprintf "%s/dotnet-fable.dll node-run %s --fable-core %s -- -c splitter.config.js"
+    sprintf "%s/dotnet-fable.dll %s %s %s"
         cliBuildDir
-        "../fable-splitter/dist/cli"
-        "force:${outDir}" // fable-splitter will adjust the path
+        "node-run ../fable-splitter/dist/cli"
+        "--fable-core force:${outDir}"  // fable-splitter will adjust the path
+        // Use overload indices instead of hashes
+        // See https://github.com/fable-compiler/Fable/issues/1472#issuecomment-403976332
+        "--experimental overload-index"
     |> run coreJsSrcDir dotnetExePath
 
 let buildCoreJs () =
--- a/src/dotnet/Fable.Compiler/AST/AST.Fable.fs
+++ b/src/dotnet/Fable.Compiler/AST/AST.Fable.fs
@@ -202,6 +202,7 @@ type ArgInfo =
 
 type ReplaceCallInfo =
   { CompiledName: string
+    OverloadSuffix: Lazy<string>
     /// See ArgIngo.SignatureArgTypes
     SignatureArgTypes: Type list
     Spread: SpreadKind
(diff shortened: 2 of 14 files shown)


alfonsogarciacaro:
Ok, let's merge this. The calculation of the overload suffix has been moved to its own module, so it shouldn't be difficult to go back to enumeration indices if needed :+1:

ncave:
@alfonsogarciacaro This seems to be broken, it generates too high an id for overloads:
```fs
let sb = new Text.StringBuilder (opNamePrefix, opNamePrefix.Length + (opLength * maxOperatorNameLength))
```
results in:
```js
const sb = (0, _System.StringBuilder$$$$002Ector$$4)("op_", "op_".length + opLength * maxOperatorNameLength);
```
but `_System.StringBuilder$$$$002Ector$$4` does not exist.

ncave:
@alfonsogarciacaro Also see #1479 is possibly relates to this PR.

alfonsogarciacaro:
Actually, this is intended though I agree it's very convoluted 😅 The index `4` is coming from the **actual** `System.Text.StringBuilder` class in the BCL because at this point Fable doesn't have access to fable-core F# types. The idea was that `StringBuilder` in fable-core would replicate the order and number of overloads of the actual types, but I acknowledge this is very error-prone. 

Anyways, I need to sleep on this to find an alternative. Probably at the end we'll have to add the fable-core F# types metadata to Fable compiler.

alfonsogarciacaro:
What about [this commit](https://github.com/fable-compiler/Fable/commit/56fb9332fcf0d668183cbfceb0a6f3e576a62209)? It uniforms the hashes for fable-core and user-defined types, but still uses a couple of attributes (`Replaces` to indicate which type is replacing when they don't have the same name, and `OverloadSuffix` to hardcode the suffix) for some specific cases (Set operators and methods). Seems to fix the issue you mention above.

ncave:
@alfonsogarciacaro I like it, less is more. But here is what gets generated:
```js
const sb = (0, _System.StringBuilder$$$$002Ector$$Z18115A39)("op_", "op_".length + opLength * maxOperatorNameLength);
```
and there is no such constructor:
```js
function StringBuilder$$$$002Ector$$Z176EF219(capacity, value) {
  return this != null ? StringBuilder.call(this, capacity, value) : new StringBuilder(capacity, value);
}
function StringBuilder$$$$002Ector$$Z524259A4(capacity$$1) {
  return StringBuilder$$$$002Ector$$Z176EF219.call(this, capacity$$1, null);
}
function StringBuilder$$$$002Ector$$Z721C83C5(value$$1) {
  return StringBuilder$$$$002Ector$$Z176EF219.call(this, 16, value$$1);
}
function StringBuilder$$$$002Ector() {
  return StringBuilder$$$$002Ector$$Z176EF219.call(this, 16, null);
}
```

alfonsogarciacaro:
That makes sense. I made a mistake and put `capacity` (int) before `value` (string). It should work by changing the order of the parameters :)

PR #1476, opened 2018-07-12
Added SystemObject type

ncave:
@alfonsogarciacaro You don't really have to merge this if you're not ok with it, but perhaps it can help reason about complicated corner cases for comparing and equality of deeply nested types. This is passing tests, but may need more work if corner cases require it.

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
@@ -1334,10 +1334,12 @@ module Util =
         [
             yield FunctionExpression([|for arg in args do yield arg :> Pattern|], BlockStatement setters)
                     |> declareModuleMember info.IsPublic info.EntityName false
-            if info.Entity.IsFSharpExceptionDeclaration
-            then yield coreValue com ctx "Types" "FSharpException" |> inherits com ctx consIdent
-            elif info.Entity.IsFSharpRecord || info.Entity.IsValueType
-            then yield coreValue com ctx "Types" "Record" |> inherits com ctx consIdent
+            if info.Entity.IsFSharpExceptionDeclaration then
+                yield coreValue com ctx "Types" "FSharpException" |> inherits com ctx consIdent
+            elif info.Entity.IsFSharpRecord || info.Entity.IsValueType then
+                yield coreValue com ctx "Types" "Record" |> inherits com ctx consIdent
+            elif info.Entity.IsClass then
+                yield coreValue com ctx "Types" "SystemObject" |> inherits com ctx consIdent
         ]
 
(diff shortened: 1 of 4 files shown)


src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
```diff
@@ -1369,6 +1371,8 @@ module Util =
             | Some(TransformExpr com ctx baseRef) -> yield inherits com ctx consIdent baseRef
             | None when info.Entity.IsValueType ->
                 yield coreValue com ctx "Types" "Record" |> inherits com ctx consIdent
+            | None when info.Entity.IsClass ->
+                yield coreValue com ctx "Types" "SystemObject" |> inherits com ctx consIdent
```
alfonsogarciacaro: I guess we need to check if the class is not already inheriting anything else.

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -601,10 +597,6 @@ let rec equals (com: ICompiler) r equal (left: Expr) (right: Expr) =
         Helper.CoreCall("Reflection", "equals", Boolean, [left; right], ?loc=r) |> is equal
     | Tuple _ ->
         Helper.CoreCall("Util", "equalArrays", Boolean, [left; right], ?loc=r) |> is equal
-    | DeclaredType(ent,_) when hasBaseImplementingBasicMethods ent ->
-        Helper.InstanceCall(left, "Equals", Boolean, [right]) |> is equal
```
alfonsogarciacaro: Maybe we can leave this for optimization? It'll be faster to call `.Equals` directly.

src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
alfonsogarciacaro: Sorry, scratch that. I thought this belonged to another file 😅

alfonsogarciacaro:
No problem, let's merge this and just assume types coming from JS may have different equality semantics :) As now all types will be deriving from something, I've added a `fable-core/Types.declare` helper so we can have better control of the inheritance.

I'm thinking that we could announce a Fable 2 beta so people can start playing with it, what do you think @ncave? Better to wait until the REPL works and we don't need to change many design decisions?

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
ncave: Right, same for classes.

ncave:
@alfonsogarciacaro Go for it, unless you want to do one last alpha.

PR #1477, opened 2018-07-12
Minor rename

ncave:


src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -599,7 +599,7 @@ let rec equals (com: ICompiler) r equal (left: Expr) (right: Expr) =
         Helper.CoreCall("Reflection", "equals", Boolean, [left; right], ?loc=r) |> is equal
     | Tuple _ ->
         Helper.CoreCall("Util", "equalArrays", Boolean, [left; right], ?loc=r) |> is equal
-    | DeclaredType(ent,_) when hasBaseImplementingBasicMethods ent ->
+    | DeclaredType(ent,_) when hasBaseImplementingBasicMethods ent || ent.IsClass ->
```
alfonsogarciacaro: I would avoid this because the class may be inheriting from a JS class (as with React components) and in this case it won't implement `.Equals`.

ncave: ok, removed.

PR #1487, opened 2018-07-17
Fixed list tests and REPL

ncave:
This PR adds some tests and fixes the REPL.

P.S. This one was a bear, it took a *looong* time to figure out why the REPL was skipping some bindings, turned out to be a problem with List.takeWhile/skipWhile. Also, we probably should do another pass at evaluating if any of the core modules generated implementations are inefficient and can be improved.

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Replacements.fs
@@ -1374,9 +1374,6 @@ let arrayModule (com: ICompiler) (_: Context) r (t: Type) (i: CallInfo) (_: Expr
     | "Item", [idx; ar] -> getExpr r t ar idx |> Some
     | "Get", [ar; idx] -> getExpr r t ar idx |> Some
     | "Set", [ar; idx; value] -> Set(ar, ExprSet idx, value, r) |> Some
-    | ("Take" | "Truncate"), [count; ar] ->
-        Helper.InstanceCall(ar, "slice", t, [makeIntConst 0; count], ?loc=r) |> Some
-    | "Skip", [count; ar] -> Helper.InstanceCall(ar, "slice", t, [count], ?loc=r) |> Some
     | "Copy", [ar] -> Helper.InstanceCall(ar, "slice", t, args, ?loc=r) |> Some
     | "ZeroCreate", [count] -> createArray count None |> Some
     | "Create", [count; value] -> createArray count (Some value) |> Some
--- a/src/dotnet/Fable.Compiler/Transforms/State.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/State.fs
@@ -96,11 +96,14 @@ type Compiler(currentFile, project: Project, options, ?fableCore: string) =
         member __.Options = options
         member __.FableCore = fableCore
         member __.CurrentFile = currentFile
-        member __.GetRootModule(fileName) =
+        member x.GetRootModule(fileName) =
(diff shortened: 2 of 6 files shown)


alfonsogarciacaro:
I'm sorry you had to spend so much time to fix the issue @ncave. At least, we're getting closer to have a fully working REPL with Fable 2. I have created a [different repository](https://github.com/fable-compiler/repl2) for it (so we can later easily compare both REPLs). It's _almost_ working, but I found a problem with FCS `BasicPatterns.Sequential` as several statements in the sequence duplicate one of the last expressions. For example the basic canvas sample:

```fsharp
let init() =
    let canvas = document.querySelector(".view") :?> HTMLCanvasElement

    let ctx = canvas.getContext_2d()
    // The (!^) operator checks and casts a value to an Erased Union type
    // See http://fable.io/docs/interacting.html#Erase-attribute
    ctx.fillStyle <- !^"rgb(200,0,0)"
    ctx.fillRect (10., 10., 55., 50.)
    ctx.fillStyle <- !^"rgba(0, 0, 200, 0.5)"
    ctx.fillRect (30., 30., 55., 50.)
```

Gets translated to this:

```js
export function init() {
  const canvas = document.querySelector(".view");
  const ctx = canvas.getContext("2d");
  ctx.fillStyle = "rgba(0, 0, 200, 0.5)";
  ctx.fillStyle = "rgba(0, 0, 200, 0.5)";
  ctx.fillStyle = "rgba(0, 0, 200, 0.5)";
  ctx.fillRect(30, 30, 55, 50);
}
```

According to the debugging, the problem is already present when getting `BasicPatterns.Sequential` from FCS, so it's not due to Fable transformations. I had some suspicions about [this code](https://github.com/ncave/FSharp.Compiler.Service/blob/2fc85f37c408493f4e527984bfb15a4e2f87c674/src/fsharp/symbols/Exprs.fs#L366-L372) but it gets translated as follows which seems to be fine:

```js
          switch ($target$$25) {
            case 0:
              {
                const e1R$$1 = FSharpExprConvert$$$ConvExpr(cenv$$13, env$$13, e1$$2);
                return FSharpExprConvert$$$ConvExprLinear(cenv$$13, env$$13, e2$$2, function ($arg$$5) {
                  return contf$$1(new E(35, "Sequential", e1R$$1, $arg$$5));
                });
                break;
              }

            case 1:
              {
                return new E(35, "Sequential", FSharpExprConvert$$$ConvExpr(cenv$$13, env$$13, x0), FSharpExprConvert$$$ConvExpr(cenv$$13, env$$13, x1$$1));
                break;
              }
```

Any ideas?

ncave:
@alfonsogarciacaro A closure issue, lust from the look of it?. Have you tried breaking that function composition? Is there a smaller repro, or is that small enough?

alfonsogarciacaro:
I tried [removing the Fable optimizations](https://github.com/fable-compiler/Fable/commit/34c7d6ed8a585c8cddf0cd2fab862b36b3f49a7d) (except uncurrying) and it works! Now I've to go one by one to pinpoint the problem :)

alfonsogarciacaro:
At the end binding beta reduction was the one to blame, [I made it more restrictive](https://github.com/fable-compiler/Fable/commit/cd0805fd5c0265f3fc4c4f089946d39f297088f2#diff-6fad46925f5049b7ad2d32380d29ba7bL255) and [REPL 2](http://fable.io/repl2/) is fully working now (even the failures in tooltips have been fixed) :tada: 💃

PR #1489, opened 2018-07-17
Restore bench project

ncave:
This PR restores the `bench` project's ability to run on dual F#/JS platform.

It's quite valuable to have the `bench` project build everything from scratch, as it is used to track performance regressions, and a two-step build is counter-productive. Its driver app code doesn't change often so it does not benefit (in fact, it's hindered) from linking to already built repl component. This two-step build is more suitable to the `testapp` project, which is probably what the CI should be running.

@alfonsogarciacaro I have left the platform sub-projects for now, but would like to remove them if possible, to keep it simple. If you can move whatever functionality you need to retain into the testapp project, that would help a lot.


alfonsogarciacaro:
Ok. Actually, I shouldn't have touched the `bench` project 😅 Anyways, I have now [something else](http://fable.io/repl2/) to play with :wink:

PR #1491, opened 2018-07-18
Bench cleanup, webpack update

ncave:


alfonsogarciacaro:
Sorry! I was making a lot of tries yesterday to make the REPL worked but didn't want to have too many commits in git history so I used `git commit --amend && git push -f` a lot (I know, I should have worked in a different branch, I'll do it next time). At least, [it's working now](http://fable.io/repl2/) 😸  so I'd prefer not to touch the REPL bundle process as the `repl2` repository depends on it (see `DownloadReplArtifact` build target).

Would it be possible to rebase/merge this PR with latest dev2.0 (I won't be changing the git history any more, I promise :wink:), remove the changes in the bundle and keep only changes for the bench app?

ncave:
@alfonsogarciacaro Ok done.

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.Util.fs
```diff
@@ -182,8 +182,7 @@ module Helpers =
 
     let unionCaseTag (ent: FSharpEntity) (unionCase: FSharpUnionCase) =
         try
-            let name = unionCase.Name
-            ent.UnionCases |> Seq.findIndex (fun uci -> name = uci.Name)
+            ent.UnionCases |> Seq.findIndex (fun uci -> unionCase.Name = uci.Name)
```
alfonsogarciacaro: Will this be OK? I'm always afraid that a property in FCS may evaluate more code than expected, that's why I was caching the name here so the `.Name` property is not evaluated multiple times.

alfonsogarciacaro:
Awesome, thank you!

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.Util.fs
ncave: Benchmark show no degradation, in fact it's a bit less allocation.

PR #1492, opened 2018-07-19
webpack for repl

ncave:
You don't have to merge this, it's just an example that it works (with uglify-es replaced by terser).
Anyway, the minified rollup size is a bit better than webpack (3.6 MB vs 3.8 MB), possibly because its tree shaking is better.


ncave:
Closing in favor of #1511

Issue #1486, opened 2018-07-16
DateTime <-> Tick roundtrip results in new values


Alxandr:
### Description

The following code should result in `a` and `b` being the same vaule:

```fsharp
let a = DateTime.MinValue.Ticks
let b = (DateTime a).Ticks

let eq = a = b
```

However, running it (with some added logging) produces the following:

```
isEq false
a: 36000000000 Long { low: 1640261632, high: 8, unsigned: false }
b: 72000000000 Long { low: -1014444032, high: 16, unsigned: false }
```

Worse still, this is time zone dependent.

### Related information

* Fable version (`dotnet fable --version`): 2.0.0-alpha.30
* Operating system: Windows 10

Zaid-Ajaj:
This can be closed too right?

PR #1499, opened 2018-07-22
Improve DateTime handling

### LINKED ISSUE #1498: (dev2.0) Bug in replacement of datetime/datetimeoffset
### Description

I'm trying to fix some issues in DateTime/DateTimeOffset wrt conversion to and from ticks (see #1486), and ran into something rather strange. I've added tests (that are currently failing) illustrating the problem with conversion to and from ticks, but ran into a weird issue that looks to me like a replacement bug. As far as I can see, the following two snippets of code should be identical:

```fsharp
// first attempt
equal d (DateTime d.Ticks)
equal d.Ticks (DateTime d.Ticks).Ticks
equal d (DateTime (d.Ticks, d.Kind))
equal d.Ticks (DateTime (d.Ticks, d.Kind)).Ticks

// second attempt
let ticks = d.Ticks
let kind = d.Kind
let fromTicks = DateTime ticks
let fromTicksWithKind = DateTime (ticks, kind)

equal d fromTicks
equal ticks fromTicks.Ticks
equal d fromTicksWithKind
equal ticks fromTicksWithKind.Ticks
equal kind fromTicksWithKind.Kind
```

yet the outcome is hugely different. In the first example, the following JS get's produced from the first test (rewritten to make it easier to read):

```js
equal(d, create(getTicks(d)));
```

Whereas, in the second one, I get the following (removing the temp vars):

```js
equal(d, fromTicks(getTicks(d)));
```

It's calling an entirely different method, and I don't understand why. As far as I can tell, the code should be identical (except for creating temp variables).

The replacement in question looks like this:

```fsharp
match i.CompiledName with
| ".ctor" ->
    match args with
    | [] -> Helper.CoreCall(moduleName, "minValue", t, [], [], ?loc=r) |> Some
    | ExprType(Builtin BclInt64)::_ ->
        Helper.CoreCall(moduleName, "fromTicks", t, args, i.SignatureArgTypes, ?loc=r) |> Some
        // let ms = Helper.CoreCall("Long","ticksToUnixEpochMilliseconds", t, [ticks], [i.SignatureArgTypes.Head])
        // Helper.CoreCall(moduleName, "default", t, (ms::rest), t::i.SignatureArgTypes.Tail, ?loc=r) |> Some
    | ExprType(DeclaredType(e,[]))::_ when e.FullName = Types.datetime ->
        Helper.CoreCall("DateOffset", "fromDate", t, args, i.SignatureArgTypes, ?loc=r) |> Some
    | _ ->
        let last = List.last args
        match args.Length, last.Type with
        | 7, EnumType(_, "System.DateTimeKind") ->
            let args = (List.take 6 args) @ [makeIntConst 0; last]
            let argTypes = (List.take 6 i.SignatureArgTypes) @ [Number Int32; last.Type]
            Helper.CoreCall("Date", "create", t, args, argTypes, ?loc=r) |> Some
        | _ ->
            Helper.CoreCall(moduleName, "create", t, args, i.SignatureArgTypes, ?loc=r) |> Some
```

### Related information

* Fable version (`dotnet fable --version`): dev2.0 branch

Alxandr:
First commit is part of #1497. I can't run tests locally without it. Also, there is a bug in replacement (not sure if I introduced it or not). See #1498 and [example](https://github.com/fable-compiler/Fable/blob/a4d582a4412d2835cd52ea692297322468df5a38/tests/Main/DateTimeTests.fs#L207).

Closes #1498, #1486

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/Replacements.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Replacements.fs
@@ -1966,13 +1966,12 @@ let dates (_: ICompiler) (_: Context) r t (i: CallInfo) (thisArg: Expr option) (
         Naming.removeGetSetPrefix i.CompiledName |> Naming.lowerFirst |> get r t thisArg.Value |> Some
     // DateTimeOffset
     | "get_DateTime" | "get_LocalDateTime" | "get_UtcDateTime" as m ->
-        let getTime = getTime thisArg.Value
         let kind =
             if m = "get_LocalDateTime" then System.DateTimeKind.Local
             elif m = "get_UtcDateTime" then System.DateTimeKind.Utc
             else System.DateTimeKind.Unspecified
             |> int |> makeIntConst
-        Helper.CoreCall("Date", "default", t, [getTime; kind], [getTime.Type; kind.Type], ?loc=r) |> Some
+        Helper.CoreCall("Date", "fromDateTimeOffset", t, [thisArg.Value; kind], [thisArg.Value.Type; kind.Type], ?loc=r) |> Some
     | "FromUnixTimeSeconds"
     | "FromUnixTimeMilliseconds" ->
         let value = Helper.CoreCall("Long", "toNumber", Number Float64, args, i.SignatureArgTypes)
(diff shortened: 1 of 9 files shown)


alfonsogarciacaro:
Thanks for this! I moved the Tick related methods out of Date.js because for Fable 1.x every tiny sample that used formatting imported Date and this imported the whole Long file. Let me check if the improvements in tree shaking have fixed this before merging :+1:

Alxandr:
It's no problem moving them back, it's just important

alfonsogarciacaro:
Anyways, let's merge it. I think the issue happened with previous versions of Webpack that didn't do tree shaking very well. We can revisit later if it's an issue. Thanks a lot for fixing this! :clap: :clap:

Alxandr:
It took me way longer than I was expecting. So many weird edge cases -.-. Never new off-by-timezone-errors was a thing 😛

PR #795, opened 2017-04-03
[WIP] Tail-rec transformExpr

cloudRoutine:
I've been having all kinds of issues trying to get the tests to run properly, and the build to work but it just doesn't seem to be going right.

There are tail calls in the function body IL, but it's really complicated so I'm not totally sure this is enough to  prevent the stackoverflow.  

I tried running the `ReproSO` target on the `repro-so` branch but it failed with

```log
> @ build C:\Users\jared\Github\Forks\Fable\src\dotnet\Fable.Client.Browser\demo
> node ../../../../node_modules/webpack/bin/webpack.js -p

Fable loader sent: C:\Users\jared\Github\Forks\Fable\src\dotnet\Fable.Client.Browser\Fable.Client.Browser.fsproj
Hash: 3bbaa81f3b37d487b4aa
Version: webpack 2.3.2
Time: 2183ms
        Asset       Size  Chunks             Chunk Names
bundle.min.js  844 bytes       0  [emitted]  main
   [0] ../Fable.Client.Browser.fsproj 324 bytes {0} [built] [failed] [1 error]

ERROR in ../Fable.Client.Browser.fsproj
Module build failed: Error: connect ECONNREFUSED 127.0.0.1:61225
Make sure Fable server is running on port 61225
    at client.send.then.catch.err (C:\Users\jared\Github\Forks\Fable\src\typescript\fable-loader\index.js:85:22)
    at process._tickCallback (internal/process/next_tick.js:109:7)

npm ERR! Windows_NT 10.0.15058
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\jared\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "run" "build"
npm ERR! node v7.8.0
npm ERR! npm  v3.10.9
npm ERR! code ELIFECYCLE
npm ERR! @ build: `node ../../../../node_modules/webpack/bin/webpack.js -p`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the @ build script 'node ../../../../node_modules/webpack/bin/webpack.js -p'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the  package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node ../../../../node_modules/webpack/bin/webpack.js -p
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs
[...]

PR #1502, opened 2018-07-23
More aggressive binding erasing

alfonsogarciacaro:
Found the problem that caused wrong compilation in the REPL: if removing bindings for accessing fields of compiler-generated bundles (happens a lot in pattern matching and lambdas with tuple arguments), we need to be sure we don't remove bindings for compiler-generated tuples. I couldn't localize an exact spot, but it seems this made some expressions disappear from the generated code. Should be working now :)

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs
@@ -248,17 +248,23 @@ module private Transforms =
         match e with
         // Don't try to optimize bindings with multiple ident-value pairs
         // as they can reference each other
-        | Let([ident, Function(args, funBody, currentName)], letBody)
-            when ident.IsCompilerGenerated && not ident.IsMutable
-                && (countReferences 1 ident.Name letBody <= 1) ->
-            if Option.isSome currentName then
-                sprintf "Unexpected named function when erasing binding (%s > %s)" currentName.Value ident.Name
-                |> addWarning com ident.Range
-            let replacement = Function(args, funBody, Some ident.Name)
-            replaceValues (Map [ident.Name, replacement]) letBody
-        | Let([ident, value], body) when ident.IsCompilerGenerated && not ident.IsMutable
-                                        && not(hasDoubleEvalRisk value) ->
-            replaceValues (Map [ident.Name, value]) body
+        | Let([ident, value], letBody) when not ident.IsMutable ->
+            match value with
+            // Match automatic destructuring of tuple arguments in inner functions
+            | Get(IdentExpr tupleIdent,_,_, _) as value
+                    when tupleIdent.IsCompilerGenerated ->
+                replaceValues (Map [ident.Name, value]) letBody
+            | Function(args, funBody, currentName)
+                    when ident.IsCompilerGenerated
+                    && (countReferences 1 ident.Name letBody <= 1) ->
+                if Option.isSome currentName then
+                    sprintf "Unexpected named function when erasing binding (%s > %s)" currentName.Value ident.Name
+                    |> addWarning com ident.Range
+                let replacement = Function(args, funBody, Some ident.Name)
+                replaceValues (Map [ident.Name, replacement]) letBody
+            | value when ident.IsCompilerGenerated && canInlineArg ident.Name value letBody ->
+                replaceValues (Map [ident.Name, value]) letBody
+            | _ -> e
         | e -> e
 
     /// Returns arity of lambda (or lambda option) types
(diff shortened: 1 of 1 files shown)


ncave:
@alfonsogarciacaro Awesome!

alfonsogarciacaro:
Thanks! This got stuck in the head, so I'm glad it's fixed :) This saves some bytes in the minified bundle (from 2,605,524 to 2,502,968) and I hope it helps a bit with the performance too. BTW @ncave, I guess you already tried using `--experimental optimize-fcs` [to build the REPL](https://github.com/fable-compiler/Fable/blob/37a695ec1f3ed3fac78bebbd762bddd1a2e0cbef/src/dotnet/Fable.JS/package.json#L4). If you did, is it close to work? Should we create a branch to the test it?

ncave:
@alfonsogarciacaro I don't know if it needs a branch, you can just run the build with the flag.
I tried adding some workarounds [here](https://github.com/ncave/FSharp.Compiler.Service/commit/e9f18ca2a08d8c5416e55bacf7b12697c86667ba) and [here](https://github.com/fable-compiler/Fable/pull/1503), it compiles but some things are still broken. Unfortunately I can't make progress as I keep hitting StackOverflow in FSharp2Fable, even when unoptimized.

ncave:
@alfonsogarciacaro I even tried compiling Fable with FCS latest source (24.0.1) instead of nuget (23.0.3), but that's crashing even earlier with StackOverflow when compiling the REPL (unoptimized or optimized, same thing).

alfonsogarciacaro:
Ah, so you're getting the SO errors too. I thought it happened only in macOS. I wonder why it runs just fine in AppVeyor 🤔

alfonsogarciacaro:
Well, actually you just [hit an SO exception in AppVeyor too](https://ci.appveyor.com/project/alfonsogarciacaro/fable/build/24.0.1-b3973) :wink: I did notice that when you check "Optimize (experimental)" in the online REPL the Fable part seems to take longer. Maybe there's something in the optimized AST that is making Fable fall in very deep loops?

ncave:
@alfonsogarciacaro It's not just the optimized AST, it also the unoptimized, when compiled with FCS 24.0.1.

alfonsogarciacaro:
Yes, the SO exception is a tough one. It started to happen in the transition to netcore and [we didn't manage to pinpoint it](https://github.com/Microsoft/visualfsharp/issues/2682). Maybe we could try using a .NET Framework-based Fable app, that seems to help, or at least it's easier to debug/profile it with VS (although the netcore debugger has improved a lot too).

Anyways, I think we've already done really good progress! Let's focus on releasing Fable 2 and leave optimizations for 2.1, MSFT-style ;)

ncave:
@alfonsogarciacaro I took out all of the optimized work-arounds from #1503. I'll try the 4.6 framework but if that doesn't work, we'll have to do it the ~~hard~~ right way, i.e. fixing the optimized AST in FCS.

alfonsogarciacaro:
@ncave BTW, if I'm not wrong the main advantages of FCS optimization are inlining and constant folding (other optimizations seem to be more .NET related). Maybe we can just [add constant folding to FableTransforms](https://github.com/fable-compiler/Fable/blob/a5eb10c5949841d3ddb8d7dea4d2816ab6fdddfd/src/dotnet/Fable.Compiler/Transforms/FableTransforms.fs#L521) and include an option to [also inline functions marked as `OptionalInline`](https://github.com/fable-compiler/Fable/blob/a5eb10c5949841d3ddb8d7dea4d2816ab6fdddfd/src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.Util.fs#L159-L160). What do you think?

ncave:
@alfonsogarciacaro Sure, more perf optimizations = better Fable.

ncave:
@alfonsogarciacaro Using the 4.6 framework did not help with the SO, it basically behaves the same.
So we'll have to fix the AST in FCS. Unfortunately, the changes introduced in FCS 24.0.1 (vs 23.0.3) add new sources of StackOverflow errors when compiling the REPL, so those will need to be fixed too. Bummer, it was so close!

alfonsogarciacaro:
Did you at least get some pointers where the SO is happening? Is it because `FSharp2Fable.transformExpr` is recursive? Maybe we could try revisiting PR #795, or I also have an idea to use a combination of stack and continuation to make it a true candidate for tail-call optimization.

ncave:
@alfonsogarciacaro Yes, the call stack is too deep, anything that helps with that would be nice. Unfortunately SO doesn't give you precise information on the reason, as it errors out in arbitrary location.

ncave:
@alfonsogarciacaro It does seem to start with `transformUnionCaseTest` but then it veers off.

alfonsogarciacaro:
I had problems with CPS because of the branching and also not all subexpressions are transformed with same context, but maybe trampolines work better 🤔

Issue #1500, opened 2018-07-22
Improve replacements for StringBuilder.


Alxandr:
I currently have a shim StringBuilder in my code so as to be able to use APIs that needs stringbuilders. This should probably be part of Fabe.Core (with replacements in place). In JS, it can be implemented as a simple `ref<string>` or similar.

ncave:
@Alxandr It's already part of Fable.Core, just use it.

Alxandr:
@ncave That's weird, cause I've been getting exceptions that no replacement was found for `StringBuilder.Append` (etc.)

ncave:
It was added recently, try it.

Alxandr:
You're right. I still get errors though 😛:

```
[!] Error: 'StringBuilder$$Append$$244C7CD6' is not exported by .fable\fable-core.2.0.0-alpha-031\System.Text.js
```

ncave:
@Alxandr The current SB implementation is a bit simplistic, probably one of the overloads is missing, as the overload mangling is now done based on argument types. Post a bit of F# code so we can see which one.

Alxandr:
I currently use the following implementation as a shim (be adviced that `Remove` and `ToString` with indexes has not been tested):

```fsharp
  type StringBuilder (s: string) =
    let mutable s = s

    new () = StringBuilder ("")

    member __.Length = s.Length

    member sb.Append (s': string) =
      s <- s + s'
      sb

    member inline sb.Append (c: char) = sb.Append (string c)

    member inline sb.Append (num: ^n) = sb.Append (sprintf "%d" num)

    member inline sb.Append (o: obj) = sb.Append (string o)

    member inline sb.AppendLine () = sb.Append System.Environment.NewLine

    member inline sb.AppendLine (s: string) =
      (sb.Append (s)).AppendLine ()

    member sb.Remove (startIndex: int, length: int) =
      if startIndex + length >= s.Length
      then s <- s.Substring (0, startIndex)
      else s <- s.Substring (0, startIndex) + s.Substring (startIndex + length)
      sb

    member inline __.ToString (startIndex: int, length: int) =
      s.Substring (startIndex, length)

    override __.ToString() = s
```

Not entirely sure which are missing, but that's at least the minimal implementation for my sake (wrt which methods are available).

ncave:
[...]

PR #1505, opened 2018-07-24
[WIP] Buffer based StringBuilder

### LINKED ISSUE #1504: TypedArray types/constructors are float based
```fsharp
    and [<AllowNullLiteral>] ArrayBuffer =
        abstract byteLength: float with get, set
        [<Emit("$0[Symbol.toStringTag]{{=$1}}")>] abstract ``[Symbol.toStringTag]``: obj with get, set
        abstract slice: ``begin``: float * ?``end``: float -> ArrayBuffer

    and [<AllowNullLiteral>] ArrayBufferConstructor =
        abstract prototype: ArrayBuffer with get, set
        [<Emit("new $0($1...)")>] abstract Create: byteLength: float -> ArrayBuffer
        abstract isView: arg: obj -> bool

    and [<AllowNullLiteral>] ArrayBufferView =
        abstract buffer: ArrayBuffer with get, set
        abstract byteLength: float with get, set
        abstract byteOffset: float with get, set
```

as well as all of the typed arrays, their `length` and constructors use floats instead of ints.

### Related information

* Fable version (`dotnet fable --version`): dev2.0 branch

Alxandr:
I'm not really sure whether or not this will be performant, so it's mostly an experiment at this stage. Also, a good way to find bugs in fable 😛.

Closes #1504 #1500

### DIFF
--- a/src/dotnet/Fable.Core/Fable.Import.JS.fs
+++ b/src/dotnet/Fable.Core/Fable.Import.JS.fs
@@ -486,480 +486,393 @@ module JS =
         abstract ``then``: ?onfulfilled: ('T->'TResult) * ?onrejected: (obj->'TResult) -> PromiseLike<'TResult>
 
     and [<AllowNullLiteral>] ArrayLike<'T> =
-        abstract length: float with get, set
+        abstract length: int with get, set
         [<Emit("$0[$1]{{=$2}}")>] abstract Item: n: int -> 'T with get, set
 
     and [<AllowNullLiteral>] ArrayBuffer =
-        abstract byteLength: float with get, set
+        abstract byteLength: int with get, set
         [<Emit("$0[Symbol.toStringTag]{{=$1}}")>] abstract ``[Symbol.toStringTag]``: obj with get, set
-        abstract slice: ``begin``: float * ?``end``: float -> ArrayBuffer
+        abstract slice: ``begin``: int * ?``end``: int -> ArrayBuffer
 
     and [<AllowNullLiteral>] ArrayBufferConstructor =
         abstract prototype: ArrayBuffer with get, set
-        [<Emit("new $0($1...)")>] abstract Create: byteLength: float -> ArrayBuffer
+        [<Emit("new $0($1...)")>] abstract Create: byteLength: int -> ArrayBuffer
         abstract isView: arg: obj -> bool
 
     and [<AllowNullLiteral>] ArrayBufferView =
         abstract buffer: ArrayBuffer with get, set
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Ah, sorry! It's really good you're updating the typed array bindings, but after merging PR #1473 you can actually just use `char[]` and `Uint16Array` will be used automatically, no need to manipulate JS typed arrays directly :)

Alxandr:
I know, but I'm using TypedArray specific apis for performance reasons (specifically, `#set` and `#subarray`).

More importantly though, #1506 is blocking me on this one and needs to be fixed. I tried rewriting to using primary constructors, and it did not help. Not only that, it forced me to write a lot of ugly code to achieve the same thing I achieve in way less code with the current implementation of the StringBuilder.

alfonsogarciacaro:
It's funny you mention because using `set` when setting typed array slices (`ar.[1..2] <- [|5;6|]` was one of the [first features of Fable](https://github.com/fable-compiler/Fable/blob/2cd894d09a796a0ca805fe9326b9d7c28a210242/src/js/fable-core/Array.ts#L82-L91) but probably nobody uses it ;)

If you're planning to use JS APIs, maybe it's better to write the replacement in JS or TypeScript. Mixing F# code with JS APIs may complicate maintenance (this has happened with the Array module for example).

Alxandr:
@alfonsogarciacaro I can definitely rewrite it into TypeScript (would be easy since the source is in C#, with the exception of the constructor overloading, which I have a few ideas how I can deal with). That being said, the `as X` and `then` feature is probably a nice thing to fix regardless, just for feature completeness 🙂

Alxandr:
Why was this closed?

alfonsogarciacaro:
I closed this? Ah, I think it's because I renamed `dev2.0` as `master` and deleted `dev2.0` to prevent confusion. Could you please resend it targeting `master`? Sorry for the hassle!

Alxandr:
Hmm. I'm not allowed to open it. Can you? I'll do a rebase later, but still blocked by #1506 though.

alfonsogarciacaro:
Ah, sorry again. Didn't notice I could do it myself. Reopen was also disabled for me but I can do it now, thanks!

MangelMaxime:
Hello I am trying to clean the old PR.

This one seems to have only one file change and in fact it has already been included in the master branch. Any reason to keep it opened ?

Alxandr:
No; you can just close it for my sake.

MangelMaxime:
Ok, thank you

PR #1507, opened 2018-07-31
Added trampoline

ncave:


alfonsogarciacaro:
Awesome, @ncave! Is this helping when compiling FCS 24 + Fable? I downloaded the Appveyor artifact and try the online repl with it. Everything seemed to work right, but funny enough now it cannot compile the pacman sample because of SO errors in the browser. Is this relying on the IL tail-call instruction? As far as I know, the trampoline technique should be good also for languages without tail-call optimization.

ncave:
@alfonsogarciacaro
- This PR is not going to help FCS 24.0.1 as it has its own SO issues when compiling the REPL, and this only fixes the Fable side.
- I'm not quite sure what CI artifact you are referring to.
- The REPL works ok with this trampoline PR, I tested it. It does have a small perf impact on the Fable side of the REPL, around 10% (as expected).
- I'm looking into adding trampoline to FableTransforms.fs so we can properly debug (that's where the SO moved to now in debug mode).

alfonsogarciacaro:
My mistake, it was actually downloading the bundle from this PR which caused the issue. I've fixed that by adding `&pr=false` to the query parameters when downloading the Appveyor artifact, so now you can use `dev2.0` or `dev2.0-trampoline` if you want to test the online REPL with either branch.

Good news is I can confirm with the trampoline I can build the REPL locally in macOS (it was causing an SO exception before), great job @ncave! :clap: :clap: It'd be great if we could manage to fix the issue with the pacman sample and then merge it.

ncave:
@alfonsogarciacaro Adding trampoline to FableTransforms.fs unfortunately results in a pretty large performance hit (half speed), so I'm not going to add it here yet.

ncave:
@alfonsogarciacaro I don't see any problems compiling the pacman sample with the REPL bench (nodejs).

alfonsogarciacaro:
No, sorry. What I mentioned [in this commit](https://github.com/fable-compiler/Fable/pull/1507#issuecomment-409337653) was a mistake, I was actually downloading the code from this PR because it was the latest successful build. I've fixed that now by [adding `&pr=false` to the query params](https://github.com/fable-compiler/repl2/blob/1964c0c6f126ca905aa3b4784be34ab84150c9a7/build.fsx#L2) when downloading the Appveyor artifact. Now you can clearly see the browser REPL cannot compile the pacman sample when using the `dev2.0-trampoline` so I'd prefer not to merge this until it can be fixed.

I also had an idea of mixing stacks and continuations that hopefully helps with performance too. But right now I'm preparing the other libs, templates, blog, etc. to announce the launch of Fable 2 beta, so it'll have to wait until that's finished.

ncave:
@alfonsogarciacaro I cannot reproduce, pacman works fine in repl2 (chrome), whether pulling from your `dev2.0-trampoline`, or manually bundling/copying repl bundle from this PR.

alfonsogarciacaro:
Interesting, maybe chrome sets the call stack size depending on the memory available. Also it seems to be more restrictive than other browsers. In Firefox and Safari it works fine (though they are slower). Can you please run the following code in your browser console?

```js
    function computeMaxCallStackSize() {
        try {
            return 1 + computeMaxCallStackSize();
        } catch (e) {
            // Call stack overflow
            return 1;
        }
    }
```

> Reference: http://2ality.com/2014/04/call-stack-size.html

In my machine I get:

- Chrome:   12 560
- Firefox:  109 244
- Safari:      36 242

ncave:
@alfonsogarciacaro
Chrome:
- first call: 25191
- second call: 33635
- third call and beyond: 41987

ncave:
@alfonsogarciacaro Ok done.

PR #1511, opened 2018-08-05
Added webpack option for REPL

ncave:
```
Minified bundle size:

ES5-rollup-uglifyjs: 3,748,680 bytes
ES6-webpack-terser:  3,751,254 bytes
```

No change, should be safe to merge.
Just added the option to bundle the REPL with webpack.


alfonsogarciacaro:
Thanks!

Issue #1521, opened 2018-08-18
Cannot resolve Microsoft.FSharp.Core.LanguagePrimitives.EnumToValue


zpodlovics:
### Description

It seems that right now there is no EnumToValue implementation exists in Fable.

### Repro code

```fsharp
type Fruits =
| Apple = 1
| Banana = 2
| Coconut = 4
let x = LanguagePrimitives.EnumToValue<Fruits,int32> Fruits.Apple
```

### Expected and actual results
Expected:
Works exactly in the same was as F# on .NET.

Actual:
```
error FABLE: Cannot resolve Microsoft.FSharp.Core.LanguagePrimitives.EnumToValue
```

### Related information

* Fable version (`dotnet fable --version`): 2.0.0-beta-001
* Operating system: Ubuntu 16.04 x86_64

alfonsogarciacaro:
Closed by #1522, thank you! I'll publish a new Fable 2 beta shortly :+1:

PR #1522, opened 2018-08-18
[WIP] Add EnumToValue implementation and test

zpodlovics:
This is a proposed fix for #1521: Right now both EnumOfValue and EnumToValue works only with NumberEnum. Is this OK? How should the StringEnum handled here?


alfonsogarciacaro:
Thanks @zpodlovics! Looks good to me, it's still WIP in the title but I think we can merge it. You don't need to worry about StringEnum, as this is Fable-specific and it's not compatible with `EnumToValue` anyways. You can get the value of a StringEnum just using the `string` operator :)

Issue #694, opened 2017-02-02
Adding byref parameter support


ncave:
Adding byref parameter support will enable some performance optimizations where needed.
Feature request, contributions welcome.

alfonsogarciacaro:
That's tricky because afaik you cannot pass byref parameters in JS. I guess the way to support it would be to wrap the function so it returns the the parameter values, but this wouldn't provide extra performance.

ncave:
I agree, probably not much of a performance gain, if any. Struct tuples FTW instead.

alfonsogarciacaro:
I've thought about this a couple of times, but I'm still unsure about what's the best way to emulate passing value types by reference in JS :

I'm closing it for now, will revisit.

ncave:
@alfonsogarciacaro

Since this is tagged with "Revisit", what is your opinion on these two options for supporting `byref/inref/outref` parameters :
- Convert all `mutable` variables and `byref/inref/outref` to `FSharpRef`. A bit heavy-handed, but simple to implement and reason about. Should be able to properly handle harder cases like mutable fields etc. Run-time backwards compatibility with older generated code might be an issue (but that probably already is an issue with Fable 2 vs 3), so perhaps enable behind a flag.
- Analyse usage of `mutable` variables and only declare as `FSharpRef` the ones that are later passed by reference (`AddressOf`), and also all `byref/inref/outref`. A bit more involving to handle harder cases, and creates two types of mutable variables, so could be either more compatible, or could be even less compatible with existing generated code. Perhaps enable behind a flag too.

Personally I like option 1 for its simplicity, but I may be missing some context, or oversimplifying things, feel free to shoot it down.
Definitely the performance impact on large real life projects needs to be evaluated so people can make an informed decision about enabling that compiler flag.

ncave:
@alfonsogarciacaro I see mutable vars are already implemented as closures (createAtom), so is it just a matter of extending that to `byref/inref/outref` and mutable fields, and implementing `AddressOf/AddressSet` to properly handle the closure function?

alfonsogarciacaro:
Hi @ncave! Sorry I don't remember well but I think I gave this a try at some point by converting variables to `FSharpRef` when passed as byref arguments and gave up (probably too quickly) because it was difficult to identify in the AST when we were accessing such refs (don't remember if it was difficult to just check for `byref` type in BasicPatterns.Value for some reason). It'd be great if you could give it another try, probably you'll come up with a better solution.

[...]

PR #1526, opened 2018-08-21
[Fable 2] Add test for  Map.TryGetValue

0x53A:
With an older version of Fable I get ``error FABLE: Cannot find replacement for Microsoft.FSharp.Collections.FSharpMap::TryGetValue``

I don't want to update right now, so let's see if this was added.

### DIFF
--- a/tests/Main/Fable.Tests.fsproj
+++ b/tests/Main/Fable.Tests.fsproj
@@ -5,6 +5,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Expecto" Version="5.1.2" />
+    <PackageReference Include="FSharp.Core" Version="4.5.2" />
   </ItemGroup>
   <ItemGroup>
     <Reference Include="Fable.Core" HintPath="../../build/fable/Fable.Core.dll" />
--- a/tests/Main/MapTests.fs
+++ b/tests/Main/MapTests.fs
@@ -36,6 +36,11 @@ let tests =
             xs.Count
             |> equal 1
 
+        testCase "Map.TryGetValue works" <| fun () ->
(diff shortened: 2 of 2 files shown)


0x53A:
AHA, TryGetValue was added with FSharp.Core 4.5 apparently, that's why it doesn't have a replacement yet, and why the tests say that it itsn't defined: https://github.com/fsharp/fslang-design/blob/master/FSharp.Core-4.5.0.0/FS-1050-trygetvalue-map.md

0x53A:
```
ERROR: the type 'byref`1' does not have a qualified name (C:/projects/fable/tests/Main/MapTests.fs)
```

this seems to be blocked by https://github.com/fable-compiler/Fable/issues/694

alfonsogarciacaro:
Yes, new F# 4.5 haven't been implemented (although funny enough `match!` seems to be supported). I guess that only `Map.TryGetValue` and `ValueOption` (which should become the same as `Option` in Fable) apply to Fable. It's not possible to pass numbers by reference in JS so we cannot support `byref` but it's possible to catch `TryGetValue` patterns in F# AST as [we already do for dictionaries](https://github.com/fable-compiler/Fable/blob/9a9d1666aaf37240741376b2ea2a0fcc96921f41/src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.Util.fs#L318-L326). I need to check what's the AST generated in this case.

alfonsogarciacaro:
Sorry, this was closed automatically because I renamed `dev2.0` as `master`. Would you mind resending it again to `master`?

0x53A:
I've changed the branch to `master`, can you please reopen the PR?

(I don't have the button)

MangelMaxime:
Done thank you @0x53A

PR #1528, opened 2018-08-21
[Fable 1] Add test for Map.TryGetValue

0x53A:
ref https://github.com/fable-compiler/Fable/pull/1526


alfonsogarciacaro:
I'd prefer not to add new features and push more releases for Fable 1.x, as this would increase the maintenance work. If it's ok with you, let's focus on Fable 2 to keep up with new F# features.

PR #1535, opened 2018-08-23
Initial work on improving "toHex"

xdaDaveShaw:
As per the discussion on #1530 there are some improvements needed to the `toHex` function.

This is the first set of changes and the behaviour in Javascript.
There are some failing tests still from running the codes in F#.

The first failure is:

```
Expected string to equal:
"-4095L: -FFF"

The string differs at index 8.
"-4095L: FFFFFFFFFFFFF001"
```

I'm not sure how to progress these now, any suggestions?

### DIFF
--- a/src/js/fable-core/String.ts
+++ b/src/js/fable-core/String.ts
@@ -1,4 +1,5 @@
 import { toString as dateToString } from "./Date";
+import Long, { toString as longToString } from "./Long";
 import { escape } from "./RegExp";
 import { isArray, toString } from "./Util";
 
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Thanks a lot @xdaDaveShaw! Let's merge this and I'll try to work on top of it. We should make tests pass both on JS and .NET when possible. So I'd say we need to change the assertion's in this case to match NET's behaviour, probably we need to do right shift longs too, I'll give it a try :+1:

PR #1543, opened 2018-08-28
Compile chars as JS strings again

### LINKED ISSUE #1524: String.join with chars return weird result
### Description

`System.String.Join` with chars returns weird result.

### Repro code

```fsharp
open Fable.Core
open System

[<Emit("debugger;")>]
let debug () : unit = jsNative

[0..10]
|> List.map (fun _ -> '*') 
|> fun chars -> String.Join("", chars)
|> fun x ->
    let y = x
    debug()
    x
|> printfn "%s"
```

*Expected:*
`"**********"`
*Actual:*
`"4242424242424242424242"`

### Related information

* Fable version (`dotnet fable --version`): 2.0.0-beta-001

### Workaround

In this case using strings instead of chars does give the expected outcome.

alfonsogarciacaro:
Also fixes #1524. Some discussion [here](https://github.com/fable-compiler/Fable/issues/1524#issuecomment-414865851).

### DIFF
--- a/src/js/fable-core/Char.ts
+++ b/src/js/fable-core/Char.ts
@@ -32,7 +32,8 @@ function getCategory() {
     categories[i / 2] = unicodeDeltas[i + 1];
   }
   // binary search in unicode ranges
-  return (cp: number) => {
+  return (s: string, index?: number) => {
+    const cp = s.charCodeAt(index || 0);
     let hi = codepoints.length;
     let lo = 0;
     while (hi - lo > 1) {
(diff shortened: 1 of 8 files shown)


src/js/fable-core/Char.ts
```diff
@@ -116,90 +117,84 @@ const isSymbolMask = 0
...
+export function isSymbol(s: string, index?: number) {
+  const test = 1 << getUnicodeCategory(s, index);
   return (test & isSymbolMask) !== 0;
 }
 
-export function isWhiteSpace(cp: number) {
-  const test = 1 << getUnicodeCategory(cp);
-  return ((test & isWhiteSpaceMask) !== 0)
-    || (0x09 <= cp && cp <= 0x0D) || cp === 0x85 || cp === 0xA0;
+export function isWhiteSpace(s: string, index?: number) {
```
ncave: @alfonsogarciacaro Shouldn't `isWhiteSpace` be using `getUnicodeCategory(s, index)` instead of regex?

alfonsogarciacaro:
@ncave I was reverting the PR but you're right, I've removed the regex. I just changed a bit the signature of `getUnicodeSignature` so it's not necessary to call `charCodeAt` twice.

Issue #1532, opened 2018-08-22
(dev2.0) Document fully numeric conversions (and make them more consistent


tomcl:
This issue is a placeholder to discuss progress on improving and documenting integer numeric conversions.

I plan to add documentation but not yet since we are still maybe going to rationalise / understand current issues in numeric types.

Here is a starting point. (To be improved)

(1) JS uses IEEE-754 64 bit floats for all numbers. These can store signed integers up to 53 bits (in FP format). Thus integers are stored in sign/magnitude form.

(2) JS BITWISE ops do 32 bit two's complement arithmetic. They convert automatically to/from the equiv (signed or unsigned?) FP integers. These ops truncate data to 32 bits. (But are the JS ints that convert to/from these unsigned: 0 - 2^32-1 or two's C range -2^31 - 2^31-1??). Details needed.

(3) Fable stores int64, uint64 as records containing two numbers (32 bits each) and a signed / unsigned boolean flag. The MS 32 bits are signed and represent top 32 two's c bits. The bottom 32 bits are unsigned and represent the lower 32 bits of a 64 bit two's C number.

(4) FABLE stores int32, uint32 as positive/negative or all positive respectively FABLE integers.

OK, so here are the expected (from fsi) .NET numeric conversions, and what Fable2.0 does.

```
(-1 |> any unsigned numeric type) should be 2^N -1 where N is type width.
(2^N-1 in any unsigned type width N -> any signed type) should be -1.

examples from FSI:

(-1 |> char |> uint) = 16383u
(-1 |> char |>int) = 16383 (char is unsigned 16 bit, so that conversion determines value)
(-1 |> uint64) = 18446744073709551615uL
(-1L |> uint64) = 18446744073709551615uL
(18446744073709551615uL |> int64) = -1L
(18446744073709551615uL |> int32) = -1
(-1 |> uint32) = 4294967295u
(4294967295u |> int) = -1
```

Basically, in .NET, -1 is preserved to and from numeric types, although it looks like 2^N-1 in an unsigned type of length N. this is pretty clean.

Fable2 gives answers that depend on BOTH source and result type. Notable from repl2 is:

uint64 -> int64 which seems to go via uint.
int -> uint64 negative numbers set to 0L

```
(-1 |> int32) -4294967296 (correct)
(-1 |> uint32) = 4294967295u (correct - and this is the most important one!)
(4294967295u |> int32) = -1 (correct, also important)
[...]

PR #1546, opened 2018-08-30
make integer conversion semantics closer to .NET as per #1532

tomcl:
New tests added and two new long.fs functions:

* fromInteger (for special cases of fromNumber - about same speed)
* toIntNumber (for special cases of toNumber - faster)


alfonsogarciacaro:
Great, thank you!

Issue #1550, opened 2018-09-01
Bitwise ops do not work on uint32


tomcl:
### Description

Bitwise operations (F# `&&&` and `|||` etc) generate signed results if the top bit of the result is 1, even if used on `uint32` data.

In JS, signed and unsigned numbers with bit 31 set - because the negative numbers get turned into JS negative floats - are different, even though in F# and .NET they are the same.

This is especially unfortunate because bitwise operations are more commonly used on unsigned quantities, and because the erroneous signed version prints (printf "%X") the same as the correct answer!

### Repro code

```
let prop = 
    let x = (0x80000000u) ||| 0u
    x = (x >>> 0)
```

This should be true, but is actually false.

### Expected and actual results

See above

### Related information

* Fable version (`dotnet fable --version`): 2.0.0-beta-003
* Operating system: windows 10

### How to cure this

Operator replacement should include (JS) >>> 0 on unsigned uint32. 

Whether there are problems with int32 as well needs more investigation, but I think this is OK.

See below for costs.

#### Alternative cure

Make all operations that care (mainly comparison) do the necessary normalisation, accepting that unsigned values can have two forms. That would in theory be more performant but it would be easy to miss cases and I suspect require a lot of miscellaneous code to be changed.

tomcl:
Proposed fix:

Inside `applyOp`, define:

* `binOpTrunc` (like `binOp` but adds truncation on `t` is unsigned)
* `unOpTrunc` (like `UnOp` but adds truncation on `t` is unsigned)

Adjust dispatch (inside `applyOp`) so that operators that need it are implemented with `binOpTrunc`, `unOpTrunc` instead of `binOp`, `unOp`.

Add a whole load of tests for this artifact so that correct operator/type cases can be adjusted.

Would this seem a good idea? If so I'll do it.

The overall affect on performance should be very small if as I hope this is not needed for any signed types (initial tests seem to indicate these are OK).

tomcl:
Precise info on what works and what not:

```


F#
```
|||: : OK
&&&: : OK
^^^: : OK
<<<: : OK
>>>: : OK
~~~: : OK
||| signed: : OK
&&& signed: : OK
^^^ signed: : OK
~~~ signed : : OK
<<< signed : : OK
>>> signed : : OK
```

Fable CLI:
```
|||: : expected 2147483648, actual -2147483648
Renderer.fs:104 &&&: : expected 2147483648, actual -2147483648
[...]

PR #1551, opened 2018-09-01
Fix truncation of large unsigned bitwise op results

tomcl:
Very small change to compiler code. Subfunction written and used on #1530 patch in addition to new usage to solve #1550.


alfonsogarciacaro:
Thanks!

PR #1554, opened 2018-09-03
Added compiler directive question

nojaf:
Added question about running code only in production.


MangelMaxime:
Is `RELEASE` directive set by default by the F# compiler ?

I believe that it's only if you use set the configuration to release so it should be mentioned.

If no, we should explain how to add it via webpack configuration.

alfonsogarciacaro:
Thanks @nojaf. I've edited the question to add more explanation and use `!DEBUG` which I think is more common. Does this look right to you?

nojaf:
Yes I've double checked in my own project. Only DEBUG is defined so `RELEASE` would indeed not work.

PR #1555, opened 2018-09-03
Fix #1542

### LINKED ISSUE #1542: Erased DU case not working with variable as key ([<Erase>] Custom of string * obj)
### Description

When using a variable for a key of an Erased DU case, the case is not erased, but produces a not erased key and an array as value. Only literal value works. I found this issue when I tried `CSSProp.Custom` from **fable-react**.

### Repro code

```fsharp
type MyType =
    | [<Erase>] Custom of string * obj

let key = "customKey"

let myList = [ Custom (key, "value") ]

// This will work as expected:
// let myList = [ Custom ("customKey", "value") ]

let myObj = keyValueList CaseRules.LowerFirst myList
printfn "%O" myObj
```

_Expected:_
`{"customKey":"value"}`
_Actual:_
`{"custom":["customKey","value"]}`

### Related information

* Fable version (`dotnet fable --version`): 1.3.17

### PS

I tried this code in Fable 2.0 REPL and it works well. So the question: Is v1 still maintained?

mvsmal:
Fix #1542

### DIFF
--- a/src/dotnet/Fable.Compiler/FSharp2Fable.fs
+++ b/src/dotnet/Fable.Compiler/FSharp2Fable.fs
@@ -183,8 +183,8 @@ let private transformNonListNewUnionCase com ctx (fsExpr: FSharpExpr) fsType uni
             // to pass custom values to keyValueList
             if hasAtt Atts.erase unionCase.Attributes then
                 match argExprs with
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Thanks @mvsmal. I will release a new Fable 1.x version :+1:

PR #1556, opened 2018-09-03
Fix AppVeyor artifacts

xdaDaveShaw:
After the rename of Fable.JS to Fable.Repl in 812168d this was missed.

I plan on sending a PR for the Repl2 repo to look in the right place if
this works.


alfonsogarciacaro:
Ups, nice catch, thank you!

PR #1557, opened 2018-09-04
Fable.Core update

ncave:


build.fsx
```diff
@@ -125,6 +125,7 @@ let buildSplitter () =
 let buildCoreJsFull () =
     Yarn.install CWD
     Yarn.run CWD "tslint" (sprintf "--project %s" coreJsSrcDir)
+    run (CWD </> "src/tools/InjectProcessor") dotnetExePath "run"
```
alfonsogarciacaro: I was thinking about this, I removed it because it made the `FableCoreJs` target take longer but it was only seldom needed, so I moved it to its own target (FableCoreJsInjects) but if you think it's useful we can keep it :)

ncave: @alfonsogarciacaro IMO it's more intuitive to have it as part of the main target, as it's very easy to forget to run it otherwise.

alfonsogarciacaro: Perfect, let's keep it then :+1: Build seems to be failing because restored info is not available for Fable.Core.JS.fsproj (I think I removed the full restore in `All` because now the Fable CLI is run with `dotnet run`), so we need run `dotnet restore src/js/fable-core` before hitting this line. Would it be possible to fix it?

alfonsogarciacaro:
Great additions, thanks a lot!

PR #1558, opened 2018-09-04
update numbers.md

tomcl:
This documentation is now a fair reflection of current state. The truncation issue could be resolved but it is not clear that is worth it. The three other issues I know about are now all resolved and don't need documentation!

Fixed:

* 64 bit conversions
* Bitwise operators
* long ranges

This is a great improvement on Fable 1.3.7.

I would not be very surprised it a few other semantic issues surface related to numbers, but I can't find any.


alfonsogarciacaro:
Thanks! I'll add a link to this document from the compatibility docs and update the web :+1:

PR #1559, opened 2018-09-06
add question which files need to be included in the nuget package

selketjah:
There were two people who asked about this on gitter in the last couple of days and the information is a bit buried in a blogpost, so I thought it would be nice to add to the FAQ


alfonsogarciacaro:
Good point, thanks a lot!

PR #1560, opened 2018-09-06
add 3 special symbols 

selketjah:
Add symbols ?  !!  and !^

I don't know how to test how it looks locally, so I'm not sure about layout.

It seems my commit message got a bit messed up because I added the special symbols to the commit, so sorry :(


alfonsogarciacaro:
No worries, this looks great. I'll adjust the layout if necessary, thank you!

PR #1564, opened 2018-09-07
Mention type providers are not supported by Fable

rbauduin:


MangelMaxime:
Actually, last time we checked with @alfonsogarciacaro  TP are supported but they need to generate Fable compatible code. 

Will let @alfonsogarciacaro confirm/infirm :)

alfonsogarciacaro:
I've edited the response and merged the commit, thanks for the PR @rbauduin!

PR #1566, opened 2018-09-08
Added missing replacement

ncave:
FCS-Fable2 was rebased to latest FCS (25.0.1), this PR adds a missing replacement.


alfonsogarciacaro:
This is great, it seems that FCS times in the REPL are the the same (or sometimes faster) than the Fable part. If I remember correctly, in Fable 1 REPL the FCS was more than 2 times slower. This probably means we've now optimization opportunities in the Fable part. I think I'll try to add an option to skip optimization passes (except for uncurrying) in DEBUG mode.

alfonsogarciacaro:
Hmm, no, I made a quick test but it seems it's not going to improve anything. Actually I get longer times if I disable some passes in FableTransforms. My guess is the uncurry transformations need to do more work if we don't do the lambda beta reduction first (and we always need the uncurry transformations).

ncave:
@alfonsogarciacaro The REPL times are more or less the same, perhaps a bit faster by a few % points, definitely not slower, but the ratio FCS to Fable hasn't changed.

ncave:
@alfonsogarciacaro The good news is the REPL compiles just fine with Fable2 using latest FCS (25.0.1), so `.paket/paket.exe update` may be in order (but I'll leave that to you as you may have other considerations).

alfonsogarciacaro:
Already did it :) everything seems to be working fine

PR #1572, opened 2018-09-13
add replacements for System.Uri

JacobChang:
It looks some tests failed, it's weired, it works on my local env

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ cache:
   yarn: true
 
 before_script:
-  - nvm install 6.11.5
-  - nvm use 6.11.5
+  - nvm install 10.10.0
+  - nvm use 10.10.0
   # # Update npm
   # - npm install -g npm
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ build:
   verbosity: detailed
 
 environment:
-  nodejs_version: "6.11.5"
+  nodejs_version: "10.10.0"
 
 cache:
  - "%LOCALAPPDATA%\\Yarn"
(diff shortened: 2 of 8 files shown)


JacobChang:
@alfonsogarciacaro For this one, As I use URL as replacements of System.Uri see [line](https://github.com/fable-compiler/Fable/pull/1572/files#diff-3a48a5025b4f5c57de2e242728c43dcbR18) , so when running tests, It expects URL as global variable , but only from version 10, Node start to export URL as global variable just like browser, so that's the reason why the tests failed because we use Node 6.11.5 for tests.

I can think of three ways to solve this

1 during tests, redeclare URL as global variable

    global.URL = require('url').URL;

2 upgrade to Node version 10

3 use PhantomJS for testing

what's your suggestion ?

alfonsogarciacaro:
This is great @JacobChang, thank you! We do have [a polyfill for the tests](https://github.com/fable-compiler/Fable/blob/b553ef7d26426d8acd01271e9d84d66a8a327361/tests/Main/js/polyfill.js) but it should be also OK to upgrade NodeJS. Can you please write the new version [here](https://github.com/fable-compiler/Fable/blob/b553ef7d26426d8acd01271e9d84d66a8a327361/appveyor.yml#L8) and [here](https://github.com/fable-compiler/Fable/blob/b553ef7d26426d8acd01271e9d84d66a8a327361/.travis.yml#L17-L18) and see if it passes CI then?

Thank you again for your contribution!

JacobChang:
@alfonsogarciacaro after I update node version to 10.10.0, All CI passes

alfonsogarciacaro:
And... tests are passing, fantastic! Thanks a lot for your contribution!

Choc13:
I found myself needing `Uri.TryCreate` today, so I've had a quick go at adding a replacement for it and I've opened a draft PR #2494. Would love any initial feedback before I go through and update all the tests and tidy up.

PR #1574, opened 2018-09-26
Minor bench updates

### LINKED ISSUE #1573: [REPL2] printf doesn't compile with optimizations enabled
__EDIT:__ it does compile, this seems to only be an issue in the editor?

--------------------
grafik

grafik

grafik

ncave:
Fixes #1573


src/dotnet/Fable.Compiler/Transforms/MonadicTrampoline.fs
```diff
@@ -1,7 +1,7 @@
 ﻿// Source: http://fssnip.net/dK/title/Monadic-Trampoline
 module MonadicTrampoline
 
-#if !FABLE_COMPILER
+#if NETSTANDARD2_0 || !FABLE_COMPILER
```
alfonsogarciacaro: Is it not enough to just use `!FABLE_COMPILER`? I'm afraid `NETSTANDARD2_0` could be activated also in Fable compilation and then we will have problems with the trampoline again.

The semantics may also be clearer if we have only the `FABLE_COMPILER` symbol. I'm using the negation operator because I didn't want to put the Fable-specific code first, but we can change the order of the branches if that's better.

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -1218,7 +1218,7 @@ let strings (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr opt
             addWarning com ctx.InlinePath r "String.Contains: second argument is ignored"
         let left = Helper.InstanceCall(c, "indexOf", Number Int32, [arg])
         makeEqOp r left (makeIntConst 0) BinaryGreaterOrEqual |> Some
-    | ("StartsWith" | "StartsWithOrdinal"), Some c, [_str] ->
+    | "StartsWith", Some c, [_str] ->
```
alfonsogarciacaro: Don't we need this anymore? It was added in #1566.

src/dotnet/Fable.Compiler/Transforms/MonadicTrampoline.fs
ncave: @alfonsogarciacaro No, not enough, sometimes Fable is compiled with both symbols on dotnet and then we want the first branch. It's fine, this is code internal to the Fable compiler. There are only 2 cases:
- Fable compiled with dotnet, with or without FABLE_COMPILER (branch 1)
- Fable compiled with Fable, with FABLE_COMPILER, but not NETSTANDARD2_0 (branch 2)

src/dotnet/Fable.Compiler/Transforms/Replacements.fs
ncave: @alfonsogarciacaro I know, it was added by mistake, I though it was missing, but it was actually caused by #1575

src/dotnet/Fable.Compiler/Transforms/MonadicTrampoline.fs
alfonsogarciacaro: Ah, ok! I remember now we had a similar problem in another place. Hmm, in that case I'd prefer to leave the `NETSTANDARD2_0`. Or we can try to clarify the meaning of `FABLE_COMPILER` which should be "this is being compiled by Fable", but I guess sometimes you're using it to mean "this is compiled by dotnet and will be used by Fable". Would it be possible to use a different symbol for the latter case?

ncave: @alfonsogarciacaro I'm afraid we already sprinkled FABLE_COMPILER all over Fable and FCS, that's how we roll.
In this particular case, I can leave only NETSTANDARD2_0 but it's safer to keep both, in case some future dotnet framework version does not define it.

alfonsogarciacaro: OK. I'm afraid to get the wrong code silently in some compilations if the build conditions change. But we can revisit this later, thank you!

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
```diff
@@ -227,6 +227,51 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) fsExpr =
...
+                 comp.FullName = "Microsoft.FSharp.Core.LanguagePrimitives.GenericEqualityComparer" ->
+        let r, typ = makeRangeFrom fsExpr, makeType com ctx.GenericArgs fsExpr.Type
+        let argTypes = [e1.Type; e2.Type] |> List.map (makeType com ctx.GenericArgs)
+        let! args = transformExprList com ctx [e1; e2]
+        let entity = comp.DeclaringEntity.Value
+        let opName = "GenericEquality"
+        let membOpt = tryFindMember com entity ctx.GenericArgs opName false argTypes
+        return (match membOpt with
+                | Some memb -> makeCallFrom com ctx r typ false argTypes None args memb
+                | None -> failwithf "Cannot find member %A.%A" (entity.FullName) opName)
```
alfonsogarciacaro: It'd be great if we could move these workarounds some where else. For Fable 2 I was trying to avoid dealing with optimizations in this pattern match. I'll see what can be done later :)

PR #1578, opened 2018-09-27
Repl edge

vbfox:
This PR remove the circular dependency that exists between `Set.fs`/`Map.fs` and `Types.js` by moving the definition of `Dictionary` and `HashSet` to a new file.

This will make the **repl2** compatible with Edge as the order of `import` statements in modules isn't defined in any specification they implemented it in a way that is unfriendly to circular dependencies

*(They plan to change it and mimic v8 behavior, it's sad I think it make a good point for using a language that doesn't allow theses like F# and then transpiling to JS).*

Related: https://github.com/fable-compiler/repl2/pull/51


vbfox:
@alfonsogarciacaro All the ones involving Types & Map should be gone, the last one is suspicious I need to check. The Array -> Array one seem weird ;)

Circular dependencies are allowed but their resolution order is undefined, v8 is doing clever hoisting so that it works in all but the most impossible cases. Edge used a more direct approach of just doing the imports in order.

Not much a difference if you don't use the imported elements directly, but fable uses `declare` everywhere and it need to really exists for the exports of other modules to be produced (As it doesn't run later when a function is called but need to run inline to produce the exports)

PR #1581, opened 2018-09-29
Make the NuGet badge show the version of dotnet-fable

inosik:
Or should we show the version of Fable.Core?

Congrats to the release of v2! 🎉


alfonsogarciacaro:
Good one, thank you!

PR #1584, opened 2018-10-01
Optimizations updates

ncave:
Small incremental steps towards better optimization.
Tests at least compile now (but they dont pass yet).
Still work to do, next work item is to fix "for x in list".


src/dotnet/Fable.Compiler/Transforms/Replacements.fs
```diff
@@ -633,15 +628,18 @@ let rec equals (com: ICompiler) r equal (left: Expr) (right: Expr) =
...
+        Helper.CoreCall("Util", "equals", Boolean, [left; right], ?loc=r) |> is equal
     | MetaType ->
         Helper.CoreCall("Reflection", "equals", Boolean, [left; right], ?loc=r) |> is equal
     | Tuple _ ->
         Helper.CoreCall("Util", "equalArrays", Boolean, [left; right], ?loc=r) |> is equal
-    | DeclaredType(ent,_) when hasBaseImplementingBasicMethods ent ->
-        Helper.InstanceCall(left, "Equals", Boolean, [right]) |> is equal
+    // unsafe optimization, left can sometimes be null
+    // | DeclaredType(ent,_) when hasBaseImplementingBasicMethods ent ->
+    //     Helper.InstanceCall(left, "Equals", Boolean, [right]) |> is equal
```
alfonsogarciacaro: Do you have an example when `left` may be null? `hasBaseImplementingBasicMethods` should only include records/unions/value types. And options are not DeclaredType in Fable AST?

ncave: @alfonsogarciacaro Boxing a union type does it sometimes, with optimize-fcs on. I haven't looked into the details, since disabling this optimization has very little perf impact if any. We can come back to it in the future, but again, the perf difference is very small.

PR #1589, opened 2018-10-03
Update Rollup plugin

inosik:
I tested it with my [example VS Code extension](https://github.com/inosik/fable-vscode-rollup-sample) locally and it worked without problems after I changed the Babel options to use the new presets and plugins.


alfonsogarciacaro:
Great, thanks! I'll release a stable version :+1:

PR #1590, opened 2018-10-03
Optimization updates

ncave:
- fixed optimized "for x in list"
- fixed optimized bigint.op_explicit
- initial work on optimized async, more to do


alfonsogarciacaro:
Thanks!

PR #1591, opened 2018-10-03
Add basic docs for #1398

inosik:
I hope the sample XML is clear enough for people who aren't that familiar with MSBuild.

One problem we have though is that we don't have any official docs where you can look up warnings/errors and their respective numbers. That's why I added the warning number to the example as a comment.


MangelMaxime:
If you think more context is need for the XML you do something like, I did for a webpack example:

```js
module.exports = {
    // ...
    output: {
        path: path.join(__dirname, CONFIG.outputDir),
    }
    // ...
};
```

The idea, is to help people understand where `output` is living and by adding `// ...` they directly understand that it's in the middle of something else.

alfonsogarciacaro:
This works @inosik, thanks a lot!

Issue #1594, opened 2018-10-06
Array.partition results are too long


ericharding:
### Array.partition result arrays are too long

Array.partition splits a source array into 2 parts.  The 'parts' both have `.length` equal to the source array but fewer elements. The remaining slots are `undefined`.  I think this function was rewritten from typescript to f# in the fable 2.0 update.

### Repro code

I modified the existing Array.partition test to fail.
```
    testCase "Array.partition works" <| fun () ->
        let xs = [|1.; 2.; 3.|]
        let ys, zs = xs |> Array.partition (fun x -> x <= 1.)
+        Array.length ys |> equal 1
+        Array.length zs |> equal 2
        ys.[0] - zs.[0] |> equal -1.
```

### Expected and actual results

  1) Arrays
       Array.partition works:

      Error: Expected: 1 - Actual: 3

### Related information

* Fable version (`dotnet fable --version`):  2.0.3
* Operating system: Windows 10
* I could not get this to fail on http://fable.io/repl.  If I 'console.log(Array.partition)` I was able to get the debug console to print out what looked like the original typescript algorithm.  Possible something was cahed? This remains a mystery.

ericharding:
Aha. Easier to repro on http://fable.io/repl2!
```
open Fable.Import.Browser
let a,b = [| 1;2;3;4 |] |> Array.partition (fun i -> i % 2 = 0)
console.log a
console.log b```

> Int32Array(4) [ 2, 4, 0, 0 ]
> Int32Array(4) [ 1, 3, 0, 0 ]

alfonsogarciacaro:
Thanks a lot for reporting and fixing this @ericharding!

PR #1595, opened 2018-10-06
Fix for #1594 Array.partition results are too long

ericharding:
This makes the algorithm more or less match the original typescript.  However it doesn't use IArrayCons so I'm not sure if we're losing something if this is applied to typed arrays.  If that's important I could switch to use 2 passes over the input to count up the size of the results.

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/ReplacementsInject.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/ReplacementsInject.fs
@@ -42,7 +42,6 @@ let fableCoreModules =
       "skipWhile", [(Types.arrayCons, 0)]
       "take", [(Types.arrayCons, 0)]
       "takeWhile", [(Types.arrayCons, 0)]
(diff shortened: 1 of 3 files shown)


Nhowka:
What if we just trim the extra elements? Like this:
```fsharp
let partition (f: 'T -> bool) (source: 'T[]) ([<Inject>] cons: IArrayCons<'T>) =
    let len = source.Length
    let res1 = cons.Create len
    let res2 = cons.Create len
    let mutable iTrue = -1
    let mutable iFalse = -1
    for i = 0 to len - 1 do
        if f source.[i] then
            iTrue <- iTrue + 1
            res1.[iTrue] <- source.[i]
        else
            iFalse <- iFalse + 1
            res2.[iFalse] <- source.[i]
    res1 |> Array.truncate iTrue, res2 |> Array.truncate iFalse
```

ericharding:
I like it.  Updated PR.

alfonsogarciacaro:
Great, thank you both!

Issue #1593, opened 2018-10-05
DateTime Format ToString("MMMM yyyy");


brase:
### Description

Formating DateTime behaves differently on server and client.

### Repro code

[Repl](http://fable.io/repl2/#?code=PYBwpgdgBAygngZwC5gLYChSSgMQIYBGANmAHQCSqIwATkqQEI3ADuCYN6JSUqYCCPAHMwAURJ8IPALxQAJsADGAV0n0Ajso5wYYEoqS0AFACIAxHwHCwJgJTpLgkeLSR6KAB5IAwsCluoAB4AWigAETwUABUASz5SADlWUijgGCQaGIghUwBZfNyoOGK4O3QgA&html=DwCwLgtgNgfAUHUBTAhgE3gAk8CSwqYDGIKATgM74C8ARAK5gBmAtABy3zAD0IqGCYACMA9mgCeWHCACMmAJZo6eChRQBzJJwCySVRqSYADlBREkIEVDRIyPWV26iJ8RL0iwgA&css=Q)

### Expected and actual results

Expected Result: "October 2018"
Actual Result: "10 2018"

voronoipotato:
````
Microsoft (R) F# Interactive version 10.2.3 for F# 4.5
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> # silentCd @".";;
- # 1 @"Untitled-1"
- ;;
> let x = System.DateTime.Now;;
val x : System.DateTime = 10/5/2018 10:53:26 PM

> x.ToString("MMMM yyyy");;
val it : string = "October 2018"

````

Reproduced expected result in FSI.

**Comment by voronoipotato** [comment]
https://momentjs.com/ might be a good thing to use here once we get to localization, though we would have to translate from .net's patterns to moment-js patterns. I'm aware we may not ~fully replicate~ DateTime.ToString here but a little probably goes a long way.

```
     Customized format patterns:
     P.S. Format in the table below is the internal number format used to display the pattern.
     Patterns   Format      Description                           Example
     =========  ==========  ===================================== ========
        "h"     "0"         hour (12-hour clock)w/o leading zero  3
        "hh"    "00"        hour (12-hour clock)with leading zero 03
        "hh*"   "00"        hour (12-hour clock)with leading zero 03
        "H"     "0"         hour (24-hour clock)w/o leading zero  8
        "HH"    "00"        hour (24-hour clock)with leading zero 08
[...]

PR #1596, opened 2018-10-06
Fix for #1593 added some date tostring options

voronoipotato:
Needs some test cases and some testing. I haven't even built fable yet because I've not gotten that far yet but this is where I am right now.

### DIFF
--- a/src/js/fable-core/Util.ts
+++ b/src/js/fable-core/Util.ts
@@ -1,3 +1,5 @@
+import { isNumber } from "./Char";
+
 // tslint:disable:ban-types
 
 // Object.assign flattens getters and setters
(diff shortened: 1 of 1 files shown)


voronoipotato:
https://github.com/fable-compiler/Fable/issues/1593

voronoipotato:
I guess feel free to close this PR

alfonsogarciacaro:
Thanks for your PR @voronoipotato! Maybe we don't need to close it in full, what's in the PR, only localization? I think it'll be difficult to support Globalization as in .NET because it's quite a big amount of work and that may increase the bundle sizes. For the same reasons, I'd prefer to avoid external dependencies in fable-core (like moment-js), but we can discuss about that.

voronoipotato:
I'll reopen for now to see if anything may be useful. I wouldn't say that it was "for" globalization but rather some of the formatting does not take into consideration other cultures. The goal I had in doing this PR was to make copy pasting F# into fable just a little bit easier but I did have concerns about it only being useful for westerners potentially mostly for Americans (I know very little about other cultures).

voronoipotato:
I added tostring options with the intention of improving compatibility with Core. I added (f: milliseconds) , (t: AM/PM), dddd Weekdays (english), ddd Weekdays Short (english), mmmm Months (english), mmm Months Short (english), in addition I added several single letter tostring options. d,D,f,F,g,G,m,M,t,T,U . The single letter ToString format for dates exists as a convenience and allows the user to quickly format a date in a useful way. `Date.ToString("d")` for example is equivalent to `Date.ToString("M/d/yyyy")` however you can see how this may be less useful for someone who doesn't use middle endian dates. There definitely are bits that would require no localization, and many of these are the default if I'm not mistaken in .Net, which wouldn't surprise me given that the Microsoft is an American company. However I am an American, and have never programmed anything in .Net that needed localization so I'm not exactly any authority on the subject.

ncave:
Perhaps later we can move all the date stuff into `Date.ts`, since `Util.ts` is getting quite big.
A few tests to cover the new options will go a long way. "*A test a day keeps the doctor away.*"

voronoipotato:
That sounds wise ncave. I'll give it a shot.

alfonsogarciacaro:
Hi @voronoipotato! What's the status of this? Will you include some tests to see what the PR is trying to solve? Thank you!

MangelMaxime:
This PR being quite old I think we can close it for now.

There is https://github.com/fable-compiler/Fable/issues/3692 which tracks the task to improve supported format for DateTime. If someone want to revive it then can either adapt the original mscord code or take inspiration from this PR too.

Issue #1587, opened 2018-10-03
Make it possible to define conditional constants in Fable REPL


alfonsogarciacaro:
I just noticed the REPL doesn't define the `DEBUG` symbol so statements like `assert false` are erased in the generated code. @ncave Would it be possible to pass such symbols to [checker.ParseAndCheckScript](https://github.com/fable-compiler/Fable/blob/0afdad2f5e6b1a3f12bf2749aab8a608ac5c8314/src/dotnet/Fable.Repl/Main.fs#L98)?

ncave:
@alfonsogarciacaro Yes, it's possible, I'll have to add it.

PR #1598, opened 2018-10-08
Repl parsing update

ncave:
- added fix for #1587 (conditional defines)
- added parsing for multiple project files

### DIFF
--- a/src/dotnet/Fable.Repl/Interfaces.fs
+++ b/src/dotnet/Fable.Repl/Interfaces.fs
@@ -18,7 +18,8 @@ type Glyph =
     | Event
 
 type Error =
-    { StartLineAlternate: int
+    { FileName: string
+      StartLineAlternate: int
       StartColumn: int
       EndLineAlternate: int
       EndColumn: int
@@ -41,9 +42,14 @@ type IChecker =
 type IParseResults =
     abstract Errors: Error[]
 
+type IParseFilesResults =
+    abstract Errors: Error[]
+    abstract GetResults: fileName: string -> IParseResults option
+
 type IFableManager =
     abstract CreateChecker: references: string[] * readAllBytes: (string -> byte[]) -> IChecker
     abstract ParseFSharpProject: checker: IChecker * fileName: string * source: string -> IParseResults
+    abstract ParseFSharpProjectFiles: checker: IChecker * projectFileName: string * fileNames: string[] * sources: string[] -> IParseFilesResults
     abstract GetParseErrors: parseResults: IParseResults -> Error[]
     abstract GetDeclarationLocation: parseResults: IParseResults * line: int * col: int * lineText: string -> Async<Range option>
     abstract GetToolTipText: parseResults: IParseResults * line: int * col: int * lineText: string -> Async<string[]>
--- a/build.fsx
+++ b/build.fsx
@@ -70,7 +70,7 @@ let coreJsBuildDir = CWD </> "build/fable-core"
 let cliSrcDir = CWD </> "src/dotnet/Fable.Compiler"
 let coreJsSrcDir = CWD </> "src/js/fable-core"
 let ncaveFcsForkRepo = "https://github.com/ncave/FSharp.Compiler.Service"
-let ncaveFcsForkBranch = "fable2"
+let ncaveFcsForkBranch = "fable"
 let ncaveFcsCodeGenTarget = "fcs/build CodeGen.Fable"
 
 // Targets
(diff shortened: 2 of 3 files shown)


alfonsogarciacaro:
Awesome, thank you!

src/dotnet/Fable.Repl/Interfaces.fs
```diff
@@ -41,9 +42,14 @@ type IChecker =
...
     abstract Errors: Error[]
 
+type IParseFilesResults =
+    abstract Errors: Error[]
+    abstract GetResults: fileName: string -> IParseResults option
+
 type IFableManager =
     abstract CreateChecker: references: string[] * readAllBytes: (string -> byte[]) -> IChecker
     abstract ParseFSharpProject: checker: IChecker * fileName: string * source: string -> IParseResults
+    abstract ParseFSharpProjectFiles: checker: IChecker * projectFileName: string * fileNames: string[] * sources: string[] -> IParseFilesResults
```
alfonsogarciacaro: Does this mean it's possible to compile multiple files with the REPL now? 😮

ncave: @alfonsogarciacaro Yes, that's exactly what it means :)

I know you said your preference is not to make the Repl into a full-blown environment, but something simple like [this layout](https://webassembly.studio/?f=pne1undq5g) with treeview on the left should now be possible.

Also, a javascript-only version of Fable is now possible too, at least for simple self-contained projects where the project cracker can just be a [carefully crafted](https://stackoverflow.com/a/1732454) regex.

alfonsogarciacaro: You're tempting me ;) I didn't want to turn the Repl into a full-blown environment so we didn't have to spend too much time with it, but if you're already doing your magic... Adding a file tree shouldn't be difficult (maybe @MangelMaxime has an idea?). How's the performance? Do you need to compile all files every time, or is it possible to recompile only single files [as dotnet-fable does](https://github.com/fable-compiler/Fable/blob/2ece26e5ea586af4ca3423d72f4d370077a9c27e/src/dotnet/Fable.Compiler/CLI/Agent.fs#L67)?

Releasing an experimental Fable compiler on node could be also interesting. Simple parsing of .fsproj shouldn't be too difficult (as long as we have an XML reader) and in fact we're [already doing that](https://github.com/fable-compiler/Fable/blob/2ece26e5ea586af4ca3423d72f4d370077a9c27e/src/dotnet/Fable.Compiler/CLI/ProjectCracker.fs#L156-L180). But it's a slippery slope because then . Maybe we can focus on the REPL for now, and later provide a Node server that can host the REPL locally and save the code or generate a JS bundle.

BTW, I realized there's a TypeScript to WebAssembly compiler in your link above. I think Babel 7 can handle Typescript, so this may be a way to compile F# modules into wasm without the .NET runtime overhead.

MangelMaxime: Adding a file tree is definitly possible and will allow me to rework the collapse/expanded sidebar to make it more user friendly. 

Some people complain because it's easy to get your mouse outside of the "hovered" menu and close the menu when you don't want.

I will take a look at it later :)

MangelMaxime: When the tree view is ready, I will ping you so we can add the multi file support in the REPL if I don't succeed alone :)

ncave: @alfonsogarciacaro There is no parse / typecheck caching at the moment, but it can be added.

I see you picked on my not so subtle cues about the wasm :)
Babel can handle TypeScript going in, but not produce it, so that doesn't help.

My [cunning plan](https://www.youtube.com/watch?v=fQk_832EAx4) is to just add back the flow types in Fable 2, slightly modified to match what AssemblyScript [expects](https://github.com/AssemblyScript/assemblyscript/wiki/Types), and remove the comment tags around them post-Babel, leaving the types.

Zaid-Ajaj: @alfonsogarciacaro If you need a Fable Xml parser, there is already [one available](https://github.com/Zaid-Ajaj/Fable.SimpleXml) that runs on [browser](https://zaid-ajaj.github.io/Fable.SimpleXml/) and node, although I am not sure how you use/bootstrap a Fable library in the Fable implementation itself but I guess you already know your way around that :smile:

alfonsogarciacaro: Thanks @Zaid-Ajaj! Yes, that should work :)

@ncave Sounds great! I've open a new issue #1601 to continue the discussion there.

PR #1605, opened 2018-10-15
fable-compiler update

ncave:
- fixed support for signatures
- minor fixes here and there
- separated errors from each different step (FCS and Fable)
- switched compiler to the simpler parsing to lower memory usage


alfonsogarciacaro:
You'll end up making the JS version faster than the .NET version 💪 This is great, thank you!

ncave:
@alfonsogarciacaro Unlikely, at least until webasm is in full force, but it doesn't seem that much slower from the [very limited preliminary] testing, perhaps 3x or so.

PR #1606, opened 2018-10-15
Improve compatibility of Fable REPL Lib files

alfonsogarciacaro:
See https://github.com/fable-compiler/repl2/pull/56/files

This PR adds some features to enable better compatibility for the REPL Lib (see link above). I'm not very happy because I has to add quite a number of extra fields to compiler and replacement options, but maybe it's not that bad.

What do you think @ncave @MangelMaxime? Is this solution OK for now or should we try to compile library sources directly in the browser?


ncave:
@alfonsogarciacaro LGTM, doesn't look that big of a change.

MangelMaxime:
Looks good to me too 👍

Issue #1389, opened 2018-04-18
Improve active pattern errors


MangelMaxime:
### Description

Considering the next code:

```fs
type Test =
    | Msg1
    | Msg2

let testToString msg =
    match msg with
    | Msg1 -> "msg 1"
```

We generate this code:

```js
import { setType } from "fable-core/Symbol";
import _Symbol from "fable-core/Symbol";
import { comparePrimitives } from "fable-core/Util";
export class Test {
  constructor(tag) {
    this.tag = tag | 0;
  }

  [_Symbol.reflection]() {
    return {
      type: "Test.Test",
      interfaces: ["FSharpUnion", "System.IEquatable", "System.IComparable"],
      cases: [["Msg1"], ["Msg2"]]
    };
  }

  Equals(other) {
    return this.tag === other.tag;
  }

  CompareTo(other) {
    return comparePrimitives(this.tag, other.tag);
  }

}
setType("Test.Test", Test);
export function testToString(msg) {
  if (msg.tag === 0) {
    return "msg 1";
  } else {
    throw new Error("test.fs", 7, 10);
  }
}
```

### Proposition

I would like to add a custom `Error` type to Fable: `MatchError` (or something like that) and generate something like:

```js
class MatchError extends Error {
  constructor(failingCase, ...params) {
    // Pass remaining arguments (including vendor specific ones) to parent constructor
    super(...params);

    // Maintains proper stack trace for where our error was thrown (only available on V8)
    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, CustomError);
    }

    // Custom debugging information
    this.failingCase = failingCase;
  }
}
```

```js
    throw new MatchError(msg, "test.fs", 7, 10);
```

### Why ?

The idea behind this proposition is to make Elmish errors understandable.

Today we have something like: 

<img width="464" alt="capture d ecran 2018-04-18 a 16 59 18" src="https://user-images.githubusercontent.com/4760796/38940204-e8734c06-4329-11e8-81e7-815d5cd96b56.png">

Tomorrow we could have:

```
Unable to process message:
Msg2

From file: 
Test.fs
```

Also, if this is a nested DU we get direct access to the whole tree:

```fs
type Test =
    | Msg1 of string
    | Msg2 of Test

printfn "%A" (Msg2 (Msg1 "deep_string"))
// Log: Msg2 (Msg1 "deep_string"))
```

This can be really useful in Elmish application where every message is a nested DU

[...]

PR #1611, opened 2018-10-19
Tackle #1389 Throw proper Error when match cases are incomplete

Zaid-Ajaj:
Addresses #1389 

This PR detects when a `MatchFailureException` exception is thrown as the last decision target on the F# AST part of things, then it replaces the *transformed* Fable target into a proper `throw Error(errorMessage)` expression where `errorMessage` contains the union case type and the file name:

> I am replacing the Fable.Expr inline from the last transformed expression because I don't know how to ree-construct F# AST in code for a `failwith msg` expression

### Example
```fs
type Shape = 
    | Circle of int 
    | Square of int 
    | Rectangle of int * int 

try 
    let unitCircle = Circle 1
    match unitCircle with
    | Rectangle(n, m) -> failwith "Should not happen" 
    | Square n -> failwith "Should not happen"
with
    | ex -> printfn "%s" ex.Message
```

### Before
```
undefined
```
### After
```
The match cases were incomplete against type of 'Shape' at <filename>
```

### Limitation 
 - Does not include the union case info of the input case being matched against because
   - Type information unavailable from runtime (specific union case is unknown at compile time) 
   - Detecting the input value from `match input with` is not trivial (for me at least) from nested `IfThenElse` statements 

### To do 
- re-write all `raise MatchFailureException()` to `failwith "The match cases were incomplete"` at the F# AST level of things

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
@@ -619,7 +619,7 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) fsExpr =
 
     // Pattern Matching
     | BasicPatterns.DecisionTree(decisionExpr, decisionTargets) ->
-        let! fableDecisionExpr = transformExpr com ctx decisionExpr 
+        let! fableDecisionExpr = transformExpr com ctx decisionExpr
         let rec transformDecisionTargets acc xs =
             trampoline {
                 match xs with
@@ -635,6 +635,7 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) fsExpr =
 
         let! fableDecisionTargets = transformDecisionTargets [] decisionTargets
         
+        // rewrite last decision target if it throws MatchFailureException
         let compiledFableTargets = 
             match isRasingMatchFailureExpr (snd (List.last decisionTargets)) with 
             | Some fileNameWhereErrorOccurs -> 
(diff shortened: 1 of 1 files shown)


MangelMaxime:
@Zaid-Ajaj Awesome :)

Can't we pass the unmatch value to `"%A"` formatter as sown in the issue. This can be important when you matching against a string or int and you didn't provide a default case.

Like that we have all the info needed:
```fs
printfn "%A" (Msg2 (Msg1 "deep_string"))
// Log: Msg2 (Msg1 "deep_string"))
```

Zaid-Ajaj:
@MangelMaxime generating call like `faillwith (sprintf "Match cases incomplete: %A" inputValue)` from the error should not be difficult per se but the hardest part is extracting `inputValue` from the `match inputValue with` because it gets converted to nested `IfThenElse` statements, so I need some way of tracing back the `inputValue` identifier so I can use it from the `sprintf` function

Maybe @alfonsogarciacaro can give me hints 🙏

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
```diff
@@ -232,6 +232,45 @@ let private transformUnionCaseTest (com: IFableCompiler) (ctx: Context) r
...
+        match methodInfo.FullName with 
+        | "Microsoft.FSharp.Core.Operators.raise" -> 
+            match value with 
+            | BasicPatterns.NewRecord(recordType, [ BasicPatterns.Const (value, valueT) ; rangeFrom; rangeTo ]) -> 
+                match recordType.TypeDefinition.FullName with 
+                | "Microsoft.FSharp.Core.MatchFailureException"-> Some (value.ToString())
+                | _ -> None
+            | _ -> None 
+        | _ -> None 
+    | _ -> None
```
alfonsogarciacaro: Maybe turning this into an active pattern and move it to the [FSharp2Fable.Patterns module](https://github.com/fable-compiler/Fable/blob/d1ef91e4a0bf8311f21e42acd3245e178a3c0b2d/src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.Util.fs#L266)?

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
```diff
@@ -232,6 +232,45 @@ let private transformUnionCaseTest (com: IFableCompiler) (ctx: Context) r
...
+            | BasicPatterns.NewRecord(recordType, [ BasicPatterns.Const (value, valueT) ; rangeFrom; rangeTo ]) -> 
+                match recordType.TypeDefinition.FullName with 
+                | "Microsoft.FSharp.Core.MatchFailureException"-> Some (value.ToString())
+                | _ -> None
+            | _ -> None 
+        | _ -> None 
+    | _ -> None  
+
+/// Creates a "throw Error({errorMessage})" expression
+let private createThrowErrorExpr (errorMessage: string) range : Fable.Expr =
```
alfonsogarciacaro: There's already a similar helper in [Replacements.Helpers.error](https://github.com/fable-compiler/Fable/blob/d1ef91e4a0bf8311f21e42acd3245e178a3c0b2d/src/dotnet/Fable.Compiler/Transforms/Replacements.fs#L97-L98), you could use it to avoid code duplication.

alfonsogarciacaro:
About accessing the `inputValue`, I'll need to play a bit with this to see how it can be done. Sometimes [we captures some values from the F# AST](https://github.com/fable-compiler/Fable/blob/d1ef91e4a0bf8311f21e42acd3245e178a3c0b2d/src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.Util.fs#L18) for this kind of situation, but I'll try to see if there's an alternative way.

alfonsogarciacaro:
Awesome, thank you! If more intensive AST manipulation is needed maybe this can be moved to the FableOptimizations pass (though this can slower the compilation a bit). Anyway, let's merge and I'll try to have a look. Thanks again! 🥇

PR #1613, opened 2018-10-20
(fable-compiler) Handle project references

ncave:
- now compiles all files and defines from all referenced projects
- retains the projects structure under the output directory (not flattened)

### DIFF
--- a/src/dotnet/Fable.Repl/bench-compiler/app.fs
+++ b/src/dotnet/Fable.Repl/bench-compiler/app.fs
@@ -6,7 +6,7 @@ open System.Text.RegularExpressions
 let references = Fable.Repl.Metadata.references false
 let metadataPath = Fable.Path.Combine(__SOURCE_DIRECTORY__, "../metadata2/") // .NET BCL binaries
 
-let parseProject projectPath =
+let parseProjectFile projectPath =
     let projectFileName = Fable.Path.GetFileName projectPath
     let projectText = readAllText projectPath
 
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Thanks @ncave! The package has been released (it doesn't include changes in this PR yet) but you'll have to download it as `npm install fable-compiler@next` because the build script automatically adds the `next` tag if the version number includes a suffix like -alpha or -beta.

I tried to download and it's working! The startup time seems to be fast than dotnet-fable. What's your vision for this? Right now it may be a bit difficult for devs as you cannot use libraries, but it shouldn't be difficult to add the Fable.Repl.Lib.dll with the precompiled JS files.

Also, if the goal is to enable F# development in a machine without dotnet SDK, we will need editor support too. This could be done either by running the Fable REPL locally or by creating a VS Code extension for F# edition that uses FCS purely in JS.


> If you have an npm account I can give you publishing rights to the package, or we can make the CI server toi publish a new prerelease version after every successful build.

ncave:
@alfonsogarciacaro Thanks, actually I'm not sure what the vision or goal is, more of a proof of concept than anything else at this point.

ncave:
@alfonsogarciacaro One clear usage of the fable-compiler is to benchmark how future changes in Fable affect the performance of the code it generates. Just pick a large diverse code base that already compiles to Fable, and bench away.

MangelMaxime:
I think we should limit ourself to use `fable-compiler` for benchmark purpose as proposed by @ncave and also for running tests for the REPL.

Because, one of the future goal is to propose a component version of the REPL or a library to use the REPL feature like auto-completion or others things. It's important that us for the REPL and users using it can run automatic tests before publishing a new versions. 

For example, current version of the REPL include `Thoth.Json` and `Fable.PowerPack` but we didn't run any tests to confirm it's working. Outside of the example that is using it but you see where I want to go ^^

Also providing a VS Code extension is out of our capabilities for now it will have a huge maintenance cost (just see Ionide maintenance cost for comparaison). Same goes with providing a local version for the REPL.

I am ok to provide support for a component version of the REPL so people can embedded it into their website and documentation because it's an environnement that we understand compared to Desktop environnement.

BTW awesome works on the `fable-compiler` package :)

PR #1614, opened 2018-10-20
fable-compiler version update

ncave:
@alfonsogarciacaro Please pull latest FCS-Fable before building/packaging/publishing.
I do have an npm account (same as github), in case you don't have time to publish.
Perhaps CI publish is simpler, but either way works, whatever is easier for you.

### DIFF
--- a/src/js/fable-compiler/package.json
+++ b/src/js/fable-compiler/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fable-compiler",
-  "version": "2.0.0-alpha-001",
+  "version": "2.0.0-alpha-002",
   "main": "index.js",
   "bin": {
     "fable": "index.js"
--- a/src/dotnet/Fable.Repl/bench-compiler/app.fs
+++ b/src/dotnet/Fable.Repl/bench-compiler/app.fs
@@ -138,7 +138,7 @@ let parseFiles projectPath outDir optimized =
 
 let parseArguments (argv: string[]) =
     // TODO: more sophisticated argument parsing
-    let usage = "Usage: fable-compiler projectPath [--options]"
+    let usage = "Usage: fable <PROJECT_PATH> [--options]"
     let opts, args = argv |> Array.partition (fun s -> s.StartsWith("--"))
     match args with
     | [| projectPath |] ->
--- a/src/js/fable-compiler/RELEASE_NOTES.md
+++ b/src/js/fable-compiler/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+### 2.0.0-alpha-002
+
+* project references support
+
 ### 2.0.0-alpha-001
 
 * First draft
--- a/src/js/fable-compiler/src/app.fs
+++ b/src/js/fable-compiler/src/app.fs
@@ -129,7 +129,7 @@ let parseFiles projectPath outDir optimized =
 
 let parseArguments (argv: string[]) =
     // TODO: more sophisticated argument parsing
-    let usage = "Usage: fable projectPath outDir [--options]"
+    let usage = "Usage: fable <PROJECT_PATH> <OUT_DIR> [--options]"
     let opts, args = argv |> Array.partition (fun s -> s.StartsWith("--"))
     match opts, args with
     | [| "--help" |], _ -> printfn "%s" usage


src/js/fable-compiler/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fable-compiler",
-  "version": "2.0.0-alpha-001",
+  "version": "2.0.0-alpha-002",
```
alfonsogarciacaro: It's better if you don't change the version in package.json because the "PublishPackages" build target checks if RELEASE_NOTES and package.json are out of sync and in that case, it publishes the package.

alfonsogarciacaro:
Thanks! I added you as a maintainer. Building the REPL bundle takes a bit of time in my machine, and given that there are many places that need to be kept in sync (your FCS fork, etc) I'm afraid it's easy for me to make mistakes. If you can publish a new version, that's great. But anyway I'll edit the build target to the bundle is taken from latest CI successful build, as it's done for the REPL.

alfonsogarciacaro:
Sorry, two quick questions:

- What exactly does the `Export.Metadata` build target in the `export` branch of your FCS fork? Create .dlls only with metadata and without actual code? If so, it's that to reduce the size or is there any other requirement for .dlls to work with Fable REPL?

- I think we already talked about this before but, it's difficult to implement `ParseAndCheckFileInProject` in the REPL? That would be very interesting to speed up recompilations.

alfonsogarciacaro:
Another question 😸 When does it make sense to use the new `ParseFSharpProjectSimple`?

ncave:
@alfonsogarciacaro 
- Yes, just to reduce DLL file size, no other reason, as long as the DLL is built against the same target framework and not too old (i.e. self-contained, no external signature files). Technically we can make the external signature files work, but they're embedded in the DLL by default nowadays, so to simplify the file loading somewhat, we're not trying to load them explicitly.

- It's not that difficult to add parsing and typecheck caches, lets open a feature request and see if the community has any interest in developing further this POC. As you pointed out, the incremental compiling only makes sense in an editor setting, and only for larger projects, but currently you need dotnet installed anyway for the editor plugins that use FCS.

- ParseFSharpProjectSimple just parses/typechecks, does not give you back the additional context for tooltips, intellisense, etc. It's enough for compilation, so that's what the fable-compiler uses, as it can handle large projects. In contrast, the non-simple one is very memory-hungry at the moment, and so is only suitable for small projects, perhaps that will improve in the future, with caching, etc.

src/js/fable-compiler/package.json
ncave: @alfonsogarciacaro Thanks, let me know once the CI can be used to trigger the publishing, as I want to get the -002 version published before people start using it.

alfonsogarciacaro:
Thanks @ncave. The two use cases for `ParseAndCheckFileInProject` in the REPL I have in mind are:

- If we add multi-file support, don't recompile/parse all files every time, just the one you're working on at that moment.
- Likewise, if we add the option to load libraries on demand (instead of a precompiled library as we do now), we would only need to compile the library files once. This should allow us to bring the "Optimize" option back.

About `ParseFSharpProjectSimple`, could this benefit dotnet-fable too? We don't need tooltips, intellisense, etc there either, so implementing something similar could have a performance benefit.

alfonsogarciacaro:
Thanks @ncave. The two use cases for `ParseAndCheckFileInProject` in the REPL I have in mind are:

- If we add multi-file support, don't recompile/parse all files every time, just the one you're working on at that moment.
- Likewise, if we add the option to load libraries on demand (instead of a precompiled library as we do now), we would only need to compile the library files once. This should allow us to bring the "Optimize" option back.

About `ParseFSharpProjectSimple`, could this benefit dotnet-fable too? We don't need tooltips, intellisense, etc there either, so implementing something similar could have a performance benefit.

alfonsogarciacaro:
Thanks @ncave. The two use cases for `ParseAndCheckFileInProject` in the REPL I have in mind are:

- If we add multi-file support, don't recompile/parse all files every time, just the one you're working on at that moment.
- Likewise, if we add the option to load libraries on demand (instead of a precompiled library as we do now), we would only need to compile the library files once. This should allow us to bring the "Optimize" option back.

About `ParseFSharpProjectSimple`, could this benefit dotnet-fable too? We don't need tooltips, intellisense, etc there either, so implementing something similar could have a performance benefit.

alfonsogarciacaro:
Thanks @ncave. The two use cases for `ParseAndCheckFileInProject` in the REPL I have in mind are:

- If we add multi-file support, don't recompile/parse all files every time, just the one you're working on at that moment.
- Likewise, if we add the option to load libraries on demand (instead of a precompiled library as we do now), we would only need to compile the library files once. This should allow us to bring the "Optimize" option back.

About `ParseFSharpProjectSimple`, could this benefit dotnet-fable too? We don't need tooltips, intellisense, etc there either, so implementing something similar could have a performance benefit.

alfonsogarciacaro:
Thanks @ncave. The two use cases for `ParseAndCheckFileInProject` in the REPL I have in mind are:

- If we add multi-file support, don't recompile/parse all files every time, just the one you're working on at that moment.
- Likewise, if we add the option to load libraries on demand (instead of a precompiled library as we do now), we would only need to compile the library files once. This should allow us to bring the "Optimize" option back.

About `ParseFSharpProjectSimple`, could this benefit dotnet-fable too? We don't need tooltips, intellisense, etc there either, so implementing something similar could have a performance benefit.

alfonsogarciacaro:
Thanks @ncave. The two use cases for `ParseAndCheckFileInProject` in the REPL I have in mind are:

- If we add multi-file support, don't recompile/parse all files every time, just the one you're working on at that moment.
- Likewise, if we add the option to load libraries on demand (instead of a precompiled library as we do now), we would only need to compile the library files once. This should allow us to bring the "Optimize" option back.

About `ParseFSharpProjectSimple`, could this benefit dotnet-fable too? We don't need tooltips, intellisense, etc there either, so implementing something similar could have a performance benefit.

ncave:
@alfonsogarciacaro Technically nothing is stopping you from linking against FCS-Fable instead of stock FCS on dotnet as well, but I would stick with the stock FCS for now as that's what everybody's using. It's very unlikely that there will be any big performance benefits, as there is a lot of internal caching going on in FCS. Also, if I'm not wrong, you need it for the hot reloading (in an editor setting).

PR #1615, opened 2018-10-22
Adding Babel types

ncave:
This is trying to bring back some of the Babel types from Fable 1.x

@alfonsogarciacaro A few questions, if you have the time to take a quick look:
- Do we need to keep generic arg list in `AST.Fable.FunctionKind` and `AST.Fable.ObjectMemberKind`?
- Obviously more types need to be properly extracted and set, e.g. for arrow func expressions, etc.
- Anything else that catches your eye.


alfonsogarciacaro:
This is great @ncave! I'll try to have a deeper look later. But sorry, I don't understand your question, `AST.Fable.FunctionKind` keeps a list of argument `Ident`s and `AST.Fable.ObjectMemberKind` doesn't keep any reference to the arguments at all (they must be contained in a Function expression within member value). What do you need exactly?

ncave:
@alfonsogarciacaro Previously (in 1.x) the `getMemberArgsAndBody` was getting the generic types as an additional parameter, and it needs to come from some func or member info in AST.Fable (or not, if you don't think so). It's used [here](https://github.com/ncave/Fable/blob/flow/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs#L369).

alfonsogarciacaro:
Ah, ok! I think probably the generic arguments can be extracted from the function type but let me check.

alfonsogarciacaro:
@ncave Probably you can use [entityRefMaybeImported](https://github.com/fable-compiler/Fable/blob/ff477c73ac29066a66334ca6dec7ff7db7c0451a/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs#L195-L205) which is used to get the constructor function reference. But it doesn't work for the "primitive" types. We'll probably need a pattern matching [like this one](https://github.com/fable-compiler/Fable/blob/ff477c73ac29066a66334ca6dec7ff7db7c0451a/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs#L590-L625) to cover all cases.

ncave:
@alfonsogarciacaro @MangelMaxime Making some good progress here, for example compiling the entire `fable-library` to TypeScript is almost there. Let me know your opinion, but IMO we can start planning for the merge of this PR, the functionality is hidden behind a compiler option, so the impact should be fairly low.

MangelMaxime:
@ncave If I understand correctly the goal of this PR is to integrate flow types in the generated JavaScript so people can get intellisense/type checking when using Fable generated JavaScript from their JS/TS code?

How can I test it out?

ncave:
@MangelMaxime Actually, the goal is to produce working Typescript output (and it does, more or less). Yes, we're using Babel flow types, but we can switch to Babel TS types if needed, it's more or less the same output.

Once you pull the `flow` branch, the quickest way (of many) to test is to use the `fable-standalone` test compiler (same as the `fable-compiler-js`, and what the Fable `repl` is using).

The test compiler project's build targets are [here](https://github.com/ncave/Fable/blob/flow/src/fable-standalone/test/bench-compiler/package.json), you can r run it directly on dotnet (quickest), or build it to node.js. Running it on dotnet produces Babel json, so there is a post-build step to convert that to JS/TS.

Debugging is available too, run `dotnet build` first if on dotnet, or the `build-dotnet` target for node.js (target naming can be a bit confusing, suggestions are welcome). Not sure about debugging on gitpod yet.

So just cd into `src\fable-standalone\test\bench-compiler` and `dotnet run ...`, see any of the build scripts in `package.json` as an example. Again, it's the same CLI as `fable-compiler-js`. The new compiler option to add types is `--typed`.

Hopefully it's not too confusing, feel free to ask. The functionality is behind the `--typed` option, so it should be ok to merge and continue improvements as we go, I don't think we need a major version change or anything like that.

ncave:
@MangelMaxime In less words, for example you can `cd src/fable-standalone/test/bench-compiler && npm run build-lib-dotnet` to see the F# part of  `fable-library` compiled to TS (almost, still wip in some places).

MangelMaxime:
@ncave Thank you for all the explanation, I will take a look at this PR.

Indeed if everything is hidden behind a flag, we could release it in order to gather feedback from the community and people who needs to interop with Fable code.

MangelMaxime:
@ncave I am starting to take a look at this PR I think it can run only against windows for now.

The command used in `"postbuild-lib-transform": "rmdir /s /q out-lib\\fable-library && xcopy /s /y ..\\..\\..\\fable-library\\*.ts out-lib\\"` are for windows.

I wanted to make some changes to it, but I still can't push to your repository. Sorry for asking again, but can you please check if you have the "Allow for maintainers" checkbox ticked? (promise I not ask again 😓)

image

If that the case, I will find another way to contribute to the PR but it can be quickly messy because this means, I will have create a fork to your repo and make a PR to your repo ^^. #repoception

Also which version of node are you using? 

If I use node 10 I have token error on the `*` in `transform.js`

If use node 12, I get:

```
internal/modules/cjs/loader.js:1174
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /workspace/Fable/src/fable-standalone/test/bench-compiler/transform.js
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1174:13)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
    at Function.Module._load (internal/modules/cjs/loader.js:901:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
    at internal/main/run_main_module.js:18:47 {
  code: 'ERR_REQUIRE_ESM'
}
```

ncave:
@MangelMaxime
- The `Allow edits from maintainers` box on the PR is checked. I unchecked, then checked it again, just in case. If all else fails, you can always fork my fork and PR the branch, I'll merge it.
- Node.js 13.x can run ES6 modules without a flag. I'm switching more and more to using ES6 modules cause it's cleaner. There is still a runtime performance penalty for running large projects using ES6 modules, but bundling eliminates that.

MangelMaxime:
Good news it seems I am allowed to push to your fork :D

I added Gitpod support and also used platform independent tools for the post process step.

In order, to use the `--typed` option I suppose we still need to update `fable-splitter` and `fable-loader` right?

I also think, we are distributing `fable-library` via `fable-compiler` (in the `bin`) folder. But the problem, is that the distributed version right now is already compiled to `.js` files.

So if we try to run Fable `--typed` against a project we will end up with the user code in TypeScript but the code coming from `fable-library` in JavaScript.

ncave:
> So if we try to run Fable --typed against a project we will end up with the user code in TypeScript but the code coming from fable-library in JavaScript.

Yes, we'll have to build, package and distribute the TS version of `fable-library` as well, and copy to destination, and leave it up to the bundler (rollup, webpack etc.) to remove the extra code. The fable-library already compiles in non-strict TS mode, and is almost there in strict.

I wonder how we can show and run the TS output in the REPL without needing an extra pane, perhaps just show TS and compile to JS and run it without showing the JS.

MangelMaxime:
> I wonder how we can show and run the TS output in the REPL without needing an extra pane, perhaps just show TS and compile to JS and run it without showing the JS.

Yes, I think we can add an option use typed mode and when active we replace the JS tab with TS tab. 

What do you think about renaming `typed` to `typescript`? My reason for that is to make things explicit for developers. Like that they should know, we are generated typescript so they have to use typescript compiler to get the runnable JavaScript.

ncave:
@MangelMaxime
Yes, running it in the browser can be something simple like that: https://jsfiddle.net/4z5tscym/

I picked `--typed` cause it's short and and generic, in theory you could be targeting other compilers (typescript, babel, flow, assemblyscript etc.). But if you like `--typescript` better, sure, it's just a flag.

MangelMaxime:
> you could be targeting other compilers (typescript, babel, flow, assemblyscript etc.).

Do you mean with the current implementation we can target the other compilers? I am asking because it is the reason why I wanted to make it explicit that we are generating typescript as an output.

I didn't want to have `--typed` for now and then we a new target is added have `--typed` for typescript and `--flow` for flow.

I hope my intention is clear ^^

ncave:
@MangelMaxime Yes, we should be able to target other compilers, but the name of the compiler option is not that important to spend much time over it. We can change it to be `--typescript`.

ncave:
Rebased to latest.

ncave:
@MangelMaxime Perhaps we can start a Fable v3.0 beta branch and merge this there.

MangelMaxime:
@ncave Sorry for the delay, I got a bit crushed by work lately. 

I will include it in Fable 2 as you mentioned it is being a flag. I am also planning on making a pass on all Fable files to fix the indentation because some files, use 2 spaces,3 spaces, 4 spaces, or a mix of both.

And, if we put your work in a separate branch it will make it even harder to fix the indentation problems^^. 

Also, with your current work on making us use `fable-spliter` to build test etc. We should be able to add a new test target for `typescript` generation.

MangelMaxime:
@ncave I am not forgetting your PR.

I am working on it while doing so I am also learning how Fable, fable-loader, fable-splitter and co works. So it will probably a bit more time for me to merge this PR.

Sorry for the delay, I am exploring areas I never explored before in Fable and they are not the easiest to understand right now ^^

ncave:
@MangelMaxime No worries, let me know if I can help with something (if I know the answer).

MangelMaxime:
@ncave Thank you.

I think I am starting to understand how things work. I will probably create a diagram to help contribute and document the infra when re-working Fable website.

Right now, I almost have a version of fable-splitter working with your branch:

image

Almost, because I have a bug when copying the fable-library-ts source files. When fixed I will share my work so we can review it.

I also saw a bug in the type generation when generating partial application:

```fs
printfn "%s : %s" "maxime" "mangel"
```

generates:

```ts
(function () {
  const clo1: (arg0: string, arg1: string) => void = toConsole(printf("%s : %s"));
  const clo2: (arg0: string) => void = clo1("maxime");
  clo2("mangel");
})();
```

image

I don't know if that's the correct definition but if I manually change the type to:

```ts
const clo1: (arg0: string) =>  (arg1: string) => void = toConsole(printf("%s : %s"));
```

then the error disappears.

ncave:
@MangelMaxime Yes, partial application is one of the few things that don't work properly yet. Producing types is still an experimental feature and needs to stay behind a flag.

Another one is union/record/value type construction, still debatable if we should stick to object constructor functions, or generate minimal classes and use `new` (only when types are enabled, otherwise it's pretty clear from benchmarking that object constructor functions have performance benefit over creating class instances with `new`).

Still, it should be fine to merge it as is and improve as we go, but I'll leave it up to you.

ncave:
Also, this PR makes some steps towards resolving #1887.

MangelMaxime:
@ncave Ahah you were quicker than me to rebase your branch

And here we go it is time to press the "red button"

Issue #1368, opened 2018-03-19
Some functions not being tail-call optimized


stkb:
Hi, I'm not sure what the current status of TCO is in Fable and if there are known limitations, but I noticed one of my functions not being optimized, so I did some experimenting in the REPL.

My actual function had the form of `fun2` here. I can change it to fun1 instead as a workaround but maybe the case is fixable.

``` fsharp
// Is optimized
let fun1 a b =
    let rec loop n =
        if n = 0 then "done"
        else loop (n - 1)

    loop b

// Not optimized
let fun2 a =
    let rec loop n =
        if n = 0 then "done"
        else loop (n - 1)

    fun b -> loop b

// Is optimized
let fun3 =
    let rec loop n =
        if n = 0 then "done"
        else loop (n - 1)

    fun a -> loop a

// Not optimized
let fun4 =
    let rec loop n =
        if n = 0 then "done"
        else loop (n - 1)

    fun a -> loop 1
```

[...]

PR #1624, opened 2018-11-02
Optimize tailcalls returning units

alfonsogarciacaro:
See [discussion](https://github.com/fable-compiler/Fable/issues/1368#issuecomment-435245905).

@GordonBGood @ncave I think I got it right now. The only concern is we may have now more useless `return` statements as not all unit functions will be actually tail called.

### DIFF
--- a/src/dotnet/Fable.Compiler/AST/AST.Babel.fs
+++ b/src/dotnet/Fable.Compiler/AST/AST.Babel.fs
@@ -189,9 +189,9 @@ type ContinueStatement(?label, ?loc) =
 // type WithStatement
 
 // Control Flow
-type ReturnStatement(?argument, ?loc) =
+type ReturnStatement(argument, ?loc) =
     inherit Statement("ReturnStatement", ?loc = loc)
-    member __.Argument: Expression option = argument
+    member __.Argument: Expression = argument
 
 // type LabeledStatement
 // type BreakStatement
--- a/src/dotnet/Fable.Compiler/AST/AST.Babel.fs
+++ b/src/dotnet/Fable.Compiler/AST/AST.Babel.fs
@@ -189,9 +189,9 @@ type ContinueStatement(?label, ?loc) =
 // type WithStatement
 
 // Control Flow
-type ReturnStatement(argument, ?loc) =
+type ReturnStatement(?argument, ?loc) =
     inherit Statement("ReturnStatement", ?loc = loc)
-    member __.Argument: Expression = argument
+    member __.Argument: Expression option = argument
 
 // type LabeledStatement
 // type BreakStatement
--- a/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/Fable2Babel.fs
@@ -325,13 +325,6 @@ module Util =
             | U2.Case2 e -> BlockStatement [|ReturnStatement e|]
         upcast FunctionExpression(args, body, ?id=id)
 
-    /// Checks if a unit function should actually have a return statement
-    /// At the moment, this happens when there's a tail call opportunity to prevent falling in infinite loops
-    /// But in the future it'd be better if we could check if the function is actually being tail-called
-    /// See https://github.com/fable-compiler/Fable/issues/1368#issuecomment-435245905
-    let willUnitFunctionActuallyReturn (ctx: Context) =
-        Option.isSome ctx.TailCallOpportunity
-
     let optimizeTailCall (com: IBabelCompiler) (ctx: Context) (tc: ITailCallOpportunity) args =
         let rec checkCrossRefs tempVars allArgs = function
             | [] -> tempVars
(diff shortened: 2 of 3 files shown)


alfonsogarciacaro:
I gave a quick try to the repl with the bundle generated by this branch. The good news is that the primes benchmark is almost 10x faster as @GordonBGood predicted :tada: The bad news is this still seems to have problems because the REPL is failing in other examples where there's a `match` on a literal value. I need to investigate.

ncave:
@alfonsogarciacaro Before the last commit, some of the match cases below were broken.
Now, after the last commit, it's looping infinitely on case 1 (TDSuccess), which is missing a return (I think).
```
        let rec accumulateTipsOfDecisionTree accBinds tree  =
            match tree with 
            | TDSwitch (_, edges, dflt, _) -> 
                assert (isNil accBinds)  // No switches under bindings
                for edge in edges do accumulateTipsOfDecisionTree accBinds edge.CaseTree
                match dflt with 
                | None -> ()
                | Some tree -> accumulateTipsOfDecisionTree accBinds tree
            | TDSuccess (es, i) -> 
                branchesToTargets.[i] <- (List.rev accBinds, es) :: branchesToTargets.[i]
            | TDBind (bind, rest) -> 
                accumulateTipsOfDecisionTree (bind::accBinds) rest 
```
compiles to:
```
    const accumulateTipsOfDecisionTree = function accumulateTipsOfDecisionTree(accBinds, tree$$11) {
      accumulateTipsOfDecisionTree: while (true) {
        switch (tree$$11.tag) {
          case 1:
            {
              const i$$14 = tree$$11.fields[1] | 0;
              const es$$15 = tree$$11.fields[0];
              branchesToTargets[i$$14] = (0, _Types.L)([(0, _List.reverse)(accBinds), es$$15], branchesToTargets[i$$14]);
              break;
            }

          case 2:
            {
              const rest$$7 = tree$$11.fields[1];
              const bind$$29 = tree$$11.fields[0];
              const $accBinds$$333 = accBinds;
              accBinds = (0, _Types.L)(bind$$29, $accBinds$$333);
              tree$$11 = rest$$7;
              continue accumulateTipsOfDecisionTree;
              break;
            }

          default:
            {
              const edges$$3 = tree$$11.fields[1];
              const dflt$$9 = tree$$11.fields[2];
              (0, _Seq.iterate)(function (edge) {
                accumulateTipsOfDecisionTree(accBinds, (0, _tast.DecisionTreeCase$$get_CaseTree)(edge));
                return;
              }, edges$$3);

              if (dflt$$9 != null) {
                const tree$$12 = dflt$$9;
                const $accBinds$$334 = accBinds;
                accBinds = $accBinds$$334;
                tree$$11 = tree$$12;
                continue accumulateTipsOfDecisionTree;
              } else {
                return;
              }
            }
        }
      }
    };
```

ncave:
@alfonsogarciacaro After commit `d2ce3be` now there is no break at the end of case 1, so it's incorrectly falling through into case 2.

alfonsogarciacaro:
Thanks a lot for the suggestion @GordonBGood! This is obviously a case where the trees didn't let me see the forest ;) Obviously just adding a `break` at the end of the `while(true)` is enough to prevent falling in an infinite loop. I tested locally and I think it's working this time. I checked the function @ncave pinpointed and the `breaks` are correctly placed now. I'll give a new try after CI is finished but this seems to be ready to merge :)

ncave:
@alfonsogarciacaro Yep, it works now, passes the "REPL bench" test :)
No measurable perf difference (which is good), so ready for merging.

GordonBGood:
> Obviously just adding a break at the end of the while(true) is enough to prevent falling in an infinite loop.

@alfonsogarciacaro, @ncave, merge away, but I'd like to do some testing of all the permutations - ie. I'm sure it works for the `unit` return case with which we are concerned, but has it been fully tested when a non-`unit` value is returned when a simple `break` at the end of the loop may not work?

What do I need to do to test this for myself?  Build the Fable compiler?  Also build repl2? How long will it be before this is released and/or the online repl2 is updated?

If it works for all my variations on the Y-combinator, it likely works for anything :)

alfonsogarciacaro:
@GordonBGood Tests are always welcome! The online repl2 is already updated but if you want to use a local build of Fable CLI that's also quite easy, just follow [these instructions](https://github.com/fable-compiler/Fable#using-your-local-build-in-your-projects). In theory (and if Fable is doing well its job) non-unit functions shouldn't be a problem because branches without tail recursion must always return.

If you can PR your Y-combinator variations to the test suite that'd be very helpful, it's always useful to add complex pieces of code there. Many times this also helps us notice when there's been a regression, even for errors that are not directly asserted in the tests.

GordonBGood:
@alfonsogarciacaro, I see that the online repl2 is already updated to latest and checked a Y-combinator as follows:
``` fsharp
// break race condition with some deferred execution - laziness...
let rec fix f = f <| fun() -> fix f 
// val fix : ((unit -> 'a) -> 'a -> 'a) = <fun>
 
let fac = fix (fun f n i -> if i < 2 then n else f () (bigint i * n) (i - 1)) <| 1I
// val fac : (int -> BigInteger) = <fun>
```
This doesn't produce TCO but then, how could it as the whole point of the Y-combinator is to implement recursion when otherwise it wouldn't be possible; this doesn't TCO in DotNet F# either, so not a problem for you.
> In theory (and if Fable is doing well its job) non-unit functions shouldn't be a problem because branches without tail recursion must always return.

The current implementation seems to work according to your tests and some extended ones from me, with maybe some performance issues but nothing major...

PR #1627, opened 2018-11-04
[WIP] Add Decimal type, fix #1620

### LINKED ISSUE #1620: Unsound decimal behavior
### Description

I know decimals in fable aren't really decimals, but this was quite a surprise for me:

### Repro code

```fsharp

let items = [
    290.8M
    290.8M
    337.12M
    6.08M
   -924.8M
]

if List.sum items = 0M then
	// do something if sum is zero
else
	// do something else

```

[Link to Fable REPL](http://fable.io/repl2/#?code=LYewJgrgNgpgBAERgYwJbAIZQFDdgFzlXxmAGc4BeOAbWzgbgCYBOABgDoAOAWXsdade-BgGZRAdg4BGJn0ZwAbBzbCFAWhZMALNz4BdXAThkIwKnAAyqMvg6nzxUmVwAHAE6oAdvgBmXuAAiAFIAQUCTMzdPH38g4N8Ih1xUX0jzajYeOHwACxgvETgPbz8AwIciCgAvGHcQAEJA7BgoMhgiktjygHVQgBVm7CA&html=DwCwLgtgNgfAUHUBTAhgE3gAm54ElgqbhgAOAtEgI4CuAlgG4C8A5AMID2AdmEj+QBUAnqSQtMAY268erXgA8wAenDQA3JJAoATgGcCTGmABm5ABwssOYLona6pMJO0ddujvYDmdLpl3aJJgAiElJdAC4lJRouUgBrTwA6KQglbVQJMAABAEYANmiINDSMsETSFzQaTLpuRIgfRIArXSCYYCVbe0crbBs7Byc7V3cvHz8A4NCIqJj4pJSSlEzyNA4IXIKaIqWVtYhyyuqwWq56xpa2jq7B+A6QVAwEYAAjDjQhSSgUN2CUUgoUh4KB8SG0bTgOFwaEYmDoaGCSCgDV0IHI-1IQS+P10iORdFR6IBVyUMIYdyUbw+8EQKkgsDgQA&css=BYFwtgNgNAUARgewCYE8AEBvGadoGYIB2IAtAM4CWAXgKYBcaAjAGwAOAHgNwwC+MMAOjABDCoRIBjIiFGEaAJ0zZcSCmVYRhKBngg0uynAHcKSEMAaMADFYCk3XGmA0KAc1CWb9w2gBWAVzIQCjwUSWkaYgYJSJAFB1xhCDdxCjiwMmjY+J9dfRJVeRoJYKJohAh-MEJuPkExVn8QJUdWYSRVQlcGAQAmAFYisAScAmJyanomNgNHEzMLNAGrDhG0EXlXMRJEEBAEMEsh2pggA)

I was expecting that the sum of the following values to be `0`.
The actual value is `1.1368683772161603e-13`.

This is quite unexpected for an F# developer where one of the language slogans is robustness.
The result is totally expected for floats, but it is a bug for decimals.

The following solutions would help avoid imprecise decimal values and gain trust in Fable:

- Disallow decimals. This sounds very radical, but since there is no equivalent replacement for decimals it is the best to give a compiler error before such kind of bugs can happen.
- Same as above, but give a warning.
- Replace decimals with something provided by Fable.

alfonsogarciacaro:
@ncave I've added a new Decimal.js file to fable-core, adapting it from big.js as you suggested. However I'm facing some difficulties. Do you think they are solvable (maybe by adding the missing functions ourselves) or should we use decimal.js-light instead?

- Missing `toNumber`
- Missing `negate`
 - Not sure how to translate [constructors with multiple arguments](https://github.com/fable-compiler/Fable/blob/5c4bf8e340fa2f52a282a2d6ad6a2ecf8c70ff30/src/dotnet/Fable.Compiler/Transforms/Replacements.fs#L1678-L1696).


ncave:
@alfonsogarciacaro I would say implement, since the constructors (and some other methods) will have to be implemented anyway.

ncave:
@alfonsogarciacaro It may be a good idea to do the operator adaptation in a `decimal.ts` file and import the `big.js`, as there is a type definition [here](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/big.js). This way we can update it or swap it for something else if needed.

alfonsogarciacaro:
@ncave That makes sense! The only drawback is we lose tree shaking as most of the functions are attached to the prototype, but for a file of this size it may not be a big problem. Also, the type definitions don't correspond to the latest version, I hope there're not many differences.

ncave:
@alfonsogarciacaro For toNumber and negate, copy from decimal.js-light, as they are trivial. For constructors, just convert to string and use the big.js string constructor (which uses parse).

alfonsogarciacaro:
@ncave We're getting close! I was very concerned about the constructors because I wasn't sure how to build a decimal from its parts and stringify in the REPL. At first [I tried to convert it to bytes](https://github.com/fable-compiler/Fable/pull/1627/commits/65276b0bafa724cfc99d417b6fc40c029ab56427#diff-a34e57b30a53951ff9b99ce4076feaa4R89) using the code from corefx, but then I noticed that decimal byte serialization in corefx and Bridge.net was different. Then [I found this](https://stackoverflow.com/a/5483418) which suggested to use the Mono implementation, so [I ported the code](https://github.com/fable-compiler/Fable/pull/1627/commits/65276b0bafa724cfc99d417b6fc40c029ab56427#diff-a34e57b30a53951ff9b99ce4076feaa4R89) and almost worked. But then I wondered if we couldn't get the string directly from the literal, and digging in your fork I realized [you're converting the string directly to decimal when parsing](https://github.com/ncave/FSharp.Compiler.Service/blob/cbd257823bcdda8289d9c03ba34affca598e3b41/src/fsharp/lex.fsl#L359-L371) and also that [you're not desugaring constants](https://github.com/ncave/FSharp.Compiler.Service/blob/cbd257823bcdda8289d9c03ba34affca598e3b41/src/fsharp/TastOps.fs#L6443-L6458) (which is where the F# compiler converts the decimal literals into `MakeDecimal` calls in the AST). This means we don't need to support `MakeDecimal` in the REPL so we can [just ignore it and the multi-argument constructors](https://github.com/fable-compiler/Fable/pull/1627/files#diff-a34e57b30a53951ff9b99ce4076feaa4R1738) to solve the problem :)

If you have a moment, could you please review the PR? It's passing CI but when testing the REPL and trying to compile a decimal (e.g. `let d = 5M`), I get the following error:

```
Unexpected type DeclaredType (decimal`1,[DeclaredType (MeasureOne,[])]) for literal 5 (System.Object) (test.fs)
```

Other notes:

- I cleaned up a bit the number conversion in the Replacements module. I hope it's ok.
- I removed Util/toString because it was conflicting with decimal printing (decimals were considered plain objects). It was not very consistent anyways and this allowed moving back some functions from Util.ts to Date.ts.
- I've ported the [decimal-byte conversion methods from Bridge.net](https://github.com/fable-compiler/Fable/pull/1627/files#diff-f4bb4ea0c393c07c4187be8ccee6e67eR87) but I haven't found where they should be used (BitConverter is not compatible with decimals).
- BTW, when going through your FCS fork, I also [found this](https://github.com/ncave/FSharp.Compiler.Service/blob/cbd257823bcdda8289d9c03ba34affca598e3b41/fcs/fcs-fable/adapters.fs#L11-L12). Given that in the REPL compilation we're not actually calling `Decimal.GetBits`, should we just remove it or make it throw to prevent silent errors in the future?

Thanks!

alfonsogarciacaro:
Never mind, I fixed the problem in the REPL (I forgot to implement decimal type testing). I'm merging this!

ncave:
@alfonsogarciacaro Awesome!

PR #1631, opened 2018-11-10
Compile fable-core BigInt from F#

alfonsogarciacaro:
Most of the operations in `BigNat` module are being tail-called with this change so hopefully we can get better performance for bigints, besides easier maintenance.

Note we need to change the namespace/module names so Fable don't try to replace the calls when compiling fable-core files.


ncave:
@alfonsogarciacaro Is it possible to reverse the white space changes? I was trying to preserve the original file formatting so they can be updated more easily if they need to change...

alfonsogarciacaro:
@nave Good point! I reverted the whitespace trimming automatically done by my editor and added `get_Zero/get_One` to BigInt.Exports. Let's see if Appveyor passes now 🤞

ncave:
@alfonsogarciacaro Perhaps when [this](https://developers.google.com/web/updates/2018/05/bigint) becomes more widely available, we can switch to it for bigint and int64.

PR #1638, opened 2018-11-14
[WIP] add replacements for System.Numerics Vector3 / Quaternion

0x53A:
This adds replacements for System.Numerics Vector3 and Quaternion.

The reason for porting System.Numerics instead of binding an existing JS lib is that I want to run the same code serverside and clientside.

If you just need something for client side, i would still recomment to use a native js lib.

In the current WIP state this is just the subset I needed for my project, I plan to complete V3 and Q.

I do NOT plan to add everything from the package, i don't really care about Plane or Vector2, or Matrix_x_.

grafik


0x53A:
V3 / Q are the first replaced types to use member fields instead of member properties, so I had to modify FSharp2Fable a little bit.

When you have some time, could you review these changes please?

If that part is OK, then adding the remaining methods is _just simple work_.

src/dotnet/Fable.Compiler/Transforms/FSharp2Fable.fs
```diff
@@ -669,6 +671,31 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) fsExpr =
             return sprintf "Cannot compile ILFieldGet(%A, %s)" ownerTyp fieldName
             |> addErrorAndReturnNull com ctx.InlinePath (makeRangeFrom fsExpr)
 
+    // read a member field (Example: Vector3.X)
+    | BasicPatterns.ILFieldGet(Some callee, calleeType, fieldName)
+       when (Replacements.isILMemberFieldSupported (makeType com ctx.GenericArgs calleeType) fieldName) ->
+        let! callee = transformExprOpt com ctx (Some callee)
```
alfonsogarciacaro: You can just use `let! callee = transformExpr com ctx callee` (same below for ILFieldSet).

alfonsogarciacaro:
Looks good! I just added a minor comment :+1:

0x53A:
after overriding equality, it now fails with this weird error:

image

0x53A:
Ok, splitting the file into two so that there actually_is_ a Vector3.js helped a bit, but equality still doesn't work:

image

0x53A:
It somehow still accessed the old file that I had split, a ``git clean`` later it now fails with

image


So some code expects a file ``Vector3.js``, some other code expects a ``System.Numerics.js`` ( ఠ ͟ʖ ఠ)

0x53A:
It generated this weird code:

__System.Numerics.js__

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.Vector3$reflection = Vector3$reflection;
exports.Vector3$$$$002Ector$$Z7138B98C = Vector3$$$$002Ector$$Z7138B98C;
exports.Vector3$$$$002Ector$$Z69591B0C = Vector3$$$$002Ector$$Z69591B0C;

// ...

var _System = require("./System.Text");

var _Vector = _interopRequireDefault(require("./Vector3"));

var _Util = require("./Util");

// ...

Vector3.prototype.Equals = function (other$$1) {
  const this$$$8 = this;

  if (other$$1 instanceof _Vector.default) {
    const vother = other$$1;
    return Vector3$$Equals$$ZF6770AE(this$$$8, vother);
  } else {
    return false;
  }
};

// ...

Vector3.prototype.CompareTo = function (other$$2) {
  const this$$$10 = this;

  if ((0, _Util.equals)(other$$2, null)) {
    return (0, _String.toFail)((0, _String.printf)("other is null")) | 0;
  } else if (other$$2 instanceof _Vector.default) {
    const vother$$1 = other$$2;
    return (0, _Util.compareArrays)([this$$$10.X, this$$$10.Y, this$$$10.Z], [vother$$1.X, vother$$1.Y, vother$$1.Z]) | 0;
  } else {
    return (0, _String.toFail)((0, _String.printf)("invalid type, expected 'other' to be a Vector, but it is '%O'"))(_Reflection.obj) | 0;
  }
};

// ...

```

Note the ``var _Vector = _interopRequireDefault(require("./Vector3"));`` and the usages of ``_Vector.default``.

so the typechecks are the issue.

alfonsogarciacaro:
Hmm, I'd have to look at the PR more carefully to find the problem but I see you're using `bclType` for the `Vector3` replacement. This was create to automatically find the JS code using the namespace of the type as the file name and then resolving the mangling of the class and member names.

Also, numeric operations are tricky because they must be resolved in different places. You probably need to add the new `BclQuaternion` and `BclVector3` to the following functions in Replacements.fs: `applyOp`, `equals`, `compare`, `compareIf`, `getZero`, `getOne`, `defaultof`

PR #1660, opened 2018-12-06
Number parsing update

ncave:
- Minor number parsing updates to make it work even when it doesn't match the boilerplate active pattern.

Side note: The dichotomy of Js Number vs Bcl numbers probably makes this part of Fable a bit harder to maintain and reason about for new contributors, but it's fine, *this is fine*.


alfonsogarciacaro:
You're right, it's not ideal. But I was trying to find a way to make it clear that some numbers are assimilated to JS `number` while others have their own types. The PR looks good to me but if you think we can represent numbers in the AST in a better way we can discuss it :+1:

Issue #1651, opened 2018-11-27
Move REPL components to this repository


alfonsogarciacaro:
We've talked several times about componentizing the REPL. I've thinking of creating a different repo to host these core components. But given that we're already generating the FCS+Fable JS compilation here and we need some parts for other packages (like the metadata assemblies for fable-compiler-js) it's probably simpler if we generate the components here and pull them from other repos.

So the idea is to put here:

- The REPL web worker that can receive messages to compile, provide tooltips, etc
- The metadata (System.dll and such) + Fable.Repl.Lib assemblies (and precompiled JS files) that must be referenced for the compilation
- The React component for the Monaco editor for F# code.

The fable-compiler/repl should pull these components and provide the rest of the UI and behavior. It will also be the base for other sites that embed the REPL, for example to create Markdown F# tutorials with result evaluation.

It shouldn't be too complicated to create the infrastructure but it's some work, I'll try to start next week, as I'm probably (the only?) person who knows how to fit all the pieces together ;)

cc @ncave @MangelMaxime @whitetigle @tpetricek

MangelMaxime:
I am adding @Krzysztof-Cieslak in the discussion.

nojaf:
>The React component for the Monaco editor for F# code.

Would be interested in this as well, I stole a lot of code to make Monaco work in [AST viewer](http://ratatosk.dynu.net/ast/#?code=Q) 😅. Although I wonder if makes more sense to have a `Fable.Import.Monaco` sort of thing.

viebel:
Please ping me when it's ready. I'd love to integrate Fable with [Klipse](https://github.com/viebel/klipse) to allow F# folks to write blog posts and tutorials with interactive code snippets.

alfonsogarciacaro:
This is complete :) Now assemblies necessary to compile F# projects with Fable are available in the `fable-metadata` npm package, while FCS + Fable compiled to JS are being shipped in `fable-standalone`, together with a web worker that can be embedded in a website.

Fable REPL already depends on these packages. Instructions are more or less like this:

- Copy the contents of `fable-metadata/lib` and `fable-standalone/dist/worker` to wherever you're serving your static files: [example](https://github.com/fable-compiler/repl/blob/d7b4218e1ad869e3879a70b3f30af013b9b50523/build.fsx#L169-L178)
[...]

PR #1666, opened 2018-12-09
Updated metadata2 binaries to 2.2.100

ncave:


ncave:
@alfonsogarciacaro Those are generated with .net core 2.2.100 sdk. No difference in performance, since it's just metadata, but it's nice to be up to date, just in case. Yes, ideally we should only have one metadata folder checked in that is used for both REPL and fable-compiler-js. I pushed a second one to not interfere with your deployments, but we only need one.

alfonsogarciacaro:
Yes, I wanted to do this for #1651 but I need to start updating the documentation and the website or I'll never do it. Priorities... 😄

PR #1681, opened 2018-12-19
Update getting started link in readme

KevinLamb:
Currently the link under the Getting Started heading leads to a doc that has broken links (Migration to Fable 2 link and REPL link). I thought it might be better to link to the Fable website so all the links work normally. Feel free to implement another solution and close this PR.


alfonsogarciacaro:
Ah, you're right! Those links only work from the web, thanks for the fix!

PR #1684, opened 2018-12-20
Pass babel options from splitter config when transforming js file.

nojaf:
I had a problem when importing a JS file that needs a Babel transform plugin in F# when using the splitter.
At first it seemed strange as I did list the plugin in my `splitter.config.js`.

```js
const path = require("path");

function resolve(relativePath) {
    return path.join(__dirname, relativePath);
}

module.exports = {
    entry: resolve("src/MyProj.fsproj"),
    outDir: resolve("out"),
    fable: {
        define: ["DEBUG"]
    },
    babel: {
        plugins: ["@babel/plugin-proposal-class-properties"]
    }
}
```

But I got:

```
C:/Users/nojaf/Projects/ronnies.be/src/RonniesClient/.fable/Fable.ReactGoogleMaps.0.6.0/mapComponent.js(1,1): error BABEL: C:\Users\nojaf\Projects\ronnies.be\src\RonniesClient\.fable\Fable.ReactGoogleMaps.0.6.0\mapComponent.js: Support for the experimental syntax '
classProperties' isn't currently enabled (60:10):

  58 | 
  59 | export class GoogleMapComponent extends React.PureComponent {
> 60 |     refs = {}
     |          ^
  61 |     searchBoxRef = {}
  62 |     state = {
  63 |       isMarkerShown: false,

Add @babel/plugin-proposal-class-properties (https://git.io/vb4SL) to the 'plugins' section of your Babel config to enable transformation.
```

This plugin is needed to deal with a newer JS feature used in [Fable.React.GoogleMaps](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.ReactGoogleMaps/mapComponent.js#L60).

This PR passes the Babel options when transforming a JS file.

### DIFF
--- a/src/js/fable-splitter/src/index.ts
+++ b/src/js/fable-splitter/src/index.ts
@@ -256,7 +256,7 @@ async function getBabelAst(path: string, options: FableSplitterOptions, info: Co
         // return Babel AST from JS file
         path = JAVASCRIPT_EXT.test(path) ? path : path + ".js";
         if (fs.existsSync(path)) {
-            ast = await getBabelAstFromJsFile(path, info);
+            ast = await getBabelAstFromJsFile(path, info, options.babel);
         } else {
             console.log(`fable: Skip missing JS file: ${path}`);
         }
@@ -264,9 +264,10 @@ async function getBabelAst(path: string, options: FableSplitterOptions, info: Co
     return ast;
 }
 
-function getBabelAstFromJsFile(path: string, info: CompilationInfo) {
+function getBabelAstFromJsFile(path: string, info: CompilationInfo, babel: Babel.TransformOptions) {
+    const transformOptions = Object.assign({}, babel, { code: false, ast: true });
     return new Promise<Babel.types.Program | null>((resolve) => {
-        Babel.transformFile(path, { code: false, ast: true }, (error, res) => {
+        Babel.transformFile(path, transformOptions, (error, res) => {
             if (error != null) {
                 const log = `${path}(1,1): error BABEL: ${error.message}`;
                 addLogs({ error: [log] }, info);


alfonsogarciacaro:
Thanks! I'll publish a new release, sorry for the delay!

nojaf:
Thanks!

alfonsogarciacaro:
Unfortunately this is causing an issue (see #1700) because we're [fixing the import paths](https://github.com/nojaf/Fable/blob/5181876b020dfdca07a6b4eaba36009a21b4e6b7/src/js/fable-splitter/src/index.ts#L349-L354) after the first Babel transform, and applying the Babel plugins at that point usually means we don't have the `import` statements anymore.

I'll see if we can reorder the operations to fix this.

alfonsogarciacaro:
I've explored different options but right now I cannot find an elegant and robust solution to this without messing the code. I will revert this change for now and publish a new package.

We'll try to think how to fix this use case later. But for Fable.ReactGoogleMaps it's probably a good idea to distribute only standard JS code (or at least approved proposals). You can precompile the JS files or just edit the code.

nojaf:
Yeah the only reason you need additional babel plugin is for [refs = {}](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.ReactGoogleMaps/mapComponent.js#L60).
Not entirely sure how to write that different, maybe adding a constructor would resolve the issue.
I'll investigate that later.

MangelMaxime:
@nojaf Can't you use `React.createRef` ?

nojaf:
Perhaps, @forki wrote that code a while ago. I had a different implementation before `Fable.ReactGoogleMaps` was a thing, there I was using JSX. So not entirely sure.

alfonsogarciacaro:
If I'm not mistaken, that's only syntax sugar. Writing the following should be equivalent:

```js
class GoogleMapComponent extends React.PureComponent {
    constructor(props) {
        super(props);
        this.refs = {};
        this.searchBoxRef = {};
        this.state = {
          isMarkerShown: false,
        };
    }
    ...
}
```

nojaf:
I was thinking the same. And `this.refs = {};` is not even used. I'll try it out this evening.

nojaf:
https://github.com/fable-compiler/fable-react/pull/134 should resolve the need for this original PR.

drk-mtr:
It's a shame there isn't an easier workaround for this.

I've just spent hooooouurs banging my head against a wall with this, so it's reassuring to know it's not just me doing it wrong but now I still don't know how to get it working. The experimental class properties are required for `react-native-collapsible` import to work.

Can't we just merge this PR and break everyone else... no one really needs tests do they? ;)

PR #1686, opened 2018-12-20
Removed dotnet fable from example

nojaf:


docs/FAQ.md
```diff
@@ -6,7 +6,7 @@ This is the section for quick questions or to miss the holes in the documentatio
 
 #### How can I add a new question?
 
-Make sure the question is not answered already either here or in the documentation. Then just [click here to edit this page in Github](https://github.com/fable-compiler/Fable/blob/master/docs/FAQ.md), add your question to the bottom and create a PR. Maintainers will write an answer and ask you if that solves the issue. If it does, the PR will be merged and the question will be added to the page so it can be referenced by other users.
+Make sure the question is not answered already either here or in the documentation. Then just [click here to edit this page in Github](https://github.com/fable-compiler/Fable/edit/master/docs/FAQ.md), add your question to the bottom and create a PR. Maintainers will write an answer and ask you if that solves the issue. If it does, the PR will be merged and the question will be added to the page so it can be referenced by other users.
```
MangelMaxime: Some people had a 404 error when using the `edit` link.

nojaf: I get 404 on the blob link, hence the change.

MangelMaxime: Seriously...

That's really strange behavior for a public website

nojaf: I think I found out what the problem is, the Fable site somehow points to `https://github.com/fable-compiler/Fable/blob/master/docs/FAQ.html`, that `.html` part leads to 404. Should be `.md`. So FAQ link does indeed not need any changes.

nojaf:
@MangelMaxime  reverted link change btw.

alfonsogarciacaro:
Thanks!

PR #1688, opened 2018-12-27
Add document to jsKeyword

### LINKED ISSUE #1687: Fable produces name collision when using parameter name `document` and using Browser.document in the same scope
### Description

If you have a function that accepts a parameter with name `document` and this function also uses the Browser.document, the produced javascript code is ambiguous and breaks.  please see the fable repl repro for further details.

### Repro code

https://bit.ly/2Q0BHAd

the only difference between myfunc and myfunc' is the name of the parameter.

I would expect an unambiguous js product for myfunc' just as myfunc has. The example code runs to js error: `document.getElementById is not a function` as it tries to call .getElementById on the string parameter and not on the browser.document.

I can imagine two fixes:
1. Fable issues compiler errors when detects such issues
2. This one requires more heuristics as Fable should resolve such issue by renaming the parameter in the produced code so it is not clashing

### Related information

* Fable version (`dotnet fable --version`): 2.0.11
* Operating system: -

MangelMaxime:
Fix #1687

### DIFF
--- a/src/dotnet/Fable.Compiler/Global/Prelude.fs
+++ b/src/dotnet/Fable.Compiler/Global/Prelude.fs
@@ -204,7 +204,7 @@ module Naming =
             // See #258, #1358
             // See https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
             // See https://twitter.com/FableCompiler/status/930725972629913600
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks!

PR #1691, opened 2018-12-31
[fable-compiler-js] Clear cache to ease GC pressure

ncave:
- Added clear cache to ease GC pressure

Now the `fable-compiler-js` can actually compile itself without running out of memory.

Compiling Fable as a benchmark (multiple runs):
- `fable-compiler` (FCS, `dotnet`): 150 sec (185 sec. cold).
- `fable-compiler-js` on `dotnet:` 132 sec (143 sec. cold).
- `fable-compiler-js` on `nodejs:` 284 sec (301 sec. cold).

So the `nodejs` version is roughly `2x` slower than `dotnet` in absolute time, not too bad.
FWIW the last 40 sec. of these times is taken by `Babel`, so without that it's more like `3x`.

Steps to reproduce benchmarks:
1) `cd src/dotnet/Fable.Repl/bench-compiler`
2) `yarn build-fable` to measure `splitter` time.
3) `yarn build-dotnet` to measure `dotnet` time.
4) `yarn build-library` to copy `fable-library` for step 5.
5) `yarn build-node` to measure `nodejs` time.


alfonsogarciacaro:
Awesome, thank you!

PR #1697, opened 2019-01-06
Add undefined

cmeeren:
I think an `undefined` literal should definitely be part of Fable. This simple change defines it in such a way that it's trivial to do "type-safe" comparisons against `undefined` for any type, e.g.:

```f#
type Foo =
  abstract Bar : int with get, set

let x = createEmpty<Foo>

let hasBar = x.Bar <> undefined
```


alfonsogarciacaro:
Good point, thank you!

PR #1706, opened 2019-01-14
Fixed #1705

### LINKED ISSUE #1705: Array.choose always returns empty array for numeric option output types
### Description
Array.choose seems to return an empty array when the output of the choose function is a numeric option type like `Option<int>` or `Option<float>`.

### Repro code

```
    let aaa = [| 1; 2; 3; 4; 5; 6; 7; 8; 9; 10 |]
    let odds_string = Array.choose (fun i -> if i % 2 = 1 then Some(string i) else None) aaa
    let odds_float  = Array.choose (fun i -> if i % 2 = 1 then Some(float i) else None) aaa
    let odds_int    = Array.choose (fun i -> if i % 2 = 1 then Some(int i) else None) aaa
    let odds_noconv = Array.choose (fun i -> if i % 2 = 1 then Some i else None) aaa
    Fable.Import.Browser.console.log("odds_string", odds_string.Length)
    Fable.Import.Browser.console.log("odds_float ", odds_float.Length)
    Fable.Import.Browser.console.log("odds_int   ", odds_int.Length)
    Fable.Import.Browser.console.log("odds_noconv", odds_noconv.Length)
```

Output:
```
odds_string 5
odds_float  0
odds_int     0
odds_noconv 0
```

All lengths should be 5.

* Fable version: 2.1.10
* Operating system: Windows 10 (1809)

ncave:
- Reverted trampoline to @alfonsogarciacaro's simpler version
- Added `RemoveAll`, `FindIndex`, `FindLastIndex` for `ResizeArray`
- Fixed #1705 (`Array.choose`)

### DIFF
--- a/src/dotnet/Fable.Compiler/Transforms/MonadicTrampoline.fs
+++ b/src/dotnet/Fable.Compiler/Transforms/MonadicTrampoline.fs
@@ -1,51 +1,23 @@
-﻿// Source: http://fssnip.net/dK/title/Monadic-Trampoline
-module MonadicTrampoline
+﻿module MonadicTrampoline
 
 #if NETSTANDARD2_0 || !FABLE_COMPILER
-type TrampolineValue<'T> =
-    | DelayValue of Delay<'T>
-    | ReturnValue of Return<'T>
-    | BindValue of IBind<'T>
 
-and ITrampoline<'T> =
-    abstract member Value : TrampolineValue<'T>
-    abstract member Run : unit -> 'T
+type Thunk<'T> =
+    | DelayValue of (unit -> Thunk<'T>)
+    | ReturnValue of 'T
 
-and Delay<'T>(f : unit -> ITrampoline<'T>) =
-    member self.Func = f
-    interface ITrampoline<'T> with
-        member self.Value = DelayValue self
-        member self.Run () = (f ()).Run()
-
-and Return<'T>(x :'T) =
-    member self.Value = x
-    interface ITrampoline<'T> with
-        member self.Value = ReturnValue self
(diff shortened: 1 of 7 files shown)


src/dotnet/Fable.Compiler/Transforms/MonadicTrampoline.fs
```diff
@@ -1,51 +1,23 @@
...
+let rec run = function
+    | DelayValue f -> f () |> run
+    | ReturnValue x -> x
 
 type TrampolineBuilder() =
-    member self.Bind(tr, f) = new Bind<'T, 'R>(tr, f) :> ITrampoline<'R>
-    member self.Delay f = new Delay<_>(f) :> ITrampoline<_>
-    member self.Return a = new Return<_>(a) :> ITrampoline<_>
-    member self.ReturnFrom a = a
+    member __.Bind(thunk, f) = DelayValue (fun () -> run thunk |> f)
```
alfonsogarciacaro: So this is working? I think for some reason it didn't work for me so I had to keep `BindValue` with the boxed parameter. If this implementation is working, we should be able to unify it with the Fable version because the problem was Fable couldn't optimize mutually recursive functions, but now there's only `run`.

alfonsogarciacaro:
Thanks! It's great that you could simplify the trampoline. I'll give it a try locally to make sure the REPL can still build with my mac :+1:

alfonsogarciacaro:
Tested it: working on my mac and also working in the REPL after removing the Fable-specific code in MonadicTrampoline.fs. Great work! 🙌

PR #1707, opened 2019-01-14
Add support for more TimeSpan.ToString options

rfrerebe:
With format being "c", "g", "G", it enables :
- ToString(format)
- ToString(format, CultureInfo.InvariantCulture)

Custom format (ex : hh:mm) are not supported


alfonsogarciacaro:
Awesome, thank you!

Issue #1541, opened 2018-08-27
(dev 2.0) 32 bit integer arithmetic does not truncate


tomcl:
### Description

Arithmetic on int32 or uint32 does not truncate mod 2^32.

This leads to errors in any code that relies on the lowest 32 bits of integer arithmetic remaining correct over repeated multiplications, for example a simple RNG:

```
let rand (s:uint32) = (s*101001u+1234759u)
```

This can be fixed by using `>>> 0` which is turned correctly into `>>> 0` or `>> 0` in the generated JS for unsigned or signed, and truncates.

Worse, lower bits will be wrong even for a single int32 multiplication:

```
0x7FFFFFFFu * 0x7FFFFFFFu
```

### Repro code

```
let z = (0x7FFFFFFFu * 0x7FFFFFFFu) &&& 0x1u
```

### Expected and actual results

z should be 1u.

In REPL 2 **and** CLI z is 0u

Theoretically truncation should be forced after every uint32 or int32 arithmetic operation +, -, `*`, `**`. It might be OK to do this only once per loop, or on function return, though that would leave pathological examples as above it would catch most of the real use cases.

`*` cannot be done in one step using JS floats if it is to deliver accurate lower bits. The correct result can be implemented as something like:

```
let (al , ah) = (a &&& 0xFFFFu), (a >>> 16)
let (bl , bh) = (b &&& 0xFFFFu, (b >>> 16)
let aTimesb = (al*bl +((al*bh+bl*ah) <<< 16)) >>> 0
```
The problem is the same for signed `int32`.

I guess those using FABLE should accept that int32 is not bit accurate, although int64 is? Clear documentation needed for this if so since it is not obvious from the embedding into floats.

I think it would still be worthwhile truncating, so that only multiply causing large results leads to loss of resolution.

### Related information

* Fable version (`dotnet fable --version`): 2.0.0-beta-001
* Operating system:windows

tomcl:
In fact multiply could probably better be implemented:

```
let limitNum = 1u <<< 20

if a < limitNum || b < limitNum then a * b >>> 0 else
   let (al , ah) = (a &&& 0xFFFFu), (a >>> 16)
   let (bl , bh) = (b &&& 0xFFFFu, (b >>> 16)
   (al*bl +((al*bh+bl*ah) <<< 16)) >>> 0
```

the much slower version only gets invoked when both a and b are large. this makes the performance cost of accurate arithmetic significantly smaller.

[...]

PR #1711, opened 2019-01-19
Info on bitwise operations on small integers

chadunit:
Documenting the fact that bitwise operations don't behave as expected for 16 bit and 8 bit types.

For example compare this in .NET vs Fable:

```
printfn "128uy <<< 1 = %i (should be 0 due to truncation to 8 bits)" (128uy <<< 1)
printfn "128uy <<< 8 = %i (should be 128 due to masking shift operand)" (128uy <<< 8)

// .NET
128uy <<< 1 = 0 (should be 0 due to truncation to 8 bits)
128uy <<< 8 = 128 (should be 128 due to masking shift operand)

// Fable
128uy <<< 1 = 256 (should be 0 due to truncation to 8 bits)
128uy <<< 8 = 32768 (should be 128 due to masking shift operand)
```

It's probably possible to fix in the compiler by masking/truncating every operation, but based on #1541 I assumed documentation would suffice until a decision is made regarding a broader effort to address data type overflows.


alfonsogarciacaro:
Thanks!

PR #1714, opened 2019-01-28
Added ParseAndCheckFileInProject

ncave:
@alfonsogarciacaro @MangelMaxime This enables implementing a multi-file tree view pane in the REPL.

### DIFF
--- a/src/dotnet/Fable.Repl/Interfaces.fs
+++ b/src/dotnet/Fable.Repl/Interfaces.fs
@@ -52,6 +52,7 @@ type IFableManager =
     abstract ClearParseCaches: checker: IChecker -> unit
     abstract ParseFSharpScript: checker: IChecker * fileName: string * source: string -> IParseResults
     abstract ParseFSharpProject: checker: IChecker * projectFileName: string * fileNames: string[] * sources: string[] -> IParseResults
+    abstract ParseFSharpFileInProject: checker: IChecker * fileName: string * projectFileName: string * fileNames: string[] * sources: string[] -> IParseResults
     abstract GetParseErrors: parseResults: IParseResults -> Error[]
     abstract GetDeclarationLocation: parseResults: IParseResults * line: int * col: int * lineText: string -> Async<Range option>
     abstract GetToolTipText: parseResults: IParseResults * line: int * col: int * lineText: string -> Async<string[]>
--- a/src/dotnet/Fable.Repl/Main.fs
+++ b/src/dotnet/Fable.Repl/Main.fs
@@ -11,20 +11,19 @@ type CheckerImpl(checker: InteractiveChecker) =
     member __.Checker = checker
     interface IChecker
 
-let mapErrors (checkProjectResults: FSharpCheckProjectResults) =
-    checkProjectResults.Errors
-    |> Array.map (fun er -> {
-        FileName = er.FileName
-        StartLineAlternate = er.StartLineAlternate
-        StartColumn = er.StartColumn
-        EndLineAlternate = er.EndLineAlternate
-        EndColumn = er.EndColumn
-        Message = er.Message
+let mapError (error: FSharpErrorInfo) =
+    {
+        FileName = error.FileName
+        StartLineAlternate = error.StartLineAlternate
+        StartColumn = error.StartColumn
+        EndLineAlternate = error.EndLineAlternate
+        EndColumn = error.EndColumn
+        Message = error.Message
         IsWarning =
-            match er.Severity with
+            match error.Severity with
             | FSharpErrorSeverity.Error -> false
             | FSharpErrorSeverity.Warning -> true
-    })
+    }
 
 type ParseResults (optimizedProject: Lazy<Project>,
                    unoptimizedProject: Lazy<Project>,
@@ -42,7 +41,7 @@ type ParseResults (optimizedProject: Lazy<Project>,
     member __.CheckProjectResults = checkProjectResults
 
     interface IParseResults with
-        member __.Errors = mapErrors checkProjectResults
+        member __.Errors = checkProjectResults.Errors |> Array.map mapError
 
 let inline private tryGetLexerSymbolIslands (sym: Lexer.LexerSymbol) =
   match sym.Text with
@@ -134,12 +133,12 @@ let makeProject projectOptions (projectResults: FSharpCheckProjectResults) optim
     project
 
 let parseFSharpScript (checker: InteractiveChecker) projectFileName fileName source =
-    let parseResults, typeCheckResults, projectResults =
+    let parseResults, checkResults, projectResults =
         checker.ParseAndCheckScript (projectFileName, fileName, source)
     let projectOptions = makeProjOptions projectFileName [| fileName |]
     let optimizedProject = lazy (makeProject projectOptions projectResults true)
     let unoptimizedProject = lazy (makeProject projectOptions projectResults false)
-    ParseResults (optimizedProject, unoptimizedProject, Some parseResults, Some typeCheckResults, projectResults)
+    ParseResults (optimizedProject, unoptimizedProject, Some parseResults, Some checkResults, projectResults)
 
 let parseFSharpProject (checker: InteractiveChecker) projectFileName fileNames sources =
     let projectResults = checker.ParseAndCheckProject (projectFileName, fileNames, sources)
@@ -147,6 +146,13 @@ let parseFSharpProject (checker: InteractiveChecker) projectFileName fileNames s
     let optimizedProject = lazy (makeProject projectOptions projectResults true)
     let unoptimizedProject = lazy (makeProject projectOptions projectResults false)
     ParseResults (optimizedProject, unoptimizedProject, None, None, projectResults)
+
+let parseFSharpFileInProject (checker: InteractiveChecker) fileName projectFileName fileNames sources =
+    let parseResults, checkResultsOpt, projectResults = checker.ParseAndCheckFileInProject (fileName, projectFileName, fileNames, sources)
+    let projectOptions = makeProjOptions projectFileName fileNames
+    let optimizedProject = lazy (makeProject projectOptions projectResults true)
+    let unoptimizedProject = lazy (makeProject projectOptions projectResults false)
+    ParseResults (optimizedProject, unoptimizedProject, Some parseResults, checkResultsOpt, projectResults)
 
 let tooltipToString (el: FSharpToolTipElement<string>): string[] =
     let dataToString (data: FSharpToolTipElementData<string>) =
@@ -253,6 +259,10 @@ let init () =
         member __.ParseFSharpProject(checker, projectFileName, fileNames, sources) =
             let c = checker :?> CheckerImpl
             parseFSharpProject c.Checker projectFileName fileNames sources :> IParseResults
+
+        member __.ParseFSharpFileInProject(checker, fileName, projectFileName, fileNames, sources) =
+            let c = checker :?> CheckerImpl
+            parseFSharpFileInProject c.Checker fileName projectFileName fileNames sources :> IParseResults
 
         member __.GetParseErrors(parseResults:IParseResults) =
             parseResults.Errors
--- a/src/dotnet/Fable.Repl/bench/app.fs
+++ b/src/dotnet/Fable.Repl/bench/app.fs
@@ -31,26 +31,27 @@ let main argv =
         | [|metadataPath; testScriptPath; compiledScriptPath|] -> metadataPath, testScriptPath, compiledScriptPath
         | _ -> metadataPath, testScriptPath, testScriptPath.Replace(".fsx", ".js")
     try
-        let optimized = false
-        // let writeAst = false
+        let optimize = false
         // let fsAstFile = Fable.Path.ChangeExtension(testScriptPath, ".fsharp.ast.txt")
         // let babelAstFile = Fable.Path.ChangeExtension(testScriptPath, ".babel.ast.json")
+        let projectFileName = "project"
+        let fileName = testScriptPath
         let source = readAllText testScriptPath
         let fable = Fable.Repl.Main.init ()
-        let createChecker () = fable.CreateChecker(references, readAllBytes metadataPath, None)
+        let createChecker () = fable.CreateChecker(references, readAllBytes metadataPath, [||], optimize)
         let ms0, checker = measureTime createChecker ()
         printfn "InteractiveChecker created in %d ms" ms0
-        let parseFSharpScript () = fable.ParseFSharpScript(checker, testScriptPath, source)
-        let parseFable (res, fileName) = fable.CompileToBabelAst(fableLibraryDir, res, fileName, optimized)
+        // let parseFSharpScript () = fable.ParseFSharpScript(checker, fileName, source)
+        let parseFSharpScript () = fable.ParseFSharpFileInProject(checker, fileName, projectFileName, [|fileName|], [|source|])
+        let parseFable (res, fileName) = fable.CompileToBabelAst(fableLibraryDir, res, fileName, optimize)
         let bench i =
-            let fileName = testScriptPath
             let ms1, parseRes = measureTime parseFSharpScript ()
             let errors = fable.GetParseErrors parseRes
             errors |> Array.iter (printfn "Error: %A")
             if errors.Length > 0 then failwith "Too many errors."
             let ms2, babelAst = measureTime parseFable (parseRes, fileName)
-            // if i = 1 && writeAst then
-            //     // let fsAstStr = fable.FSharpAstToString(parseRes, fileName, optimized)
+            // if i = 1 then
+            //     // let fsAstStr = fable.FSharpAstToString(parseRes, fileName, optimize)
             //     // printfn "%s Typed AST: %s" fileName fsAstStr
             //     // writeAllText fsAstFile fsAstStr
             //     // printfn "Babel AST: %s" (toJson babelAst)


MangelMaxime:
@ncave Awesome 🎉

And ok message received I will see what I can do about that :)

ncave:
Just to clarify, there are 3 parse and check methods now:

- `ParseFSharpScript: checker * fileName: string * source: string`
This is the original, single file only, left as is for backwards compatibility.
Despite the name, there is no support for `#load` etc.

- `ParseFSharpProject: checker * projectFileName: string * fileNames: string[] * sources: string[]`
Parses and checks the whole project, good for compilers (Fable etc.) but does not retain name resolutions and symbols which are quite memory hungry (so no intellisense etc.). Already parsed files will be cached so subsequent compilations will be faster.

- `ParseFSharpFileInProject: checker * fileName: string * projectFileName: string * fileNames: string[] * sources: string[]`
Parses and checks file in project, will compile and cache all the files up to this one if not already done before, or fetch them from cache. Returns partial project results, i.e. only up to and including the file requested. Retains resolutions and symbols for the file (only), so intellisense etc. works.

alfonsogarciacaro:
Thanks @ncave! Can we add this information as doc comments in the [actual methods](https://github.com/ncave/FSharp.Compiler.Service/blob/a1a9cfa1d6d2e66e62e8f0e8972b8fb3c5d6b987/fcs/fcs-fable/service_shim.fs#L386) and if possible also add some information in the method name itself (e.g. `ParseFSharpProjectWithoutSymbolUses`).

alfonsogarciacaro:
Another question: the `InteractiveChecker` must be recreated whenever the dll references change. Because the Fable CLI can compile several projects at once it means we need to recreate a checker every time we parse a project file in case the references change. Are these references also cached? I mean, if I load two projects and they contain the same references will they be reused (or if I discard the checker and recreate it again because the .fsproj file has changed).

If this doesn't work I guess we need to create our own cache in the `readAllBytes` function.

ncave:
@alfonsogarciacaro Raising the REPL bar from multi-file to multi-project, I see :) I like it!

Currently the `InteractiveChecker` contains the loaded references and the parse and typecheck caches, so if you discard it it's gone. It's not really meant to be reused for different projects, although that might be possible if there is no change in references. We may have to modify the caching to account for that, if that is really a desirable functionality (i.e. add project name to typecheck cache keys).

Technically you can reuse the `InteractiveChecker` without any change if you clear cache between projects.

About reusing references between multiple projects, do you mind describing the use case for it?
The `fable-compiler-js` is merging all project references into one big project, so that's probably not it.
Fable on dotnet should be using FCS as is (even when project-referenced from a ms/visualfsharp fork).

alfonsogarciacaro:
Actually I'm playing with the idea we discussed before about [using your fork directly](https://github.com/fable-compiler/Fable/blob/components/src/Fable.Transforms/Fable.Transforms.fsproj) instead of the FCS package even in the Fable CLI (dotnet). Because the CLI acts as a server and because Webpack can have multiple project entries (or you can even import another .fsproj from F# using Webpack [as the REPL itself does](https://github.com/fable-compiler/repl/blob/e9330537855a1b1d73b5dc1bd03f28b97e524f10/src/App/Main.fs#L23)) we've to support compilation of multiple projects.

Another issue I'm also having is the `InteractiveChecker` expects references to be only file names (without the .dll extension) instead of full paths. But this is tricky if the .dlls come from Msbuild resolution and are located in different folders (like different packages in the Nuget cache. I'm getting an error [here](https://github.com/ncave/FSharp.Compiler.Service/blob/c9be8bd164520c4ace88063d29ca78debdad0ae3/fcs/fcs-fable/service_shim.fs#L221-L222). Is it possible for you to get the File Name (without the parent directory) of the reference after passing the full path?

Anyways, if you think this doesn't make sense, we can go back and make Fable.Cli depend on the FCS package. I was just trying to uniform the different components and if you finally redirect your fork to visualfsharp it could be a way to get enhancements faster. But these are not showstoppers :)

ncave:
@alfonsogarciacaro You can still use my fork, or any other fork, or no fork (directly), if you are project-referencing FCS (by source from Microsoft/Visualfsharp, so you get latest). IMO Fable on dotnet should be project-referencing stock FCS, which is the officially supported tooling, used by everybody else, and guaranteed to work. Since Fable on dotnet is using dotnet anyway, and dotnet build resolutions, there is no reason not to. I don't see any benefit in trying to unify the last step (the CLI), as it inevitably will degrade quality and reliability.

`fable-compiler-js` (and REPL by extension), on the other hand, is in a different boat, as its whole point is to only use javascript. Hense we have to do regex-parsing for project files, dependencies, etc. It's very doable to add integration with `fable-splitter` to have common tooling afterwards, but that's pretty much the extend of it, as we don't want to re-implement MSBuild in javascript.

ncave:
> Another issue I'm also having is the InteractiveChecker expects references to be only file names (without the .dll extension) instead of full paths. 

The "without the .dll" part is a historic artifact, back from the olden days a few years back when FSharp.Core.dll was still distributed with a signature file on the outside. This can be changed.
Yes, we should be stripping the path from the assembly name after reading the bytes but before importing, that should fix it.

alfonsogarciacaro:
I was hoping we could take advantage of the improvements you added, like skipping the symbol uses to reduce the memory pressure, using hashes to cache the file contents, etc. With the "standard" FCS there are several things I don't understand well and I'm not sure if I'm doing things correctly. For example, for the first compilation I use `checker.ParseAndCheckProject` but for the watch compilations I use `checker.ParseAndCheckFileInProject`, but the first time I call this function is much slower than the others. Also, I need to keep a dependency map for each file myself to know which other files I need to recompile when one has changed.

In any case, for now I just added a [workaround for the issue with the dll paths](https://github.com/fable-compiler/Fable/blob/9af2f019272703a33058d714bb8bb3083137a654/src/Fable.Cli/Agent.fs#L21-L31) and it's working! :tada: I will test it tomorrow with some of my projects to see if watch compilation times improve.

ncave:
Some of those improvement are already there in recent changes in MS F# (using hashes in content cache), so they will eventually trickle down to FCS.

alfonsogarciacaro:
Hmm, I'm getting issues with the fileNames on Windows :/ So I guess we'll have to stick with "official" FCS as you said :) I'll try to compile it directly from the visualfsharp repo to see if we can support anonymous records.

ncave:
@alfonsogarciacaro What kind of issues? The paths in file names need to be normalized?
This is probably unrelated, but the assumption is that all names are unique, so for example in `fable-compiler-js` the file names are [prepended](https://github.com/fable-compiler/Fable/blob/master/src/js/fable-compiler-js/src/app.fs#L75) with the project folder.

ncave:
That may not be it the whole story, why is then Global.fs not returning.

alfonsogarciacaro:
@ncave Some more findings/questions:

- The problem with the file paths on Windows seemed to be the drive letter (like `C:`). [After removing it](https://github.com/fable-compiler/Fable/commit/ef6eaf3a037f5e50aa238f70ecda2c4f44623d4f#diff-c3bdc6d463b3cfc091315148234fc9eaR34) now it's working :tada: However, this is causing other problems. For example Webpack watch compilation is not working on Windows, seems to be related.
- Watch compilation using the code from your fork is faster and safer (it always shows whether an error happens in another file because of a change), at least on macOS where it's working.
- How can we pass other compiler options like the warning level or treating warnings as errors?

ncave:
@alfonsogarciacaro We can pass full project options instead, but then the question is do we pass it on InteractiveChecker construction, or on the parsing call, or both? And when it changes, do we want to invalidate all cache, or just some, or none? (there is a scenario that you add a file to a project and you don't want to recompile the whole thing just because of that, not sure how common that is). I guess we can separate the project file list from the rest of the options, that seems more flexible.
I'll fix the assembly name issue when I can, I also need to order properly the errors by file (they are a bit mixed right now).

alfonsogarciacaro:
Right now (in `master`) Fable just starts over when there's any change in the .fsproj so it's acceptable to do that. But I agree it makes sense to separate the file list from the rest of options to help optimization.

Also, now Fable [creates the checker just once](https://github.com/fable-compiler/Fable/blob/e79adeaa45d6df412bd1e074ce29b87a37257e3f/src/dotnet/Fable.Compiler/CLI/Agent.fs#L273) and reuses it every time. [The options are passed when parsing the project](https://github.com/fable-compiler/Fable/blob/e79adeaa45d6df412bd1e074ce29b87a37257e3f/src/dotnet/Fable.Compiler/CLI/Agent.fs#L92-L93) (or a single file). But if it helps with the caches, it's ok for me to associate one checker with each project. Actually, this is how I'm [doing it in the `components` branch](https://github.com/fable-compiler/Fable/blob/3898fc1a4940ffa11aea02fec531e17e7909a8e8/src/Fable.Cli/Agent.fs#L144-L145).

ncave:
> I'll fix the assembly name issue when I can.

I think I fxed it, get latest.

ncave:
> I'll fix the assembly name issue when I can.

I think I fxed it, get latest. You shoudn't need to remove the paths from references anymore, or remove .dll extensions. I tried to make it backwards compatible, if reference has no .dll extension, will add it, if there is, leave it as is.

alfonsogarciacaro:
Thanks @ncave! I will give it a try tomorrow... a new FCS package has been finally published so I need to give anonymous records a try ;)

About the issue with the file names on Windows. I'm totally puzzled. I browsed a bit your code and got suspicious of [implicitIncludeDir](https://github.com/ncave/FSharp.Compiler.Service/blob/ae1931dfdc94b40221a2cca01e964276451c89d0/src/fsharp/CompileOps.fs#L5392) but I tried to set it to different values and it didn't help. It's very weird, if I pass the file names without removing `C:` the returned Implementation Files miss some of them (always the same in every compilation) but I don't get what's the pattern for that 😕

alfonsogarciacaro:
No luck with anonymous records. Seems [the expression tree wasn't updated](https://github.com/Microsoft/visualfsharp/pull/4499#issuecomment-459675270) so we still have to wait :/

I'm trying to get a better grip of the FCS code base. I run `ParseAndCheckProject` multiple times with latest FCS package, but it's still quite slow in watch compilations. @ncave Not sure why but [your cache here](https://github.com/ncave/FSharp.Compiler.Service/blob/ae1931dfdc94b40221a2cca01e964276451c89d0/fcs/fcs-fable/service_shim.fs#L315-L322) it's much faster than the one from IncrementalBuilder.fs. BTW, how expensive is the [`hash source` operation](https://github.com/ncave/FSharp.Compiler.Service/blob/ae1931dfdc94b40221a2cca01e964276451c89d0/fcs/fcs-fable/service_shim.fs#L316)? I say it because Fable (and the REPL) already knows which files have change so it would be easier (and less prone to errors) to pass a structure like the following:

```fsharp
{ FilePath: string; FileContent: string; Version: int }
```

Then you can use `file.FilePath, file.Version` for the cache keys.

ncave:
@alfonsogarciacaro The cost of hashing the source is probably trivial, compared to everything else. IMO we still need to make sure the source didn't change somehow, and the hashing is not a faster way to do that but a less memory intensive way, as it does not keep sources as keys (similar to [this](https://github.com/Microsoft/visualfsharp/pull/5944)).
> Not sure why but your cache here it's much faster than the one from IncrementalBuilder.fs.

Thank you, I was going for simplicity and minimalism there (poor man's IncrementalBuilder :).

alfonsogarciacaro:
Unfortunately needs to keep the sources in memory to avoid to reread all of them from disk every time a new watch compilation is triggered, which kind of defeats the purpose. Maybe the sources could be lazy and we could pass an `invalidate` function from Fable? So only when the files as marked as invalidated, FCS retrieves the source. In the other case it just takes the cached results for that file.

The other alternative would be to pass `null` strings for the files we know that haven't changed, but these would be quite error-prone.

ncave:
@alfonsogarciacaro I'm not sure I understand what is the optimization you're going for, what is that change going to save or simplify vs the existing functionality. Perhaps an example will help.
My gut feeling is we should not be exporting the control over cache invalidation outside of the checker, except for full cache clear. If we know file changed, we call `ParseAndCheckFileInProject`, otherwise we don't and keep the previous typecheck result.

alfonsogarciacaro:
@ncave Nothing in particular. I'm just excited after testing Fable with your code :smile: The best part of web development is hot reloading and with Fable we've come pretty close to that. But using FCS for that it's been always a bit painful and I'm not entirely happy with the current solution: using `ParseAndCheckProject` when only one file has changed is slow, while using `ParseAndCheckFileInProject` forces us to keep a manual map of the dependencies of each file to recompile those too in order to detect errors (which can be slow sometimes too and doesn't always work). Your version of  `ParseAndCheckProject` seems to fix these problems (fantastic job! :clap: :clap), thus my excitement 😸 

I was just thinking aloud if there was still room for improvements, but it's not that important. I'm just trying to understand better what's going on in `service_shim.fs`. For example, you reparse the files whose content has changed, but the `checkCache` is only keyed by file name. How is it possible that you still catch errors happening in other files that haven't changed?

ncave:
@alfonsogarciacaro That's simple, all cached typecheck entries below the modified file are [removed](https://github.com/ncave/FSharp.Compiler.Service/blob/fable/fcs/fcs-fable/service_shim.fs#L308-L313).
It's not perfect, it the sense that if project file lists are changing a lot, there could in theory be some artifacts left over from previous projects. A better solution would be to copy the good entries, remove everything, then add them back.

ncave:
@alfonsogarciacaro I changed the typecheck cache to copy the good entries, remove everything, then add them back.
Also, the project errors are now ordered by file and sorted ascending by line and column.

alfonsogarciacaro:
@ncave Awesome, thank you! Interesting, I really appreciate your explanations. So contrary to my intuition, parsing takes much longer than type checking and the secret of your technique is separating the parsed and type checked caches so you can make the former more efficient 💪

ncave:
@alfonsogarciacaro I think your intuition is right, type check does take longer than parsing. Keeping separate caches for both maximizes reuse, at least in theory, but we can always build a better mouse trap if profiling shows a need for it.

I get your point that further optimizations are possible when you look at the (client + checker) as one system, at the cost of tying them tightly together, so maybe later if potential performance improvements can justify the cost of extra engineering (since FCS-Fable is just a veneer over FCS and is about doing the simplest thing possible).

Or maybe deeper changes in FCS will be needed to get that extra performance out, perhaps by not repeating the whole parsing and type checking on every insignificant change. Maybe somebody will come up with a good idea on how to substantially improve it one day.

alfonsogarciacaro:
@ncave Do you have an approximate idea of how much work would entail to adapt your service_shim to _standard_ FCS? I did [give it a try here](https://github.com/alfonsogarciacaro/visualfsharp/commit/68a230b9cf944dbb8b82cb72275cb5914d92f124) but I'm not sure how to fix the remaining errors.

ncave:
@alfonsogarciacaro Not sure, perhaps not much, but why, is it just for the full project (compiler) options? If yes, those can be added as discussed above, if no, then there are *two* forks to maintain now :)

alfonsogarciacaro:
@ncave After all the questions I ask you deserve making at least one to me 😄 I naively believed that building fcs-fable.proj from your fork without `FABLE_COMPILER` I would basically get the standard FCS behaviour plus your InteractiveChecker. But after giving it a try it doesn't seem to be the case (source list is different, etc). Everything seems to work except for the drive letter issue with paths on Windows and after some debugging it seems to have to do with the way your fork deals with paths, but both Fable and your fork assumes unix/web-style paths everywhere so it's difficult to pinpoint the problem. I was hoping that by switching to _standard_ FCS (while keeping service_shim) this could be fixed.

As you probably have noticed, I've become a bit obsessed about using your _poor man_ incremental builder in .NET too :wink: as it improves the watch compilation experience quite noticeably. All those assumptions ☝️ (single-threaded, single client, non-async, non-cancellable) are true for Fable. We use a MailboxProcessor precisely to make sure the calls to FCS don't happen concurrently to avoid overlaps and prevent issues, but once we've the F# Typed AST (aka Expression Tree) we do process files with Fable in parallel if possible.

ncave:
@alfonsogarciacaro I think I have a better understanding now, thank you. There seems to be indeed an issue when using full paths with drive names (doesn't happen when using relative or absolute paths).
I'll take a look.

ncave:
@alfonsogarciacaro I'm not sure why it's happening, but it's easy to fix, just remove the colon from the file name before sending it to the checker, i.e. `.Replace(":", "")`

ncave:
> I've become a bit obsessed about using your _poor man_ incremental builder in .NET too

@alfonsogarciacaro Alright, I created a [new FCS branch](https://github.com/ncave/FSharp.Compiler.Service/pull/14) just for you to experiment with :). It adds the `InteractiveChecker` in `service_slim.fs` and nothing else (dotnet only). The only (very minor) change vs stock FCS was to enable some constructors in `service.fsi`. There is a `fcs-test` project to test usage.
Hopefully this will enable you to play with different caching strategies, or just have fun with FCS ;)

alfonsogarciacaro:
@ncave It works and it's awesome :) I've published fable-compiler 2.2.0-beta using your second fork. Some of the watch compilations seem to be somewhat faster now but more importantly now users are always notified if a change causes an error in another file (sometimes these went missing if there was no a `import` dependency between the files in JS). Also, the code to deal with the state between watch compilations is simpler now :+1:

I only had issues with the warnings, as the slim service is not filtering them by level. I added a [custom filter that seems to work](https://github.com/fable-compiler/Fable/blob/103ed76c94c03dd3300492dfc5799c9ba42f1acd/src/Fable.Cli/Agent.fs#L242-L258) though it's currently ignoring the options from the .fsproj.

nojaf:
@alfonsogarciacaro did you try the [sample](https://github.com/fable-compiler/fable2-samples) with `2.2.0-beta-002`? It's not working for me on Windows.

alfonsogarciacaro:
@nojaf Sorry, I only tested locally and didn't notice I had to change the assembly name when loaded from the package. Can you please try with 2.2.0-beta-004?

ncave:
> I only had issues with the warnings, as the slim service is not filtering them by level.

It should, the [default level is 3](https://github.com/fsharp/FSharp.Compiler.Service/blob/8beaf7245c9e62068dd62855e0b58cf9767de53c/src/fsharp/ErrorLogger.fs#L664), everything other warn-on or warn-off needs to come via options from the .fsproj (into FSharpProjectOptions).

nojaf:
@alfonsogarciacaro it works a like a charm.
2.2.0-beta-004 is very fast as well.

One little regression I think is

```
  <PropertyGroup>
    <!-- FS0025: Incomplete pattern matches on this expression. -->
    <WarningsAsErrors>25</WarningsAsErrors>
  </PropertyGroup>
```

It creates a warn log instead of an error. 2.1 does show the error.

ncave:
@alfonsogarciacaro @nojaf Seems like compiler warning options are properly applied to the parsing phase, but not to the typecheck phase logger. I'll look into it.

alfonsogarciacaro:
In the FCS service there's a function named `ErrorHelpers.CreateErrorInfos` that is [applied after checking and filters the warnings](https://github.com/fsharp/FSharp.Compiler.Service/blob/379e1c12555b2667618ddad91d2aab706585492a/src/fsharp/service/service.fs#L2805). I tried to call it, but it needs another parameter `PhasedDiagnostic` that seems to be missing in the slim service.

alfonsogarciacaro:
@ncave Can the service_slim deal with signature .fsi files? I'm having issues to compile the REPL (now fable-standalone). It's something similar to what happened before, the implementation files returning by the checker are missing some entries (when there's a .fsi/.fs couple of files, it seems to discard the .fs). For example, [SymbolHelpers.fs file cannot be located in the map of implementation files](https://ci.appveyor.com/project/fable-compiler/fable/builds/22225436#L490). But this time is happening both on Windows and Linux (Travis).

alfonsogarciacaro:
@ncave Seems to be an issue with the deduplication, if I [comment out this line](https://github.com/ncave/FSharp.Compiler.Service/blob/b61c957d9c2b3238c2511766f3ededdb2f5ae561/fcs/FSharp.Compiler.Service/service_slim.fs#L171), then I can build the REPL.

I assume this is not a solution as it will likely cause problems in other situations, but I don't know what's the correct way to call this. The [IncrementalBuilder does something similar to service_slim](https://github.com/fsharp/FSharp.Compiler.Service/blob/8beaf7245c9e62068dd62855e0b58cf9767de53c/src/fsharp/service/IncrementalBuild.fs#L1392) but in service.fs the [return value of moduleNamesDict is discarded](https://github.com/fsharp/FSharp.Compiler.Service/blob/8beaf7245c9e62068dd62855e0b58cf9767de53c/src/fsharp/service/service.fs#L1732). I doubt that's the answer for us, because moduleNamesDict starts just as an empty Map.

ncave:
> Seems to be an issue with the deduplication

@alfonsogarciacaro I'll take a look. Same code works perfectly fine in the other fork (the one with a lot of FABLE_COMPILER magic sprinkled on it :).

ncave:
@alfonsogarciacaro
- I [fixed](https://github.com/ncave/FSharp.Compiler.Service/pull/14/commits/6c4e3c23ef4519f55bf0b3b4b9adde707766cc05) the typecheck logger (to use compiler warning options).
- The .fsi mapping issue might be on your end, since implementation files for files that have both .fs and .fsi will have their file name with .fsi (i.e. in the example above you should be looking for `SymbolHelpers.fsi`, not `SymbolHelpers.fs`).

ncave:
@alfonsogarciacaro
To maintan feature parity in FCS-Fable, I added an `InteractiveChecker` constructor that takes `FSharpProjectOptions` (the old constructor is still there):
```
InteractiveChecker.Create(references: string[], readAllBytes: string -> byte[], defines: string[], optimize: bool)
InteractiveChecker.Create(readAllBytes: string -> byte[], options: FSharpProjectOptions)
```

alfonsogarciacaro:
@ncave Awesome, thanks a lot! Fable standalone is building now in Appveyor for the `components` branch :tada:

> * The .fsi mapping issue might be on your end, since implementation files for files that have both .fs and .fsi will have their file name with .fsi

Interesting. Seems a bit counter intuitive as the implementation files should be the .fs ones (and I wonder why it was working before). In any case I fixed it following your suggestion and it's working now, thanks again!

nojaf:
```
  <PropertyGroup>
    <!-- FS0025: Incomplete pattern matches on this expression. -->
    <WarningsAsErrors>25</WarningsAsErrors>
  </PropertyGroup>
```

works again with **2.2.0-beta-006**, I see the red errors, but I think it works a bit too well.
Problem I now have is that there is an incomplete pattern match in `.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(50,8):`. As this is not my code I can't really fix this as an end-user.

With `2.1` my code compiles, even though Fable.React has an incomplete pattern match.

alfonsogarciacaro:
Hmm, strange. Fable is supposed [to hide warnings from files in `.fable` folder](https://github.com/fable-compiler/Fable/blob/d5a248c85608853259f8febdc1c33cd50a610890/src/Fable.Transforms/State.fs#L94-L98) and I didn't touch that part 🤔

ncave:
@alfonsogarciacaro Is it because those warnings are converted to errors, and Fable is compiling the "pre-compiled" packages with the same compiler options? (`<WarningsAsErrors>25</WarningsAsErrors>`)
Do `.fable` packages contain the project file as well, or just the source?

alfonsogarciacaro:
@ncave Packages in `.fable` folder are not precompiled, I'm only moving them there because Webpack has problems to locate packages from files that are outside the directory containing the `package.json`. The .fsproj files are kept but Fable is only doing a very [simplistic resolution](https://github.com/fable-compiler/Fable/blob/d5a248c85608853259f8febdc1c33cd50a610890/src/Fable.Cli/ProjectCracker.fs#L260-L274) in that case (basically only getting the source files) and they're not compiled independently. Fable merges all sources in a big project before sending it to the F# compiler.

ncave:
@alfonsogarciacaro Thanks for the information.

@nojaf Looks like you have to use compiler options that work for all the components you use, since they all get compiled as one big project.

alfonsogarciacaro:
BTW, one of the reasons to merge everything in a single project is because is not possible (or at least I didn't manage to do it) to inline functions across different projects. Fable first versions had this issue.

nojaf:
So did Fable 2.1 also merge to one project? If not that would explain the change in behaviour if so I think we are missing something here.

alfonsogarciacaro:
Sources are being merged in one single project since Fable 1 if I remember correctly :) Anyways, I think there's something different going on here, because the ReactServer file is [surrounded with `#if !FABLE_COMPILER` directives](https://github.com/fable-compiler/fable-react/blob/b89835014ae16caec51938ad8337010b36b13613/src/Fable.React/Fable.ReactServer.fs#L3) so in principle it should be ignored.

How are you compiling the project? If using fable-splitter you can pass `fable: { cli: { verbose: true } }` in the options to see the full list of arguments passed to the F# compiler.

nojaf:
I'm using fable-loader:

```
fable-compiler 2.2.0-beta-006
C:\temp\fable2-samples\minimal\src> dotnet restore App.fsproj
  Restoring packages for C:\temp\fable2-samples\minimal\src\App.fsproj...
  Generating MSBuild file C:\temp\fable2-samples\minimal\src\obj\App.fsproj.nuget.g.props.
  Restore completed in 234.1 ms for C:\temp\fable2-samples\minimal\src\App.fsproj.


fable-library: C:\temp\fable2-samples\minimal\node_modules\fable-compiler\bin\fable-library
F# PROJECT: ./src/App.fsproj
   --noframework
   --nologo
   --simpleresolution
   --nocopyfsharpcore
   --define:FABLE_COMPILER
   --define:DEBUG
   --optimize-
   --warn:3
   --fullpaths
   --flaterrors
   --target:library
   --targetprofile:netstandard
   --warnaserror:25,76
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.ValueTuple.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Threading.Tasks.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Threading.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Text.RegularExpressions.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Text.Encoding.Extensions.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Text.Encoding.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Runtime.Numerics.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Runtime.Extensions.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Runtime.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Reflection.Primitives.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Reflection.Extensions.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Reflection.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Numerics.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.IO.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Diagnostics.Tracing.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Diagnostics.Debug.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Core.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Console.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.ComponentModel.TypeConverter.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.ComponentModel.Primitives.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.ComponentModel.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Collections.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/System.Collections.Concurrent.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/netstandard.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/netstandard.library/2.0.3/build/netstandard2.0/ref/mscorlib.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/fsharp.core/4.5.2/lib/netstandard1.6/FSharp.Core.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/fable.import.browser/1.3.0/lib/netstandard1.6/Fable.Import.Browser.dll
   -r:C:/Users/SIDFLOV/.nuget/packages/fable.core/2.0.2/lib/netstandard2.0/Fable.Core.dll
   C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Import.React.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.React.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.Isomorphic.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs
   C:/temp/fable2-samples/minimal/.fable/Thoth.Json.2.0.0/Decode.fs
   C:/temp/fable2-samples/minimal/.fable/Thoth.Json.2.0.0/Encode.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/Keyboard.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/BrowserStorage.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/Result.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/Promise.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/Fetch.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/extra/PromiseSeq/Type.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/extra/PromiseSeq/Module.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/extra/PromiseSeq/Extensions.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/extra/Date/Local.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/extra/Date/Format.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.PowerPack.2.0.1/extra/IndexedDB.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.Elmish.2.0.0/prelude.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.Elmish.2.0.0/cmd.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.Elmish.2.0.0/program.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.Elmish.React.2.1.0/common.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.Elmish.React.2.1.0/react.fs
   C:/temp/fable2-samples/minimal/.fable/Fable.Elmish.React.2.1.0/react-native.fs
   C:/temp/fable2-samples/minimal/src/Types.fs
   C:/temp/fable2-samples/minimal/src/State.fs
   C:/temp/fable2-samples/minimal/src/View.fs
   C:/temp/fable2-samples/minimal/src/App.fs
Parsing ./src/App.fsproj...
fable: Compiled src\App.fsproj
fable: Compiled src\App.fs
fable: Compiled src\State.fs
fable: Compiled src\View.fs
fable: Compiled .fable\Fable.Elmish.React.2.1.0\react.fs
fable: Compiled .fable\Fable.Elmish.2.0.0\program.fs
fable: Compiled src\Types.fs
fable: Compiled .fable\Fable.Elmish.2.0.0\prelude.fs
fable: Compiled .fable\Fable.Elmish.2.0.0\cmd.fs
fable: Compiled .fable\Fable.Elmish.React.2.1.0\common.fs
‼ ｢wdm｣: Hash: b6dc7cfc9e15a14de7ce
Version: webpack 4.29.3
Time: 13597ms
Built at: 02/13/2019 11:02:33 AM
    Asset      Size  Chunks             Chunk Names
bundle.js  1.68 MiB    main  [emitted]  main
Entrypoint main = bundle.js
[0] multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj 40 bytes {main} [built]
[./.fable/Fable.Elmish.2.0.0/program.fs] 5.51 KiB {main} [built]
[./.fable/Fable.Elmish.React.2.1.0/react.fs] 2.49 KiB {main} [built]
[./node_modules/ansi-html/index.js] C:/Temp/fable2-samples/minimal/node_modules/ansi-html/index.js 4.16 KiB {main} [built]
[./node_modules/ansi-regex/index.js] C:/Temp/fable2-samples/minimal/node_modules/ansi-regex/index.js 135 bytes {main} [built]
[./node_modules/loglevel/lib/loglevel.js] C:/Temp/fable2-samples/minimal/node_modules/loglevel/lib/loglevel.js 7.68 KiB {main} [built]
[./node_modules/strip-ansi/index.js] C:/Temp/fable2-samples/minimal/node_modules/strip-ansi/index.js 161 bytes {main} [built]
[./node_modules/url/url.js] C:/Temp/fable2-samples/minimal/node_modules/url/url.js 22.8 KiB {main} [built]
[./node_modules/webpack-dev-server/client/index.js?http://localhost:8080] (webpack)-dev-server/client?http://localhost:8080 7.78 KiB {main} [built]
[./node_modules/webpack-dev-server/client/overlay.js] (webpack)-dev-server/client/overlay.js 3.58 KiB {main} [built]
[./node_modules/webpack-dev-server/client/socket.js] (webpack)-dev-server/client/socket.js 1.05 KiB {main} [built]
[./node_modules/webpack/hot sync ^\.\/log$] (webpack)/hot sync nonrecursive ^\.\/log$ 170 bytes {main} [built]
[./node_modules/webpack/hot/emitter.js] (webpack)/hot/emitter.js 75 bytes {main} [built]
[./src/App.fs] 790 bytes {main} [built]
[./src/App.fsproj] 25 bytes {main} [built]
    + 49 hidden modules

WARNING in (webpack)/buildin/global.js
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* C:\Temp\fable2-samples\minimal\node_modules\webpack\buildin\global.js
    Used by 2 module(s), i. e.
    C:\Temp\fable2-samples\minimal\node_modules\node-libs-browser\node_modules\punycode\punycode.js
* C:\temp\fable2-samples\minimal\node_modules\webpack\buildin\global.js
    Used by 1 module(s), i. e.
    C:\temp\fable2-samples\minimal\node_modules\scheduler\cjs\scheduler.development.js
 @ (webpack)/buildin/global.js
 @ C:/Temp/fable2-samples/minimal/node_modules/node-libs-browser/node_modules/punycode/punycode.js
 @ C:/Temp/fable2-samples/minimal/node_modules/url/url.js
 @ (webpack)-dev-server/client?http://localhost:8080
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj
i ｢wdm｣: Compiled with warnings.
i ｢wdm｣: Compiling...
Parsing ./src/App.fsproj...
fable: Compiled src\Types.fs
× ｢wdm｣: Hash: 31881d1fafe5ae7c3ae3
Version: webpack 4.29.3
Time: 88ms
Built at: 02/13/2019 11:02:55 AM
    Asset      Size  Chunks             Chunk Names
bundle.js  1.68 MiB    main  [emitted]  main
Entrypoint main = bundle.js
[./src/Types.fs] 401 bytes {main} [not cacheable] [built] [4 errors]
    + 63 hidden modules

WARNING in (webpack)/buildin/global.js
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* C:\Temp\fable2-samples\minimal\node_modules\webpack\buildin\global.js
    Used by 2 module(s), i. e.
    C:\Temp\fable2-samples\minimal\node_modules\node-libs-browser\node_modules\punycode\punycode.js
* C:\temp\fable2-samples\minimal\node_modules\webpack\buildin\global.js
    Used by 1 module(s), i. e.
    C:\temp\fable2-samples\minimal\node_modules\scheduler\cjs\scheduler.development.js
 @ (webpack)/buildin/global.js
 @ C:/Temp/fable2-samples/minimal/node_modules/node-libs-browser/node_modules/punycode/punycode.js
 @ C:/Temp/fable2-samples/minimal/node_modules/url/url.js
 @ (webpack)-dev-server/client?http://localhost:8080
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(471,8): (471,12) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Custom (_, _)' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(50,8): (50,12) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Custom (_, _)' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(631,8): (631,12) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Custom (_, _)' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/src/State.fs(8,10): (8,13) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Rise' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj
i ｢wdm｣: Failed to compile.
i ｢wdm｣: Compiling...
Parsing ./src/App.fsproj...
fable: Compiled src\Types.fs
× ｢wdm｣: Hash: 1ff4e47b9066a69a8978
Version: webpack 4.29.3
Time: 55ms
Built at: 02/13/2019 11:03:02 AM
    Asset      Size  Chunks             Chunk Names
bundle.js  1.68 MiB    main  [emitted]  main
Entrypoint main = bundle.js
[./src/Types.fs] 393 bytes {main} [not cacheable] [built] [3 errors]
    + 63 hidden modules

WARNING in (webpack)/buildin/global.js
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* C:\Temp\fable2-samples\minimal\node_modules\webpack\buildin\global.js
    Used by 2 module(s), i. e.
    C:\Temp\fable2-samples\minimal\node_modules\node-libs-browser\node_modules\punycode\punycode.js
* C:\temp\fable2-samples\minimal\node_modules\webpack\buildin\global.js
    Used by 1 module(s), i. e.
    C:\temp\fable2-samples\minimal\node_modules\scheduler\cjs\scheduler.development.js
 @ (webpack)/buildin/global.js
 @ C:/Temp/fable2-samples/minimal/node_modules/node-libs-browser/node_modules/punycode/punycode.js
 @ C:/Temp/fable2-samples/minimal/node_modules/url/url.js
 @ (webpack)-dev-server/client?http://localhost:8080
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(471,8): (471,12) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Custom (_, _)' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(50,8): (50,12) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Custom (_, _)' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj

ERROR in ./src/Types.fs
Module Error (from ./node_modules/fable-loader/index.js):
C:/temp/fable2-samples/minimal/.fable/Fable.React.4.0.1/Fable.Helpers.ReactServer.fs(631,8): (631,12) error FSHARP: Incomplete pattern matches on this expression. For example, the value 'Custom (_, _)' may indicate a case not covered by the pattern(s). (code 25)
 @ ./src/View.fs 1:0-66 7:19-22 11:19-22
 @ ./src/App.fs
 @ ./src/App.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/App.fsproj
i ｢wdm｣: Failed to compile.
```

The log above shows starting without errors in my code.
Then I add another case to a DU. An suddenly the errors in Fable.React also pop up.
When I fixed my own code, the Fable React errors remain.

alfonsogarciacaro:
> The log above shows starting without errors in my code. Then I add another case to a DU. And suddenly the errors in Fable.React also pop up.

Ah, so this only happens in watch compilations! You should have started from there ;) It may be that some project options are not honored in watch compilations, let me check...

alfonsogarciacaro:
@nojaf I cannot reproduce...

nojaf:
Minimal repro: https://github.com/nojaf/fable-2.2-beta-repro
If that works for you then I guess not an issue indeed.

alfonsogarciacaro:
Thanks @nojaf! I can reproduce and understand the issue now :+1: Can you please check if the latest beta fixes it?

nojaf:
Beta 007 fixes the issue, many thanks!

PR #1723, opened 2019-02-08
Repo restructure

alfonsogarciacaro:
This is an attempt to make the repository easier to maintain by rearranging the different Fable components/packages. 

The PR also removes several external tools (yarn, fake, paket) to make things simpler, though I'm still missing a few build targets. And we also build now FCS from @ncave's fork which provides several advantages when dealing with watch compilations.


src/fable-library/Fable.Library.fsproj
```diff
@@ -8,9 +8,9 @@
 
   <ItemGroup>
     <Compile Include="Global.fs" />
-    <Compile Include="BigInt/n.fsi" />
+    <!-- <Compile Include="BigInt/n.fsi" /> -->
```
ncave: @alfonsogarciacaro Not sure why those are commented, .fsi files define proper visibility and access.

inosik:
This removes the Rollup plugin from the repository. Does that mean it won't be maintained "officially" anymore?

alfonsogarciacaro:
@inosik Yes, it's been a while since I haven't touched it and it probably doesn't work with latest version. I can transfer it to you and give you publish rights for the npm package if you're interested in maintaining it.

inosik:
> it probably doesn't work with latest version.

It does, requires some manual work, though: https://github.com/inosik/fable-vscode-rollup-sample

> I can transfer it to you and give you publish rights

Can you create a new repository in the fable-compiler organization and make me a collaborator? I can split out the code and push it myself. I'll have to create an NPM account 😄

ncave:
@alfonsogarciacaro Any thoughts on renaming `fable-standalone` to `fable-bundle` or something similar?

alfonsogarciacaro:
@inosik Done! I sent you an invitation :) https://github.com/fable-compiler/rollup-plugin-fable

@ncave Hmm, standalone is not ideal but I picked it after babel-standalone to follow the idea that it allows Fable to be run in other environments (the browers, node.js) but we can change the name if there's a better alternative :+1: Though fable-bundle sounds a bit too broad for me: anything can be a bundle, and the package may also include FCS+Fable compiled also as ES2015/commonjs modules, not only in a minified bundle.

alfonsogarciacaro:
Still some things missing, but let's merge this as it becomes difficult to maintain a branch when it has diverged so much from master 💪

PR #1727, opened 2019-02-10
Replaced Fable.Repl with Fable.Standalone

ncave:
- Replaced Fable.Repl with Fable.Standalone
- Fixed repl issue (`map :> IDictionary` does not work in JS, but `dict :> IDictionary` does)
- Added `fable-compiler-js` target


alfonsogarciacaro:
Ups, I should have checked your PR earlier. I think some of my changes are overlapping yours, let me see if I can clean this.

BTW, my idea was to automatically publish fable-standalone to npm with each successful build and make both the repl worker and fable-compiler-js depend on it but apparently there's no way to get an npm token for a single package and I'm a bit reluctant to put a token there that can publish any of my packages :/

alfonsogarciacaro:
> Fixed repl issue (`map :> IDictionary` does not work in JS, but `dict :> IDictionary` does)

Actually I saw that too and took a while to figure it out. Checking your PR would have saved me time :) This seems to be a bug as it should be possible to cast Map to IDictionary.

ncave:
@alfonsogarciacaro
No worries, with the project changing so fast it's inevitable that we step on each other's toes from time to time.
I'll try to mention your name in the PR next time to attract your attention.
Anyway, I was able to salvage most of it in another PR #1731, so I'm closing this one.

PR #1728, opened 2019-02-11
[WIP] Compile fsx

nojaf:
Using [Dotnet.ProjInfo.Workspace.FCS](https://www.nuget.org/packages/Dotnet.ProjInfo.Workspace.FCS) we can compile fsx files as well.

## Possible flow

Users use paket for dependencies.
`paket load` can create a script to reference everything needed.

```fsharp
#load @".paket\load\netstandard2.0\main.group.fsx"
```

## Current limitations
- the *.fsx script requires a reference to netstandard2.0.

script.fsx:
```fsharp
#r @"C:\Users\nojaf\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\netstandard.dll"
```

- webpack-dev-server keeps compiling in watch mode, might be a loader issue. Not sure atm.

## Todo

- [ ] Fix limitations.
- [ ] Add something to Fable repo for CI, I'm currently using a test project on my machine.
- [ ] Update documentation.


MangelMaxime:
About the limitation to references `netstandard2.0` would something like used in FAKE solves the issue:

```fs
#if !FAKE
#r "Facades/netstandard"
#r "netstandard"
#endif
```

nojaf:
@MangelMaxime yeah tried that, it doesn't load all the necessary references that way.
`#r @"C:\Users\nojaf\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\netstandard.dll"` adds more references than `#netstandard`. //cc @enricosada

alfonsogarciacaro:
In fable-compiler-js we're injecting the necessary references by ourselves, so maybe we could do the same here. The main need for Msbuild is to load other files properly instead of just using regex (for example, if `#load` is surrounded by `#if !FABLE_COMPILER`).

My main concern about scripts is it's currently more cumbersome to add references to packages than generating an SDK-style fsproj.

nojaf:
Hi @alfonsogarciacaro, could you elaborate a bit more, please?

* inject references ourselves: had that idea too but not sure how to find the latest installed netstandard? Any pointers how you did that exactly?
* `more cumbersome`, what do you mean by this? Are you talking about the experience for the end-user?

alfonsogarciacaro:
> * inject references ourselves: you mean adding [references_core](https://github.com/fable-compiler/Fable/blob/master/src/dotnet/Fable.Repl/Metadata.fs#L3) by default?

Yes 😸 

> * `more cumbersome`, what do you mean by this? Are you talking about the experience for the end-user?

Also yep. Anything that requires paths to external folders in your local machine becomes problematic when shared in a repo. And the only way I know to use packages with an .fsx is to use Paket with the generate_fsx_scripts options, which requires some setup and kind-of defeats the purpose of using an .fsx to test things quickly.

enricosada:
> * the *.fsx script requires a reference to netstandard2.0.

@nojaf what's the limitation you see in this issue and `fsxBinder.GetProjectOptionsFromScriptBy` there?

nojaf:
@enricosada I think the `fsxBinder.GetProjectOptionsFromScriptBy` always assumes full framework based on what `netFwInfo.LatestVersion ()` returns.

enricosada:
You can choose the version, but only .NET full is considered. 

There isn’t an FSI for .net core yet, that’s why i didn’t support that

nojaf:
The main ux problem here is probably when you want to `open Fable.Core`. I don't think it is currently present in the `node_modules`.
So you might want to add something like `#r @"node_modules\fable-compiler\bin\Fable.Core.dll"`.

For all other deps, paket would be the only usable option atm. Maybe that can improve in [#5850](https://github.com/Microsoft/visualfsharp/pull/5850).

nojaf:
@enricosada I now check in the [Windows\Microsoft.NET](https://github.com/fable-compiler/Fable/pull/1728/commits/c2a45b871c0f5db27d1cfd7bd75c61459e901e78#diff-494bccf5a68c60580db83ca7d9144382R325) folder to add all the needed references.

As you said that .NET Core isn't a thing yet for fsi it only makes sense to include the references of full framework and Mono. I think I can find the equivalent on Mono as well. However, this seems like a piece of code that has been written before. Would you know about a function that returns the Mono folder with dlls cross-platform?

nojaf:
Discovered [fable-metadata](https://www.npmjs.com/package/fable-metadata) today!
That contains all dlls that might be missing. This seems good enough for now.

It also solves the problem of getting intellisense when using vanilla Fable.
You can do something like:
```fsharp
#r "./node_modules/fable-metadata/lib/Fable.Core.dll"

open Fable.Core
```

ncave:
@nojaf Just keep in mind those binaries in fable-metadata don't contain code, just metadata.

nojaf:
@ncave I believe that is all I need, do you see any harm in this approach @alfonsogarciacaro?

alfonsogarciacaro:
@nojaf Well, the main issue is fable-compiler package doesn't have a dependency to fable-metadata so it's not safe to assume the dlls will actually be there. Maybe you can use some basic types to detect the location of the assemblies:

```
Path.GetDirectoryName(typeof<Int32>.GetTypeInfo().Assembly.Location) // Directory to basic assemblies
typeof<Option<obj>>.GetTypeInfo().Assembly.Location // Location of FSharp.Core
```

nojaf:
@alfonsogarciacaro @enricosada I stole like a lot of code from `Dotnet.ProjInfo.Workspace.FCS` to make this work again. How could I use the Workspace.FCS nuget in combination with the custom fcs dll?

alfonsogarciacaro:
Sorry @nojaf, I missed the notification. What's the difference between Dotnet.ProjInfo and Dotnet.ProjInfo.Workspace.FCS?

src/Fable.Cli/ProjectCracker.fs
```diff
@@ -287,12 +288,446 @@ let getCrackedProjectsFromMainFsproj (projFile: string) =
...
+            |> List.map (function | ScriptOption.Dll dll -> Some dll | _ -> None)
+            |> List.choose id
+            |> fun dllRx ->
+                Array.fold (fun dllRs rf ->
+                    if not (List.exists (fun r -> Path.GetFileNameWithoutExtension(r) = rf) dllRs) then
+                       let dllPath = Path.Combine(netcoreappPath, (sprintf "%s.dll" rf))
+                       dllPath::dllRs
+                    else
+                        dllRs
+                ) dllRx Fable.Standalone.Metadata.references_core
```
alfonsogarciacaro: If you're adding here the basic System assemblies necessary to compile Fable projects, what's the rest of the code doing, just getting custom `#r` references and file `#load`s?

Also, `Metadata.references_core` only includes assembly names not their location in the machine. How is this solved.

alfonsogarciacaro:
TBH, I'm a bit reluctant to add a dependency without knowing well what it does. The README in enricosada/dotnet-proj-info doesn't mention the Dotnet.ProjInfo.Workspace packages and the name seems to suggest they are related to the Workspace functionality in VS Code which doesn't have anything to do with Fable 🤔

enricosada:
Sorry for late reply.

@alfonsogarciacaro 	the `Dotnet.ProjInfo.Workspace` and `Dotnet.ProjInfo.Workspace.FCS` are the helper libraries to parse and fsproj and load all info, to get finally the `FSharpProjectOptions` ready for Fsharp.Compiler.Service.

I have a PR open in FSAC ( https://github.com/fsharp/FsAutoComplete/pull/347 ) to replace all the code to navigate project/get fsc arguments/etc, using that library instead.
The code is pretty much the same, but refactored as library plus some fixes.
will allow to iterate faster and fix bugs, in FSAC was too entangled.
In that repo i can test all scenario, in isolation (like https://github.com/enricosada/dotnet-proj-info/blob/master/test/Dotnet.ProjInfo.Workspace.FCS.Tests/Tests.fs#L199-L412 )
 
In fable will eventually replace this code https://github.com/fable-compiler/Fable/blob/e965930ec30937ccc8d9613684a3fc98ad491196/src/Fable.Cli/ProjectCoreCracker.fs with one or two lines (same feature, but maintained in a library with fixes, and some more like choose `dotnet` path etc).

*NOTE i am dogfooding the library in FSAC atm, it's not ready yet* that's why is not publicly announced yet. i have one or two bugs more to fix.
When done i'll send a PR to fable to replace that code, so will use the same FSAC/vscode will use.

alfonsogarciacaro:
Thanks for explanations @enricosada. That's very good to hear, I'd be very happy to remove the ProjectCoreCracker file because right now I wouldn't know how to edit it in case of need. A couple of notes about the current Fable situation if you want to send a PR :)

- As said above, right now we're not using FCS from Nuget but a custom build from ncave's fork. Adding a dependency which depends on Nuget FCS is going to be problematic.
- Dealing with Nuget, MSBuild, frameworks, etc. is the most frustrating part of Fable development. Because of this, I'm not interested in supporting other scenarios than dotnet SDK format projects. The only thing we need is parsing "modern" .fsproj files (and maybe .fsx) as we do now. Ideally also being able to pass a custom "dotnet" location.
- Fable is not responsible for watching the files (clients like Webpack or fable-splitter do that) so we cannot subscribe to a notifier for file changes.

enricosada:
> * As said above, right now we're not using FCS from Nuget but a custom build from ncave's fork. Adding a dependency which depends on Nuget FCS is going to be problematic.

Final will be like this, that's why already exists `Dotnet.ProjInfo.Workspace` nupkg (without FCS reference, unlike of `Dotnet.ProjInfo.Workspace.FCS` built on top of that).
Fable will use that. I ❤️ fable 😄 , happy to help as needed.

> * Dealing with Nuget, MSBuild, frameworks, etc. is the most frustrating part of Fable development. Because of this, I'm not interested in supporting other scenarios than dotnet SDK format projects. The only thing we need is parsing "modern" .fsproj files (and maybe .fsx) as we do now. Ideally also being able to pass a custom "dotnet" location.

Np, is transparent for the library. if you want to use modern fsproj (best obv), is ok.
Custom `dotnet` is already supported.

> * Fable is not responsible for watching the files (clients like Webpack or fable-splitter do that) so we cannot subscribe to a notifier for file changes.

There is no file changes watcher. When will be added in a later version, will be optional and built on top of base functions.
Atm the notification you see are just .NET event to see progress, completly optional.

nojaf:
Closing in favor of #1780

Issue #1726, opened 2019-02-09
aDateTime.ToString("hh:mm:ss.fff") wrong


jbeeko:
### Description
DateTime.ToString("hh:mm:ss.fff") will not print decimal seconds. This seems to be true for all the decimal flags, f, ff, fff, ffff etc. 

### Repro code, Expected and actual results
In the Fable REPL evaluate `System.DateTime.Now.ToString("hh:mm:ss.fff")` this will print `12:51:34.fff` to the console. 

It should print something like `12:51:34.345`. 

In the .net version of f# you get: 
```
> System.DateTime.Now.ToString("hh:mm:ss.fff");;
val it : string = "10:23:03.347"
```

### Related information
A closely related issue is https://github.com/fable-compiler/Fable/issues/1593. But unlike that issue the issue here could be address without localization concerns. 

* Fable version: 2.0.11
* Mac OSX 10.14.2

alfonsogarciacaro:
Yeah, date printing is not a 1:1 match from .NET, I didn't know you could use `.ff` to print the decimals 😅 Would it be possible for you to send a PR [to fix that](https://github.com/fable-compiler/Fable/blob/33da3818ce42d619e3120eac13a17752c49c2793/src/js/fable-library/Date.ts#L55-L76)?.

alfonsogarciacaro:
Should be fixed by #1730. Thanks @jbeeko!

PR #1730, opened 2019-02-11
Print decimals for aDateTime.ToString("dd:mm:ss.fff")

jbeeko:
Added a case 'f' which means milliseconds can now be printed out by aDateTime.ToString("hh:mm:ss.fff"). Addresses issue [fable-compiler#1726](https://github.com/fable-compiler/Fable/issues/1726) in a very basic way. The .net format string allows you to control how many decimal should be printed but this PR does not distinguish between the following: dd:hh:mm:ss.f, dd:hh:mm:ss.ff , dd:hh:mm:ss.fff, dd:hh:mm:ss.ffff ect. All print out the un-padded millisecond value.


alfonsogarciacaro:
Thanks!

PR #1731, opened 2019-02-12
Replaced Repl with Standalone

ncave:
@alfonsogarciacaro
- Replaced some Fable.Repl with Fable.Standalone


alfonsogarciacaro:
Thanks!

Issue #1733, opened 2019-02-12
Fable + react-hot-loader (without Elmish) generates TypeError


SirUppyPancakes:
### Description

When using React (without Elmish) and `react-hot-loader` (https://github.com/gaearon/react-hot-loader), a TypeError is thrown:
```
TypeError: Property value expected type of string but got null
...
```

### Repro code

Download and extract the following Test project: [Test.zip](https://github.com/fable-compiler/Fable/files/2858101/Test.zip)

Run `yarn`, then `yarn start` in the root, and observe the error.

If you open `webpack.config.js`, you'll find a variable at the top `enableHMR`. If you set this to `false`, you can see that the app works just fine without HMR.

The specific `webpack.config.js` configuration option that causes this is setting the fable loader's babel options to have `plugins: ["react-hot-loader/babel"]`.

I believe this is the correct thing to do, based off of `react-hot-loader`'s recommended way of working with TypeScript: https://github.com/gaearon/react-hot-loader#typescript.

Additionally, for reference, here are all of the configuration options and code that should enable HMR in the project:

`webpack.config.js`:

babel options plugins...
```
if (enableHMR)
    babelOptions.plugins = ["react-hot-loader/babel"]
```
and passing `babelOptions` to `fable-loader` and `babel-loader`.

devServer setup...
```
    devServer: isDevelopment ? {
        contentBase: path.resolve("./app"),
        headers: { "Access-Control-Allow-Origin": "*" },
        port: 8080,
        hot: enableHMR,
        inline: enableHMR
    } : undefined,
```

additional entries...
```
    entry: [path.resolve("./src/Test.fsproj")].concat(
        enableHMR ? [
            "webpack-dev-server/client?http://localhost:8080",
            "webpack/hot/only-dev-server"
        ] : []
```

webpack plugins...
```
    plugins: [
        new CopyWebpackPlugin([{ from: "**/*.html", context: path.resolve("./src") },]),
        new WriteFilePlugin()
    ].concat(enableHMR ? [
        new webpack.HotModuleReplacementPlugin(),
        new webpack.NamedModulesPlugin()
    ] : [])
```

and finally the code...
```
#if DEBUG
[<Import("hot", "react-hot-loader/root")>]
let HotReload: obj -> obj = jsNative
#else
let HotReload (f: obj) = f
#endif
```
```
let CreateTest() =
    let comp() = React.ofType<Test, _, _> (obj()) []
    React.createElement(HotReload comp, obj(), [])

[...]

PR #1735, opened 2019-02-15
Adding babelOptions.filename, since it is needed by other Babel plugins (like react-hot-loader).

SirUppyPancakes:
https://github.com/fable-compiler/Fable/issues/1733

### DIFF
--- a/src/fable-loader/index.js
+++ b/src/fable-loader/index.js
@@ -35,10 +35,12 @@ function getCompiler(webpack, args) {
 
 function transformBabelAst(babelAst, babelOptions, sourceMapOptions, callback) {
     var fsCode = null;
+    var fileName = sourceMapOptions.path.replace(/\\/g, '/');
     if (sourceMapOptions != null) {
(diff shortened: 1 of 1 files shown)


src/fable-loader/index.js
```diff
@@ -35,10 +35,12 @@ function getCompiler(webpack, args) {
 
 function transformBabelAst(babelAst, babelOptions, sourceMapOptions, callback) {
     var fsCode = null;
+    var fileName = sourceMapOptions.path.replace(/\\/g, '/');
```
xdaDaveShaw: This line should probably be after the ` if (sourceMapOptions != null)` check

alfonsogarciacaro: Thanks for pointing that out @xdaDaveShaw! I'll make the change :+1:

SirUppyPancakes: Oh man, can't believe I missed that! The tunnel vision is real 😮

PR #1740, opened 2019-02-17
Add bigint xor operator, bigint byte array conversions, and Random.NextBytes

chadunit:
F# bigint supports ^^^ (but not ~~~, see below) so here it is.

```
> 2I ^^^ 3I;;
val it : System.Numerics.BigInteger = 1 {IsEven = false;
                                         IsOne = true;
                                         IsPowerOfTwo = true;
                                         IsZero = false;
                                         Sign = 1;}

> ~~~2I;;

  ~~~2I;;
  ---^^

error FS0001: The type 'System.Numerics.BigInteger' does not support the operator '~~~'
```

Edit: Also added implementations for:
* BigInteger.ToByteArray()
* BigInteger(byte array) ctor
* Random.NextBytes(byte array)

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -1887,6 +1887,8 @@ let bigints (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: 
     match thisArg, i.CompiledName with
     | None, ".ctor" ->
         match i.SignatureArgTypes with
+        | [Array _] ->
+            Helper.CoreCall("BigInt", "fromByteArray", t, args, i.SignatureArgTypes, ?loc=r) |> Some
         | [Builtin(BclInt64|BclUInt64)] ->
             Helper.CoreCall("BigInt", "fromInt64", t, args, i.SignatureArgTypes, ?loc=r) |> Some
         | _ ->
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -2355,6 +2355,12 @@ let random (_: ICompiler) (ctx: Context) r t (i: CallInfo) (_: Expr option) (arg
         Helper.CoreCall("Util", "randomNext", t, [min; max], [min.Type; max.Type], ?loc=r) |> Some
     | "NextDouble" ->
         Helper.GlobalCall ("Math", t, [], [], memb="random") |> Some
+    | "NextBytes" ->
+        let byteArray =
+            match args with
+            | [b] -> b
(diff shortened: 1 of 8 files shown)


chadunit:
Changed to WIP, I'm working on couple more replacements to fill in if you want to wait a bit and take all at once.

chadunit:
Also do you really need to preserve whitespace in the bigint files? When I saw my huge diff before commit, I checked history and found https://github.com/fable-compiler/Fable/pull/1631, so I did change it to comply, but not sure I agree because:

1. Isn't this based on legacy .NET 2.0 bigint code? I doubt there will be many ongoing changes to the origin code given that they now delegate to System.Numerics.

2. It's easy to ignore whitespace on diffs anyway. (Better than having to change editor settings just to work on specific files.)

alfonsogarciacaro:
Great addition @chadunit, thank you! I'll wait until you remove the WIP status (please ping me when you do it) before merging. But we can also merge and add the additional replacements later, as you prefer.

About the whitespace, I know it's a pain. I also usually use the trimTrailingWhitespace option and get this problem from time to time (I wish there were an option like "trim trailing whitespace only in the lines I edit). However as it was decided in #1631 to keep the whitespace, it'd be great if you could do it too if it's not much of a trouble. The code is not based on .NET (I assume that's written in C#) but in the BigInt implementation in the F# compiler. Basically it's a copy of [these files](https://github.com/Microsoft/visualfsharp/tree/54c22cdea6e069db28d8dd16723e402e364aab44/src/fsharp/FSharp.Core/math), that's why reducing the diffing can make it easier to keep them in sync in the future.

chadunit:
@alfonsogarciacaro ready for review. Cleaned up some helper functions I had for my own stuff. Might not be the best technique or performance but better than missing.

chadunit:
That MS bigint code is behind FX_NO_BIGINT compiler flag and is presumably the [old library for targeting .NET 2 apps](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/numerics.biginteger-structure-%5bfsharp%5d). I don't think it will ever be touched again since the modern replacement is BCL System.Numerics.

chadunit:
Ping @alfonsogarciacaro, any thoughts on this?

alfonsogarciacaro:
Ups, sorry. I missed the last comment, thought this was still WIP. Looks good, let's merge it. Thank you!

ncave:
It may (or may not) be feasible at some point in the future to start using the browser-native `bignit` type, once it gains more support. Access to its storage to implement things like `ToByteArray` can be round-tripped through string hex representation.

PR #1743, opened 2019-02-18
Fixed #1742

### LINKED ISSUE #1742: [fable-standalone] minor issues
Some failing tests when compiled with `fable-standalone`:
```
  isNaN (sqrt System.Double.NaN) |> equal true
  SByte.Parse("-128") |> equal -128y
  Int16.Parse("-32768") |> equal -32768s
```
Compiles (incorrectly) to:
```
  equal(true, isNaN(Math.sqrt(null))); // <-- null is incorrect
  equal(128, (0, _Int.parse)("-128", 511, false, 8)); // <-- 128 is incorrect
  equal(32768, (0, _Int.parse)("-32768", 511, false, 16)); // <-- 32768 is incorrect
```
I'll take a look.

ncave:


src/fable-compiler-js/src/util.js
```diff
@@ -94,7 +107,7 @@ export function copyFolder(from, dest) {
 export function transformAndSaveBabelAst(babelAst, filePath, projDir, outDir, commonjs) {
     try {
         // this solves a weird commonjs issue where some imports are not properly qualified
-        babelAst = JSON.parse(JSON.stringify(babelAst)); // somehow this helps with that
+        babelAst = JSON.parse(serializeToJson(babelAst)); // somehow this helps with that
```
alfonsogarciacaro: Is this still needed? Does it work by removing this line?

ncave: In what sense? This is the fix ...

alfonsogarciacaro: Yes, you're right. I'm trying several things but cannot make it work. I need to compare the ASTs before and after this line to see what's the issue. Thank you!

Issue #1757, opened 2019-02-25
`this` null check in constructors conflicts with some environments (React Native)


forki:
I'm porting my fable 1.x React Native App to Fable 2.x and I'm having problems with Array.groupBy .

        let mappedCityGroups = 
            requests
            |> Array.map (fun x ->
                match x with 
                | MeterReadingRequest x -> x.GetCity()
                | MeterChangeRequest x -> x.GetCity())

        printfn "mapped"

        let cityGroups = 
            requests
            |> Array.groupBy (fun x ->
                match x with 
                | MeterReadingRequest x -> x.GetCity()
                | MeterChangeRequest x -> x.GetCity())

the mapping part works, but during groupby it fails with "Cannot read property 'MoveNext' of undefined"

forki:
ok somthing really weird is happening.

    let ys = [| 1;2;3 |] |> Array.groupBy (fun x -> x = 2)

^^ even this fails.

forki:
![image](https://user-images.githubusercontent.com/57396/53322217-08e92080-38db-11e9-9e68-96c70320a61a.png)

btw: why is fable-splitter generating minified code?

forki:
ok I was able to repro this thing in SAFE-Nightwatch:

![image](https://user-images.githubusercontent.com/57396/53322620-37b3c680-38dc-11e9-8cd7-1c02beffdcfa.png)

compiled to:

![image](https://user-images.githubusercontent.com/57396/53322685-6631a180-38dc-11e9-93fb-3c06710da460.png)


repro: https://github.com/SAFE-Stack/SAFE-Nightwatch/pull/67 (run `build.cmd debug`)

forki:
a bit of the stacktrace:

![screenshot_20190225-090834](https://user-images.githubusercontent.com/57396/53322900-0a1b4d00-38dd-11e9-86f9-042b87fe3c9d.jpeg)

which points to:

![image](https://user-images.githubusercontent.com/57396/53322958-2cad6600-38dd-11e9-9ff6-d205c539dce4.png)

![image](https://user-images.githubusercontent.com/57396/53323159-a6455400-38dd-11e9-95b2-55b8586ac193.png)

forki:
so it fails at:

![image](https://user-images.githubusercontent.com/57396/53325067-b3187680-38e2-11e9-8df9-3e6fce136a93.png)


en is null for some reason

forki:
I found an "fix":

[...]

PR #1754, opened 2019-02-23
Fixes #1738

ncave:

PR #1758, opened 2019-02-25
Try to repro groupBy issue

forki:
Trying to repro https://github.com/fable-compiler/Fable/issues/1757


alfonsogarciacaro:
This is green, maybe #1754 has already fixed the issue?

PR #1760, opened 2019-02-26
Array sprints without braces

forki:
my chrome console output for `[|1.; 2.; 3.; 4.|]` is `1,2,3,4` - which seems wrong


alfonsogarciacaro:
It's difficult to emulate the same print formatting as in .NET, particularly with nested times, and at the same time allowing to override `ToString`. For arrays there are other complications like distinguishing at runtime between typed arrays, resize arrays or tuples.

At the moment I don't have plans to work more on string formatting. But of course contributions are always welcome!

forki:
fair enough

PR #1761, opened 2019-02-26
Added ByrefArgToTupleOptimized

ncave:


alfonsogarciacaro:
Thanks! BTW, will you send a PR to reduce the size for the distributed fable-standalone? I think I've found how to do it, so I can give it a try if you want.

ncave:
@alfonsogarciacaro I will, if you don't beat me to it, I was hitting some snags, it would be a bit easier if the worker is separate `fable-webworker` package. By all means, if you have a good approach go ahead, I can piggy back and adapt to your change in a subsequent step, as we can slim it down even further.

alfonsogarciacaro:
Ok, it should be done now :) At the end I used `importScripts` which made things easier :)

ncave:
@alfonsogarciacaro This is great, thanks!

PR #1762, opened 2019-02-27
System.Decimal.MaxValue can be compiled

forki:
let x = System.Decimal.MaxValue

gives

    error FABLE: Cannot compile ILFieldGet(DeclaredType (Decimal,[]), MaxValue)

But I saw it's defined in Decimal.js


ncave:
Closing this as tests were added in #1763.

PR #1766, opened 2019-02-28
[fable-compiler-js] Make the dist leaner

ncave:
@alfonsogarciacaro 
- Made the `fable-compiler-js` leaner by not distributing copies of `fable-library`, using the one from `fable-standalone` instead. Also, this way the compiler output only produces what is referenced.
- Updated the `fable-metadata` just a bit. This change is needed for the optimized flag.
- Updated the `fable-standalone` just a bit. Building would also capture recent fixes in `fable-library`.

I bumped the version numbers for all three packages, but I'm not sure what is the exact process to publish them (is it a two-step process?). Can you please publish the new versions? Thanks!


alfonsogarciacaro:
Thanks @ncave! Actually, we shouldn't touch the version numbers in package.json (or .fsproj) files. We update the RELEASE_NOTES, then the build script checks if the latest version in RELEASE_NOTES and the project file (package.json or fsproj) has diverged. In that case it runs the build action, bump the version in the project file and publish. For npm packages you should have run `npm login` in your machine (only once) and for the Nuget packages, you should have your key in a NUGET_KEY env var.

I'll publish the new versions anyways :+1:

PR #1773, opened 2019-03-03
Raise type resolve error for non-inlined functions requiring Type.Name

Zaid-Ajaj:
Consider the function:
```fs
let getTypeName<'t> = typeof<'t>.Name
``` 
It compiles fine but when you actually use it, the resolved it type is always `'t`:
```fs
let integerType : string = getTypeName<int> // returns "'t" instead of "Int32" 
```
The compiler should raise an error saying that the function `getTypeName` should be inlined for the generic parameters to correctly get resolved. This PR addresses the issue


alfonsogarciacaro:
Thanks!

PR #1776, opened 2019-03-05
[fable-compiler-js] Fixed duplicate project references

ncave:
@alfonsogarciacaro 
- Renamed `JS.ArrayConstructor` to `JS.Array` (as there are no constructors inside).
- Added `JS.Number` (since `Number.isNaN` is different than global `isNaN`).
- Updated `fcs-fable` to adapt to the new type resolve Fable error.
- Fixed duplicate project references in `fable-compiler-js`.

~~Note: Duplicate project references are possible in `Fable.Cli` too, so some deduping can help there as well, since just filtering out `Fable.Core` works for tests, but is not enough in the general case.~~


alfonsogarciacaro:
> Renamed JS.ArrayConstructor to JS.Array (as there are no constructors inside).

You're right, but the `Constructor` suffix is used in that file to differentiate the static and the instance parts. `.isArray` is a static member of the JS `Array` class (same goes for `Number.isNan`) so I'm not sure if this is the best solution. For now, I wanted to clean the file by removing the JS APIs that have equivalents in F# types. However, if later we want to add the instance part again we can have naming issues.

ts2fable uses the `Static` suffix. We could do that, although it'd be different than the other types in that file.

> Note: Duplicate project references are possible in Fable.Cli too, so some deduping can help there as well, since just filtering out Fable.Core works for tests, but is not enough in the general case.

Duplicated project references should be removed here: https://github.com/fable-compiler/Fable/blob/f34cd5a70fd9424741e254bab8401c57fe439e23/src/Fable.Cli/ProjectCracker.fs#L279-L290

ncave:
@alfonsogarciacaro I renamed `Array` and `Number` back to `ArrayConstructor` and `NumberConstructor`. Although by the same logic (consistency) then `Math` should also be `MathConstructor` (I know it doesn't have instance members, but still). Perhaps the Static suffix is indeed better.

alfonsogarciacaro:
I think it follows the original Typescript bindings. I'd prefer not to make many name changes to avoid breaking the Fable.Browser packages again, but we can discuss later. Thanks for the change!

PR #1786, opened 2019-03-13
Fixed #1784

### LINKED ISSUE #1784: Problem with decimal constructor
Hi @ncave! @MangelMaxime has reported [this test](https://github.com/thoth-org/Thoth.Json/blob/53eb8b17129c08e5feb2d570bf47edd3b0ceb98f/tests/Tests.Json.Encode.fs#L156-L163) is failing. The problem doesn't seem to be with Thoth but with the Decimal constructor. The `0.7833M` literal is compiled to JS as follows (which converts to string as `7833`):

```js
Decimal.fromParts(7833, 0, 0, false, 4)
```

I noticed that reverting fable-compiler to 2.1.11 makes it work. Then the decimal literal becomes:

```js
new Decimal("0.7833")
```

Between those releases there is [this commit](https://github.com/fable-compiler/Fable/pull/1720/commits/e919331b0af4577c52694c27c322aee67bc26adb) that added the decimal constructors. Could you please have a look?

ncave:


alfonsogarciacaro:
Awesome, thank you!

alfonsogarciacaro:
There's a failure when compiling the tests with fable-compiler-js but I don't think it's very important to keep all the zeroes when the value is just 0. I will deactivate that check for now if that's ok with you :+1:

```
  1) Arithmetic
       Decimal.ToString works:
      Error: Expected: 0.00000 - Actual: 0
      + expected - actual
      -0
      +0.00000
```

ncave:
@alfonsogarciacaro I'll see what can be done about this one.

ncave:
@alfonsogarciacaro Fixed in #1788.

PR #1795, opened 2019-03-19
Re-added trailing zeroes tests

ncave:
@alfonsogarciacaro These tests actually work, so they don't need to be disabled anymore.

On a related note, a new version of `fable-standalone` needs to be published, I already updated the release notes.


alfonsogarciacaro:
Ah, sorry. I think this was reverted somehow in the previous PR. Ok, I'll push a new fable-standalone release :+1: I think you have publish rights so you should be able to push it too if I'm taking too long :)

PR #1800, opened 2019-03-23
React fsx sample

nojaf:
This illustrates the difference between using an fsx file instead of fsproj.


alfonsogarciacaro:
Hmm, actually the react-todo-mvc sample is only in this repo because I use it to quickly try new Fable versions when I need something more complex than the quicktest but I should probably remove it. I've been also trying to simplify this repo as much as possible, so I'm a bit reluctant to add 14 new files and thousands of lines for a sample.

IMHO, it would be much more helpful to the community if this goes in its own repo with a post to explain how to use .fsx script files with latest fable-compiler :)

nojaf:
Would https://github.com/fable-compiler/fable2-samples be a better place for this?
How does blogging on fable.io work?

nojaf:
Moved to https://github.com/nojaf/fable-express-sample

PR #1802, opened 2019-03-23
Make dictionary throw key not found when indexing

### LINKED ISSUE #1798: Indexing a dictionary with non-existent key should throw. Returns undefined instead
### Description

Indexing a `Dictionary<'t, 'u>` with non-existent key should throw. Returns `undefined` instead with Fable which gives unexpected results for people coming from .NET

### Repro code

```fs
open System.Collections.Generic

let dict = Dictionary<int, string>()
dict.Add(1, "one")
let value = dict.[0];
printfn "%A" value // undefined
```

Running the code above in dotnet throws an exception: 

Untitled

### Related information

* Fable version: 2.2-beta-018
* Operating system: Windows

xdaDaveShaw:
Map.get in JS returns undefined, whereas in .NET using an indexer
against a key not in the dictionary throws a key not found
exception.

Fixes #1798


alfonsogarciacaro:
Thanks!

PR #1804, opened 2019-03-25
fable-standalone update

ncave:
@alfonsogarciacaro 
A few minor `fable-standalone` and `fable-compiler-js` updates.
- Fixed project dir replacement
- FCS-Fable sync
- Added TryGetValue pattern
- Fixed broken DictionaryTests


alfonsogarciacaro:
Thanks!

src/fable-compiler-js/src/app.fs
```diff
@@ -2,6 +2,18 @@ module Fable.Compiler.App
 
 open Fable.Compiler.Platform
 open Fable.Compiler.ProjectParser
+open Fable.Core.JsInterop
+
+#if TEST_LOCAL
+let [<Emit("__dirname")>] __dirname: string = jsNative
```
alfonsogarciacaro: Compiling fable-compiler-js is failing in AppVeyor with

```
C:/projects/fable/src/fable-compiler-js/src/app.fs(8,6): (8,10) error FSHARP: The type 'Emit' is not defined.
C:/projects/fable/src/fable-compiler-js/src/app.fs(8,46): (8,54) error FSHARP: The value or constructor 'jsNative' is not defined.
```

It looks like an `open Fable.Core` is missing here.

ncave: @alfonsogarciacaro Thanks, added.

ncave:
@alfonsogarciacaro The failing Dictionary test was fixed, it should pass AppVeyor now.

alfonsogarciacaro:
Hmm the Dictionary test is failing now in Appveyor. Expecto seems to give different exception messages in Appveyor and Travis for this test, probably because they've different dotnet SDK versions installed. Maybe we just need to assert that an exception is thrown and ignore the message.

ncave:
@alfonsogarciacaro It was failing before too, see your commits before this PR.

ncave:
@alfonsogarciacaro Somehow it doesn't take that test fix commit, is there a way to check if the fix makes it to the CI?

xdaDaveShaw:
The build on my PR failed on Travis (sorry, I went to bed after submitting the PR).

The tests were passing on my machine.

On travis the thrown exception's message is:
>The given key 'B' was not present in the dictionary.

On appveyor thrown exception's message is:
>The given key was not present in the dictionary.

I don't think there is one standard outcome for both - I can change the tests to not assert on the message - although in JS tests there's not much else to assert on.
Or we need to ensure the runtimes are the same and give consistent outputs.

alfonsogarciacaro:
Life is too short to waste it worrying about dotnet SDK versions (more than I already have to do now I mean) ;) I will change the test to not assert on the message :+1:

PR #1811, opened 2019-04-08
Fixed optimized patterns

ncave:
@alfonsogarciacaro This concludes a long-running effort to make the optimized AST viable.
At last, the `fable-compiler-js` can compile itself successfully with the `optimize-fcs` flag.
There are still a few minor issues using `JsInterop`, but they can be worked around.
- The *good* news is, it works!
- The *not so good* news is, it's not any faster (or smaller) than the non-optimized, as we were hoping.
Still, it does optimize F# code, so it's a nice feature, and might help some other code bases.


alfonsogarciacaro:
Awesome work @ncave! The fact that the optimized version is not much faster/smaller was somehow expected because as far as I could see most of the optimizations had the .NET IL in mind. And inlining in JS is tricky because it can produce sometimes bigger bundles (as we saw when trying to optimize getters).

Anyways, I'm sure this will open new scenarios we can explore to improve performance. It will be interesting for example to see if this makes the REPL run faster in Firefox.

src/Fable.Transforms/FableTransforms.fs
```diff
@@ -275,7 +275,8 @@ module private Transforms =
             | value when ident.IsCompilerGenerated
                     // Don't erase the binding if the compiler-generated ident is a tuple, because the getters
                     // will be erased later (see above) and there's a risk the expression gets totally removed
-                    && not(isTuple ident.Type)
+                    && not (isTuple ident.Type)
+                    && not (ident.Name.StartsWith("patternInput")) // tuple may be split into multiple getters
```
alfonsogarciacaro: The line before is already checking whether the ident is a tuple. Isn't this redundant? I'd prefer not to have too many conditions for the optimizations as this complicates maintenance. Do you have an example of when this condition is necessary (and the one above is not enough)?

src/Fable.Transforms/Replacements.fs
```diff
@@ -1215,7 +1215,7 @@ let operators (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr o
     // KeyValuePair is already compiled as a tuple
     | ("KeyValuePattern"|"Identity"|"Box"|"Unbox"|"ToEnum"), [arg] -> TypeCast(arg, t) |> Some
     // Cast to unit to make sure nothing is returned when wrapped in a lambda, see #1360
-    | "Ignore", [arg]  -> TypeCast(arg, Unit) |> Some
+    | "Ignore", [arg] -> Helper.CoreCall("Util", "ignore", Unit, [arg], ?loc=r) |> Some
```
alfonsogarciacaro: Can we use `($0, void 0)` as I just realized [Fable 1 was doing](https://github.com/fable-compiler/Fable/commit/98fa2c177f40dcaaa9003917041ae4ef97a92888)? I would like to avoid the extra call for all `ignore`s.

ncave: @alfonsogarciacaro Sure, I see you already put that in, I'll open an issue if I see something break.

src/Fable.Transforms/FableTransforms.fs
ncave: @alfonsogarciacaro No, it's not redundant. This catches the case when a tuple got split into multiple separate getters (so isTuple is false, but we still want to recognize it).

alfonsogarciacaro: Ah, ok. Thanks for confirming. Does it only happen in the optimized AST? Would it be possible to add a test case to avoid regressions? I'm concerned that the "patternInput" literal is an implementation detail from FCS that could change at any time.

ncave: @alfonsogarciacaro Yes, it is an implementation detail, but so is everything else in FCS when code is generated or transformed before building the AST. I'll try to add a test when I have some free time, but in the mean time can we leave it there, if it doesn't break anything for you?

alfonsogarciacaro: Sure :)

PR #1818, opened 2019-04-23
Quotation support

krauthaufen:
Hi, I've recently been working on Quotation support for fable in order to support dynamic shader-generation using FShade (https://fshade.org/).

The current implementation supports direct quotations like `<@ fun a -> 1 + a @>`  but currently no reflected definitions. I managed to keep all existing tests running and added very few new ones testing the quotation functionality.

Since Quotations use loads of reflection things I needed to rework fable's reflection mechanisms from ground up and all the Methods/Properties/etc. are now included in the reflection info about each type (in contrast to only Record-Fields and Union-Cases).

I'm aware that this is a rather huge diff and I don't expect you to simply merge it right away. I'd be totally happy to discuss and fix problems. One of the major caveats of the implementation is that it introduces a lot of reflection code that might (in most cases) not be needed.

Just let me know if you're even interested in quotation support...
Cheers


MangelMaxime:
I would say impressive work.

>  One of the major caveats of the implementation is that it introduces a lot of reflection code that might (in most cases) not be needed.

We indeed need to be very careful to not re-increase the bundle size. It was one of the reason for re-writing the compiler and the generated output in Fable 2.

src/fable-library/Fable.Library.fsproj
```diff
@@ -1,30 +1,23 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
-
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <DefineConstants>$(DefineConstants);FABLE_COMPILER</DefineConstants>
-    <DefineConstants>$(DefineConstants);FX_NO_BIGINT</DefineConstants>
```
alfonsogarciacaro: We should keep this

krauthaufen:
Hi, the bundle size is indeed an issue, but in fact it's just meta-information about every type/module and it could be 'compressed' by serializing it in some way...
currently I decided to keep it simple and just emit code that creates all the MehtodInfo/PropertyInfo stuff on demand (with caching implemented directly in Reflection.ts).
A simple way to reduce output size would be to drop reflection information which is never used on type-level, although I'm not entirely sure if webpack does that already.

I'll analyze the resulting bundle sizes using some projects with my fork vs. master and post the results here if you want...

alfonsogarciacaro:
Wow, fantastic work indeed! :clap: :clap: Supporting quotations would be a huge feature for Fable. I was reluctant to do it because, as you say, it implied reworking all the reflection. So I'm really impressed that you did it!

BTW, sorry, I think I broke your branch because of my bad habit of force pushing to master. I did it with the last commit. Not sure how easy is to remove 37d2439342101c22c93d8dc080591daf6177731b from your branch and merge master again :/

Let me go through the PR to understand it better (I just added a simple comment for now) and see if there's any change needed.

As far as I can see, the reflection generation methods are [still detached from the types](https://github.com/fable-compiler/Fable/blob/442a503379964d40129f960ffa23721de5f8e46f/src/Fable.Transforms/Fable2Babel.fs#L1595), which means they should be removed by tree shaking in Webpack production bundles when not used (but when used, it's likely that reflection is somewhat more expensive now). It'd be great if you could show some size comparisons of production bundles using master and your fork, I'll try to do the same.

src/fable-library/Fable.Library.fsproj
krauthaufen: might have overseen that one...
VSCode messed up all the project files and I didn't notice right away ;-)

krauthaufen:
Hey, thanks for the quick response
> BTW, sorry, I think I broke your branch because of my bad habit of force pushing to master

I tend to have the same habit, so no worries there 😁 
It should be resolved and I also added the preprocessor things again to Fable.Library

I'm glad to see that you're interested and I'd be happy to see this merged.

Some things I'd like to mention:
* I might have messed up some things on the way but I'll start cleaning everything up in the next few days. 
* Reflection Invoke/GetValue/etc. calls are a little fragile since I didn't really know what I was doing when I wrote the initial version
* Quotations.fs contains loads of unused code (e.g. the deserialize from the original implementation)


Cheers

krauthaufen:
One other thing: the `compiler-js` target keeps throwing out-of-memory exceptions and i added a 'fix' to build.fsx by simply setting the node-max-memory to 4GB. 
The memory consumption (about 3GB) might also be related to the huge `Quotations.fs`

alfonsogarciacaro:
We did have some memory problems in the past when self-compiling Fable but don't remember exactly how we solved them 😅 I wonder if separating the steps as we did before could help: first compiling to individual JS files, then bundling and optimizing. Maybe @ncave remembers something?

In any case, it'd be possible to remove the unused code in Quotation.fs (or move it to another file) to see if that helps? (Serializing quotations could be interesting though to exchange code between client and server for example)

krauthaufen:
Hi, I already removed most of the unused code from Quotations, simplified some things a little, but is still seems to cause memory problems. Btw. I reworked the MethodInfo/PropertyInfo Invoke/GetValue/SetValue and they should be clean now

krauthaufen:
Hi, I did some (very informal) tests and the bundle size (for typical things) seems to be ~1.5 times larger atm.
I used aardvark-web as a worst case tests where we have lots of generated code (see screenshot) and the reflection API is actually used.

To sum up: I actually expected it to be worse and webpack does a pretty good job removing all the  unused functions. However when a type is used via reflection all its members need to exist in the bundle. 


  | Original [kB] | Quotations [kB] | factor
-- | -- | -- | --
fable2-samples/browser | 22.8 | 35.4 | 1.55
fulma-demo | 136 | 225 | 1.65
aardvark-web (sadly private atm.) | 338 | 679 | 2.01

test

Let me know what you thing about these sizes?
Cheers

MangelMaxime:
Personally, I think it is negating the Fable 2 improvements in term of size or some work done on libraries like Fulma to minimize the bundle size.

I am not saying that I am against the quotation support but we should reduce the bundle size as much as possible. :)

I have a naive question :). Why are `fable2-samples/browser` and `fulma-demo` having all that new code added if they don't use `Quotations`?

Is it adding it by default quotation information to all the types? 

I know that for the reflection we add the type information on place when reflection is used. That is why reflection needed to use `inline` function or member with `IResolver` so Fable can add the reflection data on call place.

alfonsogarciacaro:
Thanks for checking @krauthaufen! As you say, it's not too much but also as Maxime notes, we should avoid increasing the bundle x1.5 for everybody that doesn't need the feature. I'll try to compiled the projects myself and see how the generated JS looks to see if we can help tree shaking more. For now, we should probably have an opt-in compilation switch for the feature or, if this is too difficult, keep a branch and release it as an alternative package (like `fable-compiler-quotations`) so people interested in the feature can give it a try.

krauthaufen:
> Is it adding it by default quotation information to all the types?
It's just the new reflection infos (Methods/Properties, etc.)  added to the projects.
And also the reflection library itself is a little larger.

I totally agree, that a compiler-flag would be nice. When turned off we could simply skip MethodInfos/PropertyInfos/etc. (except for record/union) things.
I'll see if I can get something like that working.

krauthaufen:
Hi, I just added an environment variable `FABLE_QUOTATIONS` which can be set in fable.config (webpack, splitter, etc.) which basically does two things:
1. it disables the generation of MethodInfo/PropertyInfo (except for record/union types)
2. it raises an error whenever a quotation is encountered, stating that one should enable it

I quickly tested Fulma with it and the bundle size is now back to 152kB (master: 136kB, with new reflection: 225kB).

So the new Reflection code increased the bundle size a little, but not too much i think.
It would be great to see the new (more flexible) reflection implementation in master since maintaining two different ones might be painful.
Cheers

alfonsogarciacaro:
Great! Yes, I agree maintaining two reflection implementations is too much work. I think we're very close but I'll like to make some experimentation before merging with master. So I've created a `quotations` branch and merged the PR there, I hope it's OK with you. I've sent you an invitation to give write rights to the repository in case you want to keep working on that branch.

BTW, would you be interesting in talking about your work in [FableConf](https://fable.io/fableconf). Would it be ok to contact you privately? If so, how?

krauthaufen:
hey, great honor to get push access to fable 😁 ...
I'd totally be interested in attending FableConf. Of course you can contact me privately via mail/gitter/signal. Whatever you prefer...

krauthaufen:
@alfonsogarciacaro @MangelMaxime thanks a lot for your help and input. 
I finally have some results I'd like to share. 

After porting parts of our [rendering engine](https://github.com/aardvark-platform/aardvark.rendering) to fable, implementing the quotations and tampering your REPL i finally got [FShade](https://github.com/krauthaufen/FShade) running in a web-browser 😍 

Here's a little video showing how it works: https://github.com/aardvark-platform/aardvark.rendering

It can also be accessed [here](https://fshade.org/repl/) but there are still many fragile parts, so don't expect too much...

MangelMaxime:
@krauthaufen This is really cool and amazing 🎉🎉🎉 

And it's running really smoothly :)

The only "bug" I have is I can't scroll in the code and need to use the scrollbar. I guess the scroll event is captured by the demo view.

krauthaufen:
Great to hear that you like it.
The whole thing is quite sketchy atm. but I'll continue to work on it...
Cheers

PR #1821, opened 2019-04-24
TimeSpan comparison works

forki:
I have a strange error in production where this is not working

### DIFF
--- a/tests/Main/ArithmeticTests.fs
+++ b/tests/Main/ArithmeticTests.fs
@@ -100,6 +100,12 @@ let tests =
         79228162514264337593543950335M |> equal System.Decimal.MaxValue
         -79228162514264337593543950335M |> equal System.Decimal.MinValue
 
+    testCase "Decimal devision works" <| fun () ->
+        let a = decimal 0.00001
+        let b = 1000.M
+        let c = decimal 0.00000001
+        a / b |> equal c
+
     testCase "Decimal.ToString works" <| fun () ->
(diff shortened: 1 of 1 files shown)


forki:
Ok I was able to reduce it to smallest repro. It seems decimal devision is broken and it returns zero .

alfonsogarciacaro:
Interestingly, this is working in the REPL. Could it be related to this change @ncave? https://github.com/fable-compiler/Fable/commit/f223ad5f9ade0cc379be1f80bae9d9c4915835fd#diff-230efdea7ee2b4c21222fbafd61ffb9c

forki:
converting things to double before division and converting back afterwards works as worksaround

ncave:
@alfonsogarciacaro @forki
Yes, the `big.js` package we use for decimals has a few problems identifying zeros and comparing same numbers with different representation.

I'll try to fix those two particular issues, but short of rewriting the whole thing (even though it's not that big), I can't say with confidence that something else will not pop up down the road. (I know, it's my NIH syndrome kicking in).

I hope we can eventually switch to something more robust and performant, perhaps based on `bigint` once that is supported enough.

forki:
Ok so in other words: it would be a good idea if I contribute more tests. That's something I can do easily.

alfonsogarciacaro:
Fixed by @ncave released in 2.3.4, thanks again!

forki:
Oh nice. Thanks

PR #1827, opened 2019-04-27
PublishUtil.fs: Fix the path of temp directory when releasing a nuget

Zaid-Ajaj:
Fix the output directory of publish utilities because in the arguments only the constant `temp` is used which is then resolved to `{rootDir}/temp` instead of `{rootDir}/{projDir}/temp`


alfonsogarciacaro:
Thanks!

PR #1835, opened 2019-05-12
fable-splitter: fix env package name in config example

stroborobo:
This package has moved and the key changed in Babel 7

Otherwise this would fail with a message like:

```
fable: Error transforming Babel AST Error: Cannot find module 'babel-preset-env' from '...'
- Did you mean "@babel/env"?
[...]
```


alfonsogarciacaro:
I'm really confused by this. Sometimes using `env` only seems to work. But anyways it's better to be on the safe side, thanks!

PR #1838, opened 2019-05-20
fcs-fable update

ncave:
@alfonsogarciacaro

Updated `fcs-fable` with the `dotnet/fsharp` [fable](https://github.com/ncave/fsharp/tree/fable) fork.

Bleeding edge, top of the food chain ... 😀


alfonsogarciacaro:
Thanks! So from now we will be forking directly the fsharp compiler instead of FCS. Should we change terminology and start talking about fsharp (compiler) instead of "fcs"? For example, changing the "fcs-fable" folder to fsharp-fable, and the like.

ncave:
@alfonsogarciacaro It's still [FCS](https://github.com/ncave/fsharp/tree/fable/fcs/FSharp.Compiler.Service) and [eventually](https://github.com/fsharp/fslang-design/pull/359) the source for packaged FCS, even though it's in the `fsharp` repo. I don't think we need to rename it, but if you want to, go ahead.

PR #1851, opened 2019-06-16
Updated fable-compiler-js references

ncave:
@alfonsogarciacaro Updated `fable-compiler-js` to use the latest `fable-metadata` and `fable-standalone`.
If you can publish it it will be great (not sure if CI does that automatically).

### DIFF
--- a/src/fable-compiler-js/RELEASE_NOTES.md
+++ b/src/fable-compiler-js/RELEASE_NOTES.md
@@ -1,6 +1,12 @@
+### 1.0.4
+
+* fable-metadata 1.1.2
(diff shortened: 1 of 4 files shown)


alfonsogarciacaro:
No, CI doesn't do it. I was looking into that but it doesn't seem to be possible to get an npm token for a specific package and I wasn't afraid of putting a token in the CI server that gives access to all my npm packages. Anyways, you've publish rights for fable-compiler-js, so it's just a matter of executing `npm run publish` after you've updated the RELEASE_NOTES :)

PR #1867, opened 2019-07-29
Fix negative timespan + milliseconds value when parsing a string

MangelMaxime:
Related to https://github.com/MangelMaxime/Thoth/issues/152

### DIFF
--- a/src/fable-library/TimeSpan.ts
+++ b/src/fable-library/TimeSpan.ts
@@ -128,6 +128,7 @@ export function parse(str: string) {
   }
   if (firstColon > 0) { // process time part
     // tslint:disable-next-line:max-line-length
+    // WIP: (-?)(((\d+)\.)?([0-9]|0[0-9]|1[0-9]|2[0-3]):(\d+)(:\d+(\.\d{1,7})?)?|\d+(?:(?!\.)))
     const r = /^(-?)((\d+)\.)?(?:0*)([0-9]|0[0-9]|1[0-9]|2[0-3]):(?:0*)([0-5][0-9]|[0-9])(:(?:0*)([0-5][0-9]|[0-9]))?\.?(\d+)?$/.exec(str);
     if (r != null && r[4] != null && r[5] != null) {
       let d = 0;
@@ -143,7 +144,7 @@ export function parse(str: string) {
         s = +r[7];
       }
       if (r[8] != null) {
-        // Depending on the number of decimals passed, we needed to adapt the numbers
+        // Depending on the number of decimals passed, we need to adapt the numbers
         switch (r[8].length) {
             case 1:
                 ms = +r[8] * 100;
@@ -168,7 +169,6 @@ export function parse(str: string) {
                 break;
             default:
                 throw new Error("String was not recognized as a valid TimeSpan.");
-                break;
         }
       }
       return sign * create(d, h, m, s, ms);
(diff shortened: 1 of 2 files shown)


src/fable-library/TimeSpan.ts
```diff
@@ -76,7 +76,8 @@ export function totalMinutes(ts: number) {
 }
 
 export function totalSeconds(ts: number) {
-  return ts / 1000;
+  // Round needed to replicated .Net behaviour
+  return Math.round(ts / 1000 * 1000000) / 1000000;
```
alfonsogarciacaro: Will this work? At the end you're dividing `ts` by 1000000000000000. And it seems there's no proper test for `.TotalSeconds`. Do you have a link to the .Net source that it's doing this operation?

alfonsogarciacaro: Ups, my memory of association rules is pretty broken, this is `(ts / 1000) * 1000000` 😅

alfonsogarciacaro: I guess you multiply by `1000000` to convert to ticks before the rounding. But if we're going to do this it's probably better to just use ticks instead of milliseconds as Fable is currently doing, and we should use longs because we risk overflowing JS numbers when converting to ticks. Also, this PR may fix `totalSeconds` but we still have `totalMinutes`, `totalHours`, etc...

I think we should do the rounding when creating the TimeSpan and there are two options:

- To be closer to .NET we can use a long to represent ticks (or just create a `TimeSpan` class).
- Keep using milliseconds if we don't mind losing a bit of precision for performance and simplicity but doing the rounding creating the timespan, here: https://github.com/fable-compiler/Fable/blob/3e544d67749f5c91d4ddf0ec89b40571468f8f26/src/fable-library/TimeSpan.ts#L19

MangelMaxime: The goal of the formula is to round+truncate the value.

Capture d’écran 2019-07-30 à 10 34 17

MangelMaxime: Ooops, you had answered me since I opened the tabs ^^

In the end, I removed my change and added tests for all the `TotalXXX` methods to check that the float representation of the value is correctly handed inside the same runtime.

It's possible that later we have to use `Long` as you said if we want to have the same precision as in .Net but for me, it's outside the scope of this PR.

alfonsogarciacaro:
Hmm, now rounding has been totally removed from the PR, is that OK? This assertion fails without rounding:

```fsharp
let ts = TimeSpan.FromDays(0.005277777778)
ts.TotalSeconds |> equal 456.
```

It works if we do the rounding when [creating the TimeSpan](https://github.com/fable-compiler/Fable/blob/3e544d67749f5c91d4ddf0ec89b40571468f8f26/src/fable-library/TimeSpan.ts#L19):

```typescript
 return Math.round(d * 86400000 + h * 3600000 + m * 60000 + s * 1000 + ms); 

```

MangelMaxime:
> Hmm, now rounding has been totally removed from the PR, is that OK?

For the original scope of the PR, I would say yes because I wanted to fix the `TimeSpan.Parse`. 

But because we are aware that there are "subtle" differences between .Net and Fable version perhaps we should indeed fix them.

> It works if we do the rounding when creating the TimeSpan:

Does this works in all case or are we just fixing one case by doing that? Because by losing precision, I think we are supporting fewer cases then. Because when doing `TimeSpan.Parse` we can pass :

> fractional seconds, consisting of one to seven decimal digits.

This corresponds to the `ms` part of the string.

alfonsogarciacaro:
Ok, I'll merge this PR and we can check the rounding issue later. I assume we can use fractional seconds but we're already losing the precision beyond milliseconds (anything smaller than 0.001 s). To prevent that we would have to use ticks instead.

Issue #1873, opened 2019-08-05
Testing: nested 'testList' does not work


0x53A:
In project Fable.Tests I can not nest ``testList``, it will fail with 

```
error FSHARP: All elements of a list must be of the same type as the first element, which here is 'obj'. This element has type 'seq<string * obj>'. (code 1)
```

PR #1874, opened 2019-08-05
add support for nested "testList", 

0x53A:
fixes https://github.com/fable-compiler/Fable/issues/1873


alfonsogarciacaro:
Yes, the `testList` alias was just an ugly hack to make the tests compatible with Expecto. But this solution is much better, thank you!

Issue #1885, opened 2019-08-29
Feature request: Allow fable-splitter to pass `usePolling` configuration to chokidar


drk-mtr:
### Description

When using vim with WSL on Ubuntu 19.04, chokidar fails to register file saves after the first file save i.e. this issue is still present: https://github.com/paulmillr/chokidar/issues/237

Not a fable issue obviously, but it looks like there have been a few unsuccessful attempts to fix this at source, so I'm wondering whether you'd consider allowing us to pass through the config for the workaround from `splitter.config.js`?

### Workaround

Manually editing the `cli.js` file in `node_modules/fable-splitter/dist` to include the `usePolling: true` option within the call to `chokidar.watch()`.

### Related information

- Tried with fable-splitter 2.1.10
- Tried with chokidar 2.1.8 and 3.0.2 - both versions miss file save events in this scenario unless `usePolling` is enabled

I could have a go at doing a PR I guess but I'd have to learn javascript first so it may take me a while. Maybe this is fairly simple and just involves adding something like below but I'm not sure how to test as I don't know how to build it ha.

```
if (findFlag(args, ["--usePolling"]))
```

alfonsogarciacaro:
Yes, adding the following to the [watcher options here](https://github.com/fable-compiler/Fable/blob/cc6a73d68ae076c88607a374a3c9e8f029dd42b2/src/fable-splitter/src/cli.ts#L212) should work:

```js
usePolling: findFlag(args, ["--usePolling"])
```

If you make a PR I can make a new release so you can test it. The only perk is at the moment `findFlag` only works with cli arguments, I should make it work with options in the splitter.config.js file too. But if that's enough for you, it should be fine for now :)

drk-mtr:
Thanks for the pointer! I'll give it a go and let you know if I get my head round it. I'll close the issue as this isn't a fable issue in my view, hope that's ok.

drk-mtr:
@alfonsogarciacaro This is working, it was actually quite easy to build and test, your documentation is really good. The hardest bit was thinking of wording for the cli help!

Shall I PR in to `master`?

Thanks again for taking a look at this.

PR #1888, opened 2019-08-31
Added usePolling flag for fable-splitter/chokidar

drk-mtr:
Workaround for issue in chokidar file watcher used by fable-splitter as per https://github.com/fable-compiler/Fable/issues/1885.

This issue only seems to show up on certain editors and the impact is that file save events are only captured on the very first file save action.

The workaround allows us to specify `--usePolling` as a command line switch to `fable-splitter`.


alfonsogarciacaro:
Thanks a lot for spotting the issue and the PR! I will release a new version of fable-splitter with the change :+1:

drk-mtr:
Quite pleased to have a commit in Fable even though it's really tiny - thank you!!

PR #1890, opened 2019-09-01
Added mutable map and set

ncave:
- Added O(1) mutable map and set, with equality comparer (using native JS Map/Set).
- Reduced `fable-library` module inter-dependency.
- Fixed `Array.zeroCreate` issue in REPL build.

The current `Dictionary`/`HashSet` implementations (using immutable map/set for non-primitive keys) are quite performant when the number of keys is small, but if you go big, you can start to see the difference.

O(1) vs O(log(n)) and native map/set translates to ~10x faster for 1e6 keys.
Anecdotally, I see 5-10% speedup in the REPL too, but of course YMMV.


alfonsogarciacaro:
Wow! 👀 Impressive work as always @ncave! Thanks a lot, I will push new releases (fable-compiler, fable-compiler-js and REPL) :+1:

PR #1893, opened 2019-09-03
conversion from array works with duplicates

forki:


alfonsogarciacaro:
This was a problem with the `count` helper that assumed it would be only be called for arrays or set/dictionaries. But it seems it was not the case anymore after @ncave performance improvements in #1890. Hopefully it's fixed now, thanks a lot for the test cases! They made it really easy to pinpoint the problem :+1:

PR #1897, opened 2019-09-09
Swap the parameter order in Testing.Assert.AreEqual

zanaptak:
`Testing.Assert.AreEqual`/`NotEqual` are currently defined as taking `expected` parameter before `actual`.

But the underlying functions take `actual` before `expected` -- [reference](https://github.com/fable-compiler/Fable/blob/1f2f48d2fdcbfb4ac2423d664102622eac011e40/src/fable-library/Util.ts#L101).

So if you obey the current definition and pass `expected` first, you get incorrect output:

```fs
Fable.Core.Testing.Assert.AreEqual("EXPECTEDVALUE", "ACTUALVALUE")

// Error: Expected: ACTUALVALUE - Actual: EXPECTEDVALUE
```

### DIFF
--- a/src/Fable.Core/Fable.Core.Util.fs
+++ b/src/Fable.Core/Fable.Core.Util.fs
@@ -22,8 +22,8 @@ module Experimental =
 
 module Testing =
     type Assert =
-        static member AreEqual(expected: 'T, actual: 'T, ?msg: string): unit = jsNative
-        static member NotEqual(expected: 'T, actual: 'T, ?msg: string): unit = jsNative
+        static member AreEqual(actual: 'T, expected: 'T, ?msg: string): unit = jsNative
+        static member NotEqual(actual: 'T, expected: 'T, ?msg: string): unit = jsNative
 
 
 module Reflection =


alfonsogarciacaro:
Thanks for spotting this @zanaptak! I think I already tried to fix it before, but for some reason I always mess up the actual and expected parameters 😅 

Originally I used that order in the public signature to follow [the one from NUnit](https://github.com/nunit/docs/wiki/Assert.AreEqual). Now that I think of it, there seem to be two ways to fix this issue.

#### 1. Change the order of parameters in Fable.Core (this PR)

- Requires users to update Fable.Core
- It's kind of a "soft" breaking change (code compiles but behavior is different)
- Users must realize by themselves the order has changed (compiler won't tell them)
- Users must swap the parameters by themselves in all call sites

#### 2. Change the order of parameters in the underlying functions in fable-library/Util

- Requires users to update fable-compiler (which must be done often to fix other bugs too)
- Profit!

It looks like the alternative solution could be easier for users as it just fixes the bug without requiring manual changes, but I'm not sure. What do you think?

MangelMaxime:
Whatever the direction we take, I think some users will have a "soft" breaking change.  For example, I know that in my project I adapted the argument order to what I was seeing in the console.

By that, I mean I didn't base myself on the name of the argument you provided by on the result in the console so either solution would lead to a breaking change in this case.

Perhaps, I am an edge case :)

If we take that of the balance, I think option 2 is better in general.

zanaptak:
I don't think option 1 is breaking. Unless I'm mistaken, it doesn't change any runtime behavior, only the UI tooltip, because these functions are just jsNative passthroughs. If users have followed the old tooltip/definition, they are already getting wrong console output, with this change they can just do nothing and continue to get exact same wrong console output. No one has to touch any existing code if they are already content with current behavior. They will simply now have a reversed tooltip, making it easier to troubleshoot in the future if they eventually notice the wrong output which has always been there.

It seems I've inadvertently pushed @Zaid-Ajaj into the same situation as @MangelMaxime with https://github.com/Zaid-Ajaj/Fable.Mocha/issues/7, since I now realize this was the underlying issue. Option 2 is going to re-break those libraries and any other user who has independently adjusted their test code according to current console output behavior, and it is probably mostly going to fix the output for users who haven't even noticed the output was reversed in the first place.

My vote would be option 1 unless I've misunderstood the above ramifications, but fine with 2 if that's the consensus of others.

MangelMaxime:
After further thinking @zanaptak I think you are right about option 1. Because, it's impacting only the displayed tooltip and if people reversed the args order, it's will now follow the right "instruction".

Perhaps, the best way to know the right solution is to have a reproduction code in order to see the effect.

alfonsogarciacaro:
Ok, then let's make the change in Fable.Core then :) @zanaptak I had to drop one of the commits in master after you created the PR. Could you please rebase with latest master before merging? Thanks!

zanaptak:
@alfonsogarciacaro Updated.

alfonsogarciacaro:
Thanks!

PR #1903, opened 2019-09-18
Updated dependencies

ncave:


alfonsogarciacaro:
Hmm, it seems that updating the dependencies is putting more strain on the FCS compilation to JS. Maybe the latest versions of Babel and/or Webpack are more memory-consuming?

ncave:
@alfonsogarciacaro I'll try to see what the difference in generated code is. FWIW it only happens with ES2015 modules, not commonjs.

ncave:
@alfonsogarciacaro Looks like the babel output is identical, and the issue is webpack-related. I'll revert the webpack change.

ncave:
@alfonsogarciacaro Can't say for sure why the latest `webpack` is doing that, perhaps cause it's using an older `terser-webpack-plugin`, and I can't force it to use a newer version. Anyway, I rolled back to the old one that still works, but I'm gravitating towards switching the `fable-standalone` build to `rollup + terser`, it works great and the bundle is even a bit smaller. Let me know if you're ok with that.

alfonsogarciacaro:
I think Webpack 5 beta is already released, maybe we could try that? Anyways I'm ok with using back rollup for fable-standalone. I changed it to Webpack for simplicity and because I was worried rollup would stop being maintained, but if you think it's a better choice here, it shouldn't be a problem to make the change 👍 

Thanks for the PR!

PR #1905, opened 2019-09-22
Fixed TypedArray binding.

Titaye:
I fixed TypedArrays bindings. 
Functions working on array items are now properly typed.
I commented out BigUint64Array definitions as I did not find biguint64 equivalent in fable.

### DIFF
--- a/src/Fable.Core/Fable.Core.JS.fs
+++ b/src/Fable.Core/Fable.Core.JS.fs
@@ -316,7 +316,7 @@ module JS =
       abstract values: unit -> obj
 
 
-    and Int8Array = TypedArray<uint8>
+    and Int8Array = TypedArray<int8>
 
     and Int8ArrayConstructor =
       [<Emit "new $0($1...)">] abstract Create: size: int -> Int8Array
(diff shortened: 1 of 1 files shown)


src/Fable.Core/Fable.Core.JS.fs
```diff
@@ -316,8 +316,7 @@ module JS =
       abstract values: unit -> obj
 
 
-    and Int8Array =
-      inherit TypedArray<Int8Array>
+    and Int8Array = TypedArray<uint8>
```
inosik: Is this right?

Titaye: No, thanks for catching it

alfonsogarciacaro:
Thanks @Titaye! I will make a new release of Fable.Core. About BigUint64Array, Fable uses big.js to represent int64/uint64 and I assume is not yet compatible with the BigUint64Array proposal. We probably should wait a bit until the proposal is official and incorporated into Babel so we can use the native bigint instead.

ncave:
To clarify, Fable uses [long.js](https://github.com/dcodeIO/long.js) to represent `int64/uint64`,  [big.js](https://github.com/MikeMcl/big.js) to represent `decimal`, and Fable-compiled `FSharp.Core` implementation for `bigint`, which can eventually be replaced with native JavaScript `bigint` when browser support is deemed sufficient (or even before that with a flag, possibly).

Issue #1807, opened 2019-03-27
incorrect Fable.Compiler.runtimeconfig.json generated during NPM install


reaysawa:
### Description

```
dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   2.2.105
 Commit:    7cecb35b92

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  18.04
 OS Platform: Linux
 RID:         ubuntu.18.04-x64
 Base Path:   /usr/share/dotnet/sdk/2.2.105/

Host (useful for support):
  Version: 2.2.3
  Commit:  6b8ad509b6

.NET Core SDKs installed:
  2.1.403 [/usr/share/dotnet/sdk]
  2.2.105 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.5 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.3 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.5 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.3 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.5 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.3 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

Upon trying to run [this project](https://github.com/cmeeren/fable-elmish-electron-material-ui-demo/) I was faced with the error message:

`
fable-compiler 2.1.12
ERROR: Could not load file or assembly 'System.ComponentModel, Version=4.0.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
`

The original `Fable.Compiler.runtimeconfig.json` was generated like this:

```
{
  "runtimeOptions": {
    "tfm": "netcoreapp2.1",
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "2.1.0"
    }
  }
}
```

[...]

PR #963, opened 2017-05-26
add `prefercliruntime` file to `dotnet-fable` package root

enricosada:
This make the `dotnet-fable` (running as cli tool) to use always the bundled version of .net core
runtime instead of the one specified in package deps (Microsoft.NETCore.App).

This matter because the Microsoft.NETCore.App is not downloaded as package during restore,
and can be not avaiaible anyway (for example some os support 1.1 but not 1.0)

fix #961

I tried the built package with

- .NET Command Line Tools (1.0.1) **where this bug was found for user**
- .NET Command Line Tools (1.0.4) 
- .NET Command Line Tools (2.0.0-preview1-005899)

on sdk2.0: 

```
e:\temp\f5>dotnet -d fable --help

...
PackagedCommandSpecFactory: Ignoring prefercliruntime file as the tool target framework (1.0.5) has a different major version than the current CLI runtime (2.0.0-preview1-002101-00)
Running C:\dotnetcli\dotnet-dev-win-x64.2.0.0-preview1-005899\dotnet.exe exec --depsfile C:\Users\e.sada\.nuget\packages\.tools\dotnet-fable\1.0.6-f1\netcoreapp2.0\dotnet-fable.deps.json --additionalprobingpath C:\Users\e.sada\.nuget\packages --additionalprobingpath C:\Users\e.sada\.dotnet\NuGetFallbackFolder C:\Users\e.sada\.nuget\packages\dotnet-fable\1.0.6-f1\lib\netcoreapp1.0\dotnet-fable.dll
Process ID: 24144
Fable F# to JS compiler (1.0.6)
Usage: dotnet fable [command] [script] [fable arguments] [-- [script arguments]]

...

```

PR #1909, opened 2019-09-24
Updated target to netcoreapp3.0

ncave:


alfonsogarciacaro:
Thanks @ncave! Have you noticed any performance improvements when upgrading to netcore 3? I was considering to postpone the upgrade until Fable 3, but maybe it's better to just do it with a minor release. I'm concerned that some users may not have netcore 3 installed yet though so latest version wouldn't run in their machines (we could add a version check before running Fable.Cli.dll to prevent this though).

ncave:
@alfonsogarciacaro Travis CI fail is possibly related to #1910, although it works fine for me on WSL/Ubuntu 18.04.
Perhaps removing [this line](https://github.com/fable-compiler/Fable/blob/master/src/Fable.Cli/Fable.Cli.fsproj#L6) will fix it?

alfonsogarciacaro:
Hmm, it seems the `RuntimeFrameworkVersion` is already removed in this PR but it's still failing in Travis. #1910 seems to be about running the current Fable version in a Linux machine with dotnet SDK 3 installed. Looks a bit like #1807.

IIRC I originally added `RuntimeFrameworkVersion` to prevent MSBuild from implicitly selecting the runtime version of the dotnet SDK in my machine which caused problems, see #963.

I hate when they release a new netcore version and then everything starts to fail... 😬

forki:
@alfonsogarciacaro I don't think it's required to move to dotnet core 3 with the project in minor. But the roll forward thing (described in https://github.com/fable-compiler/Fable/issues/1910) would be really appreciated. Right now we can't run fable on sdk 3

ncave:
@alfonsogarciacaro We may not be completely out of the woods yet, seeing that in Travis ` fable-compiler v2.4.2` works fine when targeting `netcore 2.1` on `sdk 3.0.100`, but not when targeting `netcore 3.0` on `sdk 3.0.100`.

alfonsogarciacaro:
Well, given that the app is working with the netcoreapp2.1 target now even if the user only has netcore 3 installed in their computer, I guess there's no rush to upgrade to netcoreapp3. We can keep the PR open and wait until either a new dotnet SDK or something in Travis fixes the issue. Or we can also wait until netcoreapp3.1 to get LTS support.

ncave:
@alfonsogarciacaro My point was, what if somebody is trying to compile with Fable a project that targets netcoreapp3.0 on sdk 3.0, perhaps there is still an issue.
But yes, this PR can wait until 3.1, I'll close it.

forki:
@ncave what would that do? Would fable treat them differently? IIRC I always have netstandard as target framework for fable. Is that wrong?

ncave:
@forki No, nothing like that, I was just saying that the Travis logs in this PR may be indicative of a potential issue when compiling projects targeting netcoreapp3.0 on sdk 3.0.
Somebody should just try that on a non-windows machine or VM.

alfonsogarciacaro:
Yes, I usually recommend to target netstandard2.0 (I think this also avoids issues with FSharp.Core references) as Fable projects don't need to actual dotnet apps. But I guess @ncave is thinking in Fable projects that are intended to become node apps and use the `EntryPointAttribute` (like fable-compiler-js itself). In those cases you do need to target netcoreapp.

forki:
Ah ok. Thanks

Alfonso Garcia-Caro <notifications@github.com> schrieb am Mi., 25. Sep.
2019, 22:28:

Issue #1910, opened 2019-09-24
Running fable on linux dotnet core 3 fails without details


forki:
/usr/bin/yarn webpack-cli --config src/Client/webpack.config.js -p
    ﻿﻿yarn run v1.12.3
    $ /home/gitlab-runner/builds/d5cd2d2a/0/products/msu.parking/node_modules/.bin/webpack-cli --config src/Client/webpack.config.js -p
    Bundling for production...
    fable-compiler 2.3.25
    events.js:183
          throw er; // Unhandled 'error' event
          ^

    Error: write EPIPE
        at _errnoException (util.js:992:11)
        at WriteWrap.afterWrite [as oncomplete] (net.js:864:14)
    error Command failed with exit code 1.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    Running build failed.

Unfortunately it does not say what's going wrong

forki:
![image](https://user-images.githubusercontent.com/57396/65500020-e47c1380-debe-11e9-9543-86f620da9ebe.png)

this is probably related.

forki:
I assume it needs "roll forward" trick.

Corresponding PR in Paket: https://github.com/fsprojects/Paket/pull/3635

More details: https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md

alfonsogarciacaro:
Hi @forki! Thanks for the report and the links. According to the roll forwarding rules in the document you linked, dotnet SDK will only forward the runtime minor version by default. Can you please try by setting the environmental variable `DOTNET_ROLL_FORWARD= Major`? If it works, I can release a new version that sets this directly in Fable (`RollForward` project property) so users don't need to set the environmental variable by themselves.

forki:
![image](https://user-images.githubusercontent.com/57396/65583741-0e493f00-df80-11e9-9a4a-b610d1a02ccb.png)

confirmed!

alfonsogarciacaro:
Great, thanks a lot for confirming! I published fable-compiler 2.4.0 with the `RollForward` project property. Hopefully this means you don't need to set the environmental variable by yourself :+1:

forki:
![image](https://user-images.githubusercontent.com/57396/65593163-393b8f00-df90-11e9-994a-3f258cbbec65.png)

It does not help.

forki:
[...]

Issue #1911, opened 2019-09-24
Support for implicit yields


Shmew:
[The addition of implicit yields](https://devblogs.microsoft.com/dotnet/announcing-f-4-7/) sounds really nice for Fable, as sure enough they mentioned it. Is there a thread or issue I can track to see when Fable will support this functionality? Testing it this morning gave me compile errors.

ncave:
Looks like the new language features were merged into `dotnet/fsharp` master yesterday but there are still some build issues. I can rebase to latest from it once it gets to green. It will be a new Fable compiler version, once it's released.

ncave:
@alfonsogarciacaro `fcs-fable` is rebased to latest.

alfonsogarciacaro:
Thanks @ncave! I'm getting an error when trying to build latest service_slim. Do you know what it could be?

```
     1>FSC : warning FS0075: The command-line option '--version' has been deprecated [/Users/alfonsogarciacaronunez/dev/fsharp_fable/fcs/fcs-fable/codegen/codegen.fsproj]
     1>/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(370,5): error MSB4018: The "CreateAppHost" task failed unexpectedly. [/Users/alfonsogarciacaronunez/dev/fsharp_fable/fcs/fcs-fable/codegen/codegen.fsproj]
/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(370,5): error MSB4018: System.IO.IOException: The target file '/Users/alfonsogarciacaronunez/dev/fsharp_fable/artifacts/obj/fcs/netcoreapp3.0/codegen' is a directory, not a file. [/Users/alfonsogarciacaronunez/dev/fsharp_fable/fcs/fcs-fable/codegen/codegen.fsproj]
/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(370,5): error MSB4018:    at System.IO.FileSystem.CopyFile(String sourceFullPath, String destFullPath, Boolean overwrite) [/Users/alfonsogarciacaronunez/dev/fsharp_fable/fcs/fcs-fable/codegen/codegen.fsproj]
[...]

PR #1912, opened 2019-09-25
Build netcore 2.1 on 3.0

ncave:


alfonsogarciacaro:
Thanks! I assume this is just to fix Travis and doesn't need a new release. Did you find problems with 2.4.1. Unfortunately it seems it was not enough to fix #1910 (at the least in the Linux machine where the error was originally reported).

ncave:
@alfonsogarciacaro Yeah, just trying to fix the build.

ncave:
@alfonsogarciacaro I think now the appveyor problem is VS2019 env actually has 2.1 already installed, so it uses 2.1 instead of 3.0.
I'll try fixing it in another PR.

alfonsogarciacaro:
@ncave Ah, you're right, Appveyor is failing after upgrading to dotnet SDK 3. Anyways it was failing after adding the test for implicit yields [when running the tests compiled with fable-compiler-js](https://ci.appveyor.com/project/fable-compiler/fable/builds/27652248#L654). This is because fcs-fable hasn't been updated yet because of the build error [mentioned here](https://github.com/fable-compiler/Fable/issues/1911#issuecomment-534907941).

Issue #1919, opened 2019-10-01
.fsi expected in wrong folder


krauthaufen:
### Description

When compiling a project including `.fsi` files something with paths seems to get messed up.
My current project includes `.fsi` files in a folder-structure. So the real path of the file would be something like `C:\dev\fsharp-data-adaptive\src\FSharp.Data.Adaptive\Datastructures\Index.fsi` but fable keeps looking for it in `C:\dev\fsharp-data-adaptive\Datastructures\Index.fsi` and complains that the file doesn't exist.

### Repro code

Create a repo with `.fsi` files in a project-folder or clone https://github.com/fsprojects/FSharp.Data.Adaptive and remove the $(MSBuildThisFileDirectory) from FSharp.Data.Adaptive.fsproj

when using FSharp.Data.Adaptive simply start a webpack-dev-server and the error looks like this:

```
ERROR in ../Demo/Fable/Fable.fsproj
Module build failed (from C:/dev/fsharp-data-adaptive/node_modules/fable-loader/index.js):
Error: Could not find a part of the path 'C:\dev\fsharp-data-adaptive\Datastructures\Index.fsi'.
    at Loader.command.then.data (C:\dev\fsharp-data-adaptive\node_modules\fable-loader\index.js:98:22)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

### Workaround
I can add `<Compile Include="$(MSBuildThisFileDirectory)Datastructures\Index.fsi" />` and everything works nice, but this sadly doesn't work when packaging the project as nupkg (with fable-folder included)

### Expected and actual results

expected: paths should be correct 😀
actual: they aren't 

### Related information

* Fable version: 2.4.3
* Windows 10

krauthaufen:
the problem can be avoided by setting changing the current-directory to the project folder...

alfonsogarciacaro:
Thanks for the report @krauthaufen! I've just published fable-compiler 2.4.4 with a fix, could you please try it? Cheers!

krauthaufen:
works!!! 
thank you lots 😀

PR #1920, opened 2019-10-03
WIP: Feature/add enum reflection

MangelMaxime:
Hello @alfonsogarciacaro , I have a hard time figuring two things.

In order to add support for the methods under `System.Enum` like:

- `System.Enum.GetValues(t)`
- `System.Enum.Parse(t, "0")`

I have to access the `EnumType` information from at [this place](https://github.com/fable-compiler/Fable/blob/feature/add_enum_reflection/src/Fable.Transforms/Replacements.fs#L2302-L2309). But I can't find a way to extract the enum information from the info I have access to here.

----------------

Another problem I am facing is that Enum is replaced by their number representation at compile time. That's probably the reason why Fable only support enum of type integer and not `enum<char>` etc.

So `let v = MyEnum.Foo` gives `const v = 5;`.

However, when using `t.GetEnumValues()` we get all the enum info (key + value) which is expected.

So this code fails on Fable, because `MyEnum.Foo` is a number while `extractedFoo` is the complete info about Enum `[Foo, 5]`

I found, that [this opmisation](https://github.com/fable-compiler/Fable/blob/feature/add_enum_reflection/src/Fable.Transforms/Fable2Babel.fs#L283-L290) should probably be removed for the enums. Because when doing `(t.GetEnumValues() |> Seq.cast<MyEnum> |> Seq.item 1)` it returns `0` instead of the expected `key/value`.

*Code as use to test that reflection thing (works under FSI)*

```fs
type MyEnum =
    | Foo = 0y
    | Bar = 5y

let t = MyEnum.Foo.GetType()
let extractedFoo = (t.GetEnumValues() |> Seq.cast<MyEnum> |> Seq.item 1)
let res = MyEnum.Foo = extractedFoo
// Should return true
```

Could you please help me understand what I should look at?

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -1029,8 +1029,8 @@ let defaultof com ctx (t: Type) =
     | Number _ -> makeIntConst 0
     | Boolean -> makeBoolConst false
     | Builtin BclTimeSpan
-    | Builtin BclDateTime 
-    | Builtin BclDateTimeOffset 
+    | Builtin BclDateTime
+    | Builtin BclDateTimeOffset
     | Builtin (BclInt64|BclUInt64)
     | Builtin BclBigInt
     | Builtin BclDecimal -> getZero com ctx t
@@ -2720,6 +2720,11 @@ let types (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr optio
         | "get_IsArray" ->
             match exprType with Array _ -> true | _ -> false
             |> BoolConstant |> makeValue r |> Some
+        | "get_IsEnum" ->
+            match exprType with
+            | Enum t -> true
+            | _ -> false
+            |> BoolConstant |> makeValue r |> Some
         | "GetElementType" ->
             match exprType with
             | Array t -> TypeInfo t |> makeValue r |> Some
@@ -2739,7 +2744,7 @@ let types (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr optio
         | "get_GenericTypeArguments" | "GetGenericArguments" ->
             Helper.CoreCall("Reflection", "getGenerics", t, [thisArg], ?loc=r) |> Some
         | "get_FullName" | "get_Namespace" | "get_IsArray" | "GetElementType"
-        | "get_IsGenericType" | "GetGenericTypeDefinition" ->
+        | "get_IsGenericType" | "GetGenericTypeDefinition" | "get_IsEnum" ->
             let meth = Naming.removeGetSetPrefix i.CompiledName |> Naming.lowerFirst
             Helper.CoreCall("Reflection", meth, t, [thisArg], ?loc=r) |> Some
         | "GetTypeInfo" -> Some thisArg
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -2299,16 +2299,13 @@ let unchecked (com: ICompiler) (ctx: Context) r t (i: CallInfo) (_: Expr option)
     | "Compare" -> Helper.CoreCall("Util", "compare", t, args, i.SignatureArgTypes, ?loc=r) |> Some
     | _ -> None
 
-let enums (_: ICompiler) (ctx: Context) r _ (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
+let enums (com: ICompiler) (ctx: Context) r _ (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
     match thisArg, i.CompiledName, args with
     | Some this, "HasFlag", [arg] ->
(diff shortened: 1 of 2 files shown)


MangelMaxime:
Just to complete, understanding how to access the `EnumType` in the first section of my message should allow me to move forward for several methods in theory because severals need this part. 

The second part is a bit more complex because it's about how we represent the enum at runtime, etc.

alfonsogarciacaro:
Yes, the enum values are just numbers in runtime. It seems there's some confusion between enum values and enum cases. `System.Enum.GetValues(t)` should return only the values, so in `getEnumValues` within Reflection.ts, you should do something like `return t.enumCases.map((kv) => kv[1]);`.

You can use the keys for parsing (see below for an implementation example) or to add support for [Enum.GetName](https://docs.microsoft.com/en-us/dotnet/api/system.enum.getname?view=netframework-4.8):

```typescript
export function parseEnum(t: TypeInfo, str: string): any {
    if (t.enumCases != null) {
        // Probably we should do better int parsing here,
        // as parseInt ceils floats: "4.8" -> 4
        const value = parseInt(str);
        if (isNaN(value)) {
            for (const kv of t.enumCases) {
                if (kv[0] === str) {
                    return kv[1];
                }
            }
            throw new Error(`'${t.fullname}' was not found`);
        } else {
            // .NET returns the number even if it
            // doesn't match any of the cases
            return value;
        }
    } else {
        throw new Error(`${t.fullname} is not an enum`);
    }
}
```

MangelMaxime:
> It seems there's some confusion between enum values and enum cases

Yes, that's possible because, F# use `Foo { value__ = "5"}` on .Net and perhaps I confused it with `[Foo, 5]` indeed.

I will fix this part of the code.

alfonsogarciacaro:
There may be some kind of metadata, but if I'm not mistaken enums are basically numbers in runtime too. In fact, if you only want to support enums of int32 or lower and will always encode them with the numeric value, you can replace [this code](https://github.com/thoth-org/Thoth.Json.Net/blob/b64c0ded6ed2a2c2e192c2f9879835a18126b4fa/src/Decode.fs#L1026-L1035) with something like the following that should be more performant:

```fsharp
let parseEnum<'T> (value: obj): 'T =
    let intValue = System.Convert.ToInt32(value)
    let enumType = typeof<'T>
    let underlyingType = enumType.GetEnumUnderlyingType()
    let mutable exists = false
    for enumValue in System.Enum.GetValues(enumType) do
        exists <- exists || System.Convert.ToInt32(enumValue) = intValue
    if exists then
        value :?> 'T
    else
        failwithf "Unkown %O provided for %s" value enumType.FullName
```

alfonsogarciacaro:
@MangelMaxime Do you think you will be able to finish this soon? I was planning to push a new release to fix #1919, but it'd be great to include reflection support for enums too. I think I have more or less an idea of how to do it now, so if you don't have time, I can merge this and finish it before the release. What do you think?

MangelMaxime:
I guess it will take me more times to do because I still don't know how to get the `Type` info calling `System.Enum.*` (ex `System.Enum.GetValues`).

Please feel free to merge and fix the PR. I will do a review when it's released and test it using `Thoth.Json` and add missings tests if something is missing or that I need.

MangelMaxime:
> There may be some kind of metadata, but if I'm not mistaken enums are basically numbers in .NET runtime too. In fact, if you only want to support enums of int32 or lower and will always encode them with the numeric value, you can replace [this code](https://github.com/thoth-org/Thoth.Json.Net/blob/b64c0ded6ed2a2c2e192c2f9879835a18126b4fa/src/Decode.fs#L1026-L1035) with something like the following that should be more performant:
> 
> ```fsharp
> let parseEnum<'T> (value: obj): 'T =
>     let intValue = System.Convert.ToInt32(value)
>     let enumType = typeof<'T>
>     let mutable exists = false
>     for enumValue in System.Enum.GetValues(enumType) do
>         exists <- exists || System.Convert.ToInt32(enumValue) = intValue
>     if exists then
>         value :?> 'T
>     else
>         failwithf "Unkown %O provided for %s" value enumType.FullName
> ```

Indeed, this seems more performant I will have a look. I probably like a bit too much using a pure functional approach in the internal of my libraries sometime :)

alfonsogarciacaro:
I just found there's actually an API (now also supported by Fable) to check if a numeric value corresponds to one of the cases of an enum:`Enum.IsDefined`. So there is actually a much easier alternative than my last proposal:

```fsharp
let parseEnum<'T> (value: obj): 'T =
    // In .NET runtime, this throws an error if the value is not a string or the enum underlying type
    let isDefined = Enum.IsDefined(typeof<'T>, value)
    if isDefined then value :?> 'T
    else failwithf "Unkown %O provided for %s" value typeof<'T>.FullName
```

MangelMaxime:
Nice, I need to test it :) 

Thank you for keeping me updated on that :)

PR #1921, opened 2019-10-05
Whitespace cleanup

ncave:


alfonsogarciacaro:
I'm usually a bit wary of formatting changes that touch many files, because they can interfere with the git history when using git blame. But I think now Github makes it easier to see the commit history per line so hopefully it won't be a problem. Thanks!

PR #1926, opened 2019-10-11
Trying to reproduce weird union case from gitter

forki:
@alfonsogarciacaro any idea what's going on here?

### DIFF
--- a/tests/Main/UnionTypeTests.fs
+++ b/tests/Main/UnionTypeTests.fs
@@ -80,6 +80,30 @@ open Fable.Core
 #endif
 type DU = Int of int | Str of string
 
+
+type DD = {
+        a: string
+        b: string
+    }
+
+
+type D =
+    | DA of string
+
+type EE = {
+    a: string
+    b: string
+}
+
+type E =
+    | EA of string
+
+type DDD = DD * D list
+type EEE = EE * E list
+
+type UnionType = Fable.Core.U2<DDD, EEE>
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Yes, because `Fable.Core.U2` and the like are erased, the pattern matching gets compiled to type testing. However, type testing is very weak in Fable. For example, for arrays and tuples it only checks if the value is a JS array. So the mechanism cannot tell two different tuple types apart.

The erased unions are mainly intended to interact with JS, that is to type the arguments of imported JS functions, and shouldn't as cheap replacements of `Choice`. Unfortunately, this is not very clear in the new documentation, I will fix that :+1:

forki:
Cool. Please close as "won't fix". Will report back to gitter user

Alfonso Garcia-Caro <notifications@github.com> schrieb am Fr., 11. Okt.
2019, 12:31:

alfonsogarciacaro:
Ok, thanks! I've [fixed the docs](https://github.com/fable-compiler/fable-compiler.github.io/commit/c630ecb4765f42276ae2e7dfc70d3aa34dc061b5). Closing then as this is working as intended, though may be we should [raise a warning when trying to type test agains tuples](https://github.com/fable-compiler/Fable/blob/f437117ec59e1f4a02fe7fba10a2e74082eef019/src/Fable.Transforms/Fable2Babel.fs#L861).

PR #1928, opened 2019-10-16
Fix fsx compilation on netcore 3

### LINKED ISSUE #1927: Compile fsx files fails with .NET Core 3
### Description

Ever since the release of Fable 2.4.x and the move to .NET stable F# scripts stopped compiling for me. 

### Repro code

See build at https://travis-ci.org/nojaf/fable-reactstrap/builds/589588661 or https://travis-ci.org/nojaf/fable-fsx-sample/builds/589588371.

Basically the workaround for the FSharpChecker in [ProjectCracker](https://github.com/fable-compiler/Fable/blob/master/src/Fable.Cli/ProjectCracker.fs#L142) doesn't work anymore.

`Error: File /app/src/ReactTodoMVC.fsx cannot be found in source list:`

This happens because of the checker already having lots of issues after [ParseAndCheckProject](https://github.com/fable-compiler/Fable/blob/master/src/Fable.Cli/Agent.fs#L96). No files are found in `checkedProject.AssemblyContents.ImplementationFiles`

### Related information

* Fable version: 2.4.1
* Operating system: Windows and Linux

### Solution

After some digging, I noticed that FSharpChecker it now able to deduce the correct references and there is no need for the workaround.

I will create a PR with the changes that worked for my projects both on Windows and Linux.
Maybe @rfrerebe can confirm this also works for his projects.

nojaf:
Resolved #1927.


alfonsogarciacaro:
Thanks! Can you please [remove this line](https://github.com/fable-compiler/Fable/blob/5d9c9ed1d7f56f195bfd9696420e22f1bd61a237/build.fsx#L143) which seems to be making the build fail after the Fable.Cli.Test project was removed? When the build is green we can merge  and push a new release :+1:

PR #1935, opened 2019-10-31
Add [<RequireQualifiedAccess>] to JS module

pauldorehill:
I started using Fable again after some time away (its awesome... 😄 ) and I came unstuck with `Promise`. After creating one there were no `bind`, `map` methods etc. and my `System` methods were also complaining. After some hair pulling I released it was because currently you can write:

```fsharp
open Fable.Core.JS

let myPromise : Promise<string> =  promise { return "Hello" }
```
But in opening `Fable.Core.JS` now you have some problems - so this is just a quick fix to prevent anyone else doing the same.

### DIFF
--- a/src/Fable.Core/Fable.Core.JS.fs
+++ b/src/Fable.Core/Fable.Core.JS.fs
@@ -10,7 +10,6 @@ namespace Fable.Core
 open System
 open System.Text.RegularExpressions
 
-[<RequireQualifiedAccess>]
 module JS =
     type [<AllowNullLiteral>] PropertyDescriptor =
         abstract configurable: bool option with get, set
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Hi @pauldorehill! Thanks a lot for the nice words and the PR. You're right this can be problematic, this always strikes me too because in F# when opening two modules with same name they don't shadow each other, but a type do shadow a module which is not very intuitive. Unfortunately this is also a breaking change and we need to be careful with this kind of changes in Fable.Core. I wonder if there's a way to avoid this situation or at least give a better error/warning to the user 🤔 I tried using the `ModuleSuffix` attribute on the Promise module but it wasn't helpful.

```fsharp
[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
module Promise
```

pauldorehill:
Hey Alfonso, on looking at it again I think it comes from the ` let [<Global>]` bindings:
https://github.com/fable-compiler/Fable/blob/218dca5ce4aa5230ffa31e2d59adca48b15b59c9/src/Fable.Core/Fable.Core.JS.fs#L451

For example if I do:
image

On it being a breaking change my knowledge of Fable is very limited to understand the full implications, however I'd assume that people regularly using Fable never write `open Fable.Core.JS` ? Since in opening it things like `Array`, `Map`, `Math`,  `Set` all start to complain.

alfonsogarciacaro:
Ah, you're right! It's the `let Promise` value and not the type which shadows the module. Some bindings [do open `Fable.Core.JS`](https://github.com/fable-compiler/fable-browser/blob/8c3073bd5bf45ae2d997e9a1b2abc18a093ba85b/src/WebRTC/Browser.WebRTC.fs#L5) so adding the `RequireQualifiedAccess` would break them. But I do believe that is less often to use the constructors directly. So we could move only the constructors into a new module to minimize the impact. For a while we could keep the old values too but with an Obsolete warning so people have time. For example:

```fsharp
        [<Obsolete("JS constructors have been to Fable.Core.JS.Constructors module to prevent conflicts with modules with same name")>]
        let [<Global>] Promise: PromiseConstructor = jsNative

        // ... Other constructors, also decorated with Obsolete 

        [<RequireQualifiedAccess>]
        module Constructors =
            let [<Global>] Promise: PromiseConstructor = jsNative

            // ... Other constructors
```

pauldorehill:
I think that is a better idea - as the reason I first opened that module was to remove `JS.` from the `Promise` 😄 I just changed the commit to that effect.

alfonsogarciacaro:
Looks good, thank you!

src/Fable.Core/Fable.Core.JS.fs
```diff
@@ -437,48 +437,180 @@ module JS =
         abstract warn: ?message: obj * [<ParamArray>] optionalParams: obj[] -> unit
         abstract table: ?data: obj -> unit
 
+    [<Obsolete("JS constructors are now in Fable.Core.JS.Constructors module to prevent conflicts with modules with same name")>]
     let [<Global>] Number: NumberConstructor = jsNative
+
+    [<Obsolete("JS constructors are now in Fable.Core.JS.Constructors module to prevent conflicts with modules with same name")>]
     let [<Global>] NaN: float = jsNative
```
inosik: Does it make sense to move `NaN` and many others to `Constructors`?

alfonsogarciacaro: No, you're right. Actually I proposed the name `Constructors` to avoid moving NaN and other helpers like `encodeURI` because I think devs use those more often (at least I do) and have less risk of conflict. I was planning to change it before the release, thanks for the reminder!

PR #1943, opened 2019-11-18
Added NuGet packages support

ncave:
- Added NuGet packages support in `fable-compiler-js`

@alfonsogarciacaro Looks like the recent change in Fable enums requires a new `fable-standalone` version to be published and referenced in `fable-compiler-js`.


alfonsogarciacaro:
This is great @ncave! So fable-compiler-js can be used now more or less as fable-compiler? It'd be interesting to see how they compare to each other.

I'll push new versions of fable-standalone and fable-compiler-js. Thanks again!

ncave:
Warning: Incoming WoT (huge WoT fan, the books too :)

Yes, that was always the intent to have the same functionaly as `fable-compiler`, for the use cases where it makes sense (dotnet not installed, etc.). Obviously we're not trying to replicate the complex machinery of MSBuild and package managers and resolvers with just a few lines of code, so some complicated scenarios involving different versions of binaries for referenced projects are unsupported right now.

But in theory we could be fable-compiling those projects incrementally and reference the javascript output instead. Different approach vs merging all F# source files into one project, sure, but it sounds doable, let's see if that's a really common use case and somebody needs it.

We could also look into adding support for paket.references, but usually these package references don't have versions. We probably don't want to replicate the complex piece of software that is Paket, so we can just cover the most common use case (say, pick the latest package version from the package cache). Same thing for package references with wildcards in the project, currently wildcards are unsupported, but they could be, potentially.

All these can be resolved by ingesting the `obj/project.assets.json`, but that requires `dotnet/MSBuild` to be installed, which kinda defeats the purpose. To that extend, using NuGet package store itself kind of means (or requires) dotnet already, so I'm not sure if that is really a limitation.

Adding support for other package stores is possible, say npm or GitHub packages, but I'm not sure anybody is storing their Fable packages there yet.

TL;DR yes.

PR #1944, opened 2019-11-19
specify where to run `npm run build`

Neftedollar:
I was confused with documentation and running `npm run build` inside src/fable-splitter was wrong and I fight with this for two days) Hope it will help for others


MangelMaxime:
Sorry for the hard times and thank you for the PR which should help others :)

PR #1946, opened 2019-11-21
Add globbing support for #1942

### LINKED ISSUE #1942: Can't use * in fsproj when building fable libraries
### Description

If you do something like `<Compile Include="SomeSubDirectory\*.fs" />` when building a library, when a project references that library during compile it will complain that the file `SomeSubDirectory\*.fs` does not exist. To resolve this I had to explicitly add all files in that directory.

### Repro code

Build a fable project that gets packaged into a nupkg that includes a item group property using wildcards. Then try to build a new project with a reference to that library.

For example in my [library](https://github.com/Shmew/Feliz.Plotly/blob/0.17.1/src/Feliz.Plotly/Feliz.Plotly.fsproj) gave the error, doing [this](https://github.com/Shmew/Feliz.Plotly/blob/0.17.2/src/Feliz.Plotly/Feliz.Plotly.fsproj) resolved it.

You can see very little [changed](https://github.com/Shmew/Feliz.Plotly/commit/8e23c061a22feeba7d123e100700af64cdd0ec19) between tags.

You can try it yourself with the [0.17.1](https://www.nuget.org/packages/Feliz.Plotly/0.17.1) and [0.17.2](https://www.nuget.org/packages/Feliz.Plotly/0.17.2) nuget packages.

Expected that it resolves the wild card and compiles all of the matching files.
Actual is that it complains that a single file does not exist.

Shmew:
Testing done by creating a new project, pulling down `Feliz.Plotly` at 0.17.1 which had globbing in the nupkg. Everything compiles correctly now.

Resolves #1942


alfonsogarciacaro:
Hi @Shmew! Sorry for replying late. Thanks a lot for this! Looks good :+1: Only thing is I usually try to keep the Fable dependencies to a minimum. Would it be possible to resolve the globbing pattern with the Fake.IO.Filesystem dependency? If it's not, I guess it's not so dramatic to add it to Fable.Cli.

Shmew:
Definitely can, the only thing I really need from that library is [this file](https://github.com/fsharp/FAKE/blob/master/src/app/Fake.IO.FileSystem/Globbing.fs), so we'd be forking it. 

If you are okay with adding the dependency, it does have some nice features that can make working on `Fable.Cli` easier going forward.

alfonsogarciacaro:
Well, I don't have much time to work on Fable now, but the thing is we currently have two versions: fable-compiler and fable-compiler-js. Though both of them are distributed through npm the former runs on a dotnet process while the latter runs entirely on node-js. It could be interesting to unify them in the future and for that we would have to remove the dependencies, which should be possible for Newtonsoft.Json and DotnetProj.Info, but it would get more complicated if we add more.

Shmew:
Forked the required files from [Fake.IO.FileSystem](https://github.com/fsharp/FAKE/tree/release/next/src/app/Fake.IO.FileSystem) to remove the dependency and added a [test case](https://github.com/fable-compiler/Fable/pull/1946/files#diff-70278a0311770386206c175b53e1e7e7R919)

Shmew:
Sorry about all of the spam if you were getting notifications @alfonsogarciacaro, I was having git troubles 😅

MangelMaxime:
Thank you for the PR, it has been released in fable-compiler 2.5.0 and fable-compiler-js 1.3.0

PR #1957, opened 2019-12-14
Source maps support in `fable-compiler-js`

ncave:
- Added  source maps support in `fable-compiler-js`
- Updated dependencies


alfonsogarciacaro:
Thanks! Sorry for taking so long to merge.

PR #1958, opened 2019-12-17
fixed typo in MapConstructor

Luiz-Monad:
typo was causing babel error when compiling.

``` 
Module build failed (from ./node_modules/fable-loader/index.js):
Error: unknown: BABEL ERROR: Failed to parse macro: new $0($1..) -> new $0($1..)
MACRO ARGUMENTS: $0,$1
Unexpected token (2:10)
```

Code that made it fail:
```fsharp
let sheetsManager : Map<obj, obj> = Constructors.Map.Create ()
```

### DIFF
--- a/src/Fable.Core/Fable.Core.JS.fs
+++ b/src/Fable.Core/Fable.Core.JS.fs
@@ -176,7 +176,7 @@ module JS =
         abstract values: unit -> seq<'V>
 
     and [<AllowNullLiteral>] MapConstructor =
-        [<Emit("new $0($1..)")>] abstract Create: ?iterable: seq<'K * 'V> -> Map<'K, 'V>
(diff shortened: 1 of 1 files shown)


Luiz-Monad:
Why did it fail?
 I didn't find were to put the tests for this specific interface.

alfonsogarciacaro:
Ups, good catch. Thanks! JS specific tests are in the [JsInteropTests.fs](https://github.com/fable-compiler/Fable/blob/master/tests/Main/JsInteropTests.fs) file, but anyways your fix should be enough. I'll push a new Fable.Core version. I guess we didn't catch it before because `System.Collections.Generic.Dictionary` compiles to JS Map most of the time, so people often use that instead.

alfonsogarciacaro:
Travis failed because dotnet SDK 3.0 was not installed. Probably they've already moved to 3.1.

PR #1970, opened 2020-01-28
Failing test for number formatting

forki:
I *think* this used to work and is somehow broken now


forki:
ok. so it works here and it works in REPL. Why does it not work in my app?

MangelMaxime:
Do you have the latest version of `fable-compiler` installed?

And perhaps, does the test pass if you run them on your machine? `npm install && npm run build test`

MangelMaxime:
I am merging because it's never bad to have more tests but I am not sure what we can do to help you.

forki:
Yeah. I didn't figure it out yet

PR #1997, opened 2020-03-10
Fix fable-library build

MangelMaxime:
@ncave While upgrading `typescript` and `@types/node` I broke fable-library build.

> Error: Individual declarations in merged declaration 'Long' must be all exported or all local.

I think this is because we are import `Long` and also declaring `Long` and typescript is warning us against this situation.

If I change the import to be name `LongLib` it's working again. I tested locally and the pass, so I think we can keep the fix as it is.

I am merging this PR directly into master so we don't have a broken master. But please tell me if I should revert and we should handle the problem in another way. I am still a beginner with all the typescript error etc. :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3199,7 +3199,9 @@
       }
     },
     "fable-babel-plugins": {
-      "version": "2.3.0"
+      "version": "2.3.0",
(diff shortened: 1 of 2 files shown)


ncave:
@MangelMaxime No, it works, so it's fine.

MangelMaxime:
Ok :) 

I think I will apply this philosophy now as it is not the first time you tell me that :D

PR #2019, opened 2020-04-12
Changed fable-compiler-js build

ncave:
- Now building `fable-compiler-js` with `fable-splitter`.

### DIFF
--- a/src/fable-compiler-js/splitter.config.js
+++ b/src/fable-compiler-js/splitter.config.js
@@ -1,18 +1,7 @@
 const resolve = (path) => require("path").join(__dirname, path);
 
-const testLocal = process.argv.find(v => v === "--test-local");
-if (testLocal) {
-  console.log("Using local fable-standalone and fable-metadata...")
-}
-
 module.exports = {
   cli: { path: resolve("../Fable.Cli") },
   entry: resolve("./src/fable-compiler-js.fsproj"),
-  outDir: resolve("./dist"),
-  babel: {
-    plugins: ["@babel/plugin-transform-modules-commonjs"],
-  },
-  fable: {
-    define: testLocal ? ["TEST_LOCAL"] : []
-  }
+  outDir: resolve("../../build/fable-compiler-js/out"),
 };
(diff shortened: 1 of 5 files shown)


src/fable-compiler-js/splitter.config.js
```diff
@@ -1,18 +1,7 @@
...
 module.exports = {
   cli: { path: resolve("../Fable.Cli") },
   entry: resolve("./src/fable-compiler-js.fsproj"),
-  outDir: resolve("./dist"),
-  babel: {
-    plugins: ["@babel/plugin-transform-modules-commonjs"],
-  },
-  fable: {
-    define: testLocal ? ["TEST_LOCAL"] : []
-  }
```
MangelMaxime, line 17: This definition seems to be required: https://github.com/fable-compiler/Fable/blob/master/src/fable-compiler-js/src/app.fs#L8-L17

ncave: @MangelMaxime This is actually not needed, it was replaced with npm link/unlink, see build.fsx.
I removed the code that was using it, please review again.

MangelMaxime: Thank you very much.

I have strange behaviour on my computer if I do:

- `npm install`
- `npm run build compiler`
- `npm run build test-js` <- This will report an error when compiling `fable-compiler-js` about `Fable.Core` being unknown

But if I delete the `build/fable-compiler-js` dir and re-launch:
- `npm run build test-js`

Then it is working 🤷‍♂️ This is not a big problem because the CI doesn't seem to have this problem but I wanted to mention in in case. It happens in the future.

ncave: @MangelMaxime The `test-js` target has some conditional build checks at the beginning to avoid re-compiling too much, perhaps that's what's happening for you. Perhaps we can wipe the whole `build/*` folder before `npm install`, perhaps in a `BuildAll` target.

MangelMaxime: @ncave I was working from a fresh repo cleaned using `git clean -xdf`.

I think it's more because there is a lock occurring somewhere by a process but it's strange behaviour.

For now, we can keep it like it because it should work with your changes. For example, the CI is succeeding.

Issue #2017, opened 2020-04-12
Cannot get correct union case field names using reflection


anchann:
### Description

When using reflection APIs, the name reported for a union case field seems to be some kind of autogenerated name instead of the real one in the type definition. How can I get the real name?

### Repro code

```
open System
open FSharp.Reflection

type Person = Person of FirstName: string * LastName: string

let fieldNames: seq<string> =
    FSharpType.GetUnionCases typeof<Person>
    |> Seq.collect (fun caseInfo ->
        caseInfo.GetFields()
        |> Seq.map (fun field ->
            sprintf "%s or %s" field.PropertyType.Name field.Name
        )
    )

printf "%O" fieldNames
```

### Expected and actual results

Running this in the fable REPL produces

```
seq [String or Data0; String or Data1]
```

Running it in `fsi` produces the expected

```
seq ["String or FirstName"; "String or LastName"]
```

inosik:
I'll leave some links here for anybody who's willing to tackle this:

- https://github.com/fable-compiler/Fable/blob/fa8d44f6cff1a1446078b84f9f1c72979d5fd9d5/src/fable-library/Reflection.ts#L13
  `CaseInfo.fields` should be updated to be an array of `FieldInfo`

- https://github.com/fable-compiler/Fable/blob/fa8d44f6cff1a1446078b84f9f1c72979d5fd9d5/src/Fable.Transforms/Fable2Babel.fs#L374
  The F# side of the prior function should be updated as well to build an array of `FieldInfo` objects. Must be this snippet:
  https://github.com/fable-compiler/Fable/blob/fa8d44f6cff1a1446078b84f9f1c72979d5fd9d5/src/Fable.Transforms/Fable2Babel.fs#L390-L393

- https://github.com/fable-compiler/Fable/blob/fa8d44f6cff1a1446078b84f9f1c72979d5fd9d5/src/fable-library/Types.ts#L129
  `Union` type in `Types.ts` might need to be updated too

anchann:
@inosik thank you for the guidance. I'd like to work on this as I need the fix asap for some production-facing work I'm doing, but couldn't get a local build of Fable going. Any chance you could help me with https://github.com/fable-compiler/Fable/issues/2018?

anchann:
Figured out #2018. Will proceed trying to implement this.

MangelMaxime:
It has been fixed in fable-compiler 2.6.0

PR #2023, opened 2020-04-14
Plumb through real case field names instead of generating fake ones

anchann:
Implemented as per @inosik's guidance in #2017.

### DIFF
--- a/tests/Main/ReflectionTests.fs
+++ b/tests/Main/ReflectionTests.fs
@@ -27,11 +27,6 @@ type TestType3 = class end
 type TestType4 = class end
 type TestType5 = class end
 
-type TestType6 =
-    | CaseA of string
-    | CaseB of SomeField: string
-    | CaseC of string * SomeField: string
-
 type GenericRecord<'A,'B> = { a: 'A; b: 'B }
 
 type MyEnum =
--- a/tests/Main/ReflectionTests.fs
+++ b/tests/Main/ReflectionTests.fs
@@ -27,6 +27,11 @@ type TestType3 = class end
 type TestType4 = class end
 type TestType5 = class end
 
+type TestType6 =
+    | CaseA of string
+    | CaseB of SomeField: string
+    | CaseC of string * SomeField: string
+
 type GenericRecord<'A,'B> = { a: 'A; b: 'B }
 
 type MyEnum =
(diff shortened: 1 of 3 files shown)


inosik:
Can you add one or two tests, please?

anchann:
@inosik ah, yes, sorry for the lack of discipline. Updated the existing union type reflection unit test. By the way, there were a couple unit tests failing that are unrelated to my changes.

tests/Main/ReflectionTests.fs
```diff
@@ -172,8 +172,8 @@ type TestRecord = {
 }
 
 type TestUnion =
-    | StringCase of String: string * string
-    | IntCase of Int: int
+    | StringCase of SomeString: string * string
```
anchann, line 175: I didn't want to risk some coincidental equality of the field name and the (pascal cased version of?) the type, so made the field names clearly distinct.

tests/Main/ReflectionTests.fs
```diff
@@ -364,7 +364,10 @@ let reflectionTests = [
             let types =
                 info.GetFields()
                 |> Array.map (fun field -> field.PropertyType)
-            info.Tag, info.Name, types, values)
+            let names =
```
anchann, line 367: A bit duplicative, but I figured it doesn't matter in a unit test, not worth trying to change the test structure for it.

MangelMaxime:
@anchann If the failing tests are about Date/DateTime then it is "normal". 

These tests kinds of work only if your laptop is configured on the UK timezone or something like that. We hardcoded the expected string/timezone because it was too much of an effort at the time to make the test adapt to the user machine configuration. And also, it could introduce bugs in our tests...

anchann:
@MangelMaxime Ah, got it. Yes, it's the two DateTime/DateTimeOffset related ones.

tests/Main/ReflectionTests.fs
inosik: LGTM. This code is just for testing purposes.

tests/Main/ReflectionTests.fs
```diff
@@ -27,6 +27,11 @@ type TestType3 = class end
 type TestType4 = class end
 type TestType5 = class end
 
+type TestType6 =
+    | CaseA of string
+    | CaseB of SomeField: string
+    | CaseC of string * SomeField: string
+
```
inosik: Can this be removed? This type doesn't seem to be used.

MangelMaxime:
Thank you for the PR @anchann

MangelMaxime:
Thank you too @inosik for helping implementing and reviewing it

inosik:
No problem, glad I could help 👍

anchann:
Thanks for the review, and thank you @inosik for the implementation guidance, it made for a smooth onboarding onto a new code base!

PR #2024, opened 2020-04-14
Add Length and Substring with firstIndex and length in StringBuilder

ThisFunctionalTom:
I was trying to make Hedgehog fable friendly and was missing this two functions in StringBuilder.

The ToString(firstIndex, length) is not implemented performant by I think it should be OK for the first version. Or is it?

### DIFF
--- a/src/fable-library/System.Text.fs
+++ b/src/fable-library/System.Text.fs
@@ -12,3 +12,7 @@ type StringBuilder(value: string, capacity: int) =
     member x.AppendLine() = buf.Add(System.Environment.NewLine); x
     member x.AppendLine(s: string) = buf.Add(s); buf.Add(System.Environment.NewLine); x
     override __.ToString() = System.String.Concat(buf)
+    member x.Length = buf |> Seq.sumBy String.length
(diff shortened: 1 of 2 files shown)


MangelMaxime:
Thank you for the PR.

I think the current implementation is ok.

Don't forget to add tests when adding a new feature to Fable like that we make sure they work as expected and avoid regression. :) I did it in your PR no need to do it for this one

MangelMaxime:
It has been released in fable-compiler 2.6.0

ThisFunctionalTom:
> 
> 
> Thank you for the PR.
> 
> I think the current implementation is ok.
> 
> Don't forget to add tests when adding a new feature to Fable like that we make sure they work as expected and avoid regression. :) I did it in your PR no need to do it for this one

🙏 Thank you for you help.
This was one of my first two pull requests to Fable and I am completely new to contributing to OS. Next time I will surely think on the tests.

MangelMaxime:
@ThisFunctionalTom I am really happy to see people contributing and welcome aboard :)

PR #2025, opened 2020-04-14
add using function implementation to Fable.Core.Utils

ThisFunctionalTom:
I was trying to make hedgehog library fable friendly and I found out that the using function is missing in fable.

I copied the implementation from FSharp.Core and added it to Fable.Core.Utils.

Hope this is OK.

### DIFF
--- a/src/Fable.Core/Fable.Core.Util.fs
+++ b/src/Fable.Core/Fable.Core.Util.fs
@@ -9,6 +9,10 @@ module Util =
         try failwith "JS only" // try/catch is just for padding so it doesn't get optimized
         with ex -> raise ex
 
+    let using (resource : 'T when 'T :> System.IDisposable) action =
+        try action(resource)
+        finally match (box resource) with null -> () | _ -> resource.Dispose()
+
 module Experimental =
     /// Reads the name of an identifier, a property or a type
     let inline nameof(expr: 'a): string = jsNative
--- a/src/Fable.Core/Fable.Core.Util.fs
+++ b/src/Fable.Core/Fable.Core.Util.fs
@@ -9,10 +9,6 @@ module Util =
         try failwith "JS only" // try/catch is just for padding so it doesn't get optimized
         with ex -> raise ex
 
-    let using (resource : 'T when 'T :> System.IDisposable) action =
-        try action(resource)
-        finally match (box resource) with null -> () | _ -> resource.Dispose()
-
 module Experimental =
     /// Reads the name of an identifier, a property or a type
     let inline nameof(expr: 'a): string = jsNative
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -1358,6 +1358,7 @@ let operators (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr o
     |  "LazyPattern"     // (|Lazy|_|)
     |  "Lock"            // lock
     |  "NullArg"         // nullArg
+    |  "Using"           // using
        ), _ -> fsharpModule com ctx r t i thisArg args
     // Exceptions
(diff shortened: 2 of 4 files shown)


MangelMaxime:
Hello, thank you for the PR, can you please add some tests for it.

@ncave Is it the right way to add `using` support in Fable or should it be done by the compiler?

ThisFunctionalTom:
Can you please help me a litte 🙏 

If I add the code for the using function and tests to quickcheck.fs everything works but if I put the code int Fable.Core.Util.fs and tests into JsInteropTests and run `npm test` the fable compiler returns:
C:/Users/leko.tomas/private-source/Fable/tests/Main/JsInteropTests.fs(348,8): (348,44) error FABLE: Cannot resolve Fable.Core.Util.using
C:/Users/leko.tomas/private-source/Fable/tests/Main/JsInteropTests.fs(356,12): (356,70) error FABLE: Cannot resolve Fable.Core.Util.using

What am I doing wrong?

ncave:
@ThisFunctionalTom No, that's not where it needs to be, `Fable.Core` is mainly JS bindings and interop.
- It should be moved to `src/fable-library/FSharp.Core.fs`, somewhere close to `nullArg`.
- Also, a call for it needs to be added in `src\Fable.Transforms\Replacements.fs`.
- And a test for it needs to be added in, say, `tests\Main\ComparisonTests.fs` or `tests\Main\TypeTests.fs`.
- You can search for `nullArg` as an example to see where it needs to be added.
- Please check the resulting javascript after you build (`build\fable-library\FSharp.Core.js`) to make sure it does what you want it to do.

Should be an easy one, but please do ask questions if you need to.
Thanks for contributing!

ThisFunctionalTom:
> 
> 
> @ThisFunctionalTom No, that's not where it needs to be, `Fable.Core` is mainly JS bindings and interop.
> 
>     * It should be moved to `src/fable-library/FSharp.Core.fs`, somewhere close to `nullArg`.
> 
>     * Also, a call for it needs to be added in `src\Fable.Transforms\Replacements.fs`.
> 
>     * And a test for it needs to be added in, say, `tests\Main\ComparisonTests.fs` or `tests\Main\TypeTests.fs`.
> 
>     * You can search for `nullArg` as an example to see where it needs to be added.
> 
>     * Please check the resulting javascript after you build (`build\fable-library\FSharp.Core.js`) to make sure it does what you want it to do.
> 
> 
> Should be an easy one, but please do ask questions if you need to.
> Thanks for contributing!

@ncave Thanks for your help. 

I did this as you said but I had some troubles with running the tests. I tryed running a few scripts from package.json and afterwards it worked. I guess one have to run "build-compiler-js" script before "test" script?

The gitpod build above failed with "(503) Server Unavailable. Service Unavailable" so I guess we have to try again later or maybe it will restart automatically?

Two unrelated tests are not running: "DateTime UTC from Year 1 to 99 works". I am not sure if this is a known issue. Should I do something about it?

MangelMaxime:
@ThisFunctionalTom Indeed before running `npm run build test` you need to make sure to recompile the compiler with `npm run build compiler`.

`npm run build test` do it only if it detects that the compiler has never been build. The reason, is we don't want to do it all the time because it's too long to compile ^^

About the failing tests:

> If the failing tests are about Date/DateTime then it is "normal".
> 
> These tests kinds of work only if your laptop is configured on the UK timezone or something like that. We hardcoded the expected string/timezone because it was too much of an effort at the time to make the test adapt to the user machine configuration. And also, it could introduce bugs in our tests...

The appveyor CI seems to have some problem this morning. Even, my commit to master doesn't work, I will wait a bit and re-trigger a build to see if they fixed the issue. 

If the tests for your changes were passing on your machine don't worry too much. I will run it on my computer and if that's work I will release a new version later today.

MangelMaxime:
Thank you it has been released in fable-compiler 2.8.0

PR #2029, opened 2020-04-19
Change Option.None to undefined

### LINKED ISSUE #2028: Null is erroneously generated instead of undefined for optional arguments
```ts
		/**
		 * Create a new position relative to this position.
		 *
		 * @param lineDelta Delta value for the line value, default is `0`.
		 * @param characterDelta Delta value for the character value, default is `0`.
		 * @return A position which line and character is the sum of the current line and
		 * character and the corresponding deltas.
		 */
		translate(lineDelta?: number, characterDelta?: number): Position;
```

This is from the VSCode API. Note that the type of the first argument `lineDelta` is `number | undefined`. This is what the editor tells me when I hover over it.

```fs
    /// <summary>Create a new position relative to this position.</summary>
    /// <param name="lineDelta">Delta value for the line value, default is `0`.</param>
    /// <param name="characterDelta">Delta value for the character value, default is `0`.</param>
    abstract translate: ?lineDelta: int * ?characterDelta: int -> Position
```

Here is the binding generated by `ts2fable`. I've changed `float` to `int` by hand here.

```fs
let t = startPosition.translate(characterDelta=x.Length)
```

This is how I use it in my code. I am omitting the first argument on purpose.

```js
const t = startPosition.translate(null, x$$6[0].length);
```

This is the generated code for it. Note the `null` in the first argument. When I step through it with a debugger, I get an exception raised.

```ts
	translate(change: { lineDelta?: number; characterDelta?: number; }): Position;
	translate(lineDelta?: number, characterDelta?: number): Position;
	translate(lineDeltaOrChange: number | undefined | { lineDelta?: number; characterDelta?: number; }, characterDelta: number = 0): Position {

		if (lineDeltaOrChange === null || characterDelta === null) {
			throw illegalArgument();
		}
``` 

This is how it is implemented. I omitted the extra overload you see here in the previous description.

ncave:
- Fixes #2028.

### DIFF
--- a/src/Fable.Transforms/AST/AST.Babel.fs
+++ b/src/Fable.Transforms/AST/AST.Babel.fs
@@ -99,6 +99,9 @@ type RegExpLiteral(pattern, flags_, ?loc) =
     member __.Pattern: string = pattern
     member __.Flags = flags
 
+type Undefined(?loc) =
+    inherit Identifier("undefined", ?loc = loc)
+
 type NullLiteral(?loc) =
     inherit Literal("NullLiteral", ?loc = loc)
 
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -566,7 +566,7 @@ module Util =
                 if mustWrapOption t
                 then coreLibCall com ctx r "Option" "some" [|e|]
                 else e
-            | None -> upcast NullLiteral ()
+            | None -> upcast Undefined(?loc=r)
         | Fable.EnumConstant(x,_) ->
             com.TransformAsExpr(ctx, x)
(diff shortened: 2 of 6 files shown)


Zaid-Ajaj:
Alright so might be a huge breaking change. `null | T` was a default for `Option<T>`  for quite some time. I would say if people wanted a different behaviour then it is simply by writing a function that converts `T -> T` and `null -> undefined` to get that behaviour

ncave:
@Zaid-Ajaj I agree, we should merge this as part of Fable 3.0 beta, to avoid breaking existing code.
But, it should just work most of the time, and #2028 has a valid argument on optional parameters.

ncave:
Closing in favor of #2032 (moved to a branch).

PR #2032, opened 2020-04-20
Change Option.None to undefined

### LINKED ISSUE #2028: Null is erroneously generated instead of undefined for optional arguments

ncave:
- Fixes #2028.

### DIFF
--- a/src/fable-library/Util.ts
+++ b/src/fable-library/Util.ts
@@ -587,7 +587,7 @@ const CURRIED_KEY = "__CURRIED__";
 
 export function uncurry(arity: number, f: Function) {
   // f may be a function option with None value
-  if (f == null) { return null; }
+  if (f == null) { return undefined; }
 
   // The function is already uncurried
   if (f.length > 1) {
@@ -658,7 +658,7 @@ export function curry(arity: number, f: Function): Function | undefined {
 
 export function partialApply(arity: number, f: Function, args: any[]): any {
   if (f == null) {
-    return null;
+    return undefined;
   } else if (CURRIED_KEY in f) {
     f = (f as any)[CURRIED_KEY];
     for (let i = 0; i < args.length; i++) {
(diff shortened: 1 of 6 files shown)


ncave:
>@Zaid-Ajaj commented:
Alright so might be a huge breaking change. null | T was a default for Option<T> for quite some time. I would say if people wanted a different behaviour then it is simply by writing a function that converts T -> T and null -> undefined to get that behaviour.

@Zaid-Ajaj I agree, we should merge this as part of Fable 3.0 beta, to avoid breaking existing code.
But, it should just work most of the time, and #2028 has a valid argument on optional parameters.

alfonsogarciacaro:
In principle Fable tries to turn a deaf ear into the null/undefined difference in JS and every time there's a null check, the non-strict equality operator `==` is used which will give true both for null and undefined. There was no particular reason why `null` was chosen to represent to Option.None besides .NET influence, I guess if it improves interoperability with JS it can be changed and hopefully it doesn't break anything.

We can still use `null` to output F# `null` in Nullable types. But I'd be more cautious in going further to make the distinction in runtime (e.g. by using the strict equality operator `===` for Option.None checks).

MangelMaxime:
For me, the main reason for this change is this one:

> By looking to [MDN](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Fonctions/Valeurs_par_d%C3%A9faut_des_arguments) and [this blog post](https://codeburst.io/javascript-null-vs-undefined-20f955215a2).
> 
> It indeed seems like we should emit `undefined` instead of `null`. This allows JavaScript to pick up the default value for example.

https://github.com/fable-compiler/Fable/issues/2028#issuecomment-616079794

But, yes we probably should not make the equality operator too strict as @alfonsogarciacaro said.

ncave:
> But, yes we probably should not make the equality operator too strict as @alfonsogarciacaro said.

@MangelMaxime I agree, that's why all the comparisons in this PR are left as they were (non-strict).

src/fable-library/Util.ts
```diff
@@ -671,25 +671,25 @@ export function partialApply(arity: number, f: Function, args: any[]): any {
         // Wrap arguments to make sure .concat doesn't destruct arrays. Example
         // [1,2].concat([3,4],5)   --> [1,2,3,4,5]    // fails
         // [1,2].concat([[3,4],5]) --> [1,2,[3,4],5]  // ok
-        return (a1: any) => f.apply(null, args.concat([a1]));
+        return (a1: any) => f.apply(undefined, args.concat([a1]));
```
alfonsogarciacaro, line 674: Just aesthetics, but for `apply` it seems either `null` or `undefined` will work and the examples in MDN use `null`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

alfonsogarciacaro:
I will test the branch with a couple of my projects to check there are no issues :+1:

src/fable-library/Util.ts
ncave: @alfonsogarciacaro Yes, the second note in the MDN article says `When the first argument is undefined or null ...`

alfonsogarciacaro:
Sorry, I totally forgot about this! 🤦 I just tested with one project and everything seems to be working fine, so I'm OK with merging the PR :+1:

PR #2036, opened 2020-04-28
Several fixes

alfonsogarciacaro:
**EDIT** I've removed the second commit as it seems to cause problems with fable standalone.

This fixes some issues I've been finding recently in my projects:

- `Fable.Core.Experimental.nameofLambda` didn't work in inlined functions
- ~~Local import expressions (except for `importMember`) were failing when written as a function. See "Local import with curried signatures works" in JsInteropTests.~~
- Because Fable adds a hash to class members to tell overloads apart, when the hash changes (e.g. when changing the number or type of the arguments) the files calling this member don't get recompiled by Webpack and we get warnings (see screenshot below). Now calls to such members add to the `InlineDependencies` so fable-loader recompiles the dependent files (as with inline functions).
- Don't re-mangle imported members that use the dollar sign (also see screenshot).

screenshot

### DIFF
--- a/tests/Main/DateTimeOffsetTests.fs
+++ b/tests/Main/DateTimeOffsetTests.fs
@@ -37,9 +37,9 @@ let tests =
         |> equal "2014-09-11T16:37:02.000+00:00"
 
     testCase "DateTimeOffset from Year 1 to 99 works" <| fun () ->
-        let date = DateTimeOffset(1, 1, 1, 0, 0, 0, TimeSpan.Zero)
+        let date = DateTimeOffset(1, 1, 2, 0, 0, 0, TimeSpan.Zero)
         date.Year |> equal 1
-        let date = DateTimeOffset(99, 1, 1, 0, 0, 0, TimeSpan.Zero)
+        let date = DateTimeOffset(99, 1, 2, 0, 0, 0, TimeSpan.Zero)
         date.Year |> equal 99
 
     // TODO: These two tests give different values for .NET and JS because DateTimeOffset
--- a/src/Fable.Core/Fable.Core.JsInterop.fs
+++ b/src/Fable.Core/Fable.Core.JsInterop.fs
@@ -82,7 +82,7 @@ let importSideEffects (path: string): unit = jsNative
 let importDynamic<'T> (path: string): JS.Promise<'T> = jsNative
 
 /// Imports a reference from an external file dynamically at runtime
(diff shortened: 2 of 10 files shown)


tests/Main/DateTimeOffsetTests.fs
```diff
@@ -37,9 +37,9 @@ let tests =
         |> equal "2014-09-11T16:37:02.000+00:00"
 
     testCase "DateTimeOffset from Year 1 to 99 works" <| fun () ->
-        let date = DateTimeOffset(1, 1, 1, 0, 0, 0, TimeSpan.Zero)
+        let date = DateTimeOffset(1, 1, 2, 0, 0, 0, TimeSpan.Zero)
```
MangelMaxime, line 40: @alfonsogarciacaro Were the tests failing? 

I am asking because changes to Dates tests are always tricky. And in the past, the tests were configured to only work on a machine configured on the UK timeline or something like that.

alfonsogarciacaro:
The first release I made in a long time and I managed to break everything :/ I had to revert also the [fourth commit](https://github.com/fable-compiler/Fable/commit/c35d0357a7a74d919def7f9812a4eec71334d1b0) because it broke my project when compiling. It seems the import identifiers are clashing with functions declared in the file, but [this should already be dealt with](https://github.com/fable-compiler/Fable/blob/c35d0357a7a74d919def7f9812a4eec71334d1b0/src/Fable.Transforms/Fable2Babel.fs#L1643) so I'm not entirely sure why it's happening.

tests/Main/DateTimeOffsetTests.fs
alfonsogarciacaro: It fails in Japan timezone ;)

You're right, I should have taken care of the problem. But it seems to be a very obscure error (only?) happening when creating a date 100 AD in Japan timezone. I suspect it's a mix of the fact [we already have a workaround for dates with year <= 99](https://github.com/fable-compiler/Fable/blob/d4106ca232bed645ba63de601387c8847ee05a32/src/fable-library/DateOffset.ts#L96-L98) and that Japan timezone changed from +0918 (zulu time) to +0900 [at some point in history](https://bugs.chromium.org/p/chromium/issues/detail?id=867806). So I hope it doesn't really affect users (fingers crossed).

MangelMaxime: I checked on my machine and it is working so let's keep it like that.

PR #2037, opened 2020-04-28
Fix local imports

alfonsogarciacaro:
Local import expressions (except for importMember) were failing when written as a function. See "Local import with curried signatures works" in JsInteropTests.

> The fix seems to cause an issue with fable-standalone or fable-compiler-js. Investigating.


alfonsogarciacaro:
Fixed by 5538384664289648b0bde3e2f0ec1c0eceab5fd6

Issue #2043, opened 2020-04-29
Unexpected end of JSON input at JSON.parse in dist/index.js


forki:
### Description

On our CI system we see strange flaky compile errors when building our electron app:

    C:\WORK\fdgC2beP\0\products\msu.projectmanagement\electron\node_modules\fable-compiler\dist\index.js:80
                    resolve(JSON.parse(data.substr(pattern[0].length)));
                                ^
    SyntaxError: Unexpected end of JSON input
        at JSON.parse (<anonymous>)
        at Interface.linereader.on (C:\WORK\fdgC2beP\0\products\msu.projectmanagement\electron\node_modules\fable-compiler\dist\index.js:80:30)
        at Interface.emit (events.js:198:13)
        at Socket.onend (readline.js:171:12)
        at Socket.emit (events.js:203:15)
        at endReadableNT (_stream_readable.js:1129:12)
        at process._tickCallback (internal/process/next_tick.js:63:19)
    error Command failed with exit code 1.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    Running build failed.
    Error:
    System.Exception: Error while running 'C:\Program Files (x86)\Yarn\bin\yarn.cmd' with args: run compile

This points to https://github.com/fable-compiler/Fable/blob/fe87afbc3673840f8f5c83808f23f13958f2a0b0/src/fable-compiler/src/index.ts#L92

### Related information

* Fable version:  2.8.3
* Operating system: Windows

cc @atheck @MangelMaxime @ncave @alfonsogarciacaro

alfonsogarciacaro:
Usually I see this error when I put a `printfn` by mistake somewhere and it interrupts the JSON writing in stdout. However in the current code there are only two places writing to stdout [json](https://github.com/fable-compiler/Fable/blob/35225105910eb5626bb09a00ddb5903c0d545b3d/src/Fable.Cli/Stdin.fs#L11-L15) and [logs](https://github.com/fable-compiler/Fable/blob/35225105910eb5626bb09a00ddb5903c0d545b3d/src/Fable.Cli/Util.fs#L129-L131) and they are both protected by the same lock.

[...]

PR #2044, opened 2020-04-30
Better error logging for #2043

forki:
Me and @atheck would like to have better understanding of #2043.

I hope this can give more information next time the error arises


MangelMaxime:
It has been released in `fable-compiler 2.9.0`

alfonsogarciacaro:
My only concern with this is the log is going to be quite large because the json is usually big, but well I'll guess we'll see why the json is malformed.

MangelMaxime:
Indeed, it will large because if that helps debugs the issue better than nothing.

Perhaps, we will see a pattern which allows us to only print the malformed section of the JSON in a future release.

PR #2050, opened 2020-05-06
[Not working] Prototype implementation for support TypeScript with fable-splitter

MangelMaxime:
The implementation doesn't work yet because we need a way to handle the TS imports.

Indeed, when detecting an import of an existing TypeScript file we should copy the TS file without passing via Babel and also detect all the import/require from that file to copy with and so on.

I think we need a different toolchain when handling JavaScript or TypeScript generation.

@ncave Any idea, how to get the TypeScript imports? One solution I had is to use the TypeScript compiler in order to get the AST but getting the imports list from it doesn't seem to be easy.

The idea is to remove the `fable-library` folder and have `fable-library-js` and `fable-library-ts` and Fable choose which folder to use depending on the target.

### DIFF
--- a/src/Fable.Cli/Util.fs
+++ b/src/Fable.Cli/Util.fs
@@ -80,8 +80,8 @@ type GlobalParams private (verbosity, forcePkgs, fableLibraryPath, workingDir) =
               |> Path.GetDirectoryName
             let defaultFableLibraryPaths =
                 [ "../fable-library"                         // running from npm package
-                  "../../fable-library/"                     // running from nuget package
-                  "../../../../../build/fable-library/" ] // running from bin/Release/netcoreapp2.0
+                  "../../fable-library"                     // running from nuget package
+                  "../../../../../build/fable-library" ] // running from development release located at bin/Release/netcoreapp2.0
                 |> List.map (fun x -> Path.GetFullPath(Path.Combine(execDir, x)))
             let fableLibraryPath =
                 defaultFableLibraryPaths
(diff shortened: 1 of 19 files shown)


src/Fable.Cli/Util.fs
```diff
@@ -93,7 +93,18 @@ type GlobalParams private (verbosity, forcePkgs, fableLibraryPath, workingDir) =
 
     member __.Verbosity: Fable.Verbosity = _verbosity
     member __.ForcePkgs: bool = _forcePkgs
-    member __.FableLibraryPath: string = _fableLibraryPath
+    member __.FableLibraryPath (isTypeScript : bool) : string =
```
MangelMaxime, line 96: There is also still some strange implementation/hacks that are here just for the prototype. We should try to clean this before releasing a "test" version advertising beta support for TypeScript.

ncave:
@MangelMaxime I'm not sure, pehaps some of those changes can be avoided, perhaps we can leave everything working just as it is today, only bundle an additional `fable-library-ts` folder that would be copied to the target path as `fable-library` to replace the one there (if the `--typescript` flag is enabled).

MangelMaxime:
Indeed perhaps but I don't think this will allow fable-splitter to copy/paste the files that the user code imported.

For example, if he wrote a `.ts` or `.js` file manually and used an import to import them. 

It is worth to note that I wrote the prototype also to understand how everything is wired in Fable toolchain. But I figure out it could be nice to share for brainstorming how to activate support of the now typescript generation.

PR #2053, opened 2020-05-12
Better error tracing for #2043

forki:
Another attempt for https://github.com/fable-compiler/Fable/issues/2043

### DIFF
--- a/src/fable-compiler/src/index.ts
+++ b/src/fable-compiler/src/index.ts
@@ -98,7 +98,7 @@ export default function start(cliArgs?: {}): ICompilerProxy {
                 catch(error)
                 {
                     log(cliArgs, "JSON parsing failed on " + json);
-                    throw(error);
+                    throw("JSON parsing failed on " + json +  " Error: " + error);
                 }
             }
         } else { // LOG


alfonsogarciacaro:
There has to be a better way to corner this error than pushing a new release of fable-compiler for each attempt :) I would follow @ncave's recommendation here and use `npm link` so you don't need to wait for a new release for every attempt:

- Copy `node_modules/fable-compiler` somewhere, say `temp/fable-compiler``
- Do the necessary changes on `temp/fable-compiler/dist/index.js`
- Link your local version at the beginning of your build script: `npm link temp/fable-compiler`
- Run the build
- Unlink the local version at the end of your build script: `npm unlink temp/fable-compiler`

This should allow you to check things faster and when you get with the way to spot the error you can send the PR to fable-compiler :)

ncave:
@forki @alfonsogarciacaro Or just reference the package from a [local path](https://docs.npmjs.com/files/package.json#local-paths), if you have that flexibility.

A quick reminder, just in case, that npm linking (and unlinking) is a two step process as described [here](https://docs.npmjs.com/cli/link). There is a shortcut that can combine the two steps in one for linking, but not for unlinking.

alfonsogarciacaro:
If you've upgraded to Fable 3 I hope you're not seeing this error any more, please feel free to reopen otherwise.

PR #2056, opened 2020-05-16
Added support for classes

ncave:
- Fixed some type gen for `fable-library`, now compiles to strict `typescript` without errors. 
- Added  support for compiling types as classes (optional, behind a compiler switch `--classTypes`)
- Here is the performance breakdown (compiling Fable with Fable, Node.js v14.2.0, Win10):

With factory functions (no classes):
```
   ticks  total  nonlib   name
-------------------------------------
  15335   40.9%   95.2%  JavaScript
   5359   14.3%   33.3%  GC
  21362   57.0%          Shared libraries
    770    2.1%          Unaccounted
-------------------------------------
  37467   100.0%         Total
```
With classes (class constructors):
```
   ticks  total  nonlib   name
-------------------------------------
  15516   30.9%   95.3%  JavaScript
   5489   10.9%   33.7%  GC
  33895   67.6%          Shared libraries
    761    1.5%          Unaccounted
-------------------------------------
  50172   100.0%         Total
```
This is without any effort to optimize it, so currently using classes is slower by `50172/37467 = 1.33x (33%)`. But perhaps there is some use for it in certain cases, and it can be optimized in the future.
(one possible use case is to have an alternative for using classes when compiling to typescript).


alfonsogarciacaro:
Thanks!

PR #2065, opened 2020-05-28
Fixed #2064

### LINKED ISSUE #2064: GetHashCode seems to be behaving strange in some circumstances
### Description

I seem to be getting different behaviour between Fable's built in "hash" function and Object.GetHashCode(). Hash seems to be correct, but GetHashCode() is not working comprehensively for tuples, lists, and maps. It seems to produce a non-deterministic output for each call, although the sequence seems to be.

Normally I would just go ahead and use hash, but my particular case is using an evil mutable dotnet dictionary, which I believe uses GetHashCode internally. This seems to be manifesting in some unpredictable behaviour.

Fable.Core issue I think. I think this is the right repo?

### Repro code

```
module GetHashCode

("hello", 1).GetHashCode() = ("hello", 1).GetHashCode() |> printf "%A"
(("hello", 1) |> hash)= (("hello", 1) |> hash) |> printf "%A"

["hello", 1].GetHashCode() = ["hello", 1].GetHashCode() |> printf "%A"
(["hello", 1] |> hash)= (["hello", 1] |> hash) |> printf "%A"

(Map.ofList ["hello", 1]).GetHashCode() = (Map.ofList ["hello", 1]).GetHashCode() |> printf "%A"
(Map.ofList ["hello", 1] |> hash)= (Map.ofList ["hello", 1] |> hash) |> printf "%A"
```

[fable repl example](https://fable.io/repl/#?code=LYewJgrgNgpgBAcRgFwBIEMDOALAwuGAKEIAoAibGKKEMgGjgEYBKAOiTSzwJObgF445StVoMW7FBhz4wMXnAA+APjgAHAE4BLAHbIAZnDIBSAIJlSwqjXpM+KuNi7NBJK6NsslqpznurNXQMjMwtCAG0KazEmAF1JThkePkFIkRtxeI5pbjkFB0C9QxNzUjToz1jvR2dXco9M6t9sf3VtIpDS0gBZdDVWEH0AGS1MZDh6jLi2bK5ZeRShXv7BkbGJqIbphJz5-ID24JKLEmWB4dHxyZjGKodmlyW+87WrzanbpudqwqPQoA&html=DwCwLgtgNgfAsAKAAQqaApgQwCb2ag4CdMTJcMABwFp0BHAVwEsA3AXgCIBhAewDsw6AdQAqAT0roOSAMb9BAzoIAeYAPThoAbhkhMAJwDOJNgzAAzagA4OeQhqy5EhAEY9sYu6mBq3HvD6asEA&css=Q)

Both cases should return true, at the moment the GetHashCode version returns false, which implies it is not deterministic for multiple calls.

### Related information

* Fable.Core: 3.1.5
* fable-compiler 2.8.1

Alternatively, this breaks on whatever the version running on fable.io/repl also. (see link above).

ncave:
- Fixed GetHashCode() for Decimal, Union, Record, Option, Tuple, List, Set, Map, Choice and Result.

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -714,7 +714,7 @@ let identityHash r (arg: Expr) =
     | Builtin(BclGuid | BclTimeSpan | BclDateTime | BclDateTimeOffset)
     | Builtin(FSharpSet _ | FSharpMap _ | FSharpChoice _ | FSharpResult _) ->
         Helper.CoreCall("Util", "structuralHash", Number Int32, [arg], ?loc=r)
-    | DeclaredType(ent,_) when ent.IsValueType ->
+    | DeclaredType(ent,_) when ent.IsFSharpUnion || ent.IsFSharpRecord || ent.IsValueType ->
         Helper.CoreCall("Util", "structuralHash", Number Int32, [arg], ?loc=r)
     | _ ->
         Helper.CoreCall("Util", "identityHash", Number Int32, [arg], ?loc=r)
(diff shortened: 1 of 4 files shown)


alfonsogarciacaro:
Thanks!

Issue #1314, opened 2018-01-06
Importing value as mutable makes the value a function when accessing it


Zaid-Ajaj:
### Description
When importing a value from some javascript module and annotating it with `mutable`, the value becomes a function when you want to access it. 

### Repro code
The following works:
```fs
open Fable.Core
open Fable.Core.JsInterop

let someValue : string = import "someValue" "someModule"
let value = someValue
```
Compiles to JS
```js
import { someValue as someValue_1 } from "someModule";
export const someValue = someValue_1;
export const value = someValue;
```
But as soon as you make the imported value mutable:
```fs
open Fable.Core
open Fable.Core.JsInterop

let mutable someValue : string = import "someValue" "someModule"
let value = someValue
```
The compilation is incorrect:
```js
import { someValue as someValue_1 } from "someModule";
export let someValue = someValue_1;
export const value = someValue();
```
Giving me all kinds of `someValue is not a function` errors
This error happens regardless of using `import` or `importMember`

* Fable version (`dotnet fable --version`): 1.3.7
* Operating system: Linux Ubuntu 17

alfonsogarciacaro:
This is intended as it was necessary to fix #986 but I didn't take into account importing values, I'll take a look. Thanks for the report!

alfonsogarciacaro:
Latest Fable will throw an error when trying to make an imported (or global) value both mutable and public, asking the user to make it private. Do you think this is enough @Zaid-Ajaj or should we try to enable mutable public imported values too?

Zaid-Ajaj:
I personally don't mind but I think users will expect this to just work the same way it does with normal F# on dotnet

alfonsogarciacaro:
Well, in F# dotnet you cannot import values from a JS library :)

Zaid-Ajaj:
Oh sorry my bad, I read "mutable public value" hahah. Yes, a warning is the best solution, I guess as a workaround you can import some js function that does that mutation :smile:

alfonsogarciacaro:
You can still mutate imported/global values, you just need to make the value private. I think this is safer to prevent too much exposure to mutation from JS libraries. Cool, let's close then and maybe we can revisit if some users need the feature for any reason.

PR #2066, opened 2020-06-01
[WIP] Plugins come back!

alfonsogarciacaro:
After all the problems we're having with the declaration of React components (see for example https://github.com/fable-compiler/fable-react/pull/195) I'm trying now to implement again a plugin system for Fable (as it was in the very first version), so we can declare components through an attribute (as [it's done in ReasonReact](https://reasonml.github.io/reason-react/docs/en/components)) and use some compiler magic to overcome the issues we're having to declare React function components in F#.

This new plugin system tries to solve two of the main problems we had in Fable 1:

- Plugins had access to the full Fable AST, so any change in the AST broke all the plugins.
- Assemblies containing the plugins had to be referenced in Fable options, which made it difficult for a library to depend on a plugin, as there was no way to warn the user.

In this proposal:

- Plugins only have access to a simplified version of the Fable AST that will live in a different package and (hopefully) should be more stable and easier to maintain than Fable full AST.

> The drawback is the plugins won't be able to do complex code transformations, but this may also be a good thing. The main purpose of the plugins will be to "wrap" the generated code.

- Attributes themselves can become plugins just by implementing an interface and can be declared in the library that needs them.

- Another advantage is that (I believe) we can make these plugins work with the JSified Fable so they could work in the REPL too.

-------------------------

It's not done yet, but the idea is in Fable.React we can declare an attribute similar to this:

```fsharp
open Fable.SimpleAst

type ReactComponentAttribute(?memoize: bool) =
    inherit System.Attribute()
    interface TransformDeclaration with
        member _.TransformDeclaration(logger, fullName, decl) =
            match decl with
            | ValueDeclaration _ -> logger.LogWarning("Expecting a function"); decl
            | FunctionDeclaration(args, body) ->
                // Do some magic here to turn the function into a proper React component
```

And use it as follows:

```fsharp
[<ReactComponent>]
let root model dispatch =
   // render code
```

Thoughts? @MangelMaxime @ncave @Zaid-Ajaj @forki et al


MangelMaxime:
If it can help improve Fable interop I think this is good.

I mean using the attribute + plugin combination, we can isolate a specific interop problem and take strong decision to solve as it will not break impact all the code generated by Fable.

alfonsogarciacaro:
@CedricDumont Well, you had to write a plugin for that but I think it should be possible. Though it's something I wouldn't personally recommend, as exporting mutable values is dangerous, see #1314

CedricDumont:
@alfonsogarciacaro I understand your point about mutable values (and totally agree with you), but...
- fable does compile F# to javascript 
- after this process, a webpack loader might also be applied to that generated js 
- that particular loader might want to check for `export let` variable and do something with them. 

=> that is my case 

so the generated javascript from fable might not always be the last step in the build process, and other frameworks might rely on javascript's strange behavior...

PR #2067, opened 2020-06-01
Updated class types support

### LINKED ISSUE #2073: Inheriting from secondary constructors is broken
See [this](https://github.com/fable-compiler/Fable/commit/d7f29b5dc4c4c1ab497e92e08704dd90c107142e).

ncave:
- Updated class types support.
- Renamed typeDecls flag.
- Fixed #2073.


alfonsogarciacaro:
Thanks @ncave! Two questions about compiling using classes:

- Is the performance degrade due to the fact that somewhere in the build we're compiling JS classes back to functions to make them ES5 compatible? If it is, maybe classes work well with JS engines compatible with ES6?

- One of the main problems to compile to classes was when trying to derive a class using a secondary constructor (see code below), as JS classes only support one constructor. Did you manage to fix that?

```fsharp
type A(v: int) =
    member _.Value = v
    new () = A(5)


type B() =
    inherit A()
```

ncave:
@alfonsogarciacaro 

- No, this is without down-compiling to ES5, just normal ES6 classes. Not sure where exactly the perf is being lost, somewhere in Node.js system libraries, as JS code time slice is the same, and GC portion is the same.
- No, I didn't attempt to fix that, so the wrong constructor is being called. But it's the same case without classes too, so no change there (perhaps we can open an issue to keep track of it).

alfonsogarciacaro:
> No, I didn't attempt to fix that, so the wrong constructor is being called. But it's the same case without classes too, so no change there.

Hmm, making inheritance with secondary constructors was one of my goals for Fable 2. [It seems to work in the REPL](https://fable.io/repl/#?code=C4TwDgpgBAyhDGB7AdgEwIYCcQGEUGcAKADwC4oBLZYASigF4BYAKCjamQgHcpC77YCFBmx5kRAKw0W7KAFsIcgEYRMUAPrqAdADV0AGwCu0AcRYtQkQUjRZcBHAAsK+1HwYz2VR6orBrwnZiRNLMLPoQ-sAQ+MDuTKzsEf74DAG2og7OrnyebGCYVMAAZshQAEQApBTlUPi6BsbmzNGxfEA&html=Q&css=Q), though I just noticed we didn't have a specific test for it. Do you have an example of code not working?

alfonsogarciacaro:
Ok, this is very interesting. Inheriting from secondary constructors is working in the REPL but not [in this commit](https://github.com/fable-compiler/Fable/commit/d7f29b5dc4c4c1ab497e92e08704dd90c107142e). Which means the feature got broken in one of latest commits after last repl updated and we didn't notice because I forgot to add a test 😕

ncave:
@alfonsogarciacaro Yes, I broke it [here](https://github.com/ncave/Fable/blob/master/src/Fable.Transforms/FSharp2Fable.Util.fs#L1203-L1206), I'll fix it.

ncave:
@alfonsogarciacaro I fixed #2073 in the non-class case, and added your tests, so this PR should be good to merge.

I still haven't figured out how to do it for classes, but classes are still considered an experimental feature. In the mean time, while we definitely need to make this robust and working properly for classes, but perhaps that particular type of secondary constructor usage in inherited classes is relatively rare, for example FCS compiles and runs just fine with classes, it's just slower. There are quite a few secondary constructors in FCS, just not inherited from secondary constructors.

alfonsogarciacaro:
Awesome, thank you! I'm ok with removing support for inheriting from secondary constructors, but we should warn the user if they try to do that. IIRC the main problem for that is `.IsImplicitConstructor` doesn't work when you reference a type from a signature `.fsi` file.

alfonsogarciacaro:
Seems I managed to find the implicit constructor even when referencing a type in a signature file, see #2074

PR #2075, opened 2020-06-04
Resoved conflicts

ncave:
@alfonsogarciacaro I resolved what I could, without knowing much about the original PR.
It compiles, but reflection tags are probably wrong, since some don't exist anymore, you'll see when you run the tests. For example, this is probably not valid anymore:

Fable2Babel.fs, line 800:
```fs
        | Fable.Any -> primitiveTypeInfo "obj"
```

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -105,16 +105,16 @@ module Helpers =
         ||> Naming.sanitizeIdent (fun _ -> false)
 
     let getModuleReflectionName (com: ICompiler) (ent : FSharpEntity) =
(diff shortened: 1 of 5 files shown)


alfonsogarciacaro:
Thanks a lot @ncave, this is super helpful! I will try to fix the other errors myself :+1:

alfonsogarciacaro:
Hmm, at the end solving the issues is much trickier than I hoped so I'm not sure anymore. Maybe we should just keep working on master and add more reflection support as we need in a way that plays better with tree shaking? (e.g. generating member info only where it's needed)

ncave:
I'm not sure what the scope of the original PR is, is your plan to gradually add more reflection so eventually quotations can work?

PR #2084, opened 2020-06-12
Mangle abstract members

### LINKED ISSUE #2082: Mangle name of abstract members
See #2068. Avoid name conflicts in attached members by mangling  abstract members (at least some). Rules:

- To avoid breaking changes interfaces won't be mangled by default (this means they can still conflict) unless:
    - They come from `System` namespace, with [these exceptions](https://github.com/fable-compiler/Fable/blob/596502bbe7b7c4d56dd81980bee88006e87ddaa9/src/Fable.Transforms/FSharp2Fable.Util.fs#L1038-L1048) that are used in fable-library Typescript files.
    - They are decorated with the `Mangle` attribute.

-  Members from abstract classes will always be mangled except if the class is imported (e.g. the `ReactComponent` class).

alfonsogarciacaro:
Fix #2082 

@ncave I've already started work to mangle abstract members in case you want to have a look. One thing I've noticed is we use several interface methods from Typescript in fable-library (besides `Equals` and `GetHashCode`) like `Compare` or `GetZero`. We need to decide if we want to add those as exceptions for the mangling or we change them in Replacements/fable-library.


ncave:
@alfonsogarciacaro I'm not sure, what if some other interface has a "Compare" method with, say, 3 arguments.

alfonsogarciacaro:
I'm assuming that would be ok because the `Compare` method from the other interface would be mangled. I'm concerned what of the following would be the lesser of two evils:

- Use mangled names in Typescript files of fable-library for equality and comparison
- Have too many exceptions to the mangle rule 

> Hopefully the exceptions would only be System.Object, IComparable, IComparer and IEqualityComparer (though I'm not sure if we'll have issues with the generic and untyped versions). There are also the interfaces we use for injection, but I guess we can decorate those directly with `NoMangle`:

https://github.com/fable-compiler/Fable/blob/7d02d3e690a9b15cb7ca823b90c20f46f13f2a37/src/Fable.Transforms/Transforms.Util.fs#L90-L92

src/fable-library/Util.ts
```diff
@@ -61,16 +61,6 @@ export interface IEqualityComparer<T> {
   GetHashCode(x: T): number;
 }
 
-export interface ICollection<T> {
-  readonly Count: number;
```
ncave, line 65: @alfonsogarciacaro Did that interface move somewhere else?

src/fable-library/Types.ts
```diff
@@ -35,15 +35,19 @@ export function declare(cons: any, superClass?: any) {
...
+    return this.ToString();
+  }
+
+  public ToString() {
+    return this.constructor.name;
   }
 
-  public GetHashCode(x?: any) {
-    return identityHash(x ?? this);
+  public GetHashCode() {
```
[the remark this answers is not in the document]
alfonsogarciacaro: Ah, ok. So that was the reason! I didn't know why it was like that and I wanted to clean the interfaces so I removed, but it should be ok to put it back :+1:

src/fable-library/Util.ts
alfonsogarciacaro: No, I just noticed it's not used anywhere so I removed it. If you need it we can put it back though right now all calls to `ICollection` members are replaced with something else so I don't know if there's any type implementing it.

Unfortunately I realized it's difficult to use mangled interfaces in Typescript because of the overload hash (we could try though) so I had to [add these as exceptions for the mangling](https://github.com/fable-compiler/Fable/blob/596502bbe7b7c4d56dd81980bee88006e87ddaa9/src/Fable.Transforms/FSharp2Fable.Util.fs#L1038-L1049) and tried to keep them tracked.

Thanks a lot for your comments @ncave! BTW, could you please also take a look at `Fable2Babel`? Not sure how the changes to the compilation of attached members affect the compilation with classes. At least I guess we need to check if the interface is mangled or not here: https://github.com/fable-compiler/Fable/blob/596502bbe7b7c4d56dd81980bee88006e87ddaa9/src/Fable.Transforms/Fable2Babel.fs#L1808-L1810

alfonsogarciacaro:
At the end, I'm trying to implement this in a way that avoids breaking changes. See the updated rules: https://github.com/fable-compiler/Fable/issues/2082

src/fable-library/Util.ts
ncave: @alfonsogarciacaro It's being used [here](https://github.com/fable-compiler/Fable/blob/master/src/Fable.Transforms/Fable2Babel.fs#L1747).

alfonsogarciacaro:
Wow, this was harder than expected but it finally seems to be working. I put back the interfaces so If you don't have any other @ncave we can merge to `next` :+1:

PR #2103, opened 2020-06-23
Cache for JS files

### LINKED ISSUE #2102: Cache for JS files
Related #2079 

We've been working to make watch compilations fast but the first compilation still takes a long time. In normal workflow (at least mine) developers spin the development server before touching any file, so we could cache the generated JS files from last compilation and serve them while the F# compiler/Fable are warming up.

- Implement a file cache in fable-loader as an opt-in option.
- Make project parsing in Fable not-blocking. After cracking a file, the Fable should respond quickly so the JS client is not blocked and can activate the cache.
- TODO: implement cache in fable-splitter too.

Initial tests show this can reduce the compilation time to half the time (from around 40 s to 20 s in my case).

alfonsogarciacaro:
Fix #2102


alfonsogarciacaro:
Apparently it's working, but needs some testing. I've refactored the Agent.fs file and there may be some glitches in the error reporting.

PR #2108, opened 2020-06-25
[WIP] Witnesses for STRP resolution

### LINKED ISSUE #2104: Witnesses
Related #2101 #2083 

SRTP resolution in Fable is not good and fails many times as in the issues reported above. This is because the F# TypedTree didn't expose much information and Fable had to make its own rudimentary resolution.

Fortunately this is about to change thanks to Don Syme work to include witnesses for SRTP resolution in the TypedTree and expose them in the FCS API: https://github.com/dotnet/fsharp/pull/9510

Use the new FCS to take advantage of the witnesses and use the same STRP resolution as the F# compiler does.

alfonsogarciacaro:
Fix #2104 

Take advantage of the new "witness passing" feature from F# compiler to improve (or actually fix) SRTP resolution in Fable. Discussion here: https://github.com/dotnet/fsharp/pull/9510#issuecomment-648637543


alfonsogarciacaro:
I've rebased this with `nagareyama`. I would like to focus on merging this next as witnesses are going to be a super-useful feature for Fable and we should take advantage for it in Fable 3 (better SRTP resolution, not having to inline functions from external packages). I guess it's not going to be an easy one, but if you could have a look at https://github.com/dotnet/fsharp/pull/9510#issuecomment-648637543 to see if you can find a hint to move forward that'd be of great help! 🙏

PR #2114, opened 2020-07-10
Implement boolean parsing (#2111)

### LINKED ISSUE #2111: System.Boolean.TryParse
### Description

Fable might be missing an implementation of System.Boolean.TryParse method. 
I get the following error:
"Error FABLE: Cannot resolve System.Boolean.TryParse"

Many other parsers such as Int, Date, Double exists.

### Repro code

If you try using System.Boolean.TryParse you get a compiler error.
`System.Boolean.TryParse "true"`

```
let (couldParse, parsedValue) = System.Boolean.TryParse "false";;
val parsedValue : bool = false
val couldParse : bool = true
```
Described here: [System.Boolean.TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.boolean.tryparse?view=netcore-3.1)

inosik:
Fixes #2111.

The error message is the one from .NET Core 3.1, it might be different on .NET Framework.

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -1949,9 +1949,9 @@ let optionModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: E
 
 let parseBool (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
     match i.CompiledName, args with
-    | ("Parse" | "TryParse" as method), [arg] ->
+    | ("Parse" | "TryParse" as method), str::_ ->
         let func = Naming.lowerFirst method
-        Helper.CoreCall("Boolean", func, t, [arg], [arg.Type], ?loc=r)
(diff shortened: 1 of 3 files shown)


inosik:
> The error message is the one from .NET Core 3.1, it might be different on .NET Framework.

Hm, now this is strange. The error messages seem to be different on Travis and AppVeyor 🤔

alfonsogarciacaro:
Thanks a lot @inosik! Hmm, maybe the error message is slightly different depending on the dotnet sdk version? Anyways, it's probably not so important to have the exact same error message. Would it be possible to change the test so it just checks whether the operation throws or not?c

inosik:
Looking at the logs, they both have SDK 3.1.301. Maybe it's some weird Linux vs. Windows thing, but I don't have a Windows machine right now to verify. On macOS I get the same error message as on Travis.

However, I changed the test 👍🏻

alfonsogarciacaro:
Awesome, thanks! I'll publish a new release with the fix 👏

PR #2130, opened 2020-07-26
Don't allow null or empty string when using DateTime.Parse

### LINKED ISSUE #2128: DateTime.Parse with invalid inputs give back real dates
## Description
`DateTime.Parse` with either an empty string or null give back valid dates. I would expect them to throw an exception.

## Repro Code
```fsharp
printfn "Empty String: %O" (DateTime.Parse "")
printfn "Null: %O" (DateTime.Parse null)
````

## Expected and actual results
Expected: Exceptions are throw on either statements.
Actual:

```
Empty String: Sat Jul 25 2020 00:00:00 GMT+0200 (Central European Summer Time)
Null: Thu Jan 01 1970 01:00:00 GMT+0100 (Central European Standard Time)
```

## Related Information
Fable Version: 2.10.2
Operating System: Windows 10

Zaid-Ajaj:
Fixes #2128

### DIFF
--- a/src/fable-library/Date.ts
+++ b/src/fable-library/Date.ts
@@ -164,12 +164,20 @@ export function maxValue() {
   return DateTime(253402300799999, DateKind.Unspecified);
 }
 
-export function parseRaw(str: string) {
-  let date = new Date(str);
+export function parseRaw(input: string) {
+  if (input === null) {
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks!

PR #2138, opened 2020-08-18
Make Fable AST independent of FCS

### LINKED ISSUE #2087: Better names in JS
In Fable 2 I tried to avoid name conflicts using somewhat complex rules for name mangling. This has fulfilled the purpose but it has made the variable names in the generated JS difficult to read some times. If possible we could try to improve the situation by:

- Not using so many `$`
- Using two scopes to prevent name conflicts: global (for each file) and member (for each module function), so we don't have to mangle argument names if they're repeated in different module functions.

> For the second point we need to be careful as the names of imported values are assigned in the last pass (Fable2Babel) and we need to make the sure they don't conflict with any variable in global or local scopes.

alfonsogarciacaro:
This was supposed to fix #2087 but then I started another AST cleanup, tried to make Fable.AST independent of FCS and then everything become red 😮 Finally the project builds again but I need to fix the tests. It's gonna be difficult to make a proper review because the PR is too big and the reasons behind many of the changes are only in my head atm, but if you have a moment @ncave please do have a look to see if this makes sense to you.

Next steps I think this could unblock:

- Run formatter on the files Fable.Transform. Most of the lines have been touched anyways.
- Compile to class types
- Don't use inheritance for equality/comparison as discussed here: https://github.com/fable-compiler/Fable/pull/2124#issuecomment-663030527
- Move Fable.AST to a separate project, which should be available to plugins
- Move Replacements out of Fable.Transforms, and make them look more like "regular" plugins. Hopefully this should help make the compiler more modular and easier to maintain/contribute

### DIFF
--- a/src/Fable.Core/Fable.Core.JsInterop.fs
+++ b/src/Fable.Core/Fable.Core.JsInterop.fs
@@ -34,8 +34,13 @@ let (==>) (key: string) (v: obj): string*obj = jsNative
 let createNew (o: obj) (args: obj): obj = jsNative
 
 /// Destructure a tuple of arguments and applies to literal JS code as with EmitAttribute.
-/// E.g. `emitJs "$0 + $1" (arg1, arg2)` in JS becomes `arg1 + arg2`
-let emitJs (jsCode: string) (args: obj): 'T = jsNative
+/// E.g. `emitJsExpr (arg1, arg2) "$0 + $1"` in JS becomes `arg1 + arg2`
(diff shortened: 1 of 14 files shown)


ncave:
@alfonsogarciacaro Looks good, it will eliminate a lot of impedance and unnecessary pasta code that was previously needed because Fable AST was missing some elements of FSharp AST, hence the need to keep and inspect both at later stages. In theory we should be able to resolve all errors and restore full functionality before other big changes are implemented.

Issue #2109, opened 2020-06-26
Fable 3: Remove Babel dependency?


alfonsogarciacaro:
Not sure if we should go down this rabbit hole, but I've been thinking several times to remove the Babel dependency and the release of Fable 3 could be an opportunity for it.

### Motivation

When Fable was born, it used modern JS features (particularly the module system) that weren't supported by most browsers. thanks to Babel we could output different code depending on the environment. However, now JS modules are very extended and the JS features we use are supported for most JS engines. In principle, it "shouldn't" be too difficult to create our own JS printer and remove the Babel dependency.

### Benefits

- More control.
- Pure netcore app (it seems MSFT finally fixed tool installation with dotnet SDK so we could use Nuget again for distribution).
- Less overhead. Babel is quite fast, but removing the JSON serialization, the interop, etc, should speed up things. Also, it would make the REPL less fat and maybe make it easier to embed it in other webs.

### Drawbacks

- We will still need JS interaction for bundling with Webpack (hopefully it won't require much work because the TCP server is still in place).
- Users may still need Babel when targeting older browsers like IE.
- Even when compiling for Node, if we want commonjs modules we need either to add an extra Babel step or enable commonjs output in Fable (or rely in [Node experimental ES module support](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_ecmascript_modules)).
- Source maps. This will likely be the biggest hurdle as Babel is generating the source maps for us right now. See below.

### Alternatives to source map generation

1. Disable them. It's difficult to get them right anyways (especially with computation expressions) so we may better focus the efforts in outputting more readable JS (and fulfill the original promise) that it's easier to debug.
2. The best library for source map generation [seems to be this one](https://github.com/mozilla/source-map) which is in JS, so we could spawn a node process to call it. But this will likely brings us to the same situation we have with Babel.
3. Translate the parts we need from https://github.com/mozilla/source-map to F# (it doesn't seem to be a lot of code).

> The only library I've found to **produce** source maps in .net is [this one](https://github.com/nippur72/SourceMaps.CSharp) but it's quite old and maybe not reliable.

Yay or not yay? @ncave @MangelMaxime @Zaid-Ajaj @forki

ncave:
@alfonsogarciacaro I'm on the yay side.

forki:
Sounds amazing

ncave <notifications@github.com> schrieb am Fr., 26. Juni 2020, 04:11:

MangelMaxime:
[...]

PR #2140, opened 2020-08-23
Printers for Babel AST

alfonsogarciacaro:
I got too excited when children went to the kindergarten again and I could work for seven straight hours, so I thought I could have Fable 3 ready in no time. But almost 3 months have passed since then and I haven't advance much so we need to make this more a collaborative effort.

Writing printers for Babel AST as discussed in #2109 would be a great opportunity to contribute because it doesn't depend on Fable internals. I've created a first commit as reference, but it'd be great if someone else could take this PR over. Maybe @MangelMaxime or @Zaid-Ajaj? (or anybody else who wants to make Fable 3 happen) 🙏  Basically, you only need to implement the `Print` method in all Babel nodes (check `BlockStatement` as an example) except for `MacroExpression`s which are more complicated so I can do them myself.

Fable is still working with fable-loader/fable-splitter. The only thing the new printer does is to create a new .fs.js file next to the F# source file, that you can use to check if printers are working correctly. A way to do this is, after building the library (`npm run build library`), is to open two terminals and commands bellow:

```shell
dotnet watch -p src/Fable.Cli run -- start --port 61225
```

In the other terminal:

```shell
npx fable-splitter -c src/quicktest/splitter.config.js --watch --port 61225
```

Unrelated, but another task that could be taken over is removing the inheritance in Fable-generated JS classes and instead use an `__fsharp` field or similar with the class F# kind (union, record, struct, class, etc) that can be read in the generic `equals` and `compare` methods as discussed in . Maybe it could be an interesting one for @ncave? 😊


ncave:
> another task that could be taken over is removing the inheritance in Fable-generated JS classes and instead use an __fsharp field or similar with the class F# kind (union, record, struct, class, etc)

@alfonsogarciacaro Sure, I can take that. Can you clarify what is the current state of the next (nagareyama) branch, what's already in, is there a feature tracking issue?

alfonsogarciacaro:
@ncave I'm actually using this project: https://github.com/fable-compiler/Fable/projects/2

The only thing not listed is the AST cleanup and removing its dependency with FCS, which is in preparation fo the plugin system.

It seems projects don't raise notifications in Github, so I should have mentioned it somewhere. Feel free to edit the project or, also, if a feature tracking issue results to be more convenient, we can use that instead.

alfonsogarciacaro:
Ok, printers are finished so let's merge this!

Note the build still depends on Babel, I will change it to use the new printers (and become a "pure" dotnet app) in a later PR. Fable self-compilation is broken because a problem in the Babel transformations with directives (I removed them at one point but I put them back so I don't understand the problem) but it's not worth to spend time fixing it because we will remove Babel from the pipeline next.

I've removed some of the Babel AST nodes we're not using for simplicity. I haven't written the printers yet for type annotations. This needs to be done in order to recover @ncave's work to compile to Typescript, sorry!

ncave:
@alfonsogarciacaro This is great, can you also please rebase `nagareyama` from master? Thanks!

ncave:
> ... another task that could be taken over is removing the inheritance in Fable-generated JS classes ...

@alfonsogarciacaro See #2153.

PR #2144, opened 2020-08-26
Add Guid.ToString formats

Shmew:
Adds `System.Guid.ToString` with format overloads to Fable, and expands parsing to better follow .NET behavior.

Specifier | Format of return value
-- | --
N | 32 digits: `00000000000000000000000000000000`
D | 32 digits separated by hyphens: `00000000-0000-0000-0000-000000000000`
B | 32 digits separated by hyphens, enclosed in braces: `{00000000-0000-0000-0000-000000000000}`
P | 32 digits separated by hyphens, enclosed in parentheses: `(00000000-0000-0000-0000-000000000000)`
X | Four hexadecimal values enclosed in braces, where the fourth value is a subset of eight hexadecimal values that is also enclosed in braces: `{0x00000000,0x0000,0x0000,{0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00}}`

Resolves #2055
Enhances #1637

PR #2145, opened 2020-09-01
Fix identityHash pattern matching to adapt to option/tuple/list extra arg

thinkbeforecoding:
Fable don't compile due to missing '_' in pattern matching.


alfonsogarciacaro:
Thanks @thinkbeforecoding! I think the previous version of F# was more permissive here. Actually a similar change has been done in #2144 so the fix is already in master :+1:

thinkbeforecoding:
Now it works !! 😄

Shmew:
> I think the previous version of F# was more permissive here.

https://twitter.com/Cody_S_Johnson/status/1292134342345596928?s=20

Another victim

PR #2150, opened 2020-09-05
Fixed fable-compiler-js build

ncave:
Was missing `System.Web.HttpUtility.dll`


alfonsogarciacaro:
Good catch, thank you! Hmm, I'm now thinking if adding a new dll dependency to fable-js only to have access to that method is a good idea. Maybe we can pass it as a platform dependency, we have to do it for the writer anyways. What do you think? https://github.com/fable-compiler/Fable/commit/6401e503cd8f8df11d7231bffa0da935c48e521f#diff-65e59101fe6c6bd675ab418f704fb8f4R17

ncave:
@alfonsogarciacaro `System.Web.HttpUtility.dll` is a very small DLL, so it's fine. But if you prefer it as a platform dependency, that's fine too.

ncave:
Closing as obsolete.

PR #2153, opened 2020-09-09
No base class for types that don't inherit from other types

ncave:
- Removed base class reference for most types that don't inherit from other types.
- This is done mostly to try to improve object creation performance of classes by reducing inheritance.
- The resulting performance is an improvement over the current state, although not as big as hoped.
- Unfortunately still slower than using factory functions for object construction instead of classes.
- In rough approximation and limited testing, 10-15% performance gain over the current (next) branch.
- Still a mystery why classes perform so much slower than factory functions, could be something else.


alfonsogarciacaro:
Great work @ncave, thanks a lot! This is intriguing 🤔 It wouldn't be good if we decreases the performance of the REPL and other apps compiled by Fable. What do you think we should do? Go back to the compiler flag? Or try to investigate more to see if you can find the bottleneck and fix while still outputting classes?

ncave:
@alfonsogarciacaro One obvious thing I haven't tried yet is down-transpile to ES5 to see if that helps the performance. The only other perf-related idea that comes to mind is to get rid of all getters in classes, as we do know for sure they're slower than regular methods, for whatever reason. But that's not the main culprit in this case. Depends on the timeline of the release, if we have time we can investigate, if not we can go back to the switch.

ncave:
Rebased from #2156.

PR #2154, opened 2020-09-09
[WIP] Array-based List type

ncave:
- Faster than F# linked list, when micro-benchmarked.
- Overall, still a bit slower (10% slower) than using JS linked list, possibly due to list node creation not optimized enough.


src/Fable.Transforms/FSharp2Fable.Util.fs
```diff
@@ -1192,6 +1192,10 @@ module Util =
             else
                 // Indexed properties keep the get_/set_ prefix and are compiled as methods
                 if indexedProp then memb.CompiledName, false, false
+                // performance optimization, compile get_Current as instance call instead of a getter
+                elif memb.FullName = "System.Collections.IEnumerator.get_Current" ||
+                     memb.FullName = "System.Collections.Generic.IEnumerator.get_Current"
```
alfonsogarciacaro, line 1197: I would prefer having `System.Collections(.Generic).IEnumerator` mangled as the rest of interfaces from the BCL, as mangled properties are already always compiled as methods. That way we will avid avoid having to add an exception to an exception.

ncave: Sure, if the mangled IEnumerator methods are consistent enough to be called from Seq.ts.

ncave: That is, until eventually we convert Seq module to F#.

alfonsogarciacaro: Most of the functions in Seq.ts rely on JS iterators, so I guess we will only have to change `Seq/toIterator`

ncave:
Rebased from #2156.

alfonsogarciacaro:
This is still marked as WIP, so you want to merge it?

ncave:
@alfonsogarciacaro It's ready, but it does result in about 10% perf impact, so further optimizations are needed. We can keep it on the back-burner for now.

src/Fable.Transforms/FSharp2Fable.Util.fs
ncave: Yes, any idea what the mangled IEnumerator method names should look like?

alfonsogarciacaro: Seems the generic version is already mangled. If you comment out [this line](https://github.com/fable-compiler/Fable/blob/3a24934a73b18504e5ad417e490fc691bc534497/src/Fable.Transforms/FSharp2Fable.Util.fs#L1171), compile fable library (`npm run build library`) and check `class MapTreeModule_mkIEnumerator$0027$2` you'll see the generated names are:

```
    ["System.Collections.Generic.IEnumerator`1.get_Current"]() {
    ["System.Collections.IEnumerator.get_Current"]() {
    ["System.Collections.IEnumerator.MoveNext"]() {
    ["System.Collections.IEnumerator.Reset"]() {
```

ncave: @alfonsogarciacaro OK, fixed per your recommendation.

alfonsogarciacaro:
Do you think it would help if we allocated a few slots in advance and increase it as we add more elements? (I think System.Generic.Collections.List does this)

PR #2155, opened 2020-09-09
Hasta la vista Babel!

alfonsogarciacaro:


src/Fable.Transforms/AST/AST.Babel.fs
```diff
@@ -212,481 +255,477 @@ type MacroExpression(value, args, ?loc) =
...
-    override _.Print(printer) =
-        printer.Print(value.ToString(), ?loc=loc)
+    member _.Value: float = value
+    interface Literal with
+        member _.Print(printer) =
+            let value =
+                match value.ToString() with
+                | "∞" -> "Infinity"
+                | "-∞" -> "-Infinity"
+                | value -> value
```
alfonsogarciacaro, line 343: This is the best way I could find to print numbers but it doesn't look very robust. Please feel free to edit if you can think of a better way @ncave

ncave:
@alfonsogarciacaro Just wanted to acknowledge and reiterate that this PR was some mighty great work, it went completely flawless. Thanks again!

Issue #2159, opened 2020-09-11
Performance overhead of --classTypes


chrisvanderpennen:
### Description

As noted in a few other issues, the `--classTypes` compiler option results in code that performs anywhere up to 30% worse than without that option.

Related discussions:
* https://github.com/fable-compiler/Fable/issues/2096#issuecomment-689533372
* https://github.com/fable-compiler/Fable/pull/2056
* https://github.com/fable-compiler/Fable/pull/2153

### Background

The class constructor for the Union base class is defined in TypeScript, and is transpiled to:
```javascript
constructor(tag, name, ...fields) {
    super();
    this.tag = tag | 0;
    this.name = name;
    this.fields = fields;
}
```

For Fable 2.x, union type declarations are generated as follows (sample from fsharp/src/TypedTree):
```javascript
// no class types
export const ValInline = declare(function FSharp_Compiler_TypedTree_ValInline(tag, name, ...fields) {
  this.tag = tag | 0;
  this.name = name;
  this.fields = fields;
}, Union);

// class types
export class ValInline extends Union {}
```

`declare` defers to `inherits`:
```javascript
function inherits(subClass, superClass) {
    subClass.prototype = Object.create(superClass && superClass.prototype, {
        constructor: {
            value: subClass,
            enumerable: false,
            writable: true,
            configurable: true,
        },
    });
}
```

### Observations

The default compiler output defines an ES5-style constructor function which duplicates Union's constructor body.  `declare` then splices Union.prototype onto the compiler-generated function's prototype.  This does not actually invoke Union.  As far as I can tell, these factory calls are effectively inlined by v8.

The classTypes version does not define a constructor, but v8 still creates a stub constructor from v8's builtins causing the effective class declaration to be something like:
```javascript
export class ValInline extends Union {
    constructor(...args) {
        super(...args);
    }
}
```

Comparing bottom-up traces from `perf` shows a significant increase in time spent in builtin `StoreIC`, which I believe is v8 caching generated code for future inlining.  Union is the most significant contributor to that time, but is never directly invoked by derived constructors - there is always either one or two layers of indirection, the builtin `JSConstructStubGeneric` created to stand in as the derived constructor, and builtin `ArgumentsAdaptorTrampoline`.

### Analysis

ES5 union factories don't even show up in perf, as far as I can tell.  I suspect this means v8 is able to effectively inline them.

[...]

PR #2160, opened 2020-09-11
Push union class constructors down to generated leaf types

chrisvanderpennen:
Improves performance of output generated with `--classTypes`

See #2159

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -1925,8 +1925,7 @@ module Util =
                     ClassProperty(prop.Key, ?``static``=prop.Static, ?typeAnnotation=ta) |> U2.Case2)
             else Array.empty
         // no need for constructor in unions
-        let classMethods = if ent.IsFSharpUnion then [||] else [| U2.Case1 classCons |]
-        let classBody = ClassBody([| yield! classFields; yield! classMethods |])
+        let classBody = ClassBody([| yield! classFields; yield U2.Case1 classCons |])
         let classExpr = ClassExpression(classBody, ?superClass=Some baseRef, ?typeParameters=typeParamDecl, ?loc=r)
         classExpr |> declareModuleMember r isPublic name false
 
@@ -2029,11 +2028,20 @@ module Util =
                typedIdent com ctx fieldsId |> restElement |]
         let body =
             if com.Options.classTypes then
-                [ (ident tagId) :> Expression
-                  (ident nameId) :> Expression
-                  SpreadElement(ident fieldsId) :> Expression ]
-                |> callSuperConstructor None baseRef thisExpr
-                |> ExpressionStatement :> Statement |> Array.singleton |> BlockStatement
+                let callSuper =
+                    callSuperConstructor None baseRef thisExpr []
+                    |> ExpressionStatement :> Statement |> Array.singleton
+
+                [| tagId; nameId; fieldsId |]
+                |> Array.map (fun id ->
+                    let left = get None thisExpr id.Name
+                    let right =
+                        match id.Type with
+                        | Fable.Number _ ->
+                            BinaryExpression(BinaryOrBitwise, ident id, NumericLiteral(0.)) :> Expression
+                        | _ -> ident id :> Expression
+                    assign None left right |> ExpressionStatement :> Statement)
+                |> Array.append callSuper |> BlockStatement
             else
                 [| tagId; nameId; fieldsId |]
                 |> Array.map (fun id ->
--- a/src/fable-library/Option.ts
+++ b/src/fable-library/Option.ts
@@ -112,12 +112,54 @@ export function tryOp<T, U>(op: (x: T) => U, arg: T): Option<U> {
 
 // CHOICE
 
-export class Choice<_T1, _T2> extends Union { }
-export class Choice3<_T1, _T2, _T3> extends Union { }
-export class Choice4<_T1, _T2, _T3, _T4> extends Union { }
-export class Choice5<_T1, _T2, _T3, _T4, _T5> extends Union { }
-export class Choice6<_T1, _T2, _T3, _T4, _T5, _T6> extends Union { }
-export class Choice7<_T1, _T2, _T3, _T4, _T5, _T6, _T7> extends Union { }
+export class Choice<_T1, _T2> extends Union {
+  constructor(tag: number, name: string, ...fields: any[]) {
+    super();
+    this.tag = tag | 0;
+    this.name = name;
+    this.fields = fields;
+  }
+}
+export class Choice3<_T1, _T2, _T3> extends Union {
+  constructor(tag: number, name: string, ...fields: any[]) {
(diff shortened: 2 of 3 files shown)


ncave:
@chrisvanderpennen I tested your PR, here are my results (machine-specific):

- master branch, no classes: `FCS time: 58 sec`
- master branch, --classTypes: `FCS time: 83 sec (43% slower)`
- master branch, --classTypes + this PR: `FCS time: 75 sec (29% slower)`

These results are consistent with #2153 numbers.
The good news from this PR is that base class + explicit constructors gives you the same performance as no base class.
Unfortunately 29% slower is still a lot, perhaps we can explore monomorphic union constructors to see if that can help.

chrisvanderpennen:
Which OS and node version are you testing with?  I'm not seeing the same overhead between master with no classes, and this PR with --classTypes.

chrisvanderpennen:
OH! I see what's going on.  This is interesting, it does increase FCS time but everything after that is then faster, so the total runtime of the benchmark is the same.  How strange.

ncave:
@chrisvanderpennen Node.js v14.5, Win10.

Please note these times are for the FCS portion of the compilation only (parsing + type-checking the F# source). If you look at the full Fable time, the difference is of course going to be smaller, as your denominator is much larger, and includes code transformation/generation and some external processing by Babel (albeit small).

chrisvanderpennen:
Makes sense.  I'll truncate the benchmark code to exit after `measureTime parseFSharpProject ()` and printing, and get some more perf traces.

chrisvanderpennen:
I found something.  It looks like the sole contributing factor from my side, hopefully @ncave can replicate the results.

From Types.ts:
```typescript
export const Exception = declare(function Exception(this: Exception, message?: string) {
  this.stack = Error().stack;
  this.message = message;
}, SystemObject);
```

Derived types created with `declare` don't call the base constructor - so `Error().stack` doesn't actually get hit when creating instances of derived exceptions unless `--classTypes` is used.  It looks like FCS creates a lot of derived Exception instances, e.g. `UndefinedName`, but it wraps them as the Error case of a custom `Result<,>`-like union and almost never actually throws them.

Changing the above to `this.stack = '';` sees equivalent FCS timings on my Linux desktop under Node 14.10 - 58-59 seconds on each side, over 10 iterations.

ncave:
@chrisvanderpennen

This is brilliant, thank you so much for spending the time to investigate. I've added your fix to Fable 3 in #2162, works wonders!

Now, I just can't pass the opportunity to ask you to share your process for investigating performance issues, as we have hit a wall trying to identify bottlenecks and bring the JS performance closer to .net speed, as much as possible (currently FCS time is `.NET: 24 sec`, `JS: 58 sec`, so it's not that far off). Obviously `.NET` code is highly optimized, so just looking for low-hanging fruits.

alfonsogarciacaro:
To add up to what @ncave said, we're open to change any of the generated code for Fable 3 if we can increase performance, so any advice you can give is highly appreciated. The only restrictions are:
- Try not to output too much code as this hits bundle size.
- Try to keep support of all the F# features we currently support (e.g. basic type testing). If we change the runtime representation of something but we find a workaround to make the tests pass, should be fine.
- Try to keep the current level of interoperarity (e.g. no mangling, unless explicit, of interface members, or automatic uncurrying)

chrisvanderpennen:
Excellent, glad to help.  If you're happy eliminating the call to `Error().stack` here as well I can update this PR, just let me know.

As for how I've been investigating: Node's `--prof` and `--prof-process` _should_ produce traces of node and v8 internals but I haven't been able to get it to actually do that, it's been bundling all of that time together as shared library calls.  I've been using `perf`, which is a Linux native profiler, to sample external to Node, then `perf diff` and https://github.com/KDAB/hotspot to analyse the traces.  I've also been running the unbundled output instead of `rollup`'s output, so I can see source file names in the results.

The command line I've been using to capture traces is `sudo -E perf record -q -e cpu-clock -F 1000 -g -o ~/perf/perf-prototypes.data -- ~/.config/nvm/14.10.0/bin/node --perf-basic-prof --interpreted-frames-native-stack out-node/app.js bench-compiler.fsproj out-node2` (or `-o ~/perf/perf-classes.data`).  Hotspot can produce a bottom-up flame graph from this output, so I've basically been visually comparing the two flame graphs and investigating the differences.

Once I'd cut the benchmark down to only run FCS, the additional time spent creating stack traces was clearly visible and identified the contributing JS functions and their source files.  Following one of their inheritance chains led me to the `Error().stack` call.

The earlier investigation into Union's constructor being so much slower was a bit different - it was plain from the comparative traces that a lot more time was being spent in Union, but it took a fair bit of research into ArgumentsAdaptorTrampoline and the other builtins to understand why that time was being spent and how to eliminate it.

chrisvanderpennen:
Actually, given that stack traces are only skipped for exceptions declared using `exception name of data` (see https://github.com/fable-compiler/Fable/pull/2162#issuecomment-691409932) what's the desired resolution here - fixing  `exception name of data` to include stack traces without `--classTypes` with the resulting performance penalty in FCS, or remove stack traces from exceptions everywhere?

chrisvanderpennen:
Third option - node/v8 capture the stack trace in Error's constructor, whereas .NET delays until the exception is actually thrown.  Native errors will already be `Error` or a subtype and so come with a stack trace already attached, but since we know when F# will actually `raise` an exception instance we could emit JS to capture the stack at the point it's raised, instead of during construction.

alfonsogarciacaro:
Those are great ideas @chrisvanderpennen. Wrapping the F# exceptions could be a way to meet the recommendations that we should be only throwing JS errors, but it can make it more complicated when we have to access the F# exception directly (for example for equality o type testing). I guess it's simpler to follow your first suggestion and capture the stack when raising the exception 🤔

chrisvanderpennen:
As long as it's only wrapped at the point it's thrown, and then it's unwrapped as soon as it's caught, the wrapper shouldn't ever be visible to any of the other code.  Something along the lines of:
```javascript
class ExceptionWrappingError extends Error {
  constructor(exception) {
    super();
    this.exception = exception;
  }
}

function throws() {
  // failwith "message"
  throw new ExceptionWrappingError(new Exception("message"));

  // let result: Result<_, Exception> = ...
  // match result with | Error e -> raise e | Ok ok -> ok
  let e = result.fields[0];
  throw new ExceptionWrappingError(e);
}

try {
  ...
} catch (e) {
  e = e instanceof ExceptionWrappingError ? (e.exception.stack = e.stack, e.exception) : e;
  // the rest of the generated catch block
}

alfonsogarciacaro:
@chrisvanderpennen If it's ok with you, we can close this as the change was already introduced in #2162. I have added a card to Nagareyama project to add the stack to F# exceptions: https://github.com/fable-compiler/Fable/projects/2#card-46364034

chrisvanderpennen:
Yeah no worries :)

alfonsogarciacaro:
Thanks again for all your help! 👏 👏 👏

PR #2162, opened 2020-09-12
Fixed performance with classes

ncave:
- Added @chrisvanderpennen's brilliant performance insight (see also #2160).
- Performance with classes is now exactly the same as with factory functions.

@alfonsogarciacaro Tests are passing ok, Fable-JS works fine, etc., etc. but please validate if not deriving Exception from Error has any other implications.

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -1782,10 +1782,6 @@ module Util =
     let declareClassType (com: IBabelCompiler) ctx (ent: Fable.Entity) entName (consArgs: Pattern[]) (consBody: BlockStatement) (baseExpr: Expression option) classMembers =
         let consId = Identifier "constructor"
         let typeParamDecl = makeEntityTypeParamDecl com ctx ent
-        // let baseRef =
-        //     match baseExpr with
-        //     | Some baseRef -> baseRef
-        //     | _ -> makeImportTypeId com ctx "Types" "SystemObject" :> Expression
         let consBody =
             if ent.IsFSharpExceptionDeclaration then
                 let super = callSuperConstructor None [] |> ExpressionStatement :> Statement
--- a/src/fable-library/Types.ts
+++ b/src/fable-library/Types.ts
@@ -1,5 +1,5 @@
 // tslint:disable: space-before-function-paren
-import { IEquatable, IComparable, combineHashCodes, compare, compareArrays, equalArrays, equals, identityHash, isComparable, isEquatable, isHashable, isSameType, isStringable, numberHash, structuralHash } from "./Util";
+import { IEquatable, IComparable, combineHashCodes, compare, compareArrays, equalArrays, equals, isComparable, isEquatable, isHashable, isSameType, isStringable, numberHash, structuralHash } from "./Util";
 
 export function objectToString(self: any) {
   if (isStringable(self)) {
@@ -9,24 +9,24 @@ export function objectToString(self: any) {
   }
 }
 
-export class SystemObject implements IEquatable<any> {
+// export class SystemObject implements IEquatable<any> {
 
-  public toString() {
-    return this.ToString();
-  }
+//   public toString() {
+//     return this.ToString();
+//   }
 
-  public ToString() {
-    return objectToString(this);
-  }
+//   public ToString() {
+//     return objectToString(this);
+//   }
 
-  public GetHashCode(x?: any) {
-    return identityHash(x ?? this);
-  }
+//   public GetHashCode(x?: any) {
+//     return identityHash(x ?? this);
+//   }
 
-  public Equals(x: any, y?: any) {
-    return x === (y ?? this);
-  }
-}
+//   public Equals(x: any, y?: any) {
+//     return x === (y ?? this);
+//   }
+// }
 
 function compareList<T>(self: List<T>, other: List<T>) {
   if (self === other) {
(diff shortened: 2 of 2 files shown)


alfonsogarciacaro:
Awesome! Will Nagareyama be on par with master performance-wise with this fix? About custom order exceptions the biggest concern I think is the stack trace. Do you know if it's kept? Don't remember if we have tests for this. I'll check when I'm home.

ncave:
@alfonsogarciacaro Yes, performance is now the same as master without classes.
But no, the stack is not there anymore cause Exception doesn't extent Error anymore, that's part of what was slowing it down.

alfonsogarciacaro:
Can you please also add a comment somewhere to remark we're not following MDN recommendations for creating custom errors for performance reasons? Hopefully this will prevent someone in the future to revert the change.

ncave:
@alfonsogarciacaro Ok done.

alfonsogarciacaro:
Hmm, it's a pity we're losing the stack trace for custom exceptions but maybe we can think of a solution for that. One thing I don't understand is @chrisvanderpennen [says here](https://github.com/fable-compiler/Fable/pull/2160#issuecomment-691378531) `Error().stack` is not hit, but [stack traces do work in Fable 2 for custom exceptions](https://fable.io/repl/#?code=PTAEFMA8GNwBwC4EsD2A7UBZAngUQE74r6goBmoSaCAsAFD0LZzhZ6HEBMAFJAFyVqASlABeeqEmCAFuHxIEoAMrYAzgnABbAHS4Y8ZOm5D69ADbhFZAIZIzAMWvrjYiVIT5sbqVJwEi+DwArCIAPgB8oPi2quDePqAgEJAYAERkKCipoBFRMXF0PgDuCtLJoAC04fE+UNpKCNbQANYAKtGwpnQ2do7OYZFw8tRkaQCkqqlAA&html=Q&css=Q) (although I just realized it's broken for F# exceptions). Maybe we can keep the stack traces doing the same trick (`Error().stack`) without deriving from `Error`?

Anyway, let's merge this and discuss about the stack trace later. Thanks a lot you both for the awesome work @chrisvanderpennen @ncave, please keep the performance improvements coming! :muscle:

chrisvanderpennen:
> but stack traces do work in Fable 2 for custom exceptions

If you declare `type x(i: int) = inherit System.Exception()` Fable emits `Exception.call(this)` as part of the generated constructor.  If you declare `exception x of int` it doesn't, and so the stack trace is undefined.

```fsharp
exception MyError of int

type MyError2(x: int) =
    inherit System.Exception()
```
```javascript
export const MyError = declare(function Test_MyError(Data0) {
  this.Data0 = Data0 | 0;
}, FSharpException);
export const MyError2 = declare(function Test_MyError2(x) {
  const $this$$1 = this;
  Exception.call($this$$1); // <-- this makes it work
  void null;
}, Exception);
```

alfonsogarciacaro:
Ah, ok, thanks a lot for the clarification. So it seems a bug (missing the stack trace in F# exceptions) was actually giving JS-compiled FCS a performance boost 😅

alfonsogarciacaro:
@ncave I noticed that now we have structural equality for objects of different types, which shouldn't happen often but it's not the expected behavior anyways.

```fsharp
type MyUnion1 = Foo of int * int
type MyUnion2 = Foo of int * int

type MyRecord1 = { Foo: int }
type MyRecord2 = { Foo: int }

let areEqual (x: obj) (y: obj) =
    printfn "(%s) equals (%s)? %b" (string x) (string y) (x.Equals(y))

areEqual (MyUnion1.Foo(1,2)) (MyUnion2.Foo(1,2)) // true
areEqual { MyRecord1.Foo = 1 } { MyRecord2.Foo = 1 } // true
```

I was thinking to add a field with the F# kind like: `__fsharp: union` so we can identify the type of object in equality, but then I noticed most of the issues are fixed by checking if the constructors are the same `x.constructor === y.constructor` before this line: https://github.com/fable-compiler/Fable/blob/d12cfcc7b6f8209a4148ac990d22d107a52622ea/src/fable-library/Util.ts#L390

What do you think? (We'd have to do a similar change in `compare`) A side-effect is all JS object literals get structural equality by default which didn't happen in Fable 2, but it may be a good thing. 

There's another issue when converting to string but we can check that later.

ncave:
@alfonsogarciacaro Sure, makes sense. It would be preferable if we can avoid adding new fields.

Structural equality already goes into `Types.recordEquals` (see [Replacements.equals](https://github.com/fable-compiler/Fable/blob/nagareyama/src/Fable.Transforms/Replacements.fs#L762-L765)), so we can keep Util.equals for referential and custom equality (and same thing for compare).

alfonsogarciacaro:
Ok, I added the constructor check, this should solve some situations 👍 At the moment, `Util.equals` is defaulting to structural equality (`equalObjects`) even for classes that don't implement `Equals`. Before, everything was inheriting from `SystemObject` so there was always a default implementation of `Equals` using reference equality. I'm assuming (or hoping) this won't be much of an issue because the F# compiler will disallow equality in most situations if the class doesn't implement equality.

Interestingly I did try to disable structural equality for non-union/record/struct classes but got an error when compiling the test for units of measure with fable-compiler-js. I didn't investigate much.

And yes, at one point we should probably inspect the equality/comparison attributes. I think Fable 1 did this somehow, but it was removed when refactoring the code for Fable 2 and it has remained unimplemented since.

PR #2164, opened 2020-09-13
Fixed fable-library import extensions

ncave:
- Fixed fable-library import extensions (added `.js` as TypeScript does not add them).
- Updated Fable benchmark compiler test with a --benchmark flag.


alfonsogarciacaro:
Thanks!

Issue #2139, opened 2020-08-23
Problem with object expressions


stkb:
1. Create an abstract base class with an abstract member function, where that member function is called within the constructor (in a do-expression)
    ``` fsharp
    [<AbstractClass>]
    type AbstractBase () as this =
        do this.sayHello ()
        abstract member sayHello : unit -> unit
    ```
2. Create an object expression that derives from that abstract base class, and implements/overrides the abstract member function.
    ``` fsharp
    let objectExpression () =
        { new AbstractBase () with
            override this.sayHello () = printfn "Hello %s" "Object Expression"
        }
    ````
**Expected**: In standard F# this is fine; the implemented function is called in the base class's constructor
**Actual**: With Fable this fails with error `this$.contents.sayHello is not a function`

Rest of the example code ([Repl link](https://fable.io/repl/#?code=NoHgggRgzgLgTgQwMYwMIBsFSgPgLoCwAUDAJ4AOApgASSyIoBCWNAFAJTVbUwAWAllGoBeYtXHUAJgHseAqADooCUgAlK6dLI5iJCaPGQxqAW0omIlONWVqNW6gC5qAVwB2-YwFocrjzGJidEpjaQgAK0oUAFEAD3I4Smx+aTdqDhFdcQBvajdKAHdaAwYYZig2TgLPXiyJcWkANys4fkkaPkElFXVNbU5hagT+NxgAMzSAIl6HAFIoSepJgHkIqOM4hKSoFLdJuoBfQJIKGlRUpESYSgAVU-SBupHeK09i+iNyyrqmlraaAD63TsfQeIiGrVGEyWM1k80Wk3ObkuIRodyo+yIxCRKOu6Mq1AAPr5+ABzNzSRLEMKRGLxRLJVJg4nUMkUqlEIA&html=Q&css=Q)). The `ConcreteType` is just added for comparison (that works fine).

``` fsharp
type ConcreteType () =
    inherit AbstractBase ()
    override _.sayHello () = printfn "Hello %s" "Concrete Type"

ConcreteType () |> ignore
objectExpression () |> ignore
```

Javascript. The problem is where the `extend` function is called within the `objectExpression` function; it calls the base class constructor *before* adding the member functions.
``` javascript
import { declare, FSharpRef } from "fable-library/Types.js";
import { class_type } from "fable-library/Reflection.js";
import { toConsole, printf } from "fable-library/String.js";
import { extend } from "fable-library/Util.js";
[...]

Issue #2091, opened 2020-06-13
Attached members minification in Fable 3


alfonsogarciacaro:
Being a dynamic language, JS bundlers cannot minify the name of object fields without breaking code. However this should be possible for a typed language. It was for example one of the [main features in Elm 0.19](https://elm-lang.org/news/small-assets-without-the-headache).

We decided not to try that in Fable because we were using interfaces to interact with JS. But if we're going to use the `NoMangle` attribute to mark specifically interfaces that are meant for JS, we could minify the names of other interface members (as they are already mangled anyways).

The proposal is (in non-debug compilations):

- Minify the names of interface members.
- Minify also the names of non-anonymous record fields? (for this we would have to allow the `NoMangle` attribute in records too, or disallow using records to interact with JS).
- Avoid conflicts with names of private fields in classes

Zaid-Ajaj:
Given how great Fable is to interop with Javascript, this "feature" would bring it a step backwards IMO. It is always good see that the compiled names of the records/interfaces are exactly the same at runtime so you could simply `unbox` as input to some 3rd-party library. 

During development, if I am testing fiddling with a binding and interop code until it works, suddenly it will break on a full-build because of the minification. 

An optimization like #2088 will have far more significance on the output of the generated bundle without introducing surprises for developers

alfonsogarciacaro:
Those are good points @Zaid-Ajaj :+1: Just remember for Fable 3 we're planning to mangle the name of the interfaces, so it won't be possible to use them for JS interop anyways unless they're decorated with `NoMangle`.

0x53A:
Instead of mangling by default and ``NoMangleAttribute``, would it instead make sense to add ``MangleAttribute``?

Then the compiler could earn where you have naming clashes and point you to use the attribute on one or more of the interfaces.

I also really like the good interop by default.

Zaid-Ajaj:
> for Fable 3 we're planning to mangle the name of the interfaces, so it won't be possible to use them for JS interop anyways

I don't know if this is really a good idea. It might make sense for Elm where interop is extremely limited but for Fable that is a powerful feature that will be affected. Just imaging right now how many people will update their fable-compiler versions to realise that many of their dependant libraries don't work with latest Fable because of the name mangling. 

[...]

PR #2165, opened 2020-09-14
Updated fable-library

ncave:
- Reduced `fable-library` inner dependencies.
- Added some test coverage.
- `IEnumerator` is now mangled.
- Some `Seq` improvements.

Overall some slight performance improvement on the benchmark, not much (1-2%), but still a move in the right direction.

### DIFF
--- a/src/fable-library/Seq.ts
+++ b/src/fable-library/Seq.ts
@@ -4,9 +4,10 @@ import { Option, some, value } from "./Option.js";
 import { compare, equals, IComparer, IDisposable } from "./Util.js";
 
 export interface IEnumerator<T> {
-  Current: T | undefined;
-  MoveNext(): boolean;
-  Reset(): void;
+  ["System.Collections.Generic.IEnumerator`1.get_Current"](): T | undefined;
+  ["System.Collections.IEnumerator.get_Current"](): T | undefined;
+  ["System.Collections.IEnumerator.MoveNext"](): boolean;
+  ["System.Collections.IEnumerator.Reset"](): void;
 }
 
 export interface IEnumerable<T> extends Iterable<T> {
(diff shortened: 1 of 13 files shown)


src/fable-library/Seq.ts
```diff
@@ -81,7 +85,11 @@ function makeSeq<T>(f: () => Iterator<T>): Iterable<T> {
 }
 
 export function ofArray<T>(xs: ArrayLike<T>) {
-  return delay(() => unfold((i) => i != null && i < xs.length ? [xs[i], i + 1] : undefined, 0));
+  if (Array.isArray(xs)) {
```
alfonsogarciacaro, line 88: `Array.isArray` doesn't work with typed arrays (I guess this will be less important in Fable 3 with non-typed arrays as default, though we still compile FCS with typed arrays). There's an `isArray` function in Util.ts to test both cases.

alfonsogarciacaro (approved):

alfonsogarciacaro:
Awesome, thanks a lot! I just left a small comment but that can be changed later if needed. Let's merge this!

ncave:
@alfonsogarciacaro As a side note, looks like the reason class getters/setters can be very slow in some cases is that they prevent optimization for [Functions that contain object literals that contain `__proto__`, or `get` or `set` declarations](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax).

alfonsogarciacaro:
Great catch as always @ncave! I tried to optimize this by [compiling getters with no side effects as object fields](https://github.com/fable-compiler/Fable/commit/a7f277b74e2a9b03b2bd735d72428c282e42c393) but then I remembered we have to compile objects expressions extending a class as class expressions to fix #2139, so we could do the same when we detect there's a getter/setter. Maybe that helps? I will push the change and we can see if it has any effect with the benchmarks.

ncave:
@alfonsogarciacaro But why not just compile them as normal calls instead? Anyway, I don't think there will be a noticeable impact on the benchmark, the biggest one was already fixed by the mangling of the IEnumerator in this PR.

alfonsogarciacaro:
Basically for interop, as an interface getter should be callable with `myObj.myProp` syntax from JS, since this has been the behavior so far and we decided not to change it unless the new `Mangle` attribute is specified here: https://github.com/fable-compiler/Fable/issues/2091#issuecomment-643868191

alfonsogarciacaro:
I just made [a quick test](https://github.com/fable-compiler/Fable/commit/bb55f5f08f0230dded3c87a96372e6c24689a3e2#r42676688) and indeed getters in object literals can be more than 10 times slower than fields 😮 Compiling the object as a class expression seems to fix it (`new (class {...})()`). Not sure if this pattern will affect many other places in FCS or other apps besides `IEnumerator.Current` but it's nice to have it fixed. Thanks again! :clap: 🏅

Issue #2078, opened 2020-06-09
Road to Fable 3


alfonsogarciacaro:
So far there have not been many good news in 2020 so let's to make one by releasing a new major version of Fable 3 😄 I'm collecting different threads from recent issues here to list the features that could go in Fable 3:

### Reflection support

My original hope was that #1839 would basically become Fable 3, but I've left too much time passed, the PR has lost sync with master and it's proving a bit difficult to make it work now. Besides I had some concerns about its reflection support. Although it's very interesting to be able to access to all class methods, constructors, properties and attributes (currently Fable is limited to record, union and tuple fields) for scenarios like [test generators](https://github.com/fable-compiler/Fable/issues/2027#issuecomment-638910441) (I'm also using reflection now for interacting with databases in Node with good results), I'm concerned about how this would affect the resulting bundle in web apps.

Fable tries to play well with tree shaking in JS. For example class methods are not attached to the objects and can be removed from the bundle if not actually used. However, if we keep references to the methods in order to be able to call them through reflection, tree shaking won't be able to tell unused code apart (e.g., see [this comment about Dart](https://flutter.dev/docs/development/data-and-backend/json#is-there-a-gsonjacksonmoshi-equivalent-in-flutter)).

[In some quick tests](https://github.com/fable-compiler/Fable/pull/1818#issuecomment-486185727), the increase in the bundle size was not much, but I'm not sure if reflection wasn't used much in those cases. When you include support for quotations the bundle size explodes and it could be because reflection is used more intensively.

I'm considering now to give support to Constructors, Methods, Properties and Attributes, but only through special helpers in Fable.Core.Reflection module that generate the information in compile time only where it's needed (same way as we can only get type info when the type is known at compile time). Like:

```fsharp
open Fable.Core

let cons = Reflection.generateConstructors<MyType>()
let methods = Reflection.generateMethods<MyType>()
let meth = Reflection.generateMethod<MyType>("Foo")

// For example, this would be replaced in compile time by
// an array with the attributes decorating the type
let atts = Reflection.generateCustomAttributes<MyType>()
```

Would that be a good idea? The main downside is there could be duplicated code if we make the same call in multiple places.

### Quotation support?

This also depends on #1839 Not sure how difficult would it be to add quotation support if we implement reflection differently.

### Mangling abstract members

[...]

PR #2171, opened 2020-09-21
New Fable.Core.JS Syntax

Shmew:
As I mentioned in #2078, I think this makes the syntax much nicer to work with, and if y'all like it we can add modules alongside the types so you can do things like:

```fsharp
JS.Map<int,int>()
|> JS.Map.set 2 1
|> JS.Map.entries
```

and the compiled js would be this (cutting out the junk caused by pipes):

```javascript
new Map().set(2, 1).entries()
```


MangelMaxime:
Disclaimer: I have been away from F# for a long time so I don't remember well how everything works ^^ So I am asking some basic questions probably ^^


Quoting part of your message from the issue and moving the discussion here as I think it makes sense to have it in the PR

> I'm also curious, is `AllowNullLiteral` still necessary on those types?

I guess it was used to represent `null` or `undefined` values from the Browser APIs. Like when an API returns `null` or `undefined` as a default value. But I suppose if that's the cause we should use `'T option` to represent the output of the APIs.

Could you please provide the equivalent of your example with the old syntax? Just to compare it ;)
Or is it just impacting the binding code and not how the user consumes it?

Questions that come in my mind are:

1. Is type extension still supported?

```fs
type FormDataType with
    [<Emit("new $0($1...)")>]
    member __.Create(form: HTMLFormElement): FormData = jsNative
```

2. Can we still `inherit` from several interfaces/types?

```fs
    inherit Node
    inherit ChildNode
```

3. By using interfaces we were able to use `jsEmpty`, `jsOption`, etc. to create the "options" types. I don't know if this is supported for "real types". Or perhaps, we will have a mix of interface and real types depending on if it's an "options/configurations" or API binding?

I am just trying to see if there is something that can't be supported by the proposed API and also the impact on user code. In any case, IMO, it should be included in a major release of Fable.Core ;)

Asked differently could be what was the limitation of the previous APIs?

Shmew:
> Quoting part of your message from the issue and moving the discussion here as I think it makes sense to have it in the PR
> 
> > I'm also curious, is AllowNullLiteral still necessary on those types?
> 
> I guess it was used to represent null or undefined values from the Browser APIs. Like when an API returns null or undefined as a default value. But I suppose if that's the cause we should use 'T option to represent the output of the APIs.

Yeah, it would be nice to properly define what signatures should be an `option` or not so we can get rid of that null assignment stuff. I'm not really sure much of anything here needs it, but could very possibly be wrong about that.

> Could you please provide the equivalent of your example with the old syntax? Just to compare it ;)
> Or is it just impacting the binding code and not how the user consumes it?

The example I gave above wasn't very great, but the main thing is no more `Create` methods:

```fsharp
JS.Constructors.Date.Create()
```

Which is instead simply:

```fsharp
JS.Date()
```

As well as a more expected user experience of static methods:

```fsharp
JS.Constructors.Date.now()
```

Which is now:

```fsharp
JS.Date.now()
```

The main thing is that it's a lot easier to extend both methods and static methods to add helpers as we can add our own inline methods (although we have to turn off the `Global` attribute and use `Emit` instead for all the members). Such as adding support for this kind of core api returning an array or list rather than `ResizeArray`, or the inverse where we accept a list and convert it to resize array behind the scenes.

> Is type extension still supported?

Yeah as long as we don't seal the types.

> Can we still inherit from several interfaces/types?

If it's something that's truly an interface you can still use interfaces, this change is focused really only on actual JS native objects that you generally wouldn't interface/inherit from, just like you wouldn't do that with an F# list or Map etc. That being said you can still inherit from them if you construct the base type.

> By using interfaces we were able to use jsEmpty, jsOption, etc. to create the "options" types. I don't know if this is supported for "real types". Or perhaps, we will have a mix of interface and real types depending on if it's an "options/configurations" or API binding?

I think this falls in line with my previous answer that this is really only about the more concrete structures that you wouldn't use those types of functions to build.

As far as an end user can tell from this PR in the current state is really just a couple things:

* No need for double tick syntax (which we could accomplish before with emits I suppose) like \`\`begin``
* No need for the `JS.Constructors` module, makes it easier to both construct and use static methods for these types, for example we can now add the static methods for things like `Int8Array` since it's no longer a type annotation.

If this is generally accepted by everyone we can evaluate pain points where we can add additional overloads as well as module/static methods to make piping usable for them.

MangelMaxime:
> > Can we still inherit from several interfaces/types?
> 
> If it's something that's truly an interface you can still use interfaces, this change is focused really only on actual JS native objects that you generally wouldn't interface/inherit from, just like you wouldn't do that with an F# list or Map etc. That being said you can still inherit from them if you construct the base type.

My example was taking from `Fable.Browser.Dom`

[Source file](https://github.com/fable-compiler/fable-browser/blob/master/src/Dom/Browser.Dom.fs#L22-L31)

```fs
type [<AllowNullLiteral>] CharacterData =
    inherit Node
    inherit ChildNode
    abstract data: string with get, set
    abstract length: int
    abstract appendData: arg: string -> unit
    abstract deleteData: offset: float * count: float -> unit
    abstract insertData: offset: float * arg: string -> unit
    abstract replaceData: offset: float * count: float * arg: string -> unit
    abstract substringData: offset: float * count: float -> string
```

Because I don't know if this still will be specific to Fable.JS or propagated to other bindings too.

> If this is generally accepted by everyone we can evaluate pain points where we can add additional overloads as well as module/static methods to make piping usable for them.

👍

Shmew:
Oh, yeah I'm not suggesting that we need to ditch using interfaces. It's just really painful to use interfaces when the types we're modeling are classes with constructors.

alfonsogarciacaro:
Thanks a lot for this @Shmew and sorry for not replying earlier to your comments about the Fable.Core.JS syntax. I understand is not ideal and I really appreciate your efforts to improve it, but I would be very careful with changing the APIs in Fable.Core for several reasons:

- We're trying to avoid breaking changes in the code with the move to Fable 3 and only ask the users to modify their build system.

- The relationship between Fable compiler and Fable.Core is tricky. Although it's true several features in Fable.Core require a specific version of the compiler, at the moment the versions are not in sync (Fable is v2 while Fable.Core is already v3).

- We already made breaking changes in Fable.Core a couple of times and it takes a lot of effort to update the whole ecosystem because all libraries depend on it. I'm sure package authors won't be very happy with more breaking changes even it brings some improvements.

- Although not ideal, using a separate interface for the static part of a JS type was decided [quite early on](https://github.com/fable-compiler/ts2fable/issues/1#issue-217374509) among other reasons, because it was not possible to inherit from multiple types in F# as @MangelMaxime points out. The convention is used in multiple bindings like Browser or Node and in the documentation, so changing or mixing it now could create confusion (although we could mention in the docs that it's possible to use a class for the binding too).

- About adding helpers to Fable.Core, I usually say I prefer to do that in a separate library (for example, helpers for JS promises are in the Fable.Promise package). The main reason is we're trying to keep Fable.Core as minimal and stable as possible, but also because Fable.Core is an "empty" Fable library, which means it doesn't contain actual code only metadata. When necessary, Fable injects some code or makes a compiler trick to resolve calls to Fable.Core as it does with FSharp.Core.

So in summary, I would prefer not to make breaking changes in Fable.Core (although I'm open to it if other maintainers in the ecosystem think is a good timing for it and agree with the changes). But I do encourage you to publish a package with the proposed API (just please don't use the `Fable.Core` namespace) and see how users like it so we can incorporate it to Fable.Core in the future if it becomes the preferred option.

src/Fable.Core/Fable.Core.JS.fs
```diff
@@ -1,447 +1,493 @@
...
+        static member acosh (x: float) : float = jsNative
+        static member asinh (x: float) : float = jsNative
+        static member atanh (x: float) : float = jsNative
+        static member hypot ([<ParamArray>] values: float[]) : float = jsNative
+        static member trunc (x: float) : float = jsNative
+        static member fround (x: float) : float = jsNative
+        static member cbrt (x: float) : float = jsNative
+
+    [<AllowNullLiteral;Global>]
+    type Date [<Emit("new Date($0)")>] private (inp: obj) =
```
alfonsogarciacaro, line 118: I think you don't need to use `Emit` in the constructor, Fable will detect this is a global JS object and call the constructor with `new` and the proper name thanks to the `Global` attribute. You can [test the bindings and see the generated code in the repl](https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwChcDaAPAOJLAJxIB8AurgC4Ce4UAInA2FABQMCWAYwDWAZwBcUfhAYBKKAF5cUFVEFI4o0VEgATfCgZQuohj3lLV7TmB4BWAAyygA&html=DwCwLgtgNgfAUHUBTAhgE3gAm54ElgqbhgAOAtEgI4CuAlgG4C8A5AMID2AdmEj+QBUAnqSQtMAY268erXgA8wAenDQA3JJAoATgGcCTGmABm5ABwssOYLona6pMJl3aJTAEQB9TwAkA8gDKAt5KUHQARrpK2qgSYAB0pNocaDRxdNzxEHRc8QBWuu4wwEq29o5W2DZ2Dk4ubl6+gcGeoRFRMShx5GgcEInJqemZ2bkFRSVltfAlIKgYCMDhKUKSUCi6uh4opBRSPCg5SNpFcDi4aIyYdGgeSFDZuiDkO6Tuaxtb7vePz68TSkuDBmSmWaCE8EQKkgsDgQA&css=BYFwtgNgNAUARgewCYE8AEBvGadoGYIB2IAtAM4CWAXgKYBcaAjAGwAOAHgNwwC+MMAOjABDCoRIBjIiFGEaAJ0zZcSCmVYRhKBngg0uynAHcKSEMAaMADFYCk3XGmA0KAc1CWb9w2gBWAVzIQCjwUSWkaYgYJSJAFB1xhCDdxCjiwMmjY+J9dfRJVeRoJYKJohAh-MEJuPkExVn8QJUdWYSRVQlcGAQAmAFYisAScAmJyanomNgNHEzMLNAGrDhG0EXlXMRJEEBAEMEsh2pggA) to see what works 👍

Shmew: Oh interesting. I could have sworn I tested that in the repl and it didn't do that. Good to know!

Shmew:
> We're trying to avoid breaking changes in the code with the move to Fable 3 and only ask the users to modify their build system.

A very valid point and the biggest one for sure. 

> The relationship between Fable compiler and Fable.Core is tricky. Although it's true several features in Fable.Core require a specific version of the compiler, at the moment the versions are not in sync (Fable is v2 while Fable.Core is already v3).

Ah that makes this change sound like a lot more work that I'd thought.

> We already made breaking changes in Fable.Core a couple of times and it takes a lot of effort to update the whole ecosystem because all libraries depend on it. I'm sure package authors won't be very happy with more breaking changes even it brings some improvements.

Yeah this is along the lines of the first point and very valid. For the most part the amount of actual changes would be quite minimal, however. I always get a little annoyed when I need to use this api as it's not exactly intuitive, especially when you want to use static methods. Won't said libraries need to update for the newest Fable release anyways since interfaces will be mangled at some point?

> Although not ideal, using a separate interface for the static part of a JS type was decided quite early on among other reasons, because it was not possible to inherit from multiple types in F# as @MangelMaxime points out. The convention is used in multiple bindings like Browser or Node and in the documentation, so changing or mixing it now could create confusion (although we could mention in the docs that it's possible to use a class for the binding too).

My suggestion was mostly for this specific module `Fable.Core.JS`, which shouldn't really have any cases where someone needs to inherit from them. That being said, ideally all bindings would preferably follow this style (in my opinion) as it's more user friendly. I wonder if a better solution is to have both interfaces and concrete types such that there would be a `Fable.Core.JS.Interfaces` and a `Fable.Core.JS` so that library authors could use interfaces when necessary, and the `Fable.Core.JS` concrete types would also implement them?

> About adding helpers to Fable.Core, I usually say I prefer to do that in a separate library (for example, helpers for JS promises are in the Fable.Promise package). The main reason is we're trying to keep Fable.Core as minimal and stable as possible, but also because Fable.Core is an "empty" Fable library, which means it doesn't contain actual code only metadata. When necessary, Fable injects some code or makes a compiler trick to resolve calls to Fable.Core as it does with FSharp.Core.

Yeah this makes sense, perhaps a `Fable.Extras` library is in order.

> So in summary, I would prefer not to make breaking changes in Fable.Core (although I'm open to it if other maintainers in the ecosystem think is a good timing for it and agree with the changes). But I do encourage you to publish a package with the proposed API (just please don't use the Fable.Core namespace) and see how users like it so we can incorporate it to Fable.Core in the future if it becomes the preferred option.

You've pointed out quite a few things I wasn't aware of that do make this a little tricky. I think it may indeed be better to just create a `Fable.Extras` that provides alternate bindings (this PR) and helper modules. There's a collection of libraries I *always* install for every single Fable project, and some extensions I make that I think could make it a nice package.

alfonsogarciacaro:
>  Won't said libraries need to update for the newest Fable release anyways since interfaces will be mangled at some point?

Ah, sorry. This is confusing because the discussion is scattered in several places. But at the end we decided to make interface mangling an opt-in by explicitly setting a `Mangle` attribute (vs using `NoMangle` for opting-out) to avoid the breaking changes and keep the ease of JS interop: https://github.com/fable-compiler/Fable/issues/2091#issuecomment-643868191

I will close #2078 as it can be misleading, please check the Nagareyama project to keep track of progress for the next major release: https://github.com/fable-compiler/Fable/projects/2

About the utility library, we had [Fable.PowerPack](https://github.com/fable-compiler/fable-powerpack) in the past. But we deprecated it because it became too miscellaneous, there were multiple contributions but it was difficult to make a consistent API. I would suggest to focus on a problem and create a package that solves that problem. For example, if you want to have an API similar to FSharp.Core modules for JS types you could create a package named Fable.Util.JS or Fable.Modules.JS or anything else not prefixed by Fable- ([example](https://fable.io/repl/#?code=HYQwtgpgzgDiDGEAEAxEAjANhAdAVQBcBLTHAKQGUBYAKFoHsYJhUNscBhegJ10ttoBtADwAlCAEcArkV4BFKSExEAZkQgATAILxEUKAD4AurTD0NU7EgCyIGEgC8tJC6TYCSCGBgEAnsIByAGsAGgCANwMkAC4bO0DQiKiHJC5gKAJuKXgCHigcWxhOXhACCAAKAEpnV3ckKAgPIKRwpHKwO1jChLDIysckDqKGgnLQluqaVzdGz2BM9Sg2oa744N6DfpShnGYF6CqBGgB6Y6QAFWhiYABzWlBIWARka18ABW56ACsIHKOzCxWS4ZRw1FyMZisLC4QgkI7TOplDJVUFTaauSgFOy7bx+YREeYhAkEAxg9EAHyimMKOBGSAAjEgAExk6aUpDU7F7biLIA&html=Q&css=Q)).

Shmew:
I have (tentatively) finished a library I'm going to publish as `Fable.Extras` that makes a more significant change in the JS api to enable piping which you can see [here](https://github.com/Shmew/Fable.Extras/blob/master/src/Fable.Extras/Extras.fs).

I noticed a few things along the way of writing this: 
* There are some missing APIs 
* Some cases where the type signatures are incorrect but kinda work, so changing them would result in a potentially breaking change (like the property descriptor get and set properties).
* Quite a few places where type safety can be improved such as requiring `'T : not struct` on some of the `Object` static methods and some places where `obj` isn't quite correct (such as some APIs that accept either Symbol or string)

I also added doc lines to all the methods. Please do take anything from this that y'all might find beneficial (such as copy and pasting the doc lines)!

alfonsogarciacaro:
Great work @Shmew, thanks a lot! Yes, definitely it'd be very helpful to reuse some of your work, the documentation alone would be super useful so I'll have a look before the stable release. And yes, the JS API may have some holes. When releasing Fable 2 I wanted to encourage the use of F#/.NET types that mapped to JS and removed some APIs but had to add them back later (like typed arrays). Somethings remain a bit weird, for example the JS `RegExp` constructor actually returns a .NET `Regex`.

Shmew:
An update on this, the library is now published! It has a **lot** more than what was originally written here, including things like being able to consume web assembly from Fable. The repo is [here](https://github.com/Shmew/Fable.Extras).

PR #2172, opened 2020-09-23
Updated fable-compiler-js

ncave:
- Added support for `outDir` in `fable-compiler-js`.
- Same as in Fable 2, if `outDir` is provided, compiled code is saved in the `outDir` folder, not next to the source file.


alfonsogarciacaro:
Thanks!

alfonsogarciacaro:
From a quick look I guess this is done like in Typescript (copying the full directory hierarchy so relative imports don't need to be changed) instead of flattening the directory tree as flable-splitter does, which I think now is the right way :+1: It would be nice to have this option in dotnet fable too.

ncave:
> It would be nice to have this option in dotnet fable too.

Sure, that's the next step.
> this is done like in Typescript

No, not exactly like that ;)
It's a bit hard to explain (although the code is very simple). It maps the `projectDir` to `outDir`, and all F# dependencies (project files) will be relative to it (but no higher). And non-F# dependencies are left where they are, just their imports are relative to the output.
Whereas TypeScript (from my limited observations) recreates all dependencies' folder hierarchy starting from the highest, and it copies all dependencies.

alfonsogarciacaro:
Yes, seems they do that because they don't want to touch the module imports: https://github.com/Microsoft/TypeScript/issues/5039#issuecomment-232470330

But I do agree the folder hierarchy can get too deep, I like your solution better! :clap:

PR #2175, opened 2020-09-25
Added support for outDir in Fable compiler

ncave:
- Added support for `--outDir` in Fable compiler.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -106,7 +106,7 @@
             "type": "node",
             "request": "launch",
             "name": "transform tests to JS",
-            "program": "${workspaceRoot}/build/fable-compiler-js/src/fable-compiler-js/src/app.fs.js",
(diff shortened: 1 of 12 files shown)


alfonsogarciacaro:
image

alfonsogarciacaro:
I just noticed this is not copying JS files (not translated from F#) to outDir as fable-splitter used to do, not sure if it's much of an issue.

ncave:
@alfonsogarciacaro Yes, intentionally, although it can be easily done by modifying a bit [here](https://github.com/fable-compiler/Fable/blob/nagareyama/src/Fable.Cli/Util.fs#L199-L201) and copying the (non-F#) import to the target location, something like that:

```fs
    let targetRelPath = getTargetRelPath importPath targetDir projDir outDir
    if not (importPath.EndsWith(".fs.js")) then
        <copy file from importPath to outDir + targetRelPath>
    targetRelPath
```

alfonsogarciacaro:
Ah, true. Makes total sense, thanks for the explanation!

PR #2176, opened 2020-09-26
Several fixes and optimizations

### LINKED ISSUE #2139: Problem with object expressions

### LINKED ISSUE #2167: getEnumerator missing on regex match.Groups
Using RegEx groups and getting the following warning `export 'getEnumerator' was not found in '../Proj/.fable/fable-library.2.13.0/RegExp.js'`

`match.Groups` is not iterable.

Fable version 2.13

Repo:

```fsharp

    let (|Regex|_|) pattern input =
        let m = Regex.Match(input, pattern)
        if m.Success then Some(List.tail [ for g in m.Groups -> g.Value ])
        else None
```

Workaround:

```fsharp
let (|Regex|_|) pattern input =
    let m = Regex.Match(input, pattern)
    if m.Success then
        Some [ for x in 1 .. m.Groups.Count - 1 -> m.Groups.[x].Value ]
    else 
        None
```

alfonsogarciacaro:
- Fix #2139 
- Fix #2167 
- Optimize getters in JS objects literals, [discussion](https://github.com/fable-compiler/Fable/pull/2165#issuecomment-695835444)
- Measure compilation time (F# and Fable part)
- Improve JS output: emitted JS code and don't bind `this` when it's not needed
- Improve Fable compilation time by reducing the calls to FCS `BasicPatterns`

@ncave About the last point, I did some profiling and (if I interpreted the results right) the `FSharp2Fable` step is still the one taking longest because calls to `BasicPatterns` are expensive so I tried to reduce them. For this I removed some active patterns we had built on top of `BasicPatterns`. The main changes are:

- I removed `ByrefArgToTuple` so now we have to pass a reference to replacements like `TryParse`
- I commented out the patterns to detect optimized operators. I think they are not needed because the optimization is disabled currently in nagareyama, but we can activate them again behind a compiler flag if needed.

In a simple test, when compiling FCS these changes make the Fable part run close to 20% faster in my machine (26 vs 32s).

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -45,7 +45,7 @@ Arguments:
 """
 
 type Runner =
-  static member Run(args: string list, rootDir: string, ?fsprojPath, ?watch) =
+  static member Run(args: string list, rootDir: string, ?fsprojPath, ?watch, ?testInfo) =
     let watch = defaultArg watch false
 
     fsprojPath
@@ -106,7 +106,8 @@ type Runner =
             { CliArgs = cliArgs
               ProjectCrackedAndParsed = None
               Watcher = watcher
-              WatchDependencies = Map.empty }
+              WatchDependencies = Map.empty
+              TestInfo = testInfo }
             |> startCompilation Set.empty
             |> Async.RunSynchronously
             |> function
@@ -162,6 +163,7 @@ let main argv =
         match commands with
         | ["clean"; dir] -> clean args dir
         | ["clean"] -> clean args rootDir
+        | ["test"; path] -> Runner.Run(args, rootDir, fsprojPath=path, testInfo=TestInfo())
         | ["watch"; path] -> Runner.Run(args, rootDir, fsprojPath=path, watch=true)
         | ["watch"] -> Runner.Run(args, rootDir, watch=true)
         | [path] -> Runner.Run(args, rootDir, fsprojPath=path)
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -180,7 +180,7 @@ module private Util =
             let writer = new FileWriter(com.CurrentFile, outPath, projDir, cliArgs.OutDir)
             do! BabelPrinter.run writer map babel
 
-            Log.always("Compiled " + File.getRelativePath com.CurrentFile)
+            Log.always("Compiled " + Path.getRelativePath cliArgs.RootDir com.CurrentFile)
 
             return Ok {| File = com.CurrentFile
                          Logs = com.Logs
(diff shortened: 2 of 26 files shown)


alfonsogarciacaro:
Btw, I'll be presenting about Fable 3 next month at F# eXchange, if you have some interesting benchmarks between Fable 2 & 3 @ncave that would be very helpful!

ncave:
> I commented out the patterns to detect optimized operators

It would be nice to regain support for optimized AST in some form in the future, as a lot of effort went into adding it. I can look into that when I get some time, hopefully it can be optimized to not add much to the run time.

I can confirm that this PR improves overall run time (by improving the Fable portion of the total time by 20%), so this is a definite improvement over Fable 2.

- Fable 3 (before this PR):
Total time: 73 sec
FCS time: 22 sec
Fable time: 51 sec

- Fable 3 (after this PR):
Total time: 62 sec
FCS time: 22 sec
Fable time: 40 sec

alfonsogarciacaro:
You're totally right 👍 I re-enabled the F# AST optimization in #2177 so we can keep experimenting.

Issue #2092, opened 2020-06-13
Improve type testing


alfonsogarciacaro:
Following the increased reflection support we could try to overcome the current limitations in Fable's type testing:

- Allow testing interfaces (except the `NoMangle` ones as they're supposed to represent JS objects). For that we should include the implemented interface names in the type (we already did something similar in Fable 1).
- Allow testing generic types. For that we need to inject the name of generic arguments to the constructors.

Shmew:
This would be great [look at this gross code](https://github.com/Shmew/FSharp.Control.AsyncSeq/blob/fable-support/src/FSharp.Control.AsyncSeq/AsyncSeq.fs#L794) 😅

alfonsogarciacaro:
I've tried different approaches: injecting generics as arguments, wrapping the constructors, trying to infer the generics from private fields... but I was not convinced by anyone, as none was perfect and they introduced complications without opening interesting scenarios. The only cases I've seen this is need is to cross-compile code for .net and Fable, so for now I will just enable testing generic types with a warning saying the generics are not taken into account. Hopefully that should be enough for most cases.

PR #2182, opened 2020-09-27
Use latest FSharpMap implementation

alfonsogarciacaro:
**EDIT**: Now this also includes the latest FSharpSet implementation and a bit of cleanup for the Array module. Simple benchmark [here](https://github.com/fable-compiler/Fable/pull/2182#issuecomment-700432366).

Sorry @ncave, got ahead of you with this ;) Seems the latest FSharpMap implementation is much faster so I tried using it. See https://github.com/dotnet/fsharp/pull/10188

There are some tests failing, I will fix them.

There's another PR open for FSharpSet. We could use it too or wait until the PR is merged https://github.com/dotnet/fsharp/pull/10190


ncave:
@alfonsogarciacaro Wrong target branch perhaps?

alfonsogarciacaro:
Ups, thanks!

alfonsogarciacaro:
Some simple tests:

```fsharp
measureTime <| fun () ->
    let map = Seq.init 1000000 (fun i -> Guid.NewGuid(), i) |> Map
    for _ = 1 to 1000000 do
        let g2 = Guid.NewGuid()
        let _ = Map.containsKey g2 map
        ()
```

Before: 12932ms
After: 8595ms

```fsharp
measureTime <| fun () ->
    let set = Seq.init 1000000 (fun _ -> Guid.NewGuid()) |> set
    for _ = 1 to 1000000 do
        let g2 = Guid.NewGuid()
        let _ = Set.contains g2 set
        ()
```

Before: 13122ms
After: 7997ms

ncave:
@alfonsogarciacaro Unfortunately this PR results in 5% slower performance in the FCS-JS benchmark. Please note I'm not saying it needs to be rolled back, or even that the FCS-JS benchmark is what we should be using to measure performance, just stating the fact.

I haven't looked at perf reports to see where the loss is coming from, cause most of the time it's quite hard to pinpoint it.
Unfortunately micro-benchmarks can some times be misleading. Case in point is #2154, where the `ResizeList` is more than twice as fast than the LinkedList in micro-benchmarks (on both dotnet and nodejs), but for some reason it results in 10% slower perf on the FCS-JS benchmark.

alfonsogarciacaro:
Interesting, thanks for checking this 👍 I have no problem in reverting the change, it looks indeed like some changes that look good in microbenchmark can cause deoptimizations in bigger apps. That said, I made the mistake of adding some cleanup for Array.fs in this PR, I hope it was not related. Can you please check #2187 (reverts Map/Set but not Array) to see if it fixes the performance regression?

It'd be nice to have another computation intensive app to test besides FCS as a tie-breaker in these situations. Hopefully after announcing Nagareyama some users step up as it happened with Fable 2.

ncave:
@alfonsogarciacaro There's quite a few `Reflection_typeTest` in `build/fable-library/Map.js` and `build/fable-library/Set.js`, perhaps those can be optimized.

alfonsogarciacaro:
Ah, you're right! These were part of my (unsuccessful) effort to support generic parameters in type testing #2092, we should likely compile them as `instanceof`, keeping or not keeping the warning as it can be quite verbose.

Atm `Reflection_typeTest` can only test generic parameters for arrays, lists and options, so it doesn't add a lot of value. We can remove and go back to [inline type testing as before](https://github.com/fable-compiler/Fable/blob/bd1a9944a6c16f3e7796c27a1c42a3cb1cb2d5cb/src/Fable.Transforms/Fable2Babel.fs#L1136) (without the error for generic types).

ncave:
> We can remove and go back to inline type testing as before (without the error for generic types).

@alfonsogarciacaro Yes, perhaps we should keep the type testing change as a separate PR so we can test performance impact.

PR #2191, opened 2020-10-02
Fixed import extensions

ncave:
- We're not rewriting all imports anymore so they have to have extensions to start with.


alfonsogarciacaro:
Thanks!

PR #2192, opened 2020-10-02
Updated FCS build steps

ncave:
- I've rebased my FCS fork to latest and updated the instructions.
- Updated projects to reference both `FSharp.Compiler.Service` and `FSharp.Core` from `lib/fcs`.
- Tests be broken until FCS is built (see instructions) and copied in `lib/fcs`.


alfonsogarciacaro:
Thanks, I'll do it!

alfonsogarciacaro:
It's working :) In a quick test it doesn't show a big leap in performance, although it does seem to be faster around 5-10% so that's something. Thank you!

PR #2186, opened 2020-09-29
Updated command line args

ncave:
- Minor update in command line args.
- `--optimize` is now included in project options.

PR #2194, opened 2020-10-05
Add back base Union/Record classes

alfonsogarciacaro:
When trying to compile Fable.SimpleJson with Nagareyama we realized the code that relied on Unions and Record inheriting some behavior (like `toJSON` implementations) could easily break so I've tried to apply @chrisvanderpennen solution (see #2160) to bring back the base classes for unions and records ...although it seems @Zaid-Ajaj already managed to fix Fable.SimpleJson so maybe this is already outdate 😅 

https://github.com/Zaid-Ajaj/Fable.SimpleJson/pull/54

In microbenchmarks initializing derived classes takes longer, but I couldn't see an appreciable difference when running the benchmark (see #2186) between this branch and current `nagareyama`.

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -325,6 +325,7 @@ let excludeFilesAndSignatures (exclude: string option) (files: string[]) =
         | None -> filesToCompile
 
 let rec startCompilation (changes: Set<string>) (state: State) = async {
+    // TODO: Use Result here to fail more gracefully if FCS crashes
     let cracked, parsed, filesToCompile =
         match state.ProjectCrackedAndParsed with
         | Some(cracked, parsed) ->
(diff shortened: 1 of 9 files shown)


ncave:
@alfonsogarciacaro The other option is to store the union/record base type in Reflection, so it can be properly identified if needed at runtime, instead of looking at the object shape. Using Reflection understandably has a perf cost if (over)used by default for type checking (unless we can optimize it more). If using base classes is faster than using Reflection and/or works better and doesn't impact performance, then sure, whatever makes sense.

ncave:
@alfonsogarciacaro I can confirm there is no perf impact from this PR, actually there is a small perf gain (~ 3% faster, which is small but still in the right direction).

alfonsogarciacaro:
Ah, great. Thanks for checking that! Maybe the perf gain comes because we can call now `Equals/CompareTo` directly for unions and records instead of checking the shape of the object every time. In a separate PR we could try to check if generating a specialized Equals/CompareTo method (like the F# compiler does) for unions/record/tuples used as Map/dictionary keys improves things :+1:

PR #2198, opened 2020-10-10
fable-library update

ncave:
- Split out `MapUtil.ts` from `Util.ts` to remove circular dependency.


alfonsogarciacaro:
I wanted to deprecate `keyValueList` so didn't pay too much attention to the circular dependency. But we will probably to keep it around for a while so this is a good change, thanks!

Issue #2197, opened 2020-10-10
byte[] in Fable 3 is no longer UInt8Array?


Zaid-Ajaj:
Hi @alfonsogarciacaro, I was hitting a wall when trying to send binary data to the server and it turns out Fable 3 (currently 008-alpha) doesn't treat byte[] as a typed `UInt8Array` during runtime but instead as a normal array of numbers. Is this intentional?
```fs
let input = [| byte 1; byte 2; byte 3|]
``` 
The `input` from the line above would compile to `UInt8Array` in Fable 2.x but it compiles to an array of numbers in Fable 3. 

This is workable, however it is a bit inconvenient and would cause a bunch of breaking changes to existing code that assumes the typed array representation (when uploading, downloading etc.)

alfonsogarciacaro:
Sorry for that @Zaid-Ajaj. This was changed in #2090, although the behavior was untouched, only the default value for the `--typedArrays` option changed. This was because there were bugs coming from time because of issues with JS interaction due to the typed arrays and I thought nobody was using them. I'm happy to see they were useful for you 😄 

I think we can avoid the breaking change in two ways:

- Go back to `typedArrays=true` by default. Besides JS interaction, a problem can arise when a library relies on this but a user set it to default.
- Let libraries define their own Fable compiler options properties (currently the options set by the user are used for all the sources), probably through properties in the .fsproj as Femto does.

Zaid-Ajaj:
Thanks for the follow up @alfonsogarciacaro, I don't think that libraries should provide compiler options, this goes against the idea of compiler options in the first place: imagine two libraries defining different Boolean values for a compiler option, that would be quickly ambiguous. Instead, I like option 1 more to retain backward-compat from Fable 2,x

If possible however, somehow exposing the compiler options and compiler versions to client libraries would be great, so that infra libraries can provide compatible behaviour.

alfonsogarciacaro:
3.0.0-nagareyama-alpha-009 is release with typedArrays as default again :+1: (Use `--noTypedArrays` option to disable the behaviour)

Issue #2090, opened 2020-06-13
Make typedArrays=false the default in Fable 3?


alfonsogarciacaro:
Issues like #2089 keep coming from time to time when typed arrays cause an unexpected problem when interacting with JS.

Maybe we should make `typedArrays=false` the default in Fable 3? So we can set the option only for applications that can benefit from this feature (like when self-compiling the F# compiler and Fable).

ncave:
@alfonsogarciacaro Sure, if you think it makes sense as a default. It's turned off by default for typescript output as a temporary workaround until we can make it work well with types.

alfonsogarciacaro:
As discussed in #2197, we are reverting this change to avoid breaking changes in existing code. However typed arrays will be disabled when outputting Typescript (see comment immediately above).

PR #2201, opened 2020-10-11
Set typedArrays=true as default (as in Fable 2)

alfonsogarciacaro:
See #2197 and https://github.com/fable-compiler/Fable/issues/2090#issuecomment-706692819

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -157,7 +157,6 @@ let buildStandalone(minify: bool) =
         "--outDir " + buildDir + "/bundle"
         "--fableLib " + libraryDir
         "--forcePkgs"
-        "--typedArrays"
         "--define FX_NO_CORHOST_SIGNER"
         "--define FX_NO_LINKEDRESOURCES"
         "--define FX_NO_PDB_READER"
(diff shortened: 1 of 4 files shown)


ncave:
@alfonsogarciacaro This is good, but IMO to make reasoning about the options a bit easier (since we're changing the default in the case of `--typescript`) we should do *one* of the following:

- retain the opposite option as well (i.e. `--typedArrays`), or merge the two into an option with a boolean value.
- or, untangle the two options and not change the default `typedArrays` to `false` in case of `--typescript`, but require `--noTypedArrays` explicitly.

alfonsogarciacaro:
You're right, it's uncommon to have a negative flag. I was thinking this would make it the default behavior clearer, but we already had `typedArrays` in JS-based config for fable-loader/splitter and it was not a problem.

It can also be confusing that typed arrays are automatically disabled when producing typescript, we should indeed give a better  warning to the user.

So mixing a bit your proposals:

- Keep `--typedArrays` flag and change flags behavior so you can pass a boolean to enable/disable (I think we were doing this before), so `--typedArrays false`, `--typedArrays true` or `--typedArrays` are valid. (For typed arrays the default will be true if you don't pass the flag as with Fable 2.)

- When passing `--typescript` flag, fail if typed arrays is not false. As you say, this is a bit inconvenient because you have to pass `--typedArrays false` explicitly, but hopefully it's just temporary until typed arrays work with typescript.

How does that look?

PR #2209, opened 2020-10-16
Fixed entity resolution

ncave:
- Fixed entity reference resolution.


alfonsogarciacaro:
What problems did you see? Actually I wanted to ask your opinion about the entities' map, in case you could think of a better alternative for the serialization.

I have some changes that may conflict with this in the plugins branch. Actually I was testing if the dictionary could be non-concurrent and read-only if all the entities were loaded at the beginning. But you may have found an example where this is not the case?

ncave:
> What problems did you see?

Mainly that entity map keys have the assembly in the full name, so that needed to be taken into account when searching for parent and nested entities. I'll retarget it to the plugins branch.

ncave:
@alfonsogarciacaro I rebased it against `nagareyama-plugins`.

alfonsogarciacaro:
Thanks!

Issue #2099, opened 2020-06-19
Don't include name in union constructors


alfonsogarciacaro:
I did this for Fable 2 to avoid listing all the names of big unions meant to represent JS options (as with React props) but in most cases is more efficient to list all the names only once. With this change, for example, the minified version of F#/Fable compilers bundle gets reduced by around 80kb.

PR #2215, opened 2020-10-19
Minor improvements in code generation

alfonsogarciacaro:
Unfortunately, I've noticed Nagareyama produces a bigger bundle in React apps than Fable 2 (e.g. in [fulma-demo](https://github.com/MangelMaxime/fulma-demo) 163KB vs 226K). This helps, but only a little.

### DIFF
--- a/src/Fable.AST/Babel.fs
+++ b/src/Fable.AST/Babel.fs
@@ -810,7 +810,7 @@ type ObjectMethod(kind_, key, ``params``, body, ?computed_, ?returnType, ?typePa
             printer.PrintOptional(returnType)
             printer.Print(" ")
 
-            printer.Print(body)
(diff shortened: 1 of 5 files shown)


Zaid-Ajaj:
Hmm that's funny, in my non-React projects, mainly unit test projects, the bundle size became smaller 354KB ->268KB 🤔

ncave:
@alfonsogarciacaro Yes, in other non-React projects like `FCS-Fable` the (minimized) bundle stayed more or less the same size as in Fable 2.

alfonsogarciacaro:
I thought it could be because I removed the `keyValueList` optimization, but I'm adding it back here and it doesn't have much effect as I suspected. It could be that now we're adding the `cases` method to all unions (Fable.React and Fulma have unions with many, many cases) #2099

PR #2223, opened 2020-10-23
Units of Measure Overload Resolution

Shmew:
This adds the test case to show how overload resolution doesn't function quite the same when doing some hacky stuff to make units of measure overloads work.

Here you can see Expecto runs the overload cases correctly (BTW there are a couple tests failing):

image

However when running via Mocha:

image

### DIFF
--- a/tests/Main/CustomOperatorsTests.fs
+++ b/tests/Main/CustomOperatorsTests.fs
@@ -67,6 +67,28 @@ let operatorsAsFunctions = [
         x |> equal false
 ]
 
+type [<Measure>] px
+type [<Measure>] em
+
+type D1 = D1
+type D2 = D2
+type D3 = D3
+
+type ToLength = ToLength with
+    static member (&.) (ToLength, x: int<px>) = fun D1 _ _ -> String.Join("", string x, "px")
+    static member (&.) (ToLength, x: int<em>) = fun D1 D2 _ -> String.Join("", string x, "em")
+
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
This seems to be not an issue with Fable overload suffixes as I thought at first, units of measure are correctly being represented in the hash. However because you're using the inline trick to resolve the overload, Fable gets confused. F# witnesses should help with this. They seem to have been fixed in FCS so we can try this out after merging the feature.

You're probably aware of this but if we use "normal" methods, F# compilation fails because the signatures are not different after erasing units of measure.

```fsharp
type ToLength = ToLength with
    static member Make(x: int<px>) = String.Join("", string x, "px")
    static member Make(x: int<em>) = String.Join("", string x, "em")

// Duplicate method. The method 'Make' has the same name and signature as another method in type 'ToLength' once tuples, functions, units of measure and/or provided types are erased. (code 438)
```

A trick to work around this without inlining is to add a "phantom" optional type, which works with Fable. You can try this or we can wait for F# witnesses and see.

```fsharp
type ToLength = ToLength with
    static member Make(x: int<px>, ?_phantom: int) = String.Join("", string x, "px")
    static member Make(x: int<em>, ?_phantom: int * int) = String.Join("", string x, "em")
```

Shmew:
Oh that's an interesting idea. I'll try it out when I get a chance.

PR #2227, opened 2020-10-23
Use invariant culture for consistent number formatting

### LINKED ISSUE #2226: Fable 3 outputs a wrong decimal separator depending on regional setting
### Description
Depending on the regional setting, the decimal separator output by Fable is wrong and leads to a SyntaxError when bundling.

### Repro code
```
let foo = 0.5
```

Expected output:
```
export const foo = 0.5;
```
Actual output (with e.g. regional format Sweden):
```
export const foo = 0,5;
```

Regional format set to "English (Sweden)". Setting it to "English (United States)" fixes the problem.

### Related information
* Fable version: 3.0.0-nagareyama-beta-001
* Windows 10

bklop:
Fixes #2226

### DIFF
--- a/src/Fable.AST/Babel.fs
+++ b/src/Fable.AST/Babel.fs
@@ -380,7 +380,7 @@ type NumericLiteral(value, ?loc) =
     interface Literal with
         member _.Print(printer) =
             let value =
-                match value.ToString() with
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
I had the feeling that `ToString()` was missing something ;) Thanks a lot!

Issue #2221, opened 2020-10-22
Simplify import aliases when there are no collisions in Fable 3


Zaid-Ajaj:
### Description

Imports are now always aliased even if used once and no other import has the same name. 

If we are able to simplify
```js
import { createElement as react_createElement } from 'react'
```
into just
```js
import { createElement } from 'react'
```
that would be great for when debugging the code if it is not too much work to implement

* Fable version: nagareyama-16
* Operating system: Windows 10

alfonsogarciacaro:
Well, this was actually intended, because before if you imported map from Option, Array, List, it was only `map`, `map_1`, `map_2` and I thought the origin was clearer this way. I have to write a post about the code generation in Nagareyama, I will point to this issue so users can say if they prefer the "old" or "new" way (or an alternative).

Zaid-Ajaj:
> Well, this was actually intended, because before if you imported map from Option, Array, List, it was only map, map_1, map_2 and I thought the origin was clearer this way

It makes sense for `map` because of multiple modules might implement it but that is an edge case rather than the norm. 

While I am spamming JS tooling repositories with compiled React code, I keep getting feedback that it is preferred to have the imports simplified, _especially_ when it comes to hooks which maintainers tell me that the import _must_ start with `use` (e.g. `useState`, `useEffect` etc.) 

In some cases I managed to get fast-refresh working without this restriction, however I think that users would love if this wasn't a random hit or miss when they have something more complicated. 

Here is my proposal: have some kind of ability to _invert_ the local import name so that the format becomes `{member}From{Module}` rather than `{Module}_{member}`. This way when we write custom hooks or import external ones, the import becomes
```js
import { useState_Static_1242 as  useState_Static_1242FromReactFromFeliz } from "./.fable/Feliz.1.16.0/React.fs.js"
```
instead of 
```js
import { useState_Static_1242 as Feliz_React_useState_Static_1242 } from "./.fable/Feliz.1.16.0/React.fs.js"
```
I know it looks a bit silly and ideally the alias wouldn't be needed really a lot of the times but for React hooks it will make interop better. What do you think? 

I tried looking in the code where these things happen and I came across `compiler.GetAllImports()` but not sure where is the function that actually adds the underscore between the module and the member within the import statement

alfonsogarciacaro:
[...]

PR #2232, opened 2020-10-26
Omit module prefix on module imports

Zaid-Ajaj:
As per @alfonsogarciacaro's comment [here](https://github.com/fable-compiler/Fable/issues/2221#issuecomment-716322713)


alfonsogarciacaro:
Thanks!

PR #2239, opened 2020-10-27
Updated fcs-fable to latest

ncave:
- Updated `fcs-fable` to latest.
- Updated `FSharp.Core` reference.


alfonsogarciacaro:
Thanks a ton! Do we still need to reference the custom-built FSharp.Core.dll from Fable.Cli or can we use FSharp.Core 5 from Nuget?

ncave:
@alfonsogarciacaro This is a bit tricky to answer, depends on what version of `FSharp.Core` was actually used to build `FCS`.
Yes, in theory it should be `FSharp.Core 5.0.0` from nuget, but that depends on this variable being true when you build FCS:
```xml
    <FSHARPCORE_USE_PACKAGE Condition="'$(SolutionName)' == 'FSharp.Compiler.Service'">true</FSHARPCORE_USE_PACKAGE>
```
You can try referencing `FSharp.Core 5.0.0` from nuget in `Fable.Cli`, if that doesn't work properly we need to make sure this variable is properly set when building FCS.

PR #2240, opened 2020-10-27
Witnesses for STRP resolution

### LINKED ISSUE #2104: Witnesses

alfonsogarciacaro:
Fix #2104 

Take advantage of the new "witness passing" feature from F# compiler to improve (or actually fix) SRTP resolution in Fable. Discussion here: https://github.com/dotnet/fsharp/pull/9510#issuecomment-648637543


alfonsogarciacaro:
@ncave This is passing now the first round of tests. I guess now we have to rebase `fcs-fable` too, but we can wait until https://github.com/dotnet/fsharp/pull/9510 is merged :+1:

alfonsogarciacaro:
@Shmew BTW, when testing #2223 in this branch it works!

alfonsogarciacaro:
For some reason, there's no CI data for many of the the commits in this PR so I'm not sure exactly when it started happening but I just realized there are now many warnings like the following ([Appveyor log]()) when compiling the tests with fable-compiler-js. @ncave Would you happen to have a hint?

```
C:/projects/fable/tests/Main/ApplicativeTests.fs (834,9): Warning: The struct, record or union type 'Program' is not structurally comparable because the type 'SetState<'model,'msg>' does not satisfy the 'comparison' constraint. Consider adding the 'NoComparison' attribute to the type 'Program' to clarify that the type is not comparable
C:/projects/fable/tests/Main/ApplicativeTests.fs (834,9): Warning: The struct, record or union type 'Program' does not support structural equality because the type 'SetState<'model,'msg>' does not satisfy the 'equality' constraint. Consider adding the 'NoEquality' attribute to the type 'Program' to clarify that the type does not support structural equality
C:/projects/fable/tests/Main/ArithmeticTests.fs (582,8): Warning: The value has been copied to ensure the original is not mutated by this operation or because the copy is implicit when returning a struct from a member and another member is then accessed
```

ncave:
@alfonsogarciacaro I think those may be just improved error reporting from FCS. I had something similar when rebasing FCS-Fable, but it was an error, not a warning, so I had to actually [fix it](https://github.com/fable-compiler/Fable/pull/2247/files#diff-82b4bad4aacc2795d7b160d19cc8c5facc9db7be08afaf59ba6f38027e8059d1). So this must have been introduced in FCS after 2020-10-01.

alfonsogarciacaro:
Ah ok, thanks!

PR #2248, opened 2020-11-03
Moved Choice and Result types

ncave:
- Moved Choice and Result types to their own file.


alfonsogarciacaro:
Makes sense, thank you!

PR #2266, opened 2020-11-11
+ Overload for StringBuilder.Append(StringBuilder)

gusty:
This will add the missing overload for StringBuilder


alfonsogarciacaro:
Awesome, thanks a lot!

PR #2268, opened 2020-11-12
Allow direct console output preserving colors. Fixes #2267.

### LINKED ISSUE #2267: dotnet fable does not preserve webpack console output colors in Windows Command Prompt
- `npx webpack`:

image

- `dotnet fable App --run webpack`:

image

- `npx webpack-dev-server`:

image

- `dotnet fable watch App --run webpack-dev-server`:

image

### Related information

* Fable version: 3.0.0-nagareyama-rc-004

zanaptak:
I guess redirecting stdout discards color sequences. Since we are only echoing the output to console anyway, there seems no need to redirect. We can just let the process write to console directly, which fixes the colors.

There may be some potential issue with `Fable.Verbosity.Silent` but that appears to be unused for now. If that is used in the future then either: 1) expect user to also set whatever appropriate silence flag on the child process, or 2) we can conditionally, only if Silent is enabled, wire up the redirection events and discard the output.

Fixes #2267.

### DIFF
--- a/src/Fable.Cli/Util.fs
+++ b/src/Fable.Cli/Util.fs
@@ -103,18 +103,11 @@ module Process =
         //     psi.EnvironmentVariables.[envVar.Key] <- envVar.Value
         psi.FileName <- exePath
         psi.WorkingDirectory <- workingDir
-        psi.RedirectStandardOutput <- true
-        psi.RedirectStandardError <- true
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Oh, didn't know this was possible 😅 Works great, thank you!

PR #2298, opened 2020-11-25
Ignore some base constructor calls

ncave:
- Ignore some base constructor calls (e.g. for BCL types):

```fs
type MyList<'T>() =
    inherit ResizeArray<'T>()
    member x.Length = x.Count
```

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -986,22 +986,31 @@ module Util =
         let enumerator = CallExpression(get None (Identifier "this") "GetEnumerator", [||]) :> Expression
         BlockStatement [|ReturnStatement(libCall com ctx None "Seq" "toIterator" [|enumerator|])|]
 
-    let extractBaseExprFromBaseCall com ctx baseCall =
-        let baseRef, args, hasSpread =
-            match baseCall with
-            | Fable.Call(baseRef, info, _, _) ->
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
I guess this is related to #2154. Is it a good solution to just ignore the base constructor call? Or should we try to inherit from JS array instead?

ncave:
@alfonsogarciacaro No, it's not related, I just notices that some base constructor calls are in the form of `Value` instead of a `Call`. Not sure if that's a bug in FCS or is intended, so this PR just provides an opportunity to replace the hard error with a warning.
Technically in this case, yes, the base type should be JS array, but there are other base BCL types we don't have a clear JS replacement for, so I think it's ok to just warn. Inheriting from BCL types is a fringe practice anyway, should be composition instead of inheritance most of the time, unless there is a replacement for the base type already, like in the case of `Dictionary`.

alfonsogarciacaro:
Yes, some calls to the BCL have an unexpected representation in the AST, for example `DateTime.MinValue` appears as `BasicPatterns.ILFieldGet`. We do have a function to get the reference/constructor of a native JS type, which we could try to use later. But for now, I guess the warning message is enough to let the user know something is going funny. Thanks!

https://github.com/fable-compiler/Fable/blob/e0590b20d5541ac222e69a9a527b21812434ea6d/src/Fable.Transforms/Replacements.fs#L999-L1025

PR #2301, opened 2020-11-29
Polling watcher

mlaily:
Hi,

I am using Fable from WSL2 (Because I prefer to keep Node.js there), but my IDE and my project files are on Windows.

It works more or less ok, but the file watcher doesn't (I know it would if I kept all my files on the Linux side, but that's not the point).

I didn't find a workaround for Fable 3, so I implemented a polling watcher as an alternative to the normal `FileSystemWatcher`, that can be enabled the same way it can be enabled on `dotnet watch`: by setting `DOTNET_USE_POLLING_FILE_WATCHER` to `1` or `true`.

(I ported the implementation from https://github.com/dotnet/aspnetcore/blob/master/src/Tools/dotnet-watch/src/Internal/FileWatcher/PollingFileWatcher.cs)

It works, but it could certainly be improved:
- it's pretty slow, since it's watching thousands of files (thanks to `node_modules` among others...)
- I don't feel totally confident about the way I changed `FsWatcher` (the observable and the watcher being created each time `FsWatcher.Observe` is called). I fear a little I have missed the big picture about how this is supposed to work...
- I'm not entirely satisfied about my implementation of the switch in `FsWatcher` between the polling watcher and the normal one. It might be possible to improve it by making both watchers implement a common interface...
- There are no automated tests.

Is this PR welcome in Fable?

If yes, what is your opinion about it, especially about the points I mentioned above?

PR #2305, opened 2020-12-02
Use simple F# script for building

alfonsogarciacaro:
This tries to lower the entry barrier for contributors wanting to test a local version of Fable, so they can just run `dotnet fsi build.fsx` without even running `npm install` or `dotnet tool restore`. @ncave If possible please have a look, I hope I'm not breaking any of your benchmark scripts.

The main drawback is we lose some dogfooding, but compiling the script to js has little justification now as we are not using specific js functionality besides calling npm scripts. However, I've tried to keep PublishUtils.fs compatible with node.js just in case. Hopefully we can use the common IO interfaces for other purposes, like unifying code between Fable.Cli and fable-compiler-js.

README needs to be updated, but maybe we can do that when releasing Fable 3 officially.


Fable.sln
```diff
@@ -26,14 +26,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Files", "Files", "{54BD4E45
 		build.fsx = build.fsx
 	EndProjectSection
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Fable.Standalone", "src\fable-standalone\src\Fable.Standalone.fsproj", "{D5817BB2-0664-4E0A-BBD1-06615D705E99}"
-EndProject
```
alfonsogarciacaro, line 30: I removed Fable.Standalone from the solution to avoid issues [like this one](https://github.com/fable-compiler/Fable/pull/2301#discussion_r532278149). If necessary, we can add a separate solution with fcs-fable + Fable.Standalone.

build.fsx
```diff
@@ -418,14 +406,28 @@ let githubRelease() =
...
+}}, 'fable-compiler', 'Fable', {{
+    tag_name: '{version}',
+    name: '{version}',
+    body: [{notes}].join('\n'),
+}}, (err, res) => {{
+    if (err != null) {{
+        console.error(err)
+    }}
+}})"
+"""
```
alfonsogarciacaro, line 426: Did I say that I love string interpolation already?

package.json
```diff
@@ -2,10 +2,9 @@
   "private": true,
   "scripts": {
     "tsc": "tsc",
-    "postinstall": "dotnet tool restore",
-    "build": "dotnet fable build.fsx --runScript",
-    "publish": "npm run build -- publish",
-    "test": "npm run build -- test",
+    "build": "dotnet fsi build.fsx",
+    "publish": "dotnet fsi build.fsx publish",
+    "test": "dotnet fsi build.fsx test",
```
alfonsogarciacaro, line 7: This makes this PR compatible with "old" npm commands.

build.fsx
inosik: We got some serious meta-programming right here! 😄

ncave:
@alfonsogarciacaro 

>I've tried to keep PublishUtils.fs compatible with node.js

Thank you, this is great, it keeps both use cases viable:
- Using F# build scripts without any Node.js installed (`dotnet fsi` + `PublishUtils.fs`).
- Using F# build scripts without any .NET installed (`fable-compiler-js` + `fable-publish-utils`).

PR #2310, opened 2020-12-10
[WIP] SpreadElement without parenthesis 

do-wa:
Hi,

I've splitted this merge request into two commits.
1. The oldest commit contains a new added test case which will fail when executed without the fix in the latest commit
2. The latest commit contains a (hopefully) fix for the fable compiler which will generate no parenthesis around SpreadElement Expressions

**The Problem**
Calls to JsFunc.Invoke (and maybe others) with a non-inlined obj array syntax are translated to `(...array)` calls which produces invalid Js Code due to the parenthesis.

**What is an inlined-array syntax:** `.Invoke([|...|])`
_If this is not the right term for this construct, please let me know_

I've stumbled across this problem when I was using the JsInterop Module for some TypeProvider work and noticed, that the conversion for JsFunc.Invoke calls are transformed differently when used with different syntax.

[Here is a simple repo example from the Fable Repl to see the problem in the compiled code](https://fable.io/repl/#?code=PTAEHUCcEsBcFNQGMD2ATRLKgDYoIZqj6gDO+AtgA46IBmkKFZ0GARvpALABQKV8AHagAYvja0AdAGEs8SQClSASUEJGVXvyGjxU2ZHi9eIUAHcsAa1K9asUAlKxwcABYBBSJHwBPUAAoASlAAXlBeUEjceHtHewAuXQl5A3klVXV+RVIRAFdBJFDQaGosWHccHFAAIlImeABaCnRc2mqo8J4O01hXRGhBADcUS0QkfErQAHMY0mQsQyRYHD9Yb0FSHHwEIlgUB3gnf2r8aoAaarZqwIiouMlVYdH-AG0AH1A2FAAPGtOAbk+PxqV1AbwAujceCYwGgUIdQFRGGhckh+pV4FMJqAAFZzVAYAIwKauWAAfihdgOThcvU8kB8AAVDNJDNt4EQgkVbpEqXFQIkxMkZHJshl4BpsnkCkUSlQyhUqrV6k0Wm0edF7L8wu8gb8TtVAV99aCIRqen1ikMRmMJlUZrA5gNUF54EsVg51pt2bt9nF-P5JEHvoEoR17o8bf4Q7wgA&html=Q&css=Q
) 

The test case I've written is a minimum reproduction example. I've noticed that more complex scenarios are not yielding the same results as my fable repl example and actually failed. My guess is, that the repl example is different by design and any attempt to  mimic this with actual f# functions will fail.

In addition, my current take is that even translating the JsFunc.Invoke call into a JS Spread Op isn't the final answer and it should probably be consistent with any call using the inline syntax `.Invoke([|....|])`. If you agree and someone can point me to the right direction  I'm happy to help (This is my first day navigating this awesome repo and I haven't visited every corner ;)) 

I hope this will help.
Greetings

Dominik

### DIFF
--- a/src/Fable.AST/Babel.fs
+++ b/src/Fable.AST/Babel.fs
@@ -195,6 +195,7 @@ module PrinterExtensions =
             | :? CallExpression
             | :? ThisExpression
             | :? Super
+            | :? SpreadElement
             | :? ArrayExpression -> expr.Print(printer)
             | :? ObjectExpression ->
                 match objExpr with
--- a/tests/Main/JsInteropTests.fs
+++ b/tests/Main/JsInteropTests.fs
@@ -375,7 +375,7 @@ let tests =
         style.FontSize(true) |> equal "foo"
         style.FontSize(false) |> equal "bar"
 
-    testCase "Emit wors with ParamArray" <| fun () ->
+    testCase "Emit works with ParamArray" <| fun () ->
         let style = createEmpty<TextStyle>
         style.Sum(1.5, 2, 3, 4) |> equal 10.5
 
@@ -442,6 +442,13 @@ let tests =
     testCase "Stringifying an F# type in JS works" <| fun () ->
         let addFooString (x: obj) = importMember "./js/1foo.js"
         { Float = 7. } |> addFooString |> equal "35x foo"
+
+    testCase "JsFunc.Invoke call with param array is spreaded in JS" <| fun () ->
+      let fn : Fable.Core.JsInterop.JsFunc = !!(fun (arg:int) -> arg+1) // actual implementation is not important
+      let arg = [|box 1|] // invoke wants an obj array
+      fn.Invoke(arg)  // fable compiler will spread this argument
+      |> unbox 
+      |> equal 2
 #endif
 
     testCase "Pattern matching with StringEnum works" <| fun () ->


alfonsogarciacaro (approved):

alfonsogarciacaro:
Thanks @do-wa! It does help a lot indeed 👏 👏 👏 I think this is good to merge, you're definitely right the spread shouldn't be wrapped by parens. About the following comment:

> In addition, my current take is that even translating the JsFunc.Invoke call into a JS Spread Op isn't the final answer and it should probably be consistent with any call using the inline syntax .Invoke([|....|]).

Do you mean that when passing an array to `JsFunc.Invoke` it should be applied as first argument instead of spreading it? (that is, compile `let ar = [|box 1; box 2|] in myFunc.Invoke(ar)` as `myFunc(ar)`) If that's the case, yes, using `ParamArray` with `obj[]` argument does create an ambiguity when passing an `obj[]` because we don't know if it's a single argument or it should be spread. Unfortunately, we cannot "fix" it without breaking .net/f# semantics as in this case the compiler resolves the ambiguity by spreading the array and some users may rely on this behavior. The "solution" would be to box the array when applying it: `myFunc.Invoke(box ar)`.

If you don't mind, I will merge as is because I think it's a good fix. If we need to make changes to `JsFunc` we can do them in a separate issue/PR (btw, I thought nobody was using it and almost removed it in latest release 😅).

do-wa:
Thanks for the merge :) 
The PR is closed but I'll answer here.

I should have provided an example and a better explanation, so I'll try again:

**Variant 1:**
`myFunc.Invoke([|box foo; box bar|]` is compiled to `myFunc(a,b)`
This is the right solution since `.Invoke` is exactly for this use case?

**Variant 2:**
```
let args = [|box foo; box bar|]
myFunc.Invoke(args)
```
is _now_ compiled to `myFunc(...x)` 
This is also the right solution since both variants are technically the same in JS (at least thats my understanding from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax))

So now we have technically the same outcome for both variants. What I meant was (and this is partly just my curiosity) where is the decision made to treat them differently, why are they treated differently and shouldn't both variants yield the same result (preferably the first?) 

----

I'm relying on the JsFunc.Invoke Method for a Typescript TypeProvider to replace the function calls from TP types with the actual JS function call. So please wait until I'll have a better solution before removing it 😄

PR #2312, opened 2020-12-12
Fix build

inosik:


alfonsogarciacaro:
Oh my, this is what happens when I move the test from Quicktest and push without rebuilding 😅 Thanks a lot!

inosik:
The compiler can only save us from ourselves if we indeed do compile our code from time to time 😅

PR #2319, opened 2020-12-17
Reduce the memory usage when creating char ranges

### LINKED ISSUE #2318: System.Char.MinValue .. System.Char.MaxValue range
https://github.com/lfr/fable-compiler-validation-blocks

This is a slightly modified Fable new project app, but it never finishes loading

### Related information

* Fable version: 2.13.0
* Framework: net5.0
* Windows

inosik:
Fixes #2318.

@alfonsogarciacaro ~Is it ok to use `rangeNumber` here or should I use `unfold` and `delay` as it was before?~


alfonsogarciacaro:
Thanks a lot!

PR #2324, opened 2020-12-19
Fix minor issues with the file system watcher tests on Windows

mlaily:
I also added the integration tests project to the solution.

If it was omitted on purpose, I can remove the commit.

### DIFF
--- a/Fable.sln
+++ b/Fable.sln
@@ -1,9 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26124.0
+# Visual Studio Version 16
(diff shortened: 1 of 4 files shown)


mlaily:
Ah, I didn't notice your latest commits to enable/disable running the integration tests in the CI.

This should be ok now.

mlaily:
After having a hard time understanding what was happening, I managed to fix the "access is denied error" thrown by the tests on Windows!

@alfonsogarciacaro I think this PR is ready, let me know what you think.

alfonsogarciacaro:
Awesome, thanks a lot for the quick fix! And thanks for adding the project to the solution, I just created it by copy and pasting files and forgot to edit the solution 😅

PR #2331, opened 2020-12-22
Fixed compiler option parsing

ncave:
- Minor fix for compiler option parsing
- Updated npm dependencies

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -37,8 +37,10 @@ module Util =
             m.Groups.[1].Value + version + m.Groups.[3].Value)
         |> writeFile projFile
 
-    let runTypescript projectDir =
-        // run ("npx tslint --project " + projectDir)
+    let runTSLint projectDir =
+        run ("npm run tslint -- --project " + projectDir)
+
+    let runTypeScript projectDir =
         run ("npm run tsc -- --project " + projectDir)
 
     let runFableWithArgs projectDir args =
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
Thanks!

ncave:
@alfonsogarciacaro Thanks for merging!

P.S. Dear Santa, all I want for Christmas is a release of a stable version of `fable-compiler-js` 3.0.x.
Doesn't actually have to be this year, whenever you get to it. Thanks in advance!

alfonsogarciacaro:
Sorry, I always forget to release `fable-compiler-js`, I will do it 😅 Maybe we could set it up so new releases of dotnet fable trigger new releases of `fable-compiler-js` too, but then we may have versioning issues if there are changes only in `fable-compiler-js`.

IIRC you're owner of `fable-compiler-js` too, so you can also trigger new releases by editing `fable-compiler-js/RELEASE_NOTES` and then running `dotnet fsi build.fsx publish` (you need the `NPM_TOKEN` env var).

alfonsogarciacaro:
I've pushed a new release, I almost made a mistake because I forgot to push `fable-standalone` first. I hope everything is ok now. And I wonder if we should get rid of `fable-standalone`, the main reasons for this package were:

1. The hope that it could become a component use by others to embed the F# in their websites
2. FCS took a long time to compile (and sometimes crashed my machine) so I was downloading the artifact from Appveyor instead of building locally

It seems that 1 is not happening and 2 is not a problem anymore. For the latest updates of the REPL I just pushed a locally built fable-standalone (instead of downloading it from npm. We could make the same with fable-compiler-js, I guess the main issue is we'll have to push a whole new package even if only the fable-compiler-js part changes, but I think this happens less often.

ncave:
@alfonsogarciacaro Thank you, this is great! About the standalone component separation, I think it's fine the way it is, especially now that it doesn't carry Babel with it.

Issue #2332, opened 2020-12-22
fable watch broken for certain directory structures


jwosty:
### Description

For certain project name choices, ``dotnet fable watch`` will watch the wrong directory. This seems to happen when ProjectB contains the name of ProjectA.

In my case, I have the following directory structure:

* src
  * MyProject.Core
    * MyProject.Core.fsproj
  * MyProject.Core.UI
    * MyProject.Core.UI.fsproj

### Repro code

Unzip fable-bad.zip and run ``dotnet fable watch src\MyProject.Core.UI``. 

### Expected and actual results

Fable should watch for and build changes in both projects.

Notice how it builds both projects but only watches the dependency:

	fable-test> dotnet fable watch .\src\MyProject.Core.UI\
	Fable: F# to JS compiler 3.0.4
	Thanks to the contributor! @cmeeren
	src\MyProject.Core.UI> cmd /C dotnet restore MyProject.Core.UI.fsproj
	  Determining projects to restore...
	  Restored C:\Users\jwost\source\fable-test\src\MyProject.Core.UI\MyProject.Core.UI.fsproj (in 91 ms).
	  Restored C:\Users\jwost\source\fable-test\src\MyProject.Core\MyProject.Core.fsproj (in 91 ms).
	Parsing src\MyProject.Core.UI\MyProject.Core.UI.fsproj...
	Initializing F# compiler...
	Compiling src\MyProject.Core.UI\MyProject.Core.UI.fsproj...
	F# compilation finished in 1312ms
	Compiled src\MyProject.Core.UI\Program.fs
	Compiled src\MyProject.Core\Library.fs
	Fable compilation finished in 161ms
	Watching src\MyProject.Core

Everything works if you rename "MyProject.Core.UI" to something else like "MyProject.Core". See fable-good.zip for an example of this workaround. I don't know whether or not this ever worked in Fable 2.

[fable-bad.zip](https://github.com/fable-compiler/Fable/files/5728093/fable-bad.zip)
[fable-good.zip](https://github.com/fable-compiler/Fable/files/5728094/fable-good.zip)

### Related information

* Fable version: 3.0.4
* Operating system: Windows 10

PR #2333, opened 2020-12-22
Fix fable watcher for certain directory structures

jwosty:
Fixes [#2332](https://github.com/fable-compiler/Fable/issues/2332) by appending a path separator when performing substring checks, and then removing it after the fact

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -186,8 +186,11 @@ module private Util =
     }
 
     let getCommonBaseDir (files: string list) =
+        let dirNameWithTrailingSep (file: string) = sprintf "%s%c" (IO.Path.GetDirectoryName file) IO.Path.DirectorySeparatorChar
+        // it's important to include a trailing separator, otherwise things like ["a/b"; "a/b.c"] won't get handled right
+        // https://github.com/fable-compiler/Fable/issues/2332
         files
-        |> List.map (IO.Path.GetDirectoryName)
+        |> List.map dirNameWithTrailingSep
         |> List.distinct
         |> List.sortBy (fun f -> f.Length)
         |> function
@@ -197,7 +200,7 @@ module private Util =
                 let rec getCommonDir (dir: string) =
                     if restDirs |> List.forall (fun d -> d.StartsWith(dir)) then dir
                     else
-                        match IO.Path.GetDirectoryName(dir) with
+                        match dirNameWithTrailingSep (IO.Path.TrimEndingDirectorySeparator dir) with
                         | null -> failwith "No common base dir"
                         | dir -> getCommonDir dir
                 getCommonDir dir
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -185,12 +185,11 @@ module private Util =
                             Exception = e |}
     }
 
+module FileWatcherUtil =
     let getCommonBaseDir (files: string list) =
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
This is great @jwosty, thanks a lot! Do you feel brave enough to add an [integration test](https://github.com/fable-compiler/Fable/blob/nagareyama/tests/Integration/FileWatcherTests.fs)? :wink: We probably don't need to spawn a watcher here, just pass a list of problematic paths to `getCommonBaseDir` and make sure we're getting the expected common path.

jwosty:
For sure, now that I see where they are! I'm going to bed for the night now ;) but I will be happy to do so tomorrow

jwosty:
Working on tests right now. I noticed [Prelude.fs](https://github.com/fable-compiler/Fable/blob/ba509a94a50522794d3e60f27dd826bb5602eca1/src/Fable.Transforms/Global/Prelude.fs#L576) has another definition of this function; should that one be updated too, or is it unrelated?

alfonsogarciacaro:
Oh, my. Thanks for pointing it out! Now I don't remember if I just missed that version or there was a particular reason I wrote the other one in `Fable.Cli/Main.fs` 🤔 We could try to update it too but helpers in Prelude are tricky because they must be compilable to JS, so for example we cannot invoke `System.IO` (there're custom versions of some functions like `GetDirectoryName`). Or maybe for now we can just leave a comment saying there's an alternative version in `Fable.Cli/Main.fs` to make sure we don't break anything 😅

jwosty:
Ah, I see.

This may fit be off topic for this issue (probably deserves a new one), but I've actually found myself defining my own path operators in my Fable projects that behave pretty much the same as the .Net ones. Perhaps it would be useful to proxy these in Fable itself? I'd be happy to provide a PR if you could direct me to the right area of code to put it in. I already have some implementations with lots of test cases (would definitely need some adaptations, though, since they're not perfect proxies).

jwosty:
It would probably be good to do that in a separate PR to keep this one small and to-the-point; #2332 is blocking something of mine right now :)

alfonsogarciacaro:
That'd be great! I tried [something similar](https://github.com/fable-compiler/Fable/blob/nagareyama/src/fable-publish-utils/node.js) for the build script. Although personally I would prefer if this becomes an independent library: for example, you open a namespace that shadows the `System.IO` classes.

Having the replacements built-in in Fable is tricky .NET BCL APIs have lots of overloads and semantic nuances which are not trivial to replicate. It's also difficult for users to know exactly what is supported and what not in a namespace when only part of it is implemented. Some browsers may expect this to work on environments other than node (like the browser or React native), ect.

alfonsogarciacaro (approved):

jwosty:
Thanks so much for merging and thank you for you work on Fable! Great stuff has been going on here. I'll be filing that issue to talk about Path stuff in more detail.

jwosty:
OK, interesting thought. I guess I was thinking of stuff more as Websharper-like proxies, but I don't see why that would work here too. I'll be hacking something together then (mostly extracting stuff into a library that I've already written)

alfonsogarciacaro:
In Fable 1 we had replacement plugins that worked similar to Websharper proxies (I think), we could bring them back, but I think they would have the same problems mentioned above. Shadowing a namespace may give users more immediate feedback about what's supported and what's not. Although tbh I haven't really tried so not sure if there are some quirks.

PR #2340, opened 2021-01-05
Remove glob dependency

ncave:
- Removed `glob` dependency to get to *zero [third-party] package dependency* for `fable-compiler-js`.
- Replaced `npx` with `npm run` in `build.fsx`.


alfonsogarciacaro:
The problem with `npm run` (unless it has changed) is you need to add the command to the npm scripts which is a bit tedious, but we don't have so many and it's probably a good move anyways to avoid calling a tool that we haven't installed as dev dependency 👍

PR #2348, opened 2021-01-17
Using F# List and Seq types

ncave:
- Replaces JS List type with F# List type.
- Replaces JS Seq type with F# Seq type.
- Performance is on par (or slightly faster) than the main branch.


alfonsogarciacaro:
Hmm, if we rule out the overhead of inheriting from Record, the only thing I can think of now is the transformation from IEnumerator to JS Iterator (as we've seen several times small changes in iteration make big performance difference). Right now the `Seq` module works with JS iterators, so for classes implementing `IEnumerable` a method is automatically attached like this:

```js
    [Symbol.iterator]() {
        return toIterator(this.GetEnumerator());
    }
```

This calls the following code: https://github.com/fable-compiler/Fable/blob/10cb7d4eba3c8c8b22ac317d23ebdfb28a1a0ec5/src/fable-library/Seq.ts#L57-L66

A way to test if this, could be to compiler fable-library with your changes first, then manually replace `List[Symbol.iterator]` with:

```js
    [Symbol.iterator]() {
        let cur = this;
        return {
            next: () => {
                const value = cur === null || cur === void 0 ? void 0 : cur.head;
                const done = (cur === null || cur === void 0 ? void 0 : cur.tail) == null;
                cur = cur === null || cur === void 0 ? void 0 : cur.tail;
                return { done, value };
            },
        };
    }
```

And after that, running the benchmark to see if there's any difference. If there is, it would make sense to transform Seq.ts into F# and use `IEnumerable` as Fable code most of the time will iterate over F# structures (besides arrays). In the case of JS iterables we could make the inverse conversion from JS iterator to `IEnumerator`.

ncave:
@alfonsogarciacaro Iteration doesn't seem to be a problem in micro-benchmarks, but I'll try that.

ncave:
@alfonsogarciacaro I stand corrected, replacing the iterator as you suggested above does improve performance by 20%.

ncave:
@alfonsogarciacaro
> it would make sense to transform Seq.ts into F#

Done :)

ncave:
@alfonsogarciacaro 

Performance of generated code is now at the same level as the main branch (or slightly better), so this is ready for merging.

alfonsogarciacaro:
Awesome @ncave! Fantastic work as always 👏 👏 👏 Looking at the code, it seems classes implementing `IEnumerable` will still get JS `Symbol.iterator` and getting the enumerator is still compatible with JS iterables so there shouldn't be compatibility issues :+1: Hopefully this also helps @dbrattli by making fable-library less JS-dependent.

Issue #1820, opened 2019-04-24
Regex Dotnet vs JS : a path to Regex Type Provider


evilz:
### Description

Attempt to make better Regex implementation. 
In Fable 2 two version of Regex was available Dotnet one and Js one.

Since Fable Core 3 only dotnet Regex is available. 

This can be a problem since but don't actually work exactly the same.
- JS : 
   * no group name
   * no backtraking
  ...

By the way some of those features have been implemented in V8 or Babel.
=> https://github.com/DmitrySoshnikov/babel-plugin-transform-modern-regexp#named-capturing-groups

Can/should we use these plugin to make a better implementation of Regex in `fable-library` ?

At the end we can use group name in same way as dotnet and create the Regexp Type provider \o/

MangelMaxime:
For reference, the discussion started on Gitter.

And my point of view, is we could avoid making `babel-plugin-transform-modern-regexp` a dependency of `fable-compiler` but mark it as a requirement for using the `Regexp Type provider library`.

The same way, it's required to install `@babel/plugin-proposal-class-properties` in order to use [Fable.ReactGoogleMaps](https://github.com/fable-compiler/Fable.ReactGoogleMaps).

Because the more we include dependencies the harder it get to keep tracks of everything.

evilz:
I understand your point. But in this case things are differents : it's core and in BCL. 
Then I use a regex with group name or kind of syntax that run well on dotnet I also hope this will run on Fable.

evilz:
[...]

PR #2364, opened 2021-01-25
Add support for named capture groups in Regexes

### LINKED ISSUE #2359: Regex.Replace with function and regex with named group constructs incorrect Match
### Description

Regex.Replace with function and regex with named group constructs incorrect Match

### Repro code

```fsharp
let regexIndex = Regex("0(\\d*)")
let regexNamed = Regex("0(?<number>\\d*)")
let text = "number 012345!"

let replace (m: Match) =
    console.log m
    string m.Index

let resIndex = regexIndex.Replace(text, replace)
printfn "regexIndex=%s" resIndex
let resNamed = regexNamed.Replace(text, replace)
printfn "regexNamed=%s" resNamed
```
([repl](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYDoAqYAeT0BKYA5gK4A2AhgE4CiOI1YCCAlsBAgLABQokUAGKUARuTDoAwsCboAUjF69xSKE2K4AkhAAmuKAF4oRDTgAUAIgAMZgDq2dAKgCUF58rCr1uAHKVUYDqGxiS4ljYA-AA8EKSoImDUAHz2Tq7uPCpQKHjBFrHxiVBWAIwATADMACwArACEFkqZnmpgIFQAxmBQZqgAXFAAspRIHQAWzoa8UDNQHRwIwOLo5MDEUKjTs8jUrBDrGNp6OE1ZTAhH+kbeOJc4hG2dYGY5SAA0re2UXRmMe0gAM2gFhudwMAFIEBZWhddLgPF5mH4AkFrqEcMjAg8vl0Xrh3p8nr9dhBAcCbpidBCoTDKUA&html=Q&css=Q))

Output:
> regexIndex=number 7!
> 
> regexNamed=number number 012345!!

Both should be `number 7` (with `7` being the index of the match in the input), but regexNamed instead returns the original input instead of the index.  
Named groups aren't supported in Fable, but it shouldn't introduce some incorrectness. (someone might access via indices, but copy a regex with named groups from somewhere)

### Why is this happening?

Match objects (via `console.log` in replace function):

regexIndex:
```
0: "012345"
1: "12345"
index: 7
input: "number 012345!"
length: 2
```
regexNamed:
```
0: "012345"
1: "12345"
2: 7
index: "number 012345!"
input:
  number: "12345"
length: 3
```
-> input is set to named groups, index to original input, and index is instead added as additional index (`m.[2]`).

The .Net `Match` lookalike-object is created in https://github.com/fable-compiler/Fable/blob/a7685d442c30b95db9f4fc7e9e3b2c49afe43a8f/src/fable-library/RegExp.ts#L76-L83 from the arguments passed into `replacer` function.  
These [arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_a_parameter) are:
* `match`: matched substring
* `p1`, `p2`, ...: nth capture group string
* `offset`: offset of matched substring
* `string`: whole string examined
* `groups`: named capturing groups

-> variable number of arguments (based on groups in regex)  
& `groups` is only passed if regex has any named groups

BUT: Fable always takes the last argument as original input `string`.  
-> if last argument is `groups` it takes that as `string` and the other arguments get shifted accordingly.

Fix: check if last element in arguments is of type string. Then it's original input. Otherwise it's groups.

### Related information

* Fable version: `3.1.1` (REPL version: `3.1.1`)

Note: I'm currently working on support for named capture groups and the fix is basically part of it (and how I found this bug in the first place...).  
-> Fix is in the works and should come in the next days.

Booksbaum:
implements part of #1820; fixes #2359

Adds support for named capture groups in regexes, like in:
```fsharp
let regex = Regex "(?<number>\\d+)"
let m = regex.Match "alpha 1234 beta"
let v = m.Groups.["number"].Value
assert(v = "1234")
```

Support for named groups was added some time ago in JavaScript. But named groups don't work in too old browser versions. See [Browser Compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#browser_compatibility) (search for "Named capture groups"):  
firefox >= 78; Chrome >= 64; Edge >= 79; Safari >= 11.1; Node.js >= 10; IE has no support

### `m.Groups.["name"]`
In JS: if there's a named group, the match contains a field `groups` with named groups. If there's no named group, that field doesn't exist.  
Access in Fable: `myMatch.group?.myGroup` (or actual implementation `myMatch.groups && myMatch.groups.myGroup`, because optional chaining operator not available yet) -> always one `undefined` check.  
Behaviour for non-existing named group is same as with non-existing indexed group: `undefined` (in JS), `Group` object with `Success = false`

Functions of `Groups` like `Count` or iterating over it: All named matches exist as indexed matches too. In example above: group `number` is index `1` -> Named groups are already accounted for.  
Other functions like `Keys` aren't implemented (see [example (REPL)](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYDoAqYAeT0BKYA5gK4A2AhgE4CiOI1YCCAlsBAgLABQokUAGKUARuTDoAwsCboAUjF69xSKNSgBeKEWK4oAIgAMACgD8AHgilUIsNQB8AHUcATANQBKfcrCrUmtXQAWUokAGMACwMAOWtbdUMARgAmAGYAFgBWKFYEKBBgNiRWADcwbx5eMI4EYHF0cmBiKFReRlYIJAAzaH1pUk6NAFIAQX0W9ABxamBSEAQpWc6oFYB6VahktuoO7t6ASRRqYbGoYwxp2fmoAB97WDAAR3QkYAAZXKQPKHWoSnJyC1QpFmGcRHAchAXLgPAAaKAQSioMAuKDEGZzKAAAysNjsWJyeQ60JwUESvF+7U6PQMADV-qRmCdxucphj5uh6eRGXk7g9nq8Pshvr8IMBVAg5gVqCgXBSNlS9gYANJgRDMs4XdkLVWIW73GBPF7vT4ijZiiVSmSyoA&html=Q&css=Q))

### In Replace with string (`regex.Replace(input, replacement: string`)

There's a difference in using named capture groups between JS and .NET. In `replacement` in .NET a group is accessed via `${name}`, but in JS it's `$<name>`.  
-> When calling in Fable, `${name}` is replaced with `$<name>` in `RegExp.ts::Replace` (like it's already the case with `$0` -> `$&`).

`$<name>` doesn't get escaped, but instead is kept as is. That means `${name}` and `$<name>` are same and both access the group. (It's just easier to implement (or not implement...)...)

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -107,7 +107,7 @@ module Helpers =
             Any)
 
 open Helpers
-
+open Fable.Transforms
 type BuiltinType =
     | BclGuid
     | BclTimeSpan
@@ -2678,10 +2678,7 @@ let regex com (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Exp
     let isGroup =
         match thisArg with
         | Some (ExprType (EntFullName "System.Text.RegularExpressions.Group")) -> true
-          // access to named `groups` contains a null check
-        | Some (IfThenElse (_, _, ExprType (EntFullName "System.Text.RegularExpressions.Group"), _)) -> true
         | _ -> false
-    let isGroupCollection = i.DeclaringEntityFullName = "System.Text.RegularExpressions.GroupCollection"
 
     match i.CompiledName with
     // TODO: Use RegexConst if no options have been passed?
@@ -2707,30 +2704,26 @@ let regex com (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Exp
     // Match
     | "get_Groups" -> thisArg.Value |> Some
     // MatchCollection & GroupCollection
-    | "get_Item" when isGroupCollection ->
+    | "get_Item" when i.DeclaringEntityFullName = "System.Text.RegularExpressions.GroupCollection" ->
         // can be group index or group name
         //        `m.Groups.[0]` `m.Groups.["name"]`
         match args |> List.head with
         | Value (StringConstant name, _) ->
             // name
-            (*
-              `groups` might not exist -> check first:
-                `thisArg.Value`.groups?.[`args.Head`]
-              or
-                if `thisArg.Value`.groups === undefined then
-                    undefined
-                else
-                    `thisArg.Value`.groups[`args.Head`]
+            (* `groups` might not exist -> check first:
+                (`m`: `thisArg.Value`; `name`: `args.Head`)
+                  ```ts
+                  m.groups?.[name]
+                  ```
+                or here
+                  ```ts
+                  m.groups && m.groups[name]
+                  ```
             *)
             let groups = propStr "groups" thisArg.Value
             let getItem = getExpr r t groups args.Head
 
-            IfThenElse (
-                isNull groups,
-                Value (Null Any, None),
-                getItem,
-                None
-            )
+            Operation(Logical(LogicalAnd, groups, getItem), t, None)
             |> Some
         | _ ->
             // index
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -2678,7 +2678,11 @@ let regex com (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Exp
     let isGroup =
         match thisArg with
         | Some (ExprType (EntFullName "System.Text.RegularExpressions.Group")) -> true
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
This is great @Booksbaum, thanks a lot!

PR #2368, opened 2021-01-29
Refactor Babel AST to union/records

### LINKED ISSUE #2158: Model AST.Babel with unions
When I wrote the Babel AST implementation I chose classes over unions because the structure was not purely hierarchical as some nodes could appear in different places. But after the AST simplification I think it should be possible to turn the classes into unions. Is it worth it? Would it improve the code? What do you think @ncave?

dbrattli:
Refactor Babel AST into records and unions to simplify pattern matching for Babel transformation (e.g Python).  I use records consistently (instead of tuples) since they enable keyed lookup and it's easier to add eg source location etc. The PR tries to keep the existing code style of Fable2Babel.fs and uses `.Create()` methods to support optional parameters. E.g previous code such as:

```fs
CallExpression(com.TransformImport(ctx, memberName, getLibPath com moduleName), args, ?loc=r) :> Expression
```

Now becomes:
```fs
Expression.callExpression(com.TransformImport(ctx, memberName, getLibPath com moduleName), args, ?loc=r)
```

This PR Fixes #2158

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -343,7 +343,7 @@ module Annotation =
         else
             genParams
             |> Set.toArray
-            |> Array.map (Identifier.Create >> GenericTypeAnnotation.AsTypeAnnotationInfo)
+            |> Array.map (Identifier.Create >> TypeAnnotationInfo.genericTypeAnnotation)
             |> TypeParameterInstantiation.Create |> Some
 
     let mergeTypeParamDecls (decl1: TypeParameterDeclaration option) (decl2: TypeParameterDeclaration option) =
@@ -361,7 +361,7 @@ module Annotation =
 
     let getGenericTypeAnnotation _com _ctx name genParams =
         let typeParamInst = makeTypeParamInst genParams
-        GenericTypeAnnotation.AsTypeAnnotationInfo(Identifier.Create(name), ?typeParameters=typeParamInst)
+        GenericTypeAnnotation(Identifier.Create(name), ?typeParameters=typeParamInst)
         |> TypeAnnotation.Create |> Some
 
     let typeAnnotation com ctx typ: TypeAnnotationInfo =
@@ -389,7 +389,7 @@ module Annotation =
             makeAnonymousRecordTypeAnnotation com ctx fieldNames genArgs
 
     let makeSimpleTypeAnnotation _com _ctx name =
-        GenericTypeAnnotation.AsTypeAnnotationInfo(Identifier.Create(name))
+        TypeAnnotationInfo.genericTypeAnnotation(Identifier.Create(name))
 
     let makeGenTypeParamInst com ctx genArgs =
         match genArgs with
@@ -399,7 +399,7 @@ module Annotation =
 
     let makeGenericTypeAnnotation com ctx genArgs id =
         let typeParamInst = makeGenTypeParamInst com ctx genArgs
-        GenericTypeAnnotation.AsTypeAnnotationInfo(id, ?typeParameters=typeParamInst)
+        GenericTypeAnnotation(id, ?typeParameters=typeParamInst)
 
     let makeNativeTypeAnnotation com ctx genArgs typeName =
         Identifier.Create(typeName)
@@ -439,7 +439,7 @@ module Annotation =
 
     let makeUnionTypeAnnotation com ctx genArgs =
         List.map (typeAnnotation com ctx) genArgs
-        |> List.toArray |> UnionTypeAnnotation.AsTypeAnnotationInfo
+        |> List.toArray |> UnionTypeAnnotation
 
     let makeBuiltinTypeAnnotation com ctx kind =
         match kind with
@@ -1769,7 +1769,7 @@ module Util =
                     ?superTypeParameters = e.SuperTypeParameters,
                     ?typeParameters = e.TypeParameters)
             | FunctionExpression(e) ->
-                FunctionDeclaration.AsDeclaration(
+                Declaration.functionDeclaration(
                     e.Params, e.Body, membName,
                     ?returnType = e.ReturnType,
                     ?typeParameters = e.TypeParameters)
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -529,7 +529,6 @@ module Annotation =
             let args', body' = com.TransformFunction(ctx, name, args, body)
             args', body', None, None
 
-
 module Util =
     open Lib
     open Reflection
@@ -1702,7 +1701,7 @@ module Util =
                 transformBlock com ctx None body,
                 start |> varDeclaration (typedIdent com ctx var |> IdentifierPattern) true,
                 Expression.binaryExpression(op1, identAsExpr var, limit),
-                UpdateExpression.AsExpr(op2, false, identAsExpr var), ?loc=range)|]
+                Expression.updateExpression(op2, false, identAsExpr var), ?loc=range)|]
 
     let transformFunction com ctx name (args: Fable.Ident list) (body: Fable.Expr): Pattern array * BlockStatement =
         let tailcallChance =
@@ -1769,11 +1768,11 @@ module Util =
                     ?implements = e.Implements,
                     ?superTypeParameters = e.SuperTypeParameters,
                     ?typeParameters = e.TypeParameters)
-            | FunctionExpression(e) ->
+            | FunctionExpression(_, ``params``, body, returnType, typeParameters, _) ->
                 Declaration.functionDeclaration(
-                    e.Params, e.Body, membName,
-                    ?returnType = e.ReturnType,
-                    ?typeParameters = e.TypeParameters)
+                    ``params``, body, membName,
+                    ?returnType = returnType,
+                    ?typeParameters = typeParameters)
             | _ -> varDeclaration membName' isMutable expr |> Declaration.VariableDeclaration
         if not isPublic then PrivateModuleDeclaration(decl |> Declaration)
         else ExportNamedDeclaration(decl)
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -77,36 +77,36 @@ module Reflection =
     let private transformRecordReflectionInfo com ctx r (ent: Fable.Entity) generics =
         // TODO: Refactor these three bindings to reuse in transformUnionReflectionInfo
         let fullname = ent.FullName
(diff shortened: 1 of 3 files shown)


src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -136,7 +136,7 @@ module Reflection =
             | Some t -> t
             | None ->
                 Replacements.genericTypeInfoError name |> addError com [] r
-                NullLiteral () :> Expression
+                NullLiteral.AsLiteral() |> Literal
```
dbrattli, line 139: Should use .AsExpr()

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -256,29 +256,30 @@ module Reflection =
...
+            BinaryExpression.AsExpr(BinaryEqualStrict, typeof, StringLiteral.AsExpr(primitiveType), ?loc=range)
 
         let jsInstanceof consExpr (Util.TransformExpr com ctx expr): Expression =
-            upcast BinaryExpression(BinaryInstanceOf, expr, consExpr, ?loc=range)
+            BinaryExpression.AsExpr(BinaryInstanceOf, expr, consExpr, ?loc=range)
 
         match typ with
-        | Fable.Any -> upcast BooleanLiteral true
-        | Fable.Unit -> upcast BinaryExpression(BinaryEqual, com.TransformAsExpr(ctx, expr), Util.undefined None, ?loc=range)
+        | Fable.Any -> BooleanLiteral.Create(true) |> Literal
```
dbrattli, line 277: Use .AsExpr()

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -296,7 +297,7 @@ module Reflection =
                 match expr with
                 | MaybeCasted(ExprType(Fable.DeclaredType (ent2, _)))
                         when com.GetEntity(ent2) |> FSharp2Fable.Util.hasInterface Types.idisposable ->
-                    upcast BooleanLiteral true
+                    BooleanLiteral.Create(true) |> Literal
```
dbrattli, line 300: Use .AsExpr()

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -618,29 +614,33 @@ module Util =
...
+    let identAsPattern (id: Fable.Ident) =
+        Identifier.AsPattern(id.Name, ?loc=id.Range)
 
     let thisExpr =
-        ThisExpression() :> Expression
+        ThisExpression.AsExpr()
 
     let ofInt i =
-        NumericLiteral(float i) :> Expression
+        NumericLiteral.AsLiteral(float i) |> Literal
```
dbrattli, line 632: Should use .AsExpr()

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -618,29 +614,33 @@ module Util =
 
     let addErrorAndReturnNull (com: Compiler) (range: SourceLocation option) (error: string) =
         addError com [] range error
-        NullLiteral () :> Expression
+        NullLiteral.AsLiteral()
```
dbrattli, line 617: Use .AsExpr()

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -671,95 +671,94 @@ module Util =
...
+        CallExpression.AsExpr(get None funcExpr "call", args, ?loc=r)
 
     let emitExpression range (txt: string) args =
-        EmitExpression(txt, List.toArray args, ?loc=range) :> Expression
+        EmitExpression.AsExpr(txt, List.toArray args, ?loc=range)
 
     let undefined range =
 //        Undefined(?loc=range) :> Expression
-        UnaryExpression(UnaryVoid, NumericLiteral(0.), ?loc=range) :> Expression
+        UnaryExpression.AsExpr(UnaryVoid, NumericLiteral.AsLiteral(0.) |> Literal, ?loc=range)
```
dbrattli, line 761: Use .AsExpr

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -821,28 +820,28 @@ module Util =
...
     /// Wrap int expressions with `| 0` to help optimization of JS VMs
     let wrapIntExpression typ (e: Expression) =
         match e, typ with
-        | :? NumericLiteral, _ -> e
+        | Literal(NumericLiteral(_)), _ -> e
         // TODO: Unsigned ints seem to cause problems, should we check only Int32 here?
         | _, Fable.Number(Int8 | Int16 | Int32)
         | _, Fable.Enum _ ->
-            BinaryExpression(BinaryOrBitwise, e, NumericLiteral(0.)) :> Expression
+            BinaryExpression.AsExpr(BinaryOrBitwise, e, NumericLiteral.AsLiteral(0.) |> Literal)
```
dbrattli, line 832: Use .AsExpr()

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -916,22 +915,22 @@ module Util =
...
-        | Fable.BoolConstant x -> upcast BooleanLiteral(x, ?loc=r)
-        | Fable.CharConstant x -> upcast StringLiteral(string x, ?loc=r)
-        | Fable.StringConstant x -> upcast StringLiteral(x, ?loc=r)
-        | Fable.NumberConstant (x,_) -> upcast NumericLiteral(x, ?loc=r)
-        | Fable.RegexConstant (source, flags) -> upcast RegExpLiteral(source, flags, ?loc=r)
+        | Fable.BoolConstant x -> BooleanLiteral.Create(x, ?loc=r)  |> Literal
+        | Fable.CharConstant x -> StringLiteral.AsExpr(string x, ?loc=r)
+        | Fable.StringConstant x -> StringLiteral.AsExpr(x, ?loc=r)
+        | Fable.NumberConstant (x,_) -> NumericLiteral.AsLiteral(x, ?loc=r) |> Literal
+        | Fable.RegexConstant (source, flags) -> RegExpLiteral.AsLiteral(source, flags, ?loc=r) |> Literal
```
dbrattli, line 933: Use .AsExpr()

src/Fable.Transforms/Global/Babel.fs
```diff
@@ -219,32 +219,236 @@ module PrinterExtensions =
...
+    | ObjectTypeProperty of ObjectTypeProperty
+    | ObjectTypeCallProperty of ObjectTypeCallProperty
+    | ObjectTypeInternalSlot of ObjectTypeInternalSlot
+    | TypeParameterDeclaration of TypeParameterDeclaration
+    | TypeParameterInstantiation of TypeParameterInstantiation
+
+    interface IPrintable with
+        member this.Print(printer) =
+            match this with
+            | Pattern(n) -> printer.Print(n)
```
alfonsogarciacaro, line 254: Now that we are doing this, maybe it's an opportunity to remove the `IPrintable` interface and just have `Print` methods in all the nodes, e.g.:

```fsharp
member this.Print(printer) =
    match this with
    | Pattern n -> n.Print(printer)
    ...
```

Maybe this also makes it easier to move the printer to another file to make the AST declarations cleaner.

src/Fable.Transforms/Global/Babel.fs
```diff
@@ -219,32 +219,236 @@ module PrinterExtensions =
...
+type EmitExpression =
+    { Value: string
+      Args: Expression array
+      Loc: SourceLocation option }
+
+    static member AsExpr(value, args, ?loc): Expression =
+        { Value = value
+          Args = args
+          Loc = loc }
+        |> EmitExpression
```
alfonsogarciacaro, line 447: Sorry @dbrattli, I think you mentioned this somewhere but, what was the purpose of using records instead of just union case fields? Would it be simpler to do something like:

```fsharp
type Expression =
    | EmitExpression of value: string * args: Expression array * loc: SourceLocation option
    ...
    static member NewEmitExpression(value, args, ?loc) =
        EmitExpression(value, args, loc)
```

Would it make the code simpler to do this in general and use only standalone types when they're needed?

alfonsogarciacaro (commented):
Thanks a lot for this @dbrattli, this is fantastic work! Sorry, for the late review. I left a couple of commits, could you please check?

src/Fable.Transforms/Global/Babel.fs
dbrattli: Yes at least most of them can be removed. There are still some problems related to e.g `PrintOptional`, `PrintCommaSeparatedArray` but I can see if I can fix it.

src/Fable.Transforms/Global/Babel.fs
dbrattli: This is basically just a matter of style. Records is very similar to named unions fields (except it's easier to extract values from records using dot-notation). The question is really about the extra type-level. This is great of attaching convenience methods such as `Print` and `Create`. But if you like a more compact form then I can try to reduce it. There will however be cases like `Expression` that will need to split the printing into several different functions, so not so compact anymore. But if we can more the print to a separate file (it's just another transformation AST -> string) then that could be good. I'll see what I can do.

src/Fable.Transforms/Global/Babel.fs
dbrattli: I have now removed IPrintable. Next I will try to move the print handling into BabelPrinter.fs

dbrattli: @alfonsogarciacaro But now is probably the right time to move Babel transformations i.e `IsProductiveStatement` back into Fable2Bable.fs? We don't want such transformations in the printer. I btw already did this for Babel2Python.fs

src/Fable.Transforms/Global/Babel.fs
dbrattli: Ok, I've tried to fix this. But the Babel AST is quite self referencing so it's not possible to do this for many of the types and they need to be types, e.g `Identifier`. `EmitExpression` is easy since it's not part of Babel. But I think it looks better so please tell me what you think. I would like to get back to Python sometime soon ;)

dbrattli: Ok, think I'm done now. There's quite a bit of commented code in Babel.fs that is not relevant anymore wrt the new style etc. Should we delete it?

alfonsogarciacaro:
Hi @dbrattli! Sorry for the late review, this is a great improvement to the code, thanks a lot for all this work! ❤️ Hopefully having the printer in its own file, makes it easier to reuse the Babel AST with a Python printer :muscle:

dbrattli:
@alfonsogarciacaro This is great news! Domo arigato gozaimashita!

PR #2370, opened 2021-01-31
Generate ThisExpression instead of "this" identifier.

dbrattli:
Generate Babel `ThisExpression` instead of an identifier called `this`. Makes it easier to translate the Babel AST to e.g Python where `this` needs to be translated to `self`.


alfonsogarciacaro:
Thanks @dbrattli! This makes sense. If I remember correctly, this was because Babel changes the this expression in arrow functions. But fortunately this is not an issue anymore :)

PR #2371, opened 2021-02-01
Add Object.values, Object.entries

kerams:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values


alfonsogarciacaro:
Thanks! I will push a new version :+1:

PR #2374, opened 2021-02-06
Raise an error when the name of an interface property clashes

### LINKED ISSUE #1343: Clashing member names cause runtime exception
### Repro code

In normal F#, the following is allowed - you can have a type with a member name that clashes with a member name of an interface it implements:

```
type I = 
  abstract Trick : string
type A = 
  { Trick : string }
  interface I with
    member x.Trick = x.Trick + "!"
let a = { Trick = "Hi" }    
a.Trick // Should return "Hi"
(a :> I).Trick // Should return "Hi!"
```

You get an error message at runtime:

> Uncaught TypeError: Cannot set property Trick of #<A> which has only a getter

This should either emit a warning (because you probably don't need to do this in any sensible code) or Fable should somehow mangle interface member names to make it work. I'd be quite happy with the former though.

BillHally:
Raise an error when the name of an interface property clashes with an existing member.

Fixes #1343.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -113,8 +113,12 @@ let private resolveImportMemberBinding (ident: Fable.Ident) (info: Fable.ImportI
     else info
 
 let private getAttachedMemberInfo com ctx r nonMangledNameConflicts
-                (declaringEntityName: string option) (sign: FSharpAbstractSignature) attributes =
-    let declaringEntityName = defaultArg declaringEntityName ""
+                (declaringEntity: Fable.Entity option) (sign: FSharpAbstractSignature) attributes =
+    let declaringEntityName, declaringEntityFields =
+        match declaringEntity with
+        | Some x -> x.FullName, (x.FSharpFields |> List.map (fun x -> x.Name))
+        | None   -> "", []
+
     let isGetter = sign.Name.StartsWith("get_")
     let isSetter = not isGetter && sign.Name.StartsWith("set_")
     let indexedProp = (isGetter && countNonCurriedParamsForSignature sign > 0)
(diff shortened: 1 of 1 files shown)


src/Fable.Transforms/FSharp2Fable.fs
```diff
@@ -148,11 +152,17 @@ let private getAttachedMemberInfo com ctx r nonMangledNameConflicts
...
+                        if nonMangledNameConflicts declaringEntityName name then
+                            sprintf "Member %s is duplicated, use Mangle attribute to prevent conflicts with interfaces" name
+                            // TODO: Temporarily emitting a warning, because this errors in old libraries,
+                            // like Fable.React.HookBindings
+                            |> addWarning com ctx.InlinePath r
+
+                        if declaringEntityFields |> List.contains name then
+                            sprintf "Interface member %s.%s conflicts with existing member on type %s, add the Mangle attribute to the interface to prevent conflicts"
+                                sign.DeclaringType.TypeDefinition.FullName name declaringEntityName
+                            |> addError com ctx.InlinePath r
```
alfonsogarciacaro, line 165: As the one above is still a warning, maybe have this as a warning too until we change them both. We can also merge the two branches for simplicity, like:

```fsharp
if nonMangledNameConflicts declaringEntityName name
   || declaringEntityFields |> List.contains name then
        sprintf "Member %s is duplicated, add the Mangle attribute to the interface to prevent conflicts" name
        // TODO: Temporarily emitting a warning, because this errors in old libraries, like Fable.React.HookBindings
        |> addWarning com ctx.InlinePath r
```

alfonsogarciacaro (commented):
Thanks a lot @BillHally and sorry for the late review, I just left a couple of comments, please have a look but the PR looks good to merge anyways 👍

BillHally:
>sorry for the late review

There's no way you should be apologising for reviewing a PR after a few days. I've had PRs open and unreviewed for more than a year in the past, and don't blame the maintainers at all!

I'll push a commit with your suggested changes later today - it'll probably be about 12 hours from now when I get time to do it.

Thanks again for Fable!

alfonsogarciacaro:
Awesome, thanks a lot!

PR #2381, opened 2021-02-15
Use unary expression instead of emit

dbrattli:
Check if (i.e run all tests) to see if we can avoid emit expression (non-babel) and use unary expression together with void. This means that F# code such as:

```fs
f () |> ignore
```
now will produce (new behavior):

```js
void f();
```

instead of (current behavior):

```js
const value = f();
void value;
```


alfonsogarciacaro:
Makes sense, thank you!

PR #2391, opened 2021-02-19
Fixes #2390

### LINKED ISSUE #2390: Array.choose is executing choose function two times
I noticed that when using `Array.choose` my chooser function is called two times.
Is it a design choice to prefer low allocation over execution time ?

https://github.com/fable-compiler/Fable/blob/69d11d14448c7ce1391ac6319a69ffbfdafd7e32/src/fable-library/Array.fs#L568

ncave:
- Fixed #2390
- ~optimized `Array.map` and `Array.mapi` a bit.~
 (this was rolled back as for some reason the micro-benchmark performance of mapping with `Array.from` was much, much worse than the existing implementation of preallocation + copy loop).
- updated npm dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,1534 +1,13 @@
 {
-  "name": "Fable",
-  "lockfileVersion": 2,
   "requires": true,
-  "packages": {
-    "": {
-      "dependencies": {
-        "@types/node": "^14.14.14",
-        "esm": "^3.2.25",
-        "ghreleases": "^3.0.2",
-        "mocha": "^8.2.1",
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Thanks @ncave! About the optimization, I also noticed some time ago our (naïve) map implementation is [actually faster than native Array.prototype.map](https://twitter.com/FableCompiler/status/890488515926122496). It seems the reason is native functions need to take sparse arrays into account to conform to ecmascript specs while in our implementation we can just ignore them: https://github.com/codemix/fast.js/tree/master#how

PR #2393, opened 2021-02-22
Log in same line

alfonsogarciacaro:
See https://github.com/SAFE-Stack/SAFE-template/pull/415#issuecomment-783003956

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -153,7 +153,7 @@ module private Util =
                 else path
             member _.Dispose() = stream.Dispose()
 
-    let compileFile (cliArgs: CliArgs) dedupTargetDir (com: CompilerImpl) = async {
+    let compileFile (cliArgs: CliArgs) dedupTargetDir logger (com: CompilerImpl) = async {
         try
             let babel =
                 FSharp2Fable.Compiler.transformFile com
@@ -193,7 +193,7 @@ module private Util =
             do! BabelPrinter.run writer mapGen babel
             do! mapPrinter outPath
 
-            Log.alwaysInSameLine("Compiled " + File.getRelativePathFromCwd com.CurrentFile)
+            logger <| "Compiled " + File.getRelativePathFromCwd com.CurrentFile
 
             return Ok {| File = com.CurrentFile
                          Logs = com.Logs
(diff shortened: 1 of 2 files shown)


src/Fable.Cli/Main.fs
```diff
@@ -193,7 +193,7 @@ module private Util =
             do! BabelPrinter.run writer mapGen babel
             do! mapPrinter outPath
 
-            Log.always("Compiled " + File.getRelativePathFromCwd com.CurrentFile)
+            logger <| "Compiled " + File.getRelativePathFromCwd com.CurrentFile
```
isaacabraham, line 196: Consider

```fsharp
logger $"Compiled {File.getRelativePathFromCwd com.CurrentFile}"
```

instead?

alfonsogarciacaro: Yes, I have a tendency to just use concatenation in simple cases because formatting/interpolation is more expensive. But shouldn't really matter in this case :)

alfonsogarciacaro:
@ncave If we make the "Compiled Foo.fs" logs less noisy, it would be nice to have also some kind of progress indicator for the F# compilation part. Do you think it would make sense to pass a reference to the one-line logger here?https://github.com/ncave/fsharp/blob/43b7223423cf2b93af3bf60aaf1c36789b27361a/src/fsharp/FSharp.Compiler.Service/service_slim.fs#L230

Logging after each `ParseFile` should be easy (now that I think of it, can this be parallelized?) but if most of the work is spent in `TypeCheckClosedInputSet` l guess we'll still have to wait a few seconds without a progress indicator.

ncave:
@alfonsogarciacaro I don't see why not, you can hook up a notification of which file is [about to be typechecked](https://github.com/ncave/fsharp/blob/43b7223423cf2b93af3bf60aaf1c36789b27361a/src/fsharp/FSharp.Compiler.Service/service_slim.fs#L201), before or after it's done. I haven't seen other compilers do progress indicators, but it sounds good. If file-level notifications are not fine-grained enough, you can pass in a custom `TcResultsSink` (in the same line) to see what declarations are being processed etc.

alfonsogarciacaro:
Ah, ok, didn't check that function, thanks! I will try to send a PR then :+1: A message per file parsed and/or checked should work. It's true this is not common for CLI compilers, I'm thinking mainly in the psychological effect: when you stare at the screen without any output it feels as it time goes more slowly 😅 For example, Bucklescript (and other JS tools like npm/yarn) has this kind of progress indicator. Everybody keeps saying that it's super fast though I found it to be on par with Fable 3 for a big project, so I believe showing fast messages on the console help leave this impression on the user.

isaacabraham:
A percentage / progress indicator would be great - you could just start with something simple e.g. "4/22 files" or something?

Issue #2394, opened 2021-02-22
interpolate doesn't unescape %% correctly


thinkbeforecoding:
### Description

When using sprintf or interpolated strings, % has to be escaped as %%. It should be returned as %, but remains as %%.

### Repro code

try in the repl:
printfn $"{100}%%"

https://fable.io/repl/#?code=A4JwlgdgLgZhAEASARAbwIwAZMF8CkeyQA&html=DwCwLgtgNgfAsAKAAQqaApgQwCb2ag4CdMTJcMABwFp0BHAVwEsA3AXgCIBhAewDsw6AdQAqAT0roOSAMb9BAzoIAeYAPThoAbhkhMAJwDOJNgzAAzagA4OeQhqy5EhAEY9sYu6mBq3HvD6asEA&css=Q

### Expected and actual results

Expected: 100%
Actual: 100%%

### Related information

Fable version: 3.1.5
Also on the Repl

KaBaJIeP:
doesn't work

```
let getUrl url = $"/{url}"
```

```
F# <TargetFramework>net5.0</TargetFramework>
```
```
(27,28) error FSHARP: This token is reserved for future use (code 1160)
```
```
fable-compiler 2.13.0
```

works 
```
let getUrl url = sprintf "/%s" url
```

PR #2395, opened 2021-02-22
Fix: #2394 interpolate doesn't unescape %% correctly

thinkbeforecoding:
This is a fix for #2394


alfonsogarciacaro:
Awesome, thanks a lot!

PR #2397, opened 2021-02-24
Fix: Accessing named capture group in Regex only works with string constant

Booksbaum:
Support for named capture groups in Regexes was introduced in #2364.

But there's an issue: Accessing a group by name can only be done via string constant, not via string variable (or any other string expression):
```fsharp
let regex = Regex "(?<num>\d+)"
let text = "Text 123 Text"
let m = regex.Match text

let gc = m.Groups.["num"].Value
printfn "gc=%s" gc  // gc=123

let g = "num"
let gv = m.Groups.[g].Value
printfn "gc=%s" gv  // gv=

let ga = m.Groups.["nu" + "m"].Value
printfn "ga=%s" ga  // ga=
```
([repl](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYDoAqYAeT0BKYA5gK4A2AhgE4CiOI1YCCAlsBAgLABQv5YJFCbFcUALxQionFABEACgD8AHgilUAPgA6AEwDUASjn9BUFHgnzslgIwAmAMxQbSEzwFDUVkbnQBZSiQAYwALc1wkXlMhYmCrDABxamBSEAR0AG05dVQ5AF10ADVKclIwXkZWCCQAM2g5OPEAUgQ5KDioKAB6bqgAd2BqAGtuPg8zYiscjXdPDoA3BPRk1PSs4kKSsoqeKpr6+WIFlrbFrt6oXWBmCAByIUGR6InYymXVtIzs9Xb9eTyW1K5Uq1GqdQaxEop3aUIufWutweAyGw14QA&html=Q&css=Q))

-> I only added handling of `StringConstant` (https://github.com/fable-compiler/Fable/blob/d8f8ab1642ebda8890b6658ef269e8d043d5a019/src/Fable.Transforms/Replacements.fs#L2711), but not anything that returns a String.

-> fixed. Should now work with everything stringy.

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -2707,8 +2707,8 @@ let regex com (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Exp
     | "get_Item" when i.DeclaringEntityFullName = "System.Text.RegularExpressions.GroupCollection" ->
         // can be group index or group name
         //        `m.Groups.[0]` `m.Groups.["name"]`
-        match args |> List.head with
-        | Value (StringConstant name, _) ->
+        match (args |> List.head).Type with
+        | String ->
             // name
             (* `groups` might not exist -> check first:
                 (`m`: `thisArg.Value`; `name`: `args.Head`)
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Thanks a lot for the fix and for the tests!

Issue #2401, opened 2021-03-03
Use build configuration other than Debug for Constants from fsproj


stroborobo:
### Description

If you have defined constants in your project files (or imported props) that are meant for a Debug build like this:

```xml
<PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
  <DefineConstants>FOOBAR</DefineConstants>
</PropertyGroup>
```

And in  your code use `#if` to implement things differently:

```fsharp
#if FOOBAR
"FOOBAR is defined"
#else
"FOOBAR is NOT defined"
#endif
```

it seems to be impossible to build the project with `dotnet fable` for a non-Debug configuration. The constant FOOBAR will always be set.

### Repro code

[Repository with example](https://github.com/stroborobo/FableDefineIssueDemo)

### Expected and actual results

Expected: no constants are defined that I don't apply myself using --define (except FABLE_COMPILER)

Actual: constants are loaded from project file with assumed configuration Debug

It would be nice though to have --configuration in addition to --define

### Related information

* Fable version: 3.1.5
* Operating system: macOS 11.2.2

PR #2402, opened 2021-03-03
ProjectCracker: support setting a configuration

stroborobo:
Hello wonderful Fable team

For #2401

Technically this changes behaviour since it defaulted to not being set at all. With this change it would default to "Debug" and also not define DEBUG anymore for watch, since that would also deviate from what the project might define. Not sure if that is actually a correct approach.
 
Might be a very naive implementation, since I don't really know how most of these things work from the inside and don't know the code base. Please let me know if I need to change anything.

### DIFF
--- a/src/Fable.AST/Plugins.fs
+++ b/src/Fable.AST/Plugins.fs
@@ -14,6 +14,7 @@ type CompilerOptions =
       abstract ClampByteArrays: bool
       abstract Typescript: bool
       abstract Define: string list
+      abstract Configuration: string
       abstract DebugMode: bool
       abstract OptimizeFSharpAst: bool
       abstract Verbosity: Verbosity
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -56,6 +56,7 @@ Arguments:
   -s|--sourceMaps   Enable source maps
 
   --define          Defines a symbol for use in conditional compilation
+  --configuration   The configuration to use for building the project, the default is 'Debug'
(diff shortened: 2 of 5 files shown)


alfonsogarciacaro:
Hi @stroborobo, thanks a lot for the PR! I think it's good to have the option to set the configuration particularly in cross-platform projects where you may want to have different options depending whether you're targeting .NET or Fable. I remember we had some issues because DotnetProj.Info returns slightly different values depending on the configuration you pass but hopefully it's not a problem.

Code looks good to me but it'd be great if @ncave could also have a look :)

ncave:
Looks good to me too.

stroborobo:
@alfonsogarciacaro I think I see what you mean. Using the Debug configuration doesn't actually define it in a blank project from Fable's POV, but building with dotnet does. Is that what you meant? I guess it only applies those that are explicit?

Should I try to imitate the default defines for Release/Debug targets? I *think* that's only TRACE and DEBUG, none for Release, but I need to read up on that.

stroborobo:
So this is what a `dotnet build` on a net5.0 project defines:

Configuration | DefineConstants
--------------|-----------------
Debug | TRACE;DEBUG;NET;NET5_0;NETCOREAPP
Release | TRACE;RELEASE;NET;NET5_0;NETCOREAPP
asdfg | TRACE;ASDFG;NET;NET5_0;NETCOREAPP

I'll add a define for the capitalized configuration name for now.

alfonsogarciacaro:
Thanks a lot!

PR #2405, opened 2021-03-06
Fixed type issues with getters

ncave:
- Minor cleanup.
- Fixed some type issues with tuple and union getters.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -56,7 +56,13 @@
         {
             "type": "node",
             "request": "launch",
-            "name": "Run bench-compiler tests",
+            "name": "Run bench-compiler JS test",
+            "program": "${workspaceFolder}/src/fable-standalone/test/bench-compiler/out-test/src/test.js"
+        },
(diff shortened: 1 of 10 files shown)


alfonsogarciacaro:
Thanks @ncave! We need to be now a bit careful with the AST because Fable.AST is published as a separate package and to avoid breaking the plugins. Actually, after releasing Fable 3 I wanted to name all the union fields but noticed this broke the union constructors or something else and refrained from that.

There are not many plugins out there yet and the build is passing for the PR which means the `ReactComponent` plugin is still working (it's tested in the React tests) so I guess we could publish it as a new Fable.AST major version. Though maybe it's better to put AST changes in a separate branch and wait to make a major release until we're sure we're ironed out all the issues?

ncave:
@alfonsogarciacaro I removed the proposed AST changes (they'll go to another PR), so this can be merged.

alfonsogarciacaro:
Thanks!

PR #2420, opened 2021-03-29
Added byref parameters support

### LINKED ISSUE #694: Adding byref parameter support

ncave:
- Fixes #694
- Adds general `byref/inref/outref` parameter support.

Note:
- `byref/inref/outref` args are converted to `FSharpRef` on the fly at each call site, in order to not change the generated code. This is how Fable already works, this PR just broadens the `byref` support beyond just `TryGetValue/TryParse/DivRem`.
- This means there are no performance improvements related to this PR (it's more or less the same as the status quo).
- In order to get some performance wins, we'll have to change how mutable variables are represented, instead of capturing them as `FSharpRef` at each call site. They should be created as `FSharpRef` from the beginning, but that is for another PR.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -72,17 +72,17 @@
             "request": "launch",
             "name": "Run bench-compiler (Node)",
             "program": "${workspaceRoot}/src/fable-standalone/test/bench-compiler/out-node/app.js",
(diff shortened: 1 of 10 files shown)


alfonsogarciacaro (approved):
This is great, thanks a lot @ncave!

PR #2430, opened 2021-04-10
Add a --sourceMapsRoot CLI option to set source maps sourceRoot

### LINKED ISSUE #2427: Customize generated source maps
*This is a follow up for https://github.com/delneg/source-map-sharp/pull/5 (Fix option serialization).*

I would like to be able to override the default settings used by Fable to generate source maps.

In particular, as a Fable user, I would like to be able to set the `sourceRoot` property to "file:///" so that source maps work out of the box without requiring another tool (e.g. webpack or another bundler).

I have verified this would work in Chrome on Windows.

It wouldn't currently work in Firefox on Windows, because of something looking like a bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1702920). A workaround would be to url-encode the colon following the drive letter (It seems Firefox interprets `C:/` as an url scheme, making the source path absolute and preventing the `sourceRoot` from being used), so I wonder if Fable could also allow to somehow customize the source paths in its source maps... (Simply always url-encoding colons might be a solution, though I'm not sure it's a good idea)

- Supposing I was to make a PR, how would you like me to expose the `sourceRoot` property from the Fable CLI? (Should I add a CLI option? Is there another extension mechanism? A settings file maybe?)
- What's your opinion about the suggested workaround for the Firefox bug?
  - Would url-encoding the source paths be acceptable? (I don't know how webpack and co handle these urls, so it might maybe break things?)
  - As an alternative, I was thinking of maybe exposing a setting when generating source maps to make Fable choose between simple absolute paths (like currently), or relative paths (either the nearest common folder for all sources, or a manually configured folder, I'm not sure...)

I can start by making a PR to expose just the `sourceRoot` property for now, but I'd really like your opinion about all this before I submit anything... :)

mlaily:
Closes #2427

By setting `--sourceMaps --sourceMapsRoot "file:///"`, this PR makes it possible to use source maps without a bundler :) (does not work on Firefox for now, see #2427 for more details)

I went the simplest way that worked.

I did not use a macro-like syntax like you suggested, since the source maps `sourceRoot` property is not supposed to be a complete path. Maybe you were suggesting this syntax so that it could be used to set the source maps `sources` paths instead, bypassing the `sourceRoot` property completely?
(It would have been a workaround for the Firefox bug, but I didn't feel it was the right thing to do...)

Tell me if you think I missed something, or if I should have done differently.

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -72,6 +72,7 @@ Arguments:
   --noCache         Recompile all files, including sources from packages
   --exclude         Don't merge sources of referenced projects with specified pattern
                     (Intended for plugin development)
+  --sourceMapsRoot  Set the value of the `sourceRoot` property in generated source maps
 
   --optimize        Compile with optimized F# AST (experimental)
   --typescript      Compile to TypeScript (experimental)
@@ -173,6 +174,7 @@ type Runner =
               Configuration = configuration
               OutDir = argValueMulti ["-o"; "--outDir"] args |> Option.map normalizeAbsolutePath
               SourceMaps = flagEnabled "-s" args || flagEnabled "--sourceMaps" args
+              SourceMapsRoot = argValue "--sourceMapsRoot" args
               NoRestore = flagEnabled "--noRestore" args
               NoCache = flagEnabled "--noCache" args || flagEnabled "--forcePkgs" args // backwards compatibility
               Exclude = argValue "--exclude" args
--- a/src/Fable.Cli/Fable.Cli.fsproj
+++ b/src/Fable.Cli/Fable.Cli.fsproj
@@ -37,7 +37,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Dotnet.ProjInfo" Version="0.44.0" />
-    <PackageReference Include="source-map-sharp" Version="1.0.5" />
+    <PackageReference Include="source-map-sharp" Version="1.0.7" />
     <!-- <PackageReference Include="FSharp.Core" Version="5.0.0" /> -->
   </ItemGroup>
 </Project>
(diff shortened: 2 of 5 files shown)


alfonsogarciacaro:
Looks good to me, thank! :+1: Though there seems to be an error in the Appveyor build related to the source-map-sharp upgrade. @ncave Seems that latest source-map-sharp cannot be compiled with Fable because of the new FSharpConverters. Would it be possible to have a look?

ncave:
@alfonsogarciacaro See [PR#6](https://github.com/delneg/source-map-sharp/pull/6), we can update the version once it goes through.

mlaily:
Ah sorry about that, I added this code! (but wouldn't ignoring the converters when compiling with Fable yield an incorrect json result? That said, generating source maps from Fable-translated code might not be a use case worth fixing...)

delneg:
> @alfonsogarciacaro See [PR#6](https://github.com/delneg/source-map-sharp/pull/6), we can update the version once it goes through.

Merged.

Starting to think that we might need to test Fable version separately to keep source-map-sharp compatible with both .net & fable...

ncave:
@mlaily
- We're not just ignoring the Json converters, we're ignoring the whole `System.Text.Json.Serialization`, as it is not implemented by Fable's BCL.
- Generating source maps from Fable-translated code is most definitely a valid use case, it is already being used in `fable-compiler-js`, which for now just uses [JSON.stringify](https://github.com/fable-compiler/Fable/blob/nagareyama/src/fable-compiler-js/src/util.js#L38) to serialize the source maps.

mlaily:
> Generating source maps from Fable-translated code is most definitely a valid use case, it is already being used in `fable-compiler-js`, which for now just uses [JSON.stringify](https://github.com/fable-compiler/Fable/blob/nagareyama/src/fable-compiler-js/src/util.js#L38) to serialize the source maps.

Ah thanks for the pointer, I hadn't realized `option` types were erased properly by Fable out of the box! (Though I faintly recall seeing that somewhere in the docs a long time ago :))

Thank you all for your quick responses. I updated my PR to use source-map-sharp 1.0.7.

alfonsogarciacaro:
Thank you all @mlaily @delneg @ncave!

PR #2432, opened 2021-04-12
Minor Fable AST cleanup

ncave:
- Just so it doesn't get lost in the sands of time, a modest proposal to make `Fable.Set` a bit more symmetric with `Fable.Get`.


alfonsogarciacaro:
We can create a new `next` branch for next major release and merge this there. Other changes we may want to do to the AST:

- Give names to all the union fields. This way plugins can pattern match union fields by name instead of position so hopefully they don't break if we add a new field to the union case.
- Add union case tag to `Get(_, UnionField(index, t), _, _)`, as in the F# AST, so we can get the union case and field from the entity reference. See https://github.com/fable-compiler/Fable/pull/2447#discussion_r638436247
- Add metadata to all elements so we can "extend" the AST (at the cost of some type safety) without modifying it.

ncave:
@alfonsogarciacaro Perhaps also add entity reference to `AnonymousRecordType`:
```fs
| AnonymousRecordType of EntityRef * fieldNames: string [] * genericArgs: Type list
```

src/Fable.AST/Fable.fs
```diff
@@ -266,12 +267,18 @@ type KeyKind =
 type GetKind =
     | ByKey of KeyKind
     | TupleIndex of int
+    | FieldGet of Field * index: int
```
alfonsogarciacaro, line 270: Should we add `Field` to the AST elements or maybe just the name so we can retrieve the field from the entity reference when needed as in here? https://github.com/fable-compiler/Fable/pull/2447#discussion_r640307157

alfonsogarciacaro: Same for `SetKind` and `KeyKind` (which already contain a direct reference to `Field` but maybe we can remove it).

alfonsogarciacaro (commented):
Sorry @ncave I created the `next` branch and merged this too quickly but I've left a comment later. Could you please check? Thanks!

src/Fable.AST/Fable.fs
ncave: @alfonsogarciacaro I think adding Field is fine, but whatever yo prefer.

Issue #2439, opened 2021-05-09
Consider several NuGet packaging enhancements


cartermp:
Hey fine fable folks 👋 

There are a few modifications to packaging for NuGet that you may want to consider:

* Adding an icon to packages
* Adding a license expression
* Publishing symbols

Additionally, it's considered a best practice to have an organization be the owner and publisher of a package or packages when it maps to an organization on GitHub. Individuals with publishing rights would be members of the organization, so it wouldn't materially impact their ability to publish things.

An example of all of the above is the [XPlot.Plotly](https://www.nuget.org/packages/XPlot.Plotly/) package.

Happy to send some PRs for the first three items if you're interested. Any org-level stuff would have to be taken care of by someone with permissions, though.

MangelMaxime:
Hello @cartermp,

I agree with you.

It shouldn't be too hard to add the different information to the different project using a `Directory.Build.props` we could kind of copy it in the different repository.

> Publishing symbols

I am kind of waiting for 5.0.300 to be released as there is a new NuGet `DotNet.ReproducibleBuilds` which seems to configure it for us (source link). Unless you are speaking about something else?

MangelMaxime:
Also, @nojaf started doing that on 2 projects if you want to review one of the PR

https://github.com/fable-compiler/fable-browser/pull/62 

Like that we can take inspiration from it for others repositories :)

How does managing the organization works ? Can several people be admin so if someone his not available anymore someone else can add members etc.?

alfonsogarciacaro:
Hi @cartermp! Thank you and @MangelMaxime for the ideas. Yes, it'd be good to improve the publishing of the Nuget packages, so far I've been quite sloppy with this. (Every time I see the licence warning I want to fix but then I follow the provided link and cannot find the solution I give immediately up 😅). There are also some ideas [here](https://github.com/fable-compiler/fable-compiler.github.io/issues/86#issuecomment-823200266) to improve the search results for the packages. Basically adding tags and a readme, I think.

It'd be great a to get a PR with the proposals that we can use as a base for other packages. And having just a single file like `Directory.Build.props` as @MangelMaxime says that we can just copy/paste to other repos would be ideal. If some automation is required, this is the function that I usually use for publishing: https://github.com/fable-compiler/Fable/blob/caa715f1156be29c8dd9b866a03031a1852b3186/src/fable-publish-utils/PublishUtils.fs#L547

alfonsogarciacaro:
[...]

PR #2441, opened 2021-05-12
Update package properties to adhere to some best practices

cartermp:
Fixes https://github.com/fable-compiler/Fable/issues/2439 by adding:

* PackageLicenseExpression for the license
* Icon instead of IconURL
* SourceLink properties so that symbol packages (`.snupkg`) can also be published

Now when you pack each of the package directories, it will produce two packages under `bin/release`:

* `package-name.nupkg`
* `package-name.snupkg`

You'll want to upload both to NuGet when updating packages.


src/Directory.Build.props
```diff
@@ -2,9 +2,22 @@
...
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+        <!-- For SourceLink. See: https://github.com/dotnet/sourcelink#using-source-link-in-net-projects -->
+        <PublishRepositoryUrl>true</PublishRepositoryUrl>
+        <EmbedUntrackedSources>true</EmbedUntrackedSources>
+        <IncludeSymbols>true</IncludeSymbols>
+        <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     </PropertyGroup>
+    <ItemGroup>
+        <None Include="..\..\fable_logo.png" Pack="true" PackagePath="\" />
```
alfonsogarciacaro, line 18: It seems the fable_logo.png icon is missing from this repo, we should probably add it. It'd be also great if the directory is not relative (I guess we can use the MsBuildThisFileDirectory constant) so we can easily copy the file to other repos without depending on a specific structure.

alfonsogarciacaro (commented):
Thanks a lot @cartermp! If it's not too much to ask, would it be possible to also edit these lines in this PT so the publishing script automatically pushes the correct files as you mention in the description? https://github.com/fable-compiler/Fable/blob/caa715f1156be29c8dd9b866a03031a1852b3186/src/fable-publish-utils/PublishUtils.fs#L569-L588

src/Directory.Build.props
cartermp: Yep, it's included in the PR.

cartermp:
@alfonsogarciacaro absolutely! Will take a look. It's starting to get into the evening now on my side, so I'll likely pick this back up tomorrow. I need to unwind a bit, otherwise I won't fall asleep easily!

src/Directory.Build.props
```diff
@@ -2,9 +2,22 @@
     <PropertyGroup>
         <PackageProjectUrl>http://fable.io</PackageProjectUrl>
         <RepositoryUrl>https://github.com/fable-compiler/Fable.git</RepositoryUrl>
-        <PackageLicense>https://raw.githubusercontent.com/fable-compiler/Fable/master/LICENSE</PackageLicense>
-        <PackageIconUrl>https://raw.githubusercontent.com/fable-compiler/Fable/master/docs/img/fable_logo.png</PackageIconUrl>
+        <PackageLicenseExpression>MIT</PackageLicenseExpression>
```
nojaf, line 5: @MangelMaxime would you prefer the actual license file here as well?

src/Directory.Build.props
MangelMaxime: Instead of using a relative link, I propose to use `$(MSBuildThisFileDirectory)`

If find it easier to read and it will works for all the packages in the repository independently of their level.

For example, in Fable.Form I do it like that for the license file

```xml
<PropertyGroup>
    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
</PropertyGroup>
<ItemGroup>
    <None Include="$(MSBuildThisFileDirectory)\LICENSE.txt" Pack="true" PackagePath="$(PackageLicenseFile)" />
</ItemGroup>
```

src/Directory.Build.props
MangelMaxime: It is true that I prefer the licence file in general.

My idea, is that using the licence file means that if a user read the code from Github or NuGet he will see the same source of truth ==> The same LICENSE.txt file

Related to one of this https://github.com/fable-compiler/fable-browser/pull/62 @nojaf sent

cartermp: Done.

src/Directory.Build.props
cartermp: Done.

cartermp:
@alfonsogarciacaro @MangelMaxime Addressed all feedback. Since I don't have the ability to publish to NuGet, you'll have to verify the publish script works as expected, but I expect it will be fine.

Here's what the package will look like:

image

Note that determinism/compiler flags were always a red x mark, so nothing has regressed there. Critically, the icon, license, and symbols are all correct.

MangelMaxime:
Looks good to me 

@alfonsogarciacaro I let you publish the new versions 😉

src/fable-publish-utils/PublishUtils.fs
```diff
@@ -584,7 +584,21 @@ module Publish =
...
+                        then Some(tempDir </> path)
+                        else None)
+                    |> function
+                        | Some x -> x
+                        | None -> failwithf "Cannot find .snupgk for %s" projDir
+
                 runList ["dotnet nuget push"; nupkg; "-s nuget.org -k"; nugetKey]
+
+                // Publish snupkg after the nupkg, since you need to already have a corresponding nupkg to match the symbols against
+                runList ["dotnet nuget push"; snupkg; "-s nuget.org -k"; nugetKey]
```
alfonsogarciacaro, line 601: @cartermp I got this when trying to push a package so it seems the first `dotnet nuget push` command is already detecting and pushing the symbols (which makes the second command fail because they're already published) so I have commented out this line.

```
> dotnet nuget push src/Fable.AST/temp/Fable.AST.3.1.1.nupkg -s nuget.org -k ...
Pushing Fable.AST.3.1.1.nupkg to 'https://www.nuget.org/api/v2/package'...
  PUT https://www.nuget.org/api/v2/package/
  Created https://www.nuget.org/api/v2/package/ 1690ms
Your package was pushed.
Pushing Fable.AST.3.1.1.snupkg to 'https://www.nuget.org/api/v2/symbolpackage'...
  PUT https://www.nuget.org/api/v2/symbolpackage/
  Created https://www.nuget.org/api/v2/symbolpackage/ 725ms
Your package was pushed.
> dotnet nuget push src/Fable.AST/temp/Fable.AST.3.1.1.snupkg -s nuget.org -k ...
Pushing Fable.AST.3.1.1.snupkg to 'https://www.nuget.org/api/v2/symbolpackage'...
  PUT https://www.nuget.org/api/v2/symbolpackage/
  Conflict https://www.nuget.org/api/v2/symbolpackage/ 843ms
To skip already published packages, use the option --skip-duplicate
```

cartermp: Makes sense! Interesting that it all just works with the snupkg.

alfonsogarciacaro: If we're not sure what are the conditions to trigger the symbol publishing maybe it's better to still issue the second command with `-skip-duplicate` just to be sure 🤔

PR #2443, opened 2021-05-17
Updated fcs-fable

ncave:
- Udated `fcs-fable` to latest
- Updated `fable-standalone`

PR #2444, opened 2021-05-17
Merge branch 'next1' into update-fcs

ncave:
- Merged #2443  here.


alfonsogarciacaro:
Thanks @ncave! Seems we still have the warnings about inheriting from primary constructors and the following error. Any idea how to fix this?
```
src/fcs-fable/src/fsharp/TypedTree.fs(1,1): error EXCEPTION: The exception has been reported. This internal exception should now be caught at an error recovery point on the stack. Original message: Exception of type 'FSharp.Compiler.ConstraintSolver+ConstraintSolverError' was thrown.)
```

ncave:
@alfonsogarciacaro I'm not sure, there are quite a few ConstraintSolver errors being thrown when I debug it. And some are thrown when compiling code that was not recently touched. I'll keep looking.

ncave:
@alfonsogarciacaro The error above seems to only show when we don't compensate for witness errors, so it's clearly related to the error handling when calling `GetWitnessArgs`.

PR #2442, opened 2021-05-14
Update FCS

alfonsogarciacaro:

PR #2446, opened 2021-05-20
Updated binaries

ncave:
@alfonsogarciacaro This fixes the FCS issues and updates a few other binaries.
- Main issue: After scratching my head for a while, turns out the issue was in the FCS binary in the `lib/fcs` folder.
I rebuilt it and it works now.
- (unrelated) Updated BCL metadata binaries to latest.
- (unrelated) Avoid throwing exceptions for non-F# delegates.
- (unrelated) Fixed a minor problem with Range.
- (unrelated) Fixed an issue with FileSystem.

~The latest `fcs-fable` has an outstanding tail calls issue [here](https://github.com/ncave/fsharp/blob/main/src/fsharp/symbols/Exprs.fs#L1186-L1193) (affecting only `fable-standalone`/`fable-compiler-js`) with Node.js maximum call stack size exceeded on large files, but that shouldn't affect this PR.~
It's a pity compiling large files like `pplex.fs` using the `fable-compiler-js` no longer works with the default Node.js `--stack_size=984`, but bumping it to `--stack_size=1200` makes it work.


alfonsogarciacaro:
This is great, thanks a lot! I will merge #2442 and push a new version :+1:

Issue #2339, opened 2021-01-04
Python Support for Fable


dbrattli:
### Description

This issue is a discussion about [Python](https://www.python.org) support for Fable. The Babel AST currently used by Fable is close enough to generate Python code. Especially now that classes are supported. 

A POC has been made to show that this is possible. However, quite a lot of work is needed to:

1. Add decent support for Python 
2. Make Fable more JS agnostic and support multiple target languages.

The [Expression](https://github.com/cognitedata/Expression/tree/fable) library enables functional programming in Python inspired by F#. It provides it's own implementation of data types such as `option`, `result`, `seq`, `map`, `list` (called FrozenList), mailbox processor, ... Thus it should be suited as the Python equivalent of the Fable library. In Python you use Expression to do F#'ish code. In Fable you generate Python code using Expression.

## Use cases:

- F# running in Jupyter notebook with Python kernel, i.e similar to what is done with e.g [Hy](https://github.com/hylang/hy) and [Calysto Hy](https://github.com/Calysto/calysto_hy).
- F# scripting using Python for reduced friction
- More easily support embedded environments like the [micro:bit](https://microbit.org) and [Rasberry PI](https://www.raspberrypi.org).
- Sharing data model between F# and Python code

## Things to discuss and decide:

- Should Python for F# be more Pythonic than .NET? There is already .NET support for Jupyter which we should not try to replace or compete with. Fable for Python should perhaps target Python instead of .NET (if possible) and be a more natural fit for Python developers looking to F#.
- Should we try to be compatible with Babel AST or should we diverge into our own Python AST. This is perhaps inevitable, but Babel gives us a good starting point.
- Should we for Fable (as with [Peeble](https://github.com/thinkbeforecoding/peeble)) or should we stay within Fable and have Fable support multiple languages. Having our own repo gives freedom and speed (by not needing to care about JS), but we will risk being left behind and obsoleted over time. My feeling is that we should become part of Fable.
- Python basic types is also different from F#. E.g `int` has an arbitrary length. Should we emulate .NET (machine) types to be compatible or should we expose the Python `int` type as an F# int?

## Installing the POC

The source code of the POC currently lives here:

- https://github.com/dbrattli/Fable/tree/python (https://github.com/dbrattli/Fable/pull/1)
- https://github.com/cognitedata/Expression/tree/fable

1. Install the latest Python 3.9 from https://www.python.org or `brew install python@3.9` on Mac.
[...]

PR #2448, opened 2021-05-23
Language selection

dbrattli:
Add language selection to enable compilation to other languages like Python and PHP. Ref #2339 and #2447

### DIFF
--- a/src/fable-standalone/src/Main.fs
+++ b/src/fable-standalone/src/Main.fs
@@ -1,6 +1,7 @@
 module Fable.Standalone.Main
 
 open System
+open Fable
 open Fable.AST
 open Fable.Transforms
 open Fable.Transforms.State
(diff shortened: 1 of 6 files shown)


src/fable-standalone/src/Main.fs
```diff
@@ -265,7 +266,8 @@ let init () =
                 if x.StartsWith("--define:") || x.StartsWith("-d:")
                 then x.[(x.IndexOf(':') + 1)..] |> Some
                 else None) |> Array.toList
-            let options = Fable.CompilerOptionsHelper.Make(define=define, ?typedArrays=typedArrays, ?typescript=typescript)
+            let language = typescript |> Option.map (fun ts -> match ts with | true -> TypeScript | _ -> JavaScript) |> Option.defaultValue JavaScript
+            let options = Fable.CompilerOptionsHelper.Make(define=define, ?typedArrays=typedArrays, language=language)
```
ncave, line 270: @dbrattli Is this missing the `?` on the language arg?

dbrattli: No, don't think so. The `language` is not an option as I use `defaultValue`. Btw. Tried first to propagate language settings but problems with deps and where the Language union should live to be visible for all projects. Currently it's in `Fable.AST/Plugins.fs` which makes it hard to get it into e.g Worker.fsproj. So decided to keep the `typescript: bool` for now for fable-standalone.

ncave: @dbrattli That's fine, I'll fix it after this gets merged. Thanks!

dbrattli: @ncave Sounds great!

alfonsogarciacaro:
Looks good, thanks a lot!

thinkbeforecoding:
There is still some work for some specific constructs, and working out imports in a more php idiomatic way.. But globaly, it works and generate nice php code 😁

sandeepc24:
Compilation to Dart language might be a good option, just saying :-)

PR #2454, opened 2021-05-26
Fix events issues and implement FSharpEvent`2

### LINKED ISSUE #1950: can't use event exposed on generic interface
### Description

Event registration doesn't work when done through a generic interface.

### Repro code

```fsharp
let [<Literal>] willRun = false
type IMyType<'t> =
    [<CLIEvent>] abstract member OnI : IEvent<int>
    abstract member Send : 't -> unit
type MyType () =
    let e = new Event<_>()
    
    member x.Send (i:int) =
        e.Trigger i
        
    interface IMyType<int> with
        [<CLIEvent>] member x.OnI = e.Publish
        member x.Send i = e.Trigger i
printfn "hello"
let t : IMyType<int> = MyType() :> _
try

  if willRun then () else
  printfn "subscribing"
  let subscription = t.OnI.Subscribe(fun i -> printfn "on i worked %i" i)
  ()
with e -> 
  printf "oops subscription %A" e
```

the `OnI` event can't be subscribed to from the generic interface, skipping the registration and the program runs to completion.

I expect the registration to work the same as if it was on the concrete instance, it currently causes

> Uncaught TypeError: Object(...) is not a function

### Related information

`$ yarn info fable-compiler version`
> yarn info v1.16.0
> fable2.4.11

chkn:
Hello! For my first PR to Fable, there are 3 things related to events, described briefly in the sections below.

## Fix CLI events

As mentioned in #1950, there is some nifty logic in `(|CreateEvent|_|)` to match what the F# compiler emits to wrap CLI events. [This SO answer](https://stackoverflow.com/a/13387792) has some nice details about CLI vs non-CLI events. There are a couple implications for Fable:

1. For a CLI event called `Event`, the F# compiler emits `add_Event` and `remove_Event` methods. These are the only members in the contract of a CLI event, but an `Event` property is also emitted that exposes the `IEvent`. In the cases where the type is statically known, we can find that `Event` property, which is what `(|CreateEvent|_|)` attempts to do. Previously it wasn't working in all cases because it was just emitting `foo.Event` instead of calling the mangled property getter function, so this PR fixes that.

2. When a CLI event is in an interface, we cannot find the property because it is not in the contract, so we need an implementation for `CreateEvent`, which wraps the `add_Event` and `remove_Event` methods into an anonymous `IEvent`. This PR adds that.

Fixes #1950 

## Implement ``FSharpEvent`2``

Currently, Fable only supports ``FSharpEvent`1``. It turns out it's fairly straightforward to modify the existing TypeScript `Event` class to serve as both ``FSharpEvent`1`` and ``FSharpEvent`2``. We can simply put both types of delegates into the same array and eliminate the `Map`, simplifying the implementation.

## Streamline Events module

As part of the above, I took a look at the functions in the Events module. These seemed to all depend on implementation details of `Event`, when they should really work on any `IEvent` (such as the ones returned by the new `createEvent` implementation). So I rewrote these to more closely match the [implementations in `FSharp.Core`](https://github.com/dotnet/fsharp/blob/main/src/fsharp/FSharp.Core/eventmodule.fs) - and the implementations are now much simpler and easier to understand IMHO.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -667,7 +667,14 @@ module Patterns =
                 Lambda(_callback, NewDelegate(_, Lambda(_delegateArg0, Lambda(_delegateArg1, Application(Value _callback',[],[Value _delegateArg0'; Value _delegateArg1'])))))])
           when createEvent.FullName = Types.createEvent ->
             let eventName = addEvent.CompiledName.Replace("add_","")
-            Some (callee, eventName)
+            match addEvent.DeclaringEntity with
+            | Some klass ->
+                klass.MembersFunctionsAndValues
+                |> Seq.tryFind (fun m -> m.LogicalName = eventName)
+                |> function
+                | Some memb -> Some (callee, memb)
+                | _ -> None
+            | _ -> None
         | _ -> None
 
     let (|ConstructorCall|_|) = function
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -493,9 +493,10 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) fsExpr =
 
     | FSharpExprPatterns.Let((var, value), body) ->
         match value, body with
-        | CreateEvent(value, eventName), _ ->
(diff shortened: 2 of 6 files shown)


alfonsogarciacaro:
This is great @chkn! I implemented support for Observables but Events came from a contribution and I hadn't really touched them since 😅 Thanks a lot for improving them!

PR #2456, opened 2021-06-01
Adds ModuleDecl in AST

ncave:
- Adds `ModuleDecl` in AST for languages that support them.


alfonsogarciacaro:
Makes sense, thanks!

PR #2459, opened 2021-06-02
Check for correct types in Anonymous Record when assigning to Interface with `[<EmitIndexer>]` via `!!`

Booksbaum:
Checks assignment of Anonymous Records to an interface (with `!!`) when interface contains `[<EmitIndexer>]`, see https://github.com/fable-compiler/ts2fable/issues/415#issuecomment-836084087 . (There's already a check for normal members in Interface)

-> Checks stuff like:
```fsharp
type MyInterface =
  [<EmitIndexer>]
  abstract Item: key:string -> string

let v: MyInterface = !!{| Value = "foo"; AnotherValue = 42 |}
```
-> Warning for `AnotherValue`:
> FABLE: Expected type 'System.String' for field 'AnotherValue' because of Indexer 'Item' in interface 'MyInterface', but is 'System.Int32'

<br/>
<br/>

A couple of caveats & issues:
* Errors are just Compiler Warnings, not compiler errors. Same behaviour as already existing checks for interface members.
* Previously there was only one error/warning per anon record displayed; now there's one error/warning for every incorrect Field (either missing or incorrect type)
* Handles `[<EmitIndexer>]` (-> 'index signature' in TS), NOT F# indexer. (Though in most cases they are probably the same member)
* Currently it's just assumed a Member with `[<EmitIndexer>]` is valid. In TS: string and/or int input, and some return type. That's not checked.  
  Might be reasonable for another PR, but not here. Location to check would probably be https://github.com/fable-compiler/Fable/blob/155bd5e0e446a11ad010d4869a04ca59785f183a/src/Fable.Transforms/FSharp2Fable.Util.fs#L1416
* Erased Unions are reduced in Fable to Type `Any` (https://github.com/fable-compiler/Fable/blob/155bd5e0e446a11ad010d4869a04ca59785f183a/src/Fable.Transforms/FSharp2Fable.Util.fs#L856) -> quite a lot checks involve `Any` either on interface member side (lhs) or anon record field side (rhs), like:
  ```fsharp
  [<Erase>] type E = E of string
  type I1 =
    [<EmitIndexer>]
    abstract Item: string -> E
    //                       ^ Any

  let i1: I1 = !! {| Value = 42 |}
  //                 ^^^^^^^^^^
  //                 `int` fits into `Any` -> no error
  ```
  There's one special case that's handled: `Ux<...>` in interface:
  ```fsharp
  type I1 =
    [<EmitIndexer>]
    abstract Item: string -> U2<string,int>

  let i1: I1 =
      !! {|
          V1 = 42
          //   ^^ int -> ok
          V2 = "foo"
          //   ^^^^^ string -> ok
          V3 = 3.14
          //   ^^^^ float -> error
      |}
  ```
  But because there are valid values on the rhs that are reduced to `Any`, those cases have to be accepted too:
  ```fsharp
  let i1: I1 =
      !! {|
          V5 = U2.Case2 42  // valid: type `U2<string,int>`
          //   ^^ Any -> Don't know if correct type -> accepted
          V5 = obj ()       // invalid: type `obj`
          //   ^^^ Any -> don't know if correct type -> accepted
          V6 =              // invalid: type `U3<string,int,float>`
              let v: U3<string, int, float> = U3.Case3 3.14
              v
          //  ^ Any -> don't know if correct type -> accepted
      |}
  ```
  Reason lhs (interface) can be handled, but rhs (actual value) cannot: For interface its F# type is available, while the rhs is already reduced to its Fable Type (and `Type.ErasedUnion` was removed some time ago (#2112))  
  Additional:
  * `Ux<...>` behaviour was added to normal interface member check too
  * `Ux<...> option` received special treatment too (`U2<string,int> option = U2<string option, int option>`). `option` behaviour is same as normal interface member check (`int option` => both allowed: `Some 42` and just `42`)
* There were no tests for compiler messages -> Added test project `tests/Compiler`
  * Uses `Fable.Standalone` for compiling. 
  * Creating Checker takes some time -> running tests seems like it's doing nothing for some time...
  * Run tests:
    * `dotnet fsi build.fsx test-compiler`
    * in `test/Compiler`: `dotnet run`
    * executed as part of all `test`s (`dotnet fsi build.fsx test`)

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -939,38 +939,318 @@ module TypeHelpers =
                 else false)
         | _ -> None
 
-    let fitsAnonRecordInInterface _com (argExprs: Fable.Expr list) fieldNames (interface_: Fable.Entity) =
+    [<RequireQualifiedAccess; Flags>]
+    type private Allow =
+    | TheUsual      = 0b0000
+      /// Enums in F# are uint32
+      /// -> Allow into all int & uint
+    | EnumIntoInt   = 0b0001
+      /// Erased Unions are reduced to `Any`
+      /// -> Cannot distinguish between 'normal' Any (like `obj`) and Erased Union (like Erased Union with string field)
+      ///
+      /// For interface members the FSharp Type is available
+      /// -> `Ux<...>` receive special treatment and its types are extracted
+      /// -> `abstract Value: U2<int,string>` -> extract `int` & `string`
+      /// BUT: for Expressions in Anon Records that's not possible, and `U2<int,string>` is only recognized as `Any`
+      /// -> `{| Value = v |}`: `v: int` and `v: string` are recognized as matching,
+      ///    but `v: U2<int,string>` isn't: only `Any`/`obj` as Type available
+      /// To recognize as matching, we must allow all `Any` expressions for `U2` in interface place.
+      ///
+      /// Note: Only `Ux<...>` are currently handled (on interface side), not other Erased Unions!
+    | AnyIntoErased = 0b0010
+      /// Unlike `AnyIntoErased`, this allows all expressions of type `Any` in all interface properties.
+      /// (The other way is always allow: Expression of all Types fits into `Any`)
+    | AlwaysAny     = 0b0100
+    let fitsAnonRecordInInterface
+        (_com: IFableCompiler)
+        (argExprs: Fable.Expr list)
+        (fieldNames: string array)
+        (interface_: Fable.Entity)
+        (range: SourceLocation option)
+        =
         match interface_ with
         | :? FsEnt as fsEnt ->
             let interface_ = fsEnt.FSharpEntity
+            let interfaceMembers =
+                getAllInterfaceMembers interface_
+                |> Seq.toList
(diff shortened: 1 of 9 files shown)


src/Fable.Transforms/FSharp2Fable.Util.fs
```diff
@@ -939,38 +939,318 @@ module TypeHelpers =
...
+    | AnyIntoErased = 0b0010
+      /// Unlike `AnyIntoErased`, this allows all expressions of type `Any` in all interface properties.
+      /// (The other way is always allow: Expression of all Types fits into `Any`)
+    | AlwaysAny     = 0b0100
+    let fitsAnonRecordInInterface
+        (_com: IFableCompiler)
+        (argExprs: Fable.Expr list)
+        (fieldNames: string array)
+        (interface_: Fable.Entity)
+        (range: SourceLocation option)
```
alfonsogarciacaro, line 969: I usually put the range as second argument between compiler and type, so most calls look like `doSomething com r t ...` but that's a minor styling issue we can change later :)

alfonsogarciacaro:
Wow, it's great you added so many integration tests. Thanks a lot for this! :clap: :clap: :clap:

Issue #2321, opened 2020-12-17
Resolve System.Type.GetMethod(name) at compile time


lfr:
After having worked around most limitations, it appears I cannot make [FSharp.ValidationBlocks](/lfr/FSharp.ValidationBlocks) compatible with Fable which is a shame as it's a match made in heaven for web dev for form validation.

I found workarounds to read the content of the blocks, but I can't create them because the information required to create the blocks (such as their generic parameters) is stored in the interface they implement, and the limited reflection support that I gather from [these tests](https://github.com/fable-compiler/Fable/blob/nagareyama/tests/Main/ReflectionTests.fs) doesn't include GetInterfaces().

I'd love to contribute but I'm not myself a web developer and I don't think the solution to this is pure F#. Is better interface reflection support planned or should I give up?

lfr:
I found a workaround to my problem by limiting the block types to blocks of non-DU types, which I think it's an ok compromise since 99% of the time people are going to be validating strings, ints, etc.

Since I already had a ticket open about reflection, I'll close this one.

alfonsogarciacaro:
In the early stages of Fable we picked interfaces to easily type objects coming from JS, and we cannot change this now because a lot of bindings depend on this, that's why we cannot do too many things with interfaces because Fable sees them just as a contract of the object shape and doesn't expect the object to contain any information about the interfaces it implements, so we cannot do interface testing or similar. Though sometimes we do dirty runtime checks like `typeof x.Dispose === "function"`.

However, I guess it would be possible to implement `.GetInterfaces()` for types known at compile time. We should only be careful because interface declarations are erased so we would have to generate the interface information in the call site. Which means that if you call `.GetInterfaces()` multiple times on the same type in different places Fable will output duplicated code.

alfonsogarciacaro:
Ok, adding support for `.GetInterface` was not too problematic when types can be resolved at compile time, so in the next release you'll be able to do something [like this](https://github.com/fable-compiler/Fable/blob/13104ae650d4ec32acf684709dc183e10a09ddee/tests/Main/ReflectionTests.fs#L499-L505). I hope it helps!

lfr:
@alfonsogarciacaro it'll help if we also get the following, not sure if it's possible:

✅ Type.MakeGenericType
🟩 Type.GetMethods
🟩 MethodInfo.MakeGenericMethod
🟩 MethodBase.Invoke

Do you think this is realistic?

alfonsogarciacaro:
[...]

PR #2465, opened 2021-06-11
Extend Fable AST to calculate overload suffixes

alfonsogarciacaro:
This PR was originally intended to handle tail-call optimization directly in the Fable AST so I added an optional label to `While` loops and a new `NativeInstruction` expression for instructions like `break`, `throw` or `debugger`. @dbrattli you can use this to identify when an exception is thrown. Also @thinkbeforecoding, if you have a moment could you please add the code to handle these instructions? https://github.com/fable-compiler/Fable/blob/d60f08a9d87b6b31e0831c2147e10c63a4fc8688/src/Fable.Transforms/Php/Fable2Php.fs#L512-L514

After updating the AST I realized, however, that porting the code for tail-call optimization from Fable2Babel to Fable.Transforms is not trivial so I didn't do it at the end 😅 I wanted to merge the PR as is for now but before that I had the idea of updating the `OverloadSuffix` module so it can handle the Fable AST and doesn't depend on FCS. That proved to be not trivial either but I think I managed to do it. For this it has also been necessary to extend the AST with information to tell the type of some arguments apart like struct option/tuples or numbers with units of measure.

Why is this useful? First it's another step towards being able to serialize the AST (the next one will be to also use the Fable AST to resolve inline functions) and should also help solve this issue: https://github.com/fable-compiler/Fable/issues/2321#issuecomment-759200517

BTW, the test for the `ReactComponent` plugin is finally failing because of the changes in the AST, so I've disabled it for now.


thinkbeforecoding:
@alfonsogarciacaro , I created #2467 to implement it for Php.

PR #2479, opened 2021-07-12
Fix dropped DEBUG constant

### LINKED ISSUE #2477: DefineConstants "DEBUG" and "TRACE" are dropped
### Description

Hey lovely Fable people,

(This is kind of a follow up to #2401 and #2402.)

I have a project configuration called "Test" that also defines "DEBUG" and "TRACE". When building with Fable these constants are stripped after the project cracker had found them in the project file.

### Repro code

In the project file:

```xml
<PropertyGroup Condition=" '$(Configuration)' == 'Test' ">
  <DefineConstants>DEBUG</DefineConstants>
</PropertyGroup>
```

```fsharp
printfn "DEBUG %s"
#if DEBUG
    "yes"
#else
    "no"
#endif
```

Build with `--configuration Test`

- Expected (DotNet): prints "DEBUG yes"
- Actual (Fable): prints "DEBUG no"

After some digging I've found a function that explicitly drops `--define:DEBUG` and `--define:TRACE`, 

https://github.com/fable-compiler/Fable/blob/4ecab554/src/Fable.Cli/ProjectCracker.fs#L272

Maybe the outer match is obsolete since the change with the configuration parameter in #2401 / #2402?

### Related information

* Fable 3.2.9
* macOS 11.4

stroborobo:
- don't strip DEBUG and TRACE
- don't define the configuration ourselves
- rely on Dotnet.ProjInfo to return constants based on configuration
- add tests:
  - Debug with extra constants
  - custom configuration "Test"
  - Release without extra constants

Fixes #2477

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -352,8 +352,28 @@ let testMocha() =
 
     runMocha buildDir
 
+let testDefineConstants() =
+    [ "tests/DefineConstants/DebugWithExtraDefines", "Debug"
+      "tests/DefineConstants/CustomConfiguration", "Test"
+      "tests/DefineConstants/ReleaseNoExtraDefines", String.Empty ]
+    |> List.iter (fun (projectDir, configuration) ->
+        let buildDir = "build/"+ projectDir
+
+        cleanDirs [ buildDir ]
+        runFableWithArgs projectDir [
+            "--outDir " + buildDir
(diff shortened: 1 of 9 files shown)


alfonsogarciacaro:
Looks good! Thanks a lot for the fix and the tests :clap: :clap:

PR #2482, opened 2021-07-16
Python fixes

dbrattli:
- Add list tests for Python
- Add map tests for Python
- Fix bug with for-loop when counting down (Python is exclusive for range limits)
- Fix build script to build python library instead of js library for python targets
- Fix pyNative to nativeOnly


alfonsogarciacaro:
Thanks!

PR #2, opened 2016-02-10
Make buildable on Windows

davidpodhola:
+ add missing functions from List module

PR #2486, opened 2021-07-18
Python fixes #2

dbrattli:
- Use Python `__str__` instead `ToString`. Use Python string protocol instead of `ToString()` members. I.e members are named `__str__` instead of `ToString`. That way we can string any object using `str(obj)`. This is the same as we do with iterators being named `__iter__`. 
- Fix Python entry point by adding `if __name__ == "__main__":` before calling entry point function with `sys.args[1:]`

### DIFF
--- a/src/Fable.Transforms/Python/Fable2Python.fs
+++ b/src/Fable.Transforms/Python/Fable2Python.fs
@@ -1991,12 +1991,15 @@ module Util =
         let args = Arguments.arguments(args |> List.map Arg.arg)
         args, body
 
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Thank you!

alfonsogarciacaro:
Sorry @dbrattli! Forgot to ask you to target `beyond` branch instead. I'll rebase the commit.

dbrattli:
@alfonsogarciacaro Thanks, I'll start targeting the beyond branch instead. Have some more fixes coming soon!

PR #2494, opened 2021-07-22
Add replacement for System.Uri.TryCreate.

Choc13:
I've had a crack at adding support for `System.Uri.TryCreate` here to follow on from #1572. I've not updated the tests yet, but just wanted to open this now for some very early feedback on whether or not I'm heading in the right direction here as this is my first time getting stuck into Fable internals like this.

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -2904,7 +2904,8 @@ let guids (com: ICompiler) (ctx: Context) (r: SourceLocation option) t (i: CallI
 
 let uris (com: ICompiler) (ctx: Context) (r: SourceLocation option) t (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
     match i.CompiledName with
-    | ".ctor" -> Helper.LibCall(com, "Uri", "default", t, args, i.SignatureArgTypes, isJsConstructor=true, ?loc=r) |> Some
+    | ".ctor" -> Helper.LibCall(com, "Uri", "Uri.create", t, args, i.SignatureArgTypes, ?loc=r) |> Some
+    | "TryCreate" -> Helper.LibCall(com, "Uri", "Uri.tryCreate", t, args, i.SignatureArgTypes, ?loc=r) |> Some
     | "UnescapeDataString" -> Helper.LibCall(com, "Util", "unescapeDataString", t, args, i.SignatureArgTypes) |> Some
     | "EscapeDataString"   -> Helper.LibCall(com, "Util", "escapeDataString", t, args, i.SignatureArgTypes) |> Some
     | "EscapeUriString"    -> Helper.LibCall(com, "Util", "escapeUriString", t, args, i.SignatureArgTypes) |> Some
--- a/src/fable-library/Uri.ts
+++ b/src/fable-library/Uri.ts
@@ -41,20 +41,6 @@ export class Uri {
       case UriKind.Absolute:
         return Uri.isAbsoluteUri(uri)
           ? ok(new Uri({ value: new URL(uri), kind }))
-          : uri.startsWith("/")
-          ? uri.startsWith("//")
-            ? ok(
-                new Uri({
-                  value: new URL(
-                    `file://${uri.substring(
-                      Array.from(uri).findIndex((c) => c !== "/"),
-                      uri.length
-                    )}`
-                  ),
-                  kind,
-                })
-              )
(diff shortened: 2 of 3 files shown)


src/Fable.Transforms/Replacements.fs
```diff
@@ -2904,7 +2904,8 @@ let guids (com: ICompiler) (ctx: Context) (r: SourceLocation option) t (i: CallI
 
 let uris (com: ICompiler) (ctx: Context) (r: SourceLocation option) t (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
     match i.CompiledName with
-    | ".ctor" -> Helper.LibCall(com, "Uri", "default", t, args, i.SignatureArgTypes, isJsConstructor=true, ?loc=r) |> Some
+    | ".ctor" -> Helper.LibCall(com, "Uri", "Uri.create", t, args, i.SignatureArgTypes, ?loc=r) |> Some
```
Choc13, line 2907: I'm particularly interested in feedback on this bit as it's the most foreign to me. Specifically, is it OK to call the static `Uri.create` method in JS even though this is a constructor in .NET. The reason I made it a static create method in the JavaScript implementation was because it allowed me to better share the creation logic with `TryCreate` and also allowed me to make better use of TypeScript's type system to eliminate some invalid parameter combinations.

src/fable-library/Uri.ts
```diff
@@ -1,127 +1,208 @@
...
   RelativeOrAbsolute = 0,
   Absolute = 1,
   Relative = 2,
 }
 
+type State =
+  | { value: URL; kind: UriKind.Absolute }
+  | { value: string; kind: UriKind.Relative | UriKind.RelativeOrAbsolute };
+
+type Result<T> = { tag: "ok"; value: T } | { tag: "error"; error: string };
```
Choc13, line 13: I'm wondering whether this construct (or similar) is defined else where in here so that I don't need to define this internal helper type just this usage.

src/Fable.Transforms/Replacements.fs
alfonsogarciacaro: The Replacements module and fable-library are still a bit messy. It's great that you managed to find your way through it! :clap: Ideally we want to progressively update it so most of the code is in F#, but given the Uri module is still in Typescript and create/tryCreate is similar to parse/tryParse in other modules, your implementation looks good to me 👍

Choc13: Thanks. I'll crack on and write the tests too then.

src/fable-library/Uri.ts
alfonsogarciacaro: The Result type is defined now in F#... in the Choice.fs file :/ So it's ok to have your custom `Result` here internally :+1:

Choc13: Cool, just wanted to check that there wasn't another internal `Result` defined somewhere else for the typescript stuff.

Choc13:
I think I've got this covered with tests now and I was able to get them to all pass locally. I believe it's testing from both JS and .NET because I had them working for mocha but then they were failing when run with .NET runner for a while. However, let me know if that's not the case as I'm not that familiar with writing tests that are run from both Fable and .NET.

Choc13:
Ah I see what the test errors are here. Unfortunately `Uri.TryCreate("/test", UriKind.Absolute)` works differently depending on the platform. For Windows it's an error, but for mac/Linux it assumes it's a file path and adds "file://" to the front. I guess we need to decide which platform to mimic here. I'm tempted to say Windows because I think as far as usage in Fable goes it's unlikely people will be trying to refer to file Uris without specifying them explicitly with the file protocol as in `Uri.TryCreate("file://test", UriKind.Absolute)`. 

What do you think?

Choc13:
I've fixed the tests so they work on all platforms. Basically, I've chosen to emulate the Windows behaviour and omit testing the edge case of creating an absolute URI from a string with a leading `/`, e.g. `Uri.TryCreate("/test", UriKind.Absolute)`.

src/fable-library/Uri.ts
alfonsogarciacaro: Now that I think of it, maybe we could put it in `Types.ts` with a note saying it's only an internal type. But I can do that :+1:

alfonsogarciacaro (approved):

alfonsogarciacaro:
Sorry @Choc13, I just realized I missed your previous comment last week. Thanks a lot for this! I will push a new release tomorrow after netconf :+1:

Issue #2462, opened 2021-06-03
Php Support for Fable


thinkbeforecoding:
### Description

Currently, the Import node in AST contains only the path and selector of the target function. This is fine for JS and Python that have a similar import system.
In those languages, files are considered as module that contains exported functions, and a user of the module just has to add a directive like this to make use of this module function:
``` js
import modulefunction as localfunction from "module";
```
The module system provides an isolation to avoid name clashes when importing several modules.

This doesn't exist in Php. In php, there is a include/require/require_once directive that just "includes" the content in the scope:
``` php
require_once('module.php');
```

To avoid name clashes, Php has a namespace mechanism. The module.php file for instance can start with a namespace directive to scope all the definitions to the namespace:
``` php
namespace MyModule;

function myfunction() { return 0; }
```
A user of the module has to import it using require_once, but must then know the name of the namespace to use the functions:
``` php
require_once('module.php');

$x = \MyModule\myfunction();
```

To avoid repeating module prefixes, it is possible to define an alias:

``` php
require_once('module.php');
use \MyModule\myfunction;

$x = myfunction();
```
or even 
``` php
require_once('module.php');
use \MyModule\myfunction as myfunc;

$x = myfunc();
```

But in both cases, knowning the namespace is mandatory.

Moreover, Php 7.2+ has a decent class construct, and it would be more idiomatic to compile class members as such. Fable today flattens all members as functions in the global scope. The Php transform could promote those functions as static or instance members. But the lack of information in Import make it impossible to generate the call:
``` fsharp
module MyModule

type MyType(x: int) =
      member this.Value = x
      static member DefaultValue() = MyType 0

let increment x = x+1
```
Will generate imports to selectors like "MyType__Value" , "MyType__DefaultValue". But the call in php should look different:
``` php
$v = $instance->Value(); // or ->get_Value();
$d = \MyModule\MyType::DefaultValue();
$x = \MyModule\increment(1);
```
There is currently now way to make the distinction.

### Repro code

This can be observed when transpiling fable-library to Php

thinkbeforecoding:
It seems this is also request for Python to support classes. @dbrattli

dbrattli:
Yes, it would be great to be able to generate more idiomatic Python by having (static) members inside the class definition. I don't think it's a blocker for Python but would make it possible to generate nicer code.

thinkbeforecoding:
[...]

PR #2509, opened 2021-08-21
[WIP] Lua Language support

alexswan10k:
early early wip


alexswan10k:
Hi @alfonsogarciacaro, @dbrattli, @thinkbeforecoding , been following your work closely with respect to Fable extended language support for Python, Php, Typescript, Dart etc. I thought for laughs I would have a crack at a Lua provider for.. reasons.. :) I originally was deliberating Rust but I figured I would start with something "easier". As you can imagine, I am struggling a little so any help/advice would be appreciated. I haven't got very far yet but here is where I am at!

* Firstly which AST should I be basing this on? I noticed that the babel AST is probably closest to what I want, but the extended AST seems to be the preferred approach. The dart transpiler seems to be incomplete, however, which is the only consumer of this. I noticed the Python and Php providers basically build their own AST. Should I be doing this? Lua seems to have a lot in common with js in terms of the fact it is not expression based, although it is missing some higher level bits such as restructuring etc.
* What is the best approach for standard library compilation? Is this also something I can defer as this is presumably a non-trivial exercise. I understand that a lot of the modules are already in f# so if theory if they can be transpiled, you get them for free. I appreciate there may need to be some language specific libraries for integration such as the python example.
* How do I go about setting up the testing project? I have been experimenting with the QuickTest, which is nice to get a sample, but is obviously a far-cry from real tests. I imagine the outputted .lua files will need to be executed on the actual lua runtime.
* Debugging the AST? I was directed to https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQ2wAuYVYAEZhmYALxgAOgCdJyCBGEBqEWAA%2BAPjCIA5slbTYMuQrAAWAEyQQAXyA

I have also read https://github.com/fable-compiler/Fable/issues/2462 and https://github.com/fable-compiler/Fable/issues/2339 but I am aware this is a moving target and things might have changed.

Anyway, any feedback/guidance would be great. Happy to raise a ticket to discuss architectural details etc. Happy to have a crack at this if it is helpful/interesting, although I do have a day job so cannot really commit to anything more. Thanks!

thinkbeforecoding:
Speaking about php, the AST is a way to represent the Php code in a structured way before serialization to text.
The Php transpiler converts the fable AST to Php AST. For php, the bable AST had to much things converted to JS, so a loss of information important to produce better php.

dbrattli:
Very nice and welcome to the club! 🎉 

I actually started with `Babel2Python.fs` but now use a `Fable2Python.fs`. Using Babel was much easier in the beginning, but I got into some hard problems after a while related to e.g member/property access (`obj.value`) vs indexed (`obj["value"]`). Working with the Fable AST gives you more control, and you also know better what the original purpose was and it's possible to output more Python like code. 

For standard library you need to start with a minimal `fable-library-lua`. The code will generate imports to this by itself and you need to rewrite them to match your language ([`rewriteFableImport`](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Transforms/Python/Fable2Python.fs#L429). Start minimal and add to the equivalent of e.g `types.py` and `string.py` one function at a time.  For the Python standard library I'm now making an external NuGet package `Fable.Python`. 

For testing I decided to make a separate test project for Python, starting with a single unit-test and then adding one by one. Start with e.g `TestString.fs` and get a single test running. For Python, the tests first runs as F# using Xunit, then use `pytest` for testing the generated Python code. Pytest expects tests to be functions named `test_` which was easier to generate with XUnit than Expecto. I might reconsider to use the original Expecto tests when the code is more mature.

alexswan10k:
Thanks, both, appreciate the input. I figure I probably am going to be creating my own intermediary AST, I guess I was just hoping there would be a shortcut.

I am going to start by cloning the babel AST, transforms (which I have already experimented with a little), and start modifying it to fit the language features.

alexswan10k:
Got a few trivial test cases up and running, was quite painful getting that first pass done, but it is definitely getting easier now the foundations are in.

Lots of good examples from the other languages, so I have tried to copy existing patterns where they made sense etc. Let me know if things are not generally correct though, drastic changes are still possible!

There is something weird going on with my import paths in the library dir. Do I need to add an extra transform to the paths or something?

alfonsogarciacaro:
> It sounds like Lua might actually be a good candidate to test the statement-focused intermediary AST you are working on @alfonsogarciacaro. Perhaps I should hold off, and try to integrate this, rather than plough on with the current approach?

Not sure yet when I'll be able to finish a prototype, but if you're not in a hurry it could be a good idea to wait so se can evolve the common AST together to target several languages :+1:

alexswan10k:
I am really not in a hurry so more than happy to explore this. This is kind of a learning exercise/experiment for me anyway. 

I am getting to a point where quite a moderate amount of the basic language features are there, so perhaps it would be a good thing soon to compare and contrast, with the intent the Lua AST can potentially be completely swapped out. I haven't started on the libraries yet, as there are some fiddly bits around module loading, namely, relative paths do not work in the traditional sense. Would it be worth opening a ticket for language support to track these details, along with steps to get to mvp?

On a side note, I would love to look at Rust longer term, I wonder if this is another language that can be taken into account for a general AST. I think the big divergence with Rust, is it is basically an ML family language, so it is probably closer to F# than any of the current compilation targets (in the sense it is an expression language over a statement language). I know the borrow checker is going to be a huge hurdle also, but I have had some ideas around patterns, auto-cloning, wrapping stuff in Rc/Arc/RwLock when shared usage can be inferred. Would be good to discuss if people are interested in this. An unmanaged target would be really exciting for F# (same with nelua).

ncave:
@alexswan10k There is already an ongoing effort to add Rust, see [here](https://github.com/ncave/Fable/pull/4). It needs to be retargeted to `beyond` branch and made into an official PR. Rust AST and printer are already done, now it's just the slow trudge to map features. Any contributions would be very welcome, even if it's just an opinion.

alexswan10k:
@ncave this is super exciting, how did I miss this! Let me have a good look through when I get time, and see if there is anything I can help with.

alexswan10k:
So I have hit a bit of a roadblock, not sure how to proceed.

Lua can accept relative modules, but these cannot be outside of the root directory of the entrypoint file unless added to the path (they can only be subdirectories). In the libraries example I have fudged this by copying the libraries into the tests output directory as a subfolder. This seems to work for trivial cases, but relative paths to subsequent imports are then all wrong.

The biggest problem seems to be coming from the fact I am importing files that are outside of the project dir (to share common code for things like Array.fs/List.fs, same as the Python example). This creates weird output paths because I think it is trying to calculate the relative imports based on where it is on disk:

```
<Compile Include="Native.fs" />
<Compile Include="../../fable-library/Choice.fs" />
```

Any ideas on what the best approach is here?

ncave:
@alexswan10k
The Fable-generated code either stays co-located with the source files (when no `--outDir` is used), or
when `--outDir` is used, the way imports work is to copy all F#-transpiled (project) dependencies under the target `outDir`, retaining the folder structure to the extent possible relative to the `outDir` but below it).
E.g. this assumes a separate step to copy the `fable-library` under `outDir`.

Perhaps I've misunderstood what the issue is, are you trying to share common source code (library) between projects?
Can you show an example of an import that would be a problem, or perhaps a small test project sample?

alexswan10k:
Shared modules such as Array/Seq, that is exactly it. The path is outside of the project (in order not to have to copy the file) so the import path will also be outside. This is normally what you want, but I suppose in the context of a library you want it almost to redirect that path to the same folder. I can do this with regex but its a bit messy.

I will try and get an example up if I can get one of them working.

ncave:
@alexswan10k Perhaps you just need to modify the [existing step](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Cli/ProjectCracker.fs#L486-L494) that copies the `fable-library` into the `outDir` for your language. That's what Fable normally does, it copies dependencies into the `outDir`.

voronoipotato:
I might give this a try, this is HUGE in terms of getting to use F# for game dev in several programming environments. Lua is a staple in modding / game dev so I'll be watching this. Thanks for all your work!

alexswan10k:
Hi @voronoipotato glad this interests you.

Unfortunately, this is probably not quite in a useable state yet. I have every intention to come back to this at some point, but there have been quite a few things distracting me currently! It was my thought also that for game modding, this would be a big win.

The big-ticket items that still need attention are:
* Merge from beyond (some structural changes have happened and need resolving)
* Figure out how to support different Lua versions, or pick one. The latest officially supported is 5.4.4, but a lot of games/apps use older versions such as 5.2. This is somewhat problematic because quite a few operators are not supported and need to be stubbed, and not 100% sure there is even backward compatibility here. 
* Add missing support for IEnumerable/classes etc, which will allow the list/array modules to be built out.
* Rough edges around path imports

voronoipotato:
Definitely item two can be addressed by just doing whatever is useful to you and stubbing out a version flag. Getting anything is better than nothing. 
Item 3 is nice to have but I'm okay with a mvp without it.

alfonsogarciacaro:
Sorry @alexswan10k! This PR got closed when I deleted `beyond` branch and now GH gives me an error when I try to retarget it. If you want to continue working on this, could you please send a new PR to `snake_island` branch? Sorry for the hassle!

alexswan10k:
No worries. This needed quite a bit of work anyway to be mergeable. There were a couple of quite fundamental issues that needed addressing first such as exactly what target version to use and how it would work with imports etc. I will come back to this at some point and raise a new PR, probably best for me to focus on Rust for now though and help get at least something over the line 😀

voronoipotato:
can we reopen this though so that we have a placeholder for it? I almost lost it.

smoothdeveloper:
This target would be awesome for targetting Repear / Reascript (mostly done in lua): https://www.extremraym.com/cloud/reascript-doc/

There is some chance that https://typescripttolua.github.io/ could help in meantime of a proper Fable2Lua target, but it doesn't succeed in compiling most of the typescript target for fable_modules into lua as a starter.

image

MangelMaxime:
It seems like Reascript has a Python API.

As Python is maintained and Lua target is kinda dead, perhaps it is worth trying to consume the API using Python instead.

smoothdeveloper:
@MangelMaxime the python API is not embedded in Reaper runtime:

> Python must be downloaded and installed separately (see [requirements](https://www.reaper.fm/sdk/reascript/reascript.php#reascript_req_py)) and as such, will never be as portable between users as EEL or Lua. Within REAPER, Python performance is not as good as EEL or Lua, and does not offer any UI or graphic features. However, Python does benefit from a vast number of third-party resources, libraries, tutorials, and code examples.

https://www.reaper.fm/sdk/reascript/reascript.php

It is hence limited for the context of code that needs to run fast in reaper, on top of the deployment headache.

The other route is rust bindings for the C API: https://github.com/helgoboss/reaper-rs, which I should give a look at but would not be as trivial as simple lua files to deploy (needs to build binaries for each target).

Issue #2093, opened 2020-06-14
Async.Sequential does not exist


0x53A:
In the repl:

![image](https://user-images.githubusercontent.com/4236651/84596048-a31c0180-ae5b-11ea-8132-d7498696120d.png)

TerenceHinrichsen:
I have the same - project compiles, but warning:
"export 'sequential' was not found in '../xxxxxxxx/.fable/fable-library.2.13.0/Async.js'

MangelMaxime:
Hum, I think that `Async.sequential` is not implemeted in fable-library because when searching for `sequential` in [Async.ts](https://github.com/fable-compiler/Fable/blob/master/src/fable-library/Async.ts) it yield zero results.

I guess we miss either implementing it or adding the standard Fable message: `Fable can't find replacements for ....`

PR #2515, opened 2021-08-24
add Async.Sequential

0x53A:
fixes https://github.com/fable-compiler/Fable/issues/2093

### DIFF
--- a/src/fable-library/Async.ts
+++ b/src/fable-library/Async.ts
@@ -15,6 +15,11 @@ function emptyContinuation<T>(_x: T) {
   // NOP
 }
 
+// see AsyncBuilder.Delay
+function delay<T>(generator: () => IAsync<T>) {
+  return protectedCont((ctx: IAsyncContext<T>) => generator()(ctx));
+}
+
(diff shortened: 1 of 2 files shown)


0x53A:
Both the existing Async.Parallel and this new Async.Sequential differ from .NET in that they are eager - they immediately start their computations.

Is Async being cold and lazy something that should be preserved in Fable, or does it not really matter?

Looking at the other code, it might be enough to wrap 'em in ``protectedCont`` or something similar.

alfonsogarciacaro:
Thanks a lot for this @0x53A! We made promise computation expressions to match native JS behavior but given that `Async` is F# only I would try to keep the same behaviour as in .NET, so it'd be great if Async.Parallel/Sequential can be fixed. Do you want to try it in this PR or can I merge already?

0x53A:
I am not deeply familiar with either the Async machinery specifically or even computation expressions in general, but ``AsyncBuilder.Delay`` should be exactly what we need, right?

I pushed one commit, if that isn't correct, then I'd rather merge-as-is and take a good look when I have time (or let someone else do the honors).

alfonsogarciacaro:
Looks good to me! Thanks again for this!

PR #2516, opened 2021-08-24
Fix #2505 - String.Split(null)

### LINKED ISSUE #2505: Stirng.Split.(null) throws exception, should return argument string
Example code, tested in the Fable REPL.
```fs
open Browser.Dom

try 
    let x = "ABC".Split(null)
    console.log($"{x}")
with
    |x -> console.log($"{x}")
```

should print "ABC" to the console but instead gives exception `str is null`.

.NET docs specify that `null` is a valid argument for String.Split, and this is verified in a .NET fiddle:

```
using System;
					
public class Program
{
	public static void Main()
	{
		Console.WriteLine("{0}", "ABC".Split( (char[]) null, StringSplitOptions.RemoveEmptyEntries) );
	}
}
```

This was encountered in the port from `FSharp.Formatting.Markdown` to Fable:

```fs
        let splits = header.Split((null : char array), StringSplitOptions.RemoveEmptyEntries)
        match splits |> Seq.tryFind (fun _ -> true) with
        | None -> "", ""
        | Some langString ->
            let ignoredString = header.Substring(header.IndexOf(langString) + langString.Length)
            langString, if String.IsNullOrWhiteSpace ignoredString then "" else ignoredString
```

The idea appears to be to turn "" into `[| |]` and everything other value "X" into `[| "X" |]`. I can work around this code for now in Fable.

jwosty:
Fixes #2505.

``"foo bar".Split(null)`` should behave the same as ``"foo bar".Split()``, as it does in .NET.

### DIFF
--- a/src/fable-library/String.ts
+++ b/src/fable-library/String.ts
@@ -469,7 +469,7 @@ export function split(str: string, splitters: string[], count?: number, removeEm
       splitters[key - 1] = arguments[key];
     }
   }
-  splitters = splitters.map((x) => escape(x));
+  splitters = splitters.map((x) => escape(x ?? " "));
   splitters = splitters.length > 0 ? splitters : [" "];
   let i = 0;
   const splits: string[] = [];
--- a/tests/Main/StringTests.fs
+++ b/tests/Main/StringTests.fs
@@ -515,6 +515,8 @@ let tests =
             |> (=) [|"a";"b";"c";"";"d"|] |> equal true
             "a b c  d ".Split()
             |> (=) [|"a";"b";"c";"";"d";""|] |> equal true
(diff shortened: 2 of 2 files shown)


alfonsogarciacaro:
Thanks!

inosik:
This should also split by other whitespace characters, see [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=net-5.0#remarks) for reference:

>  If no delimiting characters are specified, the string is split at white-space characters.

@jwosty Care to change this to the regex pattern `/\s/`? Otherwise I'll submit a PR.

jwosty:
@inosik good catch

PR #2518, opened 2021-08-25
Make String.Split split on all whitespaces by default to mirror .NET

jwosty:
``String.Split()`` and ``String.Split(null)`` are supposed to split on *any* unicode whitespace character (\s, \t, \r, \n, EM space, etc) according to the .NET implementation.

See https://github.com/fable-compiler/Fable/pull/2516#issuecomment-905198074

### DIFF
--- a/tests/Main/StringTests.fs
+++ b/tests/Main/StringTests.fs
@@ -515,8 +515,20 @@ let tests =
             |> (=) [|"a";"b";"c";"";"d"|] |> equal true
             "a b c  d ".Split()
             |> (=) [|"a";"b";"c";"";"d";""|] |> equal true
+            "a\tb".Split()
+            |> (=) [|"a";"b"|] |> equal true
+            "a\nb".Split()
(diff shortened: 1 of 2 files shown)


tests/Main/StringTests.fs
```diff
@@ -515,8 +515,20 @@ let tests =
...
+            |> (=) [|"a";"b"|] |> equal true
+            "a\rb".Split()
+            |> (=) [|"a";"b"|] |> equal true
+            "a\u2003b".Split() // em space
+            |> (=) [|"a";"b"|] |> equal true
             "a b c  d".Split(null)
             |> (=) [|"a";"b";"c";"";"d"|] |> equal true
+            "a\tb".Split(null)
+            |> (=) [|"a";"b"|] |> equal true
+            "a\u2003b".Split() // em space
```
inosik, line 530: This test case is duplicated. Did you mean to do this instead?

```suggestion
            "a\u2003b".Split(null) // em space
```

PR #2523, opened 2021-09-01
Rust language support - MVP

ncave:
This PR adds support for the Rust language.
Still a very early prototype, lots of work ahead:

- [x] Rust AST
- [x] Rust AST Printer
- [x] `Fable2Rust` MVP: 
  - [x] imports
  - [x] modules
  - [x] primitives
    - [x] bool
    - [x] char    
    - [x] string
    - [x] integers
    - [x] floats
  - [x] control flow
  - [x] functions
  - [x] lambdas
  - [x] records
  - [x] unions
  - [x] arrays
  - [x] lists
  - [x] tuples
  - [x] options
  - [x] sequences
  - [x] enums (some)
  - [x] classes (some)
  - [x] interfaces (some)
- [x] `fable-library-rust` MVP:
  - [x] Array (some)
  - [x] List (some)
  - [x] Seq (some)
  - [x] Option (some)

-----
Post-MVP (i.e. not in this PR):

- `Fable2Rust`:
  - enums (full)
  - classes (full)
  - interfaces (full)
  - exceptions
  - conversions
  - comparers
  - formatting
  - reflection
  - async
  - threading
- `fable-library-rust`:
  - Array
  - ResizeArray
  - List
  - Seq
  - Option
  - Map
  - Set
  - Dictionary
  - HashSet
  - Decimal
  - BigInt
  - Guid
  - DateTime
  - DateTimeOffset
  - Timer
  - String
  - Unicode
  - RegExp
  - Choice
  - Event
  - Observable
  - other `FSharp.Core` modules
  - other `.NET BCL` implementations


kerams:
What's your goal here? Just fun? This could be very interesting but also incredibly difficult, what with the borrow checker and other things that would be hard to expressin F# (traits, etc.).

ncave:
@kerams Yes, just for fun. For now the plan is to use let binding for locals, and [rust-gc](https://github.com/Manishearth/rust-gc) for (potentially circular) references. There is no need to express all the Rust features, as it is not meant to write idiomatic Rust, just transpile F# code. If you need to map the other direction (Rust to F#), I use a few simple regex [replacements](https://github.com/ncave/rust2fsharp), that gets me there 95%.

alexswan10k:
@kerams, I agree the borrow checker is the biggest nut to crack. I think there may be things we can do (on a sliding scale of complexity).

As for traits, my thoughts are that some traits such as Copy, Clone, Debug, should perhaps even be invisible to the F# world. The compiler needs to keep track of them to decide what to do when passing things over function boundaries. Traits to some extent can be approximated by interfaces however, although there are things traits can do that interfaces cannot model.

I have a few ideas around the more complex borrow checker solutions, but they get quite complex. and I need to think about how to convey them. Here is a rough go at it: 
The bottom line though is you can start with a dumb copy everywhere, then move to references in, ownership out for function calls. Further to that you could then begin tracking multiple borrows and implicitly wrapping things (and unwrapping where necessary) in Rc or Arc depending on the context. The big one then is move and multithreading, which presumably can be solved with tracking, and again wrapping in a RWLock. I kind of feel like we could start really dumb (copy everywhere) and just gradually work up, as ultimately this is an optimization.

Interop with external code though is another story with this approach!

Anyway this is really cool. Maybe when you have the test harness up I can help you write a few tests and flesh out the std library.

ncave:
@alexswan10k Added Rust test harness. Please note that `Fable2Rust` is still a WIP, so a lot of features are not there yet.
To run tests: `npm run build test-rust`

ncave:
@alfonsogarciacaro @alexswan10k 
This is ready for review and merge, 300 tests passing. Obviously lots of work ahead, but let's see if we can get this in so we have a common base for all languages so far.

alexswan10k:
@ncave great work. Apologies have been busy with a few other things recently, but if there is anything I can do to help get this over the line, happy to help. I will pull and run locally again too.

alfonsogarciacaro:
Thanks a lot for this @ncave! Changes in the "common" files seem to be minimal so I'm ok with merging this (it's hard for me to review Rust-specific code as my knowledge of Rust is limited), we can see later if there's a better way to modularize each language target :+1:

Is there any showcase for this? For example, an F# console app that's compiled to Rust and takes much less space that a self-contained .NET app (and/or it's more performant). Can this be used to compile F# to WebAssembly using Rust toolchain?

ncave:
@alfonsogarciacaro 
> Is there any showcase for this? For example, an F# console app that's compiled to Rust and takes much less space that a self-contained .NET app (and/or it's more performant). Can this be used to compile F# to WebAssembly using Rust toolchain?

Not yet as it's not feature-rich enough for general purpose code, but anecdotally numeric-heavy code like [NBody](https://github.com/ncave/Fable/blob/684c11758e57a210b7af2c2fae3e45343cde56ea/tests/Rust/tests/src/NBodyTests.fs) runs at native speed, and should be able to compile to WebAsm (but I haven't tried yet).

It will be great if we can somehow have an alpha Repl for `beyond` with a language toggle, that gets re-deployed on merge in `beyond`, to showcase the different languages.

ncave:
@alfonsogarciacaro Here is a [small demo](https://github.com/ncave/fable-raytracer) to showcase the Rust support in Fable 4.

Just a few notes on performance (for this specific demo):
- F# to Rust to native is faster than F# on .NET 6.0 (as it should be, since this is numeric-heavy code).
- F# to Rust to webasm is faster than F# to JS (as expected, normally webasm is at least 30% faster than JS).

alfonsogarciacaro:
Thanks a lot for this @ncave! It's amazing to see the same F# code compiled to different platforms and compare how it behaves 🙌  I got some interesting results so I will share them with you:

- `npm run test-js`: Around 1470ms
- `npm run test-wasm`: Around 1150ms
- `npm run test-rust`: Around 1300ms
- `npm run test-dotnet`: Around 900ms

So in my case the baseline is dotnet and rust is even slower that wasm :thinking: This makes me think that something may be wrong with my Rust installation. I'm running Windows on a Core i7 11th gen CPU, maybe net6 is more optimized than Rust for these settings? `npm run test-web` is consistent with Node and also with the Raytracer in the sample when rendering an image of size 1024x1024px. As you said, the wasm version is around 30% faster (actually somewhat less in my case).

BTW, I also got this error when running `dotnet tool restore`, very weird. I had to install Fable in an upper directory.
```
/tmp/fnywtra5.yea/restore.csproj : error NU1101: Unable to find package fable. No packages exist with this id in source(s): dotnet-experimental, dotnet-public

Package "fable" failed to restore, due to Microsoft.DotNet.ToolPackage.ToolPackageException: The tool package could not be restored.
   at Microsoft.DotNet.Tools.Tool.Install.ProjectRestorer.Restore(FilePath project, PackageLocation packageLocation, String verbosity)
   at Microsoft.DotNet.ToolPackage.ToolPackageInstaller.InstallPackageToExternalManagedLocation(PackageLocation packageLocation, PackageId packageId, VersionRange versionRange, String targetFramework, String verbosity)
   at Microsoft.DotNet.Tools.Tool.Restore.ToolRestoreCommand.InstallPackages(ToolManifestPackage package, Nullable`1 configFile)

Restore failed.
```

ncave:
@alfonsogarciacaro

Sorry about the error, I fixed the `nuget.config`, hopefully that helps.

The first run for `test-rust` specifically is always slower, try running it several times and pick the best time.
Same for the other tests, they're CPU-bound so they are heavily dependent on current CPU load. Results can vary between runs if you have other things going on your machine.

Here are my results on an older machine (so they're only comparable to themselves):
(Ubuntu 20.04 on WSL2, Node.js 17.3, .NET 6.0.1) 
```
test-js     - 3025 ms
test-web    - 1800 ms
test-wasm   - 1805 ms
test-rust   - 1120 ms
test-dotnet - 1262 ms
```

The differences will get more pronounced if you render at 2048x2048, but still more or less proportional.

Side note:
I tried the (still experimental) .NET 6 AOT to see how far it has progressed, as I've gotten good results with it in the past (e.g. it cut the FCS time something like 30-40%, despite the huge binary it produced). But at least for this particular demo the .NET AOT native speed was not greater than jitted .NET, and the binary sizes are still enormously gigantic. In this regard the Rust native binaries (and webasm) are much smaller and can be trimmed further with options.

alfonsogarciacaro:
Thanks for fixing the nuget.config @ncave! For some reason I had trouble installing rust in WSL so I tried in a partition where I have Ubuntu installed and these are the best times I got after some runs (above were more or less averages):

```
Ubuntu 20.04, Node.js 16.13, .NET 6.0.1, rustc 1.58.0

test-js     - 1595 ms
test-web    -  829 ms (Chrome)
test-wasm   -  812 ms
test-rust   -  567 ms
test-dotnet -  887 ms
```

Rust result is utterly impressive now, this is fantastic work @ncave! We can envision now a world where you design your algorithm using the Python interpreter, and then compile it to rust/wasm to get the best performance in the deployment (which may well be a Fable React app) :rocket: 

> Note: It seems rust compilation needs fable-library-rust to be in ../Fable/build/fable-library-rust/

ncave:
@alfonsogarciacaro

> Note: It seems rust compilation needs fable-library-rust to be in ../Fable/build/fable-library-rust/

I just pre-compiled and checked in `fable-library-rust` to lower the barrier for testing, so you don't need Fable just to test.
To clarify, this is just a transpiler demo, it's not meant to be performant or use good practices, that's not the point.
Rust support is still a work in progress, just wanted to have something out that showcases it. Thanks for testing!

delneg:
Just for the reference, now it's possible to do some interesting stuff like this - fable raytracer compiled to a static lib & imported into iOS application 
https://github.com/delneg/fable-raytracer-ios-net6
Most interesting bits are in https://github.com/delneg/fable-raytracer-ios-net6/blob/master/FableRustRaytracerNet6iOS/AppDelegate.fs

kant2002:
> I tried the (still experimental) .NET 6.0 AOT compiler to see how far it has progressed

@ncave how can I replicate AOT experiment of you. Definitely would like to check if there something which can be improved. Did you just add ILCompiler package, or it was something more involved?

ncave:
@kant2002 See some explanation [here](https://github.com/ncave/fable-raytracer/issues/1#issuecomment-1021227894).

Issue #2525, opened 2021-09-04
Should we support infinit arguments for curry/uncurry?


MangelMaxime:
Hello,

When using Fable.Form, someone reported an error about having an error if they go more than 8 fields.

https://github.com/MangelMaxime/Fable.Form/issues/22

Their issue has a workaround by rewrite the code a bit, but it is never fun to see this kind of "cryptic" errors.

It seems like it should be possible to support any number of arguments for curry/uncurry.

Should we try to do it?

https://javascript.info/currying-partials

alfonsogarciacaro:
I've sent a PR to fix this. Seems to be working although I'm not sure if it can have any impact on performance. Should we benchmark it first before merging? Also, we should probably add tests, it'd be nice to use the code reported in Fable.Form, though that would imply adding several dependencies to the test project 🤔

MangelMaxime:
> Seems to be working although I'm not sure if it can have any impact on performance.

I have the same interrogation ^^

> Also, we should probably add tests, it'd be nice to use the code reported in Fable.Form, though that would imply adding several dependencies to the test project thinking

Adding Fable.Form can be tricky because as you said, it is split in 3 different packages.

I manage to write a code in the REPL which looks similar to Fable.Form and generate curried function.

Also the code should be testable because we can just test the generated tuple. We should probably test arity 1, 2, 8 and 10 for example. 

```fs
let addString (value : string) (f : string -> 'T) =
    f value

let addBool (value : bool) (f : bool -> 'T) =
    f value


let res =

    let onSubmit =
        fun (arg1 : string)
            (arg2 : string)
            (arg3 : bool)
            (arg4 : string)
            (arg5 : string)
            (arg6 : string)
            (arg7 : string)
            (arg8 : string)
            (arg9 : string) ->

            (arg1, arg2, arg3)

    onSubmit 
        |> addString "name"
        |> addString "password"
        |> addBool true
        |> addString "dummmyValue"
        |> addString "dummmyValue"
        |> addString "dummmyValue"
        |> addString "dummmyValue"
        |> addString "dummmyValue"
        |> addString "dummmyValue"

printfn "%A" res
```

[...]

PR #2529, opened 2021-09-09
Add tests for curry utils (testing infinity arguments)

MangelMaxime:
Related to #2525

@alfonsogarciacaro I don't know how to generate code which use `uncurry` :man_shrugging: 

Also, I wasn't able to generate a test which generate code using `curry(1, ...)` but I don't think this is possible. I don't know when a function of 1 arguments would be curried.

### DIFF
--- a/tests/Main/ApplicativeTests.fs
+++ b/tests/Main/ApplicativeTests.fs
@@ -1403,6 +1403,98 @@ module FSharpPlus =
             sequence (seq [Some 1; Some 2]) |> equal expected
     ]
 
+module Curry =
+    let addString (value : string) (f : string -> 'T) =
+        f value
+
+    let addBool (value : bool) (f : bool -> 'T) =
+        f value
+
+    let tests =
+        [
+            // Test the smallest arity I managed to create
+            testCase "curried function of arity 2 works" <| fun _ ->
+                let res =
+
+                    let onSubmit =
+                        fun (arg1 : string)
+                            (arg2 : bool) ->
+
+                            (arg1, arg2)
+
+                    onSubmit
+                        |> addString "name"
+                        |> addBool true
+
+                equal ("name", true) res
+
+            // Test old maximum limit for curried functions
+            testCase "curried function of arity 8 works" <| fun _ ->
+                let actual =
+                    let onSubmit =
+                        fun (arg1 : string)
+                            (arg2 : bool)
+                            (arg3 : string)
+                            (arg4 : string)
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Ups, sorry! I should have told you 😅 Testing what caused the issue in Fable.Form should be enough but just in case, the "secret" rules of uncurrying in Fable is that functions are expected to be uncurried at certain points and not others (because I couldn't do it well). The most common cases are passing functions as argument (uncurried) and returning them (curried). So passing a function of arity-2 or higher as argument and returning it is a way to force currying at the moment ([repl](https://fable.io/repl/#?code=DYUwLgBAJgrgtnAnhAFAMwFwQJYDtIC0AfDvhMaWAJQQC8AUBExGvfaJGCAM5go0NmEDhDgwwAQwBGoCAA86EAAyNmItACZFsBMnQxc8iMgoKA1MaqqmCgDwEIARggB6FxADK2KCAgg0aCAAxpAAVjC8pBBBEty+YAD20Njc0rIJAA5g2HDYAF4S2Qm43NbC4CwAzIqaEAAsZXYOjmVo1ZVAA&html=DwCwLgtgNgfAUHUBTAhgE3gAm54ElgqbhgAOAtEgI4CuAlgG4C8A5AMID2AdmEj+QBUAnqSQtMAY268erXgA8wAenDQA3JJAoATgGcCTGmABm5ABwt4wFagwJgAIw5ohkqCl26mAIhSkKUjwodFxI2t5YOMBojJh0aD5IUBB0uiDkfqQR1jEMVkpOLvCIKpCwcEA&css=Q)).

> In principle functions of arity-1 shouldn't be curried or uncurried but it may happen in some situations (for example if it returns a generic because it can return a function) but I don't remember well.

MangelMaxime:
> Testing what caused the issue in Fable.Form should be enough

The tests I added are doing that.

The code you shown is testing the same thing as my tests I think. It is using `import { curry } from "fable-library/Util.js";`

What I don't know how to test is `import { uncurry } from "fable-library/Util.js";`. If you know how to test it, we can add a test for it just to be sure to test everything.

Otherwise, for me, this PR cover the bug reported in #2525

alfonsogarciacaro:
You could do the opposite to force uncurrying: get a function returned by another function and then pass it as argument to another function, but hopefully some cases (at least for arity less than 8) are covered by the current tests. We can add new tests later if necessary :+1: Thank you!

PR #2533, opened 2021-09-13
Allow import with a single argument

dbrattli:
Allow import with only the package. Selector defaults to "default" so `Import("my-package")` is the same as `Import("default", "my-package")`.  This is so we can import single modules in Python and looks much better than `ImportDefault("json")` since Python does not have the concept of a default import, and still make sense for JS as well.

Adding you to the PR since it touches outside the Python code. What do you think?

With this PR I can write much better looking Python bindings. e.g:

```fs
type IExports =
    abstract dumps : obj: obj -> string
    abstract loads : string -> obj

[<Import("json")>]
let json: IExports = nativeOnly
```

Full source for example: https://github.com/dbrattli/Fable.Python/blob/main/src/Json.fs

### DIFF
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -43,6 +43,9 @@ type GlobalAttribute() =
 type ImportAttribute(selector: string, from: string) =
     inherit Attribute()
 
+    /// Same as `Import("default", "my-package")`
+    new (from: string) = ImportAttribute ("default", from)
+
 /// Takes the member name from the value it decorates
 type ImportMemberAttribute(from: string) =
     inherit Attribute()
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -1472,6 +1472,8 @@ module Util =
                     ImportAtt("default", path.Trim()) |> Some
                 | Atts.importMember, [(:? string as path)] ->
                     ImportAtt(Naming.placeholder, path.Trim()) |> Some
+                | _, [(:? string as path)] ->
+                    ImportAtt("default", path.Trim()) |> Some
                 | _, [(:? string as selector); (:? string as path)] ->
                     ImportAtt(selector.Trim(), path.Trim()) |> Some
                 | _ -> None


alfonsogarciacaro:
Hmm, we already have `ImportDefault` that accepts one argument (and `ImportAll/ImportMember`). If we have a constructor for `Import` accepting one argument it can be confusing when targeting JS because it's not obvious if it will be translated as `import * as foo` or ` import foo` (many devs unused to JS imports already have trouble with this). I guess in Python there's no difference between import defaults or namespaces, so maybe you can just accept either. So for example you can write:

```fsharp
[<ImportAll("json")>]
let json: IExports = nativeOnly
```

> `ImportAll` sounds to me a bit semantically closer to the idea of importing a whole package, but you can also use `ImportDefault` if you prefer.

dbrattli:
@alfonsogarciacaro In Python `import json` is not importing the whole contents of the module, just making the module itself available. To import the full contents of the module you would need to write e.g `from json import *` which will import everything in the module and is closer to `ImportAll`.  That is why I didn't use `ImportAll`. There is no default import in Python so it will be very confusing to use `ImportDefault` for Python. That's why I need a `Import` with a single argument.  Perhaps we should split and move to e.g `Fable.Core.Python` instead so Python has it's own set of imports?

alfonsogarciacaro:
Yes, it probably makes sense. There's already a different `PyInterop` and Python bindings, and we cannot expect the imports to work the same for every language. so probably we will need to use different Fable.Core surfaces for each language. The only question I have is whether we should use different namespaces or just use `Fable.Core` for all Fable.Core.XXX packages. The reason for that would be that many mechanisms in FSharp2Fable are triggered by types in Fable.Core matched by name. So in your example above, you can add the match for Import with single argument but this wouldn't affect users of Fable.Core(.Js). It can also help keeping consistency of attributes that are indeed common like `Emit`.

dbrattli:
So do I understand you correctly that it would be different NuGets for each language, but that they use the same namespace i.e Fable.Core? But then how do the compiler load the right nuget for each language when running e.g Fable.Cli?

Happypig375:
What about the function version? `let json: IExports = importAll "json"`

dbrattli:
@Happypig375 Not sure how that would be in Python. You can write `from json import *` but you cannot write `from json import * as alias`. Thus for Python we should probably not use ImportAll as `... import *` so perhaps we should use it for module import as @alfonsogarciacaro suggested i.e `import module as alias`  and if module and alias is the same then it becomes `import module`. That could be the solution, to use importAll after all. Thanks.

alfonsogarciacaro:
Good that we have an alternative 😸 About having multiple Fable.Core.XXX packages using the same namespace, I think it shouldn't be an issue as far as you don't reference two Fable.Core.XXX packages at the same time. Not sure if that's an scenario that could happen. I assume that any library that's compilable to multiple languages should just use standard F# and not Fable.Core because the Fable tricks (import, emit, erase...) won't apply exactly the same across languages 🤔

MangelMaxime:
> I assume that any library that's compilable to multiple languages should just use standard F# and not Fable.Core because the Fable tricks (import, emit, erase...) won't apply exactly the same across languages thinking

I think that it can happen that you have to resort to `emit` etc. or write a specific function in native language because it is not possible or really hard to do in F#.

I suppose, it means that the library will be using compiler directives kind of like we do today when support Fable, .NETCore, .NETFramework, etc.

But yes, each language should have its specific function under their respective module `Fable.Core.JS`, `Fable.Core.Python`, etc.

dbrattli:
Thanks for great feedback! I'm closing this PR, and want to say that the discussion have helped me move the Python efforts one step forward with regards to interop with existing Python libraries and F# (Fable Python) nugets. Still a lot of work left, but  at least moving forward.

PR #2536, opened 2021-09-15
Fix isSubclassOf to walk up the inheritance chain

chkn:
This fixes `Type.IsSubclassOf` to work the same way as .NET. To repro the issue, run this snippet on [Fable REPL](https://fable.io/repl/#?code=C4TwDgpgBAYg9nAFASigXigYwDYEMDO+UEAdgCYBQokUAQrgE4rpQCWJAFhA68LAiirho9AF7MM7Ljz70myChTA8SwAGYkoAIjlsi+AK4AjHASJw1-OAH4oAUiNaoiahAsAeOQD4AdAEl8AGVjU0IAeTUXYQ94OC9kBWV2dU0dXFE9KEMTPEIoCytbBycoyA8xXwDgnLMI0rc1d1j45CA&html=Q&css=Q) and in FSI:

```fsharp
type Foo() = class end
type Bar() = inherit Foo()
type Baz() = inherit Bar()

printfn "Bar is subclass of Foo? %b" (typeof<Bar>.IsSubclassOf(typeof<Foo>))
printfn "Baz is subclass of Foo? %b" (typeof<Baz>.IsSubclassOf(typeof<Foo>))
```

FSI output:

```
Bar is subclass of Foo? true
Baz is subclass of Foo? true
```

Fable output:

```
Bar is subclass of Foo? true
Baz is subclass of Foo? false
```

### DIFF
--- a/src/fable-library/Reflection.ts
+++ b/src/fable-library/Reflection.ts
@@ -187,7 +187,7 @@ export function isEnum(t: TypeInfo) {
 }
 
 export function isSubclassOf(t1: TypeInfo, t2: TypeInfo) {
-  return t1.parent?.Equals(t2) ?? false;
+  return t1.parent != null && (t1.parent.Equals(t2) || isSubclassOf(t1.parent, t2));
 }
 
 /**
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Seems the typescript compiler still complains with the options set for the build ;) But I can do that, thanks a lot for the fix!

chkn:
Ah thanks! Sorry about the compiler errors.. it always happens when you think it's such a simple fix that you don't need to check :)

PR #2540, opened 2021-09-20
2097-async-start-child-timeout

njlr:
* Adds System.TimeoutException
 * Adds timeout to Async.StartChild


src/fable-library/SystemException.ts
```diff
@@ -0,0 +1,9 @@
+import { Exception } from "./Types.js";
+
+export declare class SystemException extends Exception {
+    constructor();
+}
+
+export declare class TimeoutException extends SystemException {
+    constructor();
+}
```
alfonsogarciacaro, line 9: Because we already have the F# version we probably can remove this one. It's likely that it just gets overwritten by the F# one during build.

alfonsogarciacaro (approved):

src/fable-library/SystemException.ts
njlr: I tried removing it but seems to be required by `Async.ts`

alfonsogarciacaro: Ah, I see, thanks! Well, then the next move would be to convert Async.ts to F# :)

alfonsogarciacaro:
Thanks!

PR #2544, opened 2021-09-23
issue-2530-system-collections-generic-stack

njlr:
* Adds Stack and various tests


Happypig375:
Dynamic resizing looks like ResizeArray's job.

njlr:
> Dynamic resizing looks like ResizeArray's job.

Many ways to do this I'm sure. This implementation is based on the one in [FCS](https://github.com/dotnet/fsharp/blob/main/src/fsharp/utils/prim-parsing.fs#L88-L111) and extended to match the `System.Collections.Generic.Stack` API. If there's a reason to leverage `ResizeArray` I can update it.

src/Fable.Transforms/Replacements.fs
```diff
@@ -2474,6 +2474,17 @@ let hashSets (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr op
...
+let stacks (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
+    match i.CompiledName, thisArg, args with
+    | ".ctor", None, args ->
+        let moduleName, mangledName = getMangledNames i thisArg
+        Helper.LibCall(com, moduleName, mangledName, t, args, i.SignatureArgTypes, isJsConstructor=true, ?thisArg=thisArg, ?loc=r) |> Some
+    | ("Push" | "Pop" | "Peek" | "get_Count" | "Clear" | "Contains" |
+        "TryPeek" | "TryPop" | "ToArray" as meth), Some c, args ->
+        let moduleName, mangledName = getMangledNames i thisArg
+        Helper.LibCall(com, moduleName, mangledName, t, args, i.SignatureArgTypes, ?thisArg=thisArg, ?loc=r) |> Some
+    | _ -> None
```
alfonsogarciacaro, line 2486: We have the `bclType` function to automatically map  types from the BCL (see `System.Text.StringBuilder` for example). Maybe could be used here too?

alfonsogarciacaro (approved):

alfonsogarciacaro:
Yes, `ResizeArray` should work too, but it's a good idea to keep the implementation as close as possible to FCS to reduce maintenance. I've also got good results when preallocating arrays in JS and this may help when compiling to other languages too.

src/Fable.Transforms/Replacements.fs
njlr: Yep, `bclType` worked nicely!

alfonsogarciacaro:
Thanks!

PR #2545, opened 2021-09-23
Python module handling and project cracker fixes

dbrattli:
Fix project cracker for python so we include the fable library in a directory called `fable` in the output directory instead of `.fable`. This way we can import fable library code directly from the source files without having to install the library in some e.g (virtual) environment. Note, this is because Python modules cannot contain any dots.

Also add `OutputType` to the compiler and extract from project file since we need to make imports in Python relative for libraries and absolute for programs (Exe). This might be beneficial for other language translations as well. Please review this @alfonsogarciacaro @ncave so I know if this looks acceptable or not. Getting such project options available for the compiler was surprisingly hard.

All files with package references are compiled with the same compiler (options) even if they are from referenced project files (in e.g nuget packages). Thus such files would normally also get the OutputType of the main project. This PR fixes so that all code in `fableHiddenDir` will be compiled as "Library" even if the main project is "Exe".


src/Fable.Cli/ProjectCracker.fs
```diff
@@ -459,17 +471,20 @@ let createFableDir (opts: CrackerOptions) =
...
-            failwith ("Source directory is missing: " + source)
-        let source = source.TrimEnd('/', '\\')
-        let target = target.TrimEnd('/', '\\')
-        for dirPath in IO.Directory.GetDirectories(source, "*", IO.SearchOption.AllDirectories) do
-            IO.Directory.CreateDirectory(dirPath.Replace(source, target)) |> ignore
-        for newPath in IO.Directory.GetFiles(source, "*.*", IO.SearchOption.AllDirectories) do
-            IO.File.Copy(newPath, newPath.Replace(source, target), true)
+        copyDir source target
 
 let copyFableLibraryAndPackageSources (opts: CrackerOptions) (pkgs: FablePackage list) =
```
dbrattli, line 489: @alfonsogarciacaro If you think this is too messy with all the language matches, I could perhaps make a separate `copyFableLibraryAndPackageSourcesPy` function and have match before calling!?

dbrattli: Added a separate copy function for Python to avoid the 3 messy language matches

PR #2546, opened 2021-09-23
Implement Type.IsInstanceOfType

chkn:
Hello, I've added an implementation of `Type.IsInstanceOfType`, but it might warrant some discussion, since it doesn't work for all types. Here's a comparison of the new `Type.IsInstanceOfType` and the existing `:?` operator:

**Legend**: ✅ fully supported ❌ not supported (always returns false)

| Type | New `Type.IsInstanceOfType` | Existing `:?` |
| ------|-------------------------------|-------------|
|Class types | ✅ | ✅|
| Nominal union types | ✅ | ✅ |
| Nominal record types | ✅ | ✅ |
| Anonymous records | ❌ | ❌ |
| Functions | ❌ | `typeof x === "function"` |
| Primitive string | ❌ | ✅ |
| Primitive number | ❌ | `typeof x === "number"` |

I've added some failing tests for the above cases, but I'm not sure if it's possible to get them all passing. It might be better to try to align the behavior with `:?`, but I'm not sure the best way to go about it. Perhaps we could add a type test function to `TypeInfo` that would use `typeof` for functions and primitives? On the other hand, it's already useful as it is for many use cases.

Let me know what you think. Thanks!

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -3119,7 +3119,7 @@ let types (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr optio
         | "get_FullName" | "get_Namespace"
         | "get_IsArray" | "GetElementType"
         | "get_IsGenericType" | "GetGenericTypeDefinition"
-        | "get_IsEnum" | "GetEnumUnderlyingType" | "GetEnumValues" | "GetEnumNames" | "IsSubclassOf" ->
+        | "get_IsEnum" | "GetEnumUnderlyingType" | "GetEnumValues" | "GetEnumNames" | "IsSubclassOf" | "IsInstanceOfType" ->
             let meth = Naming.removeGetSetPrefix i.CompiledName |> Naming.lowerFirst
             Helper.LibCall(com, "Reflection", meth, t, thisArg::args, ?loc=r) |> Some
         | _ -> None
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Yes, this one is tricky. Type testing in Fable is very weak by design (so we don't have to wrap all number primitives, for example). I wouldn't worry too much about function because it's quite hacky and doesn't really work with `:?` either. If you want it to work with strings and numbers too maybe you can do something like this:

```typescript
export function isInstanceOfType(t: TypeInfo, o: any) {
  switch (typeof o) {
    case "boolean":
      return t.fullname === bool_type.fullname;
    case "string":
      return t.fullname === string_type.fullname;
    case "number":
      return [
        int8_type.fullname,
        uint8_type.fullname,
        int16_type.fullname,
        uint16_type.fullname,
        int32_type.fullname,
        uint32_type.fullname,
        float32_type.fullname,
        float64_type.fullname,
      ].indexOf(t.fullname) >= 0;
    default:
      return t.construct != null && o instanceof t.construct;
  }
}
```

chkn:
@alfonsogarciacaro Thanks for the guidance! I've taken your suggestion and just added enums (which I'd forgotten :p), so I think the behavior of `Type.IsInstanceOfType` would now be the same as the `:?` operator on a boxed value, which is probably the best we can ask for right now.

alfonsogarciacaro:
Thanks, let's merge this then!

Issue #2031, opened 2020-04-20
Seq.cache does not work on infinite sequences


jwosty:
### Description

Seq.cache tries to iterate the whole sequence, so if its passed an infinite sequence, it'll crash when it eventually hits the memory limit. Native F# does not behave in this way. Also see #1143

### Repro code

```
open System
let guids = Seq.initInfinite (fun _ -> Guid.NewGuid()) |> Seq.cache
printfn "guid 0: %A" (Seq.item 0 guids)
printfn "guid 1: %A" (Seq.item 1 guids)
printfn "guid 0 again: %A" (Seq.item 0 guids)
```

### Expected and actual results

Code should not fail, and guid 0 should be the same both times

ncave:
@jwosty The repro provided above is invalid.
But you are correct, `Seq.cache` and seq printing are currently not implemented to support infinite sequences (`Seq.cache` is just implemented with `Seq.toArray` for now).

jwosty:
Yep, I see that now, I've fixed the repro

njlr:
It seems this has been fixed already. 

I added your repro as a test-case here: https://github.com/fable-compiler/Fable/pull/2547

njlr:
OK to close @jwosty ?

alfonsogarciacaro:
Closing as @njlr already submitted a test for this and it's working 👍

PR #2547, opened 2021-09-24
 * Adds extra test for Seq.initInfinite and Seq.cache

njlr:


njlr:
Adds repro from https://github.com/fable-compiler/Fable/issues/2031 as a test-case to ensure it doesn't regress.

alfonsogarciacaro:
Thank you!

Issue #2229, opened 2020-10-24
Bug:  System.DateTime.TryParse returns a inconsistent tuple (true, >"Invalid Date"<) 


funkr:
### Description

If the returned date is invalid, than the boolean flag must be false.
Must have the same behavoir as on the interactive shell.

### Repro code

    open System

    let validGermanDate = "31.10.2020" 
    let invalidAmericanDate = "13/1/2020" 
    //let d = invalidAmericanDate
    let d = validGermanDate
    let r, date = System.DateTime.TryParse d
    if r then 
        printfn "Result: %b but the date is: %s -> this not good!" r (date.ToShortDateString ())
    else 
        printfn "Result: %b than the date is: %s -> this is ok!" r (date.ToShortDateString ())

### Expected and actual results

If TryParse gives back a true than the date must be usable.

On the interactive shell TryParse behaves correct!

### Related information

Reproducible on [REPL](https://fable.io/repl/) with the given code.

Zaid-Ajaj:
Please checkout [this comment](https://github.com/fable-compiler/Fable/issues/2003#issuecomment-682077951) for a workaround which implements a small module to do the parsing

PR #2550, opened 2021-09-27
issue-2229-datetime-tryparse-inconsistent-tuple

njlr:
This is a _partial_ fix for https://github.com/fable-compiler/Fable/issues/2229. 

There are two issue raised there: 

 1. Fable date-time parser does not accepts dates that .NET does (e.g. `"31.10.2020"`) 
 2. Fable `DateTime.TryParse` sometimes returns `(true, new Date("Invalid date"))` for invalid dates

This PR fixes (2) by adding an additional safety check to `parseRaw`.

### DIFF
--- a/src/fable-library/Date.ts
+++ b/src/fable-library/Date.ts
@@ -219,7 +219,12 @@ export function parseRaw(input: string) {
     } else {
       throw new Error("The string is not a valid Date.");
     }
+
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks! Yes, it's difficult to emulate .NET behaviour completely (moreover if we start with globalization) but this should be already very helpful :+1:

PR #2553, opened 2021-09-30
Fix typo and add message about Node.js minimum version

MangelMaxime:


MangelMaxime:
We could be even more strict by adding `"node": "^12.20.0 || ^14.13.1 || >=16.0.0"` to the Fable package.json.

I think that this would emit some warning when running the package.

alfonsogarciacaro:
I think that's the engines section, but not being the main package.json of the app not sure how it will work. We can leave it as is for now and check later. Thanks!

MangelMaxime:
Yes, this is for the engine section.

I remember that when Nacara as a dependency, I was getting some warning when not using the right version of Node.js.

But yes, for now adding the instructions should be a first step in the right direction without any chance of breaking anything

PR #2556, opened 2021-09-30
Proposed rewording of the Femto info message

Zaid-Ajaj:


MangelMaxime:
I like the new wording. 

I would probably replace `metadata` with `information` as it is a simpler term.

alfonsogarciacaro:
In principle it should be fine, but the issue with the message is it's going to be displayed **every time** a package with `NpmDependencies` tag is detected. Fable doesn't check if the npm packages are actually installed or not (if it did, it would be doing Femto's job).

So it may be better not to imply that npm dependencies need to be installed, as it's very likely they already are. That's why I was trying to keep the message in the lines: "Reminder that Femto exists and can make your life better" :)

MangelMaxime:
Another proposition could be to say "manage":

> Some Nuget packages contain information about the required NPM packages. You can use Femto (https://github.com/Zaid-Ajaj/Femto) to automatically manage these NPM packages

This is a less direct action compare to "install".

alfonsogarciacaro:
@MangelMaxime Ok, let's do it like this :+1:

PR #2557, opened 2021-09-30
Replace .fsproj with .fableproj in fable_modules

alfonsogarciacaro:
This is to avoid conflicts with other F# tooling that scans the workspace for .fsproj files (Paket, Ionide), see https://github.com/fable-compiler/fable-react-native/issues/64#issuecomment-931304787

### DIFF
--- a/src/Fable.Cli/ProjectCracker.fs
+++ b/src/Fable.Cli/ProjectCracker.fs
@@ -463,7 +463,14 @@ let createFableDir (opts: CrackerOptions) =
 
     fableDir
 
-let copyDirIfDoesNotExist (source: string) (target: string) =
+// Replace the .fsproj extension with .fableproj for files in fable_modules
+// We do this to avoid conflicts with other F# tooling that scan for .fsproj files
+let changeFsprojToFableproj (path: string) =
+    if path.EndsWith(".fsproj") then
+        IO.Path.ChangeExtension(path, Naming.fableProjExt)
+    else path
+
+let copyDirIfDoesNotExist replaceFsprojExt (source: string) (target: string) =
     if isDirectoryEmpty target then
         IO.Directory.CreateDirectory(target) |> ignore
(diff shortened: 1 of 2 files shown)


ncave:
@alfonsogarciacaro Is this really needed, it's a little bit odd. Why does the `fable_modules` folder need to be ignored?

alfonsogarciacaro:
Yes, it's true it looks a bit like a hack, but it avoids the need of having to patch any other F# tooling that scans for .fsproj files (see link on comment of top). Don't remember what Paket is doing but I remember Ionide used to drive crazy when it found all the .fsproj files.

In principle, this should be completely transparent for the user because the folder is .gitignored and the files are managed automatically by Fable. We could just delete the .fsproj and keep the list of sources from the packages somewhere, although it requires a bit more changes.

alfonsogarciacaro:
Let's try like this for now and we can fix it more properly later. I've added issue #2558 for that.

ncave:
@alfonsogarciacaro Why can't package sources be referenced from the package folder directly? That's what `fable-compiler-js` [does](https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/src/ProjectParser.fs#L59-L82).

alfonsogarciacaro:
Because then they're outside the workspace and if one of the .fs sources from a packages references an npm package (say "react" in Fable.React) JS building tools won't be able to locate the npm package. And we cannot just put the compiled .js files in fable_modules, because some packages also contains their own js/css files so we need to keep the original .fs and the compiled .fs.js next to each other ([related](https://github.com/fable-compiler/repl/issues/141#issuecomment-930759941)).

Maybe... we could just copy everything from the package that's not an .fs(proj) and then put the compiled .js files there too. This would require some more changes but could be done along the work to split compilation for package and "user" sources, see #2527

ncave:
@alfonsogarciacaro Ok, thanks for the explanation.

PR #2559, opened 2021-10-03
Implementation of System.Collections.Generic.Queue<T>

davedawkins:
- Based on the recent addition of `System.Collections.Generic.Stack<T>` (same changes to `Replacements.fs`)
- Includes tests (`tests/Main/QueueTests.fs`)

There are some complications:

- There is a shim for `Queue<T>` in `src/fcs-fable/System.Collections.fs` - this has been removed but not tested. How to test this? Will rebuilding the REPL suffice, or do we then need to use Queue in the REPL? 

- The ApplicativeTests include tests that depend on `FSharp.Data.Adaptive`, which itself supplies a [definition of Queue<T>](https://github.com/fsprojects/FSharp.Data.Adaptive/blob/8d751c441aa4b43f590692bcd5f527038134137e/src/FSharp.Data.Adaptive/FableHelpers.fs#L25)

Unfortunately, this definition eclipses the new built-in Fable definition of Queue.  To allow the new tests (`QueueTests.fs`) to work, I have refactored `ApplicativeTests.fs` to isolate the dependencies into `AdaptiveTests.fs`. The project file `Fable.Tests.fsproj` defines `FableDefinesQueue` and then conditionally adds/compiles `FSharp.Data.Adaptive` and `AdaptiveTests.fs` based on its value.

```xml
  <FableDefinesQueue>True</FableDefinesQueue>


 <Choose>
    <When Condition=" '$(FableDefinesQueue)'=='True' ">
      <PropertyGroup>
        <DefineConstants>$(DefineConstants);FABLE_DEFINES_QUEUE</DefineConstants>
      </PropertyGroup>
    </When>
    <When Condition=" '$(FableDefinesQueue)'=='False' ">
    <ItemGroup>
        <PackageReference Include="FSharp.Data.Adaptive" Version="1.2.3" />
    </ItemGroup>
    </When>
 </Choose>
```

- If this PR gets merged, some way of patching `FSharp.Data.Adaptive` will be required. Is there any possibility of code like this being added to that package in FableHelpers.fs?
```fsharp
#if FABLE_VERSION < 3.2.NN
type Queue<'T> = ...
#endif
```

- Finally: one of the tests for DateTime was already failing before any of my changes, so I had to (temporarily) disable those tests in order to check that everything else passed.

I hope this submission is of use. Let me know if there's anything I need to change.


davedawkins:
One more thing: I see there are 3 commits queued here. I made a mistake with my first commit while trying to create a branch, and so reverted it. I'm hoping you don't have to include that mistake and the reversion too. I could delete my repo and make  a clean one (with a single commit of the changes) if that helps.

alfonsogarciacaro:
Oh, my, finding the shim in FSharp.Adaptive must have been painful. Probably the easiest solution for now is to put the FSharp.Adaptive tests in a different test project. I can do that.

About testing the REPL, in CI Fable compiles itself to create fable-standalone (which also powers the REPL) and then compiles the tests again. So if CI is passing we know `Queue` will be working in the REPL too (after updating it).

Don't worry about the commits, we can just squash when merging 👍

src/Fable.Transforms/Replacements.fs
```diff
@@ -3394,6 +3395,14 @@ let tryBaseConstructor com ctx (ent: Entity) (argTypes: Lazy<Type list>) genArgs
             | _ -> failwith "Unexpected stack constructor"
         let entityName = FSharp2Fable.Helpers.cleanNameAsJsIdentifier "Stack"
         Some(makeImportLib com Any entityName "Stack", args)
+    | Types.queue ->
+        let args =
+            match argTypes.Value, args with
+            | [], _ ->
+                []
+            | _ -> failwith "Unexpected queue constructor"
```
alfonsogarciacaro, line 3403: I just realized that Stack is doing the same, but there are actually other overloads for the constructor so probably we shouldn't do a hard crash here.

davedawkins: I was puzzled by this code, but had just copied it from the stack code. I felt it should be related to the 3 constructors. Shall I take a pass at reworking that in a separate commit?

davedawkins:
Thanks for the merge :-)

Any thoughts on how we can conditionally compile the Queue in `FSharp.Adaptive...`? We could just submit a PR that removes it and document that the latest version of it requires the latest Fable.

alfonsogarciacaro:
I've just put the Adaptive tests in a different test project :) https://github.com/fable-compiler/Fable/commit/387487703765d6ffb39ec6210e202e499abc6fbc

davedawkins:
Ok, let's leave it for now. What we have to look out for is puzzled devs who can't understand why they can't construct a Queue with an initial capacity or an existing collection.

PR #2560, opened 2021-10-03
Remove class restrictions for ParamObject

alfonsogarciacaro:


ncave:
@alfonsogarciacaro This alters the AST, so we probably need it in `beyond`.

alfonsogarciacaro:
Yes, I was wondering if I should add `IsFSharpModule` but it didn't break the plugins (unless you try to implement Fable.Entity) so I added it at the end (I think there are also a couple of other interface members that are not in the published Fable.AST package). Anyways, I do need to sync `main` and `beyond`. I started to do it, but there were some conflicts, I'll try to finish it his week 👍

PR #2573, opened 2021-10-28
Use FCS nuget package

alfonsogarciacaro:
@ncave If I'm understanding correctly latest improvements in the F# compiler, FCS already does some similar caching to service_slim so may be it's a good opportunity to use the Nuget package directly and reduce the work of maintaining a fork. Only thing is we still need to maintain/update fcs_fable for the repl and fable-compiler-js... and by "we" I mean "you" ;) What do you think?


alfonsogarciacaro:
@ncave You're probably already into it, but I [already tried to fix fcs-fable](https://github.com/fable-compiler/Fable/pull/2573/commits/9b2af473d3e224b9e81fb1e33d1713afb240c8e9) (better to hide whitespace to see the diff, I just added `|> Seq.toList` wherever the compiler was complaining).

> Actually I'm a bit puzzled I had to fix some build errors in fcs-fable because these files weren't touched 🤔 

However, when trying to build Fable.Standalone I still get this error:

```
src\fcs-fable\src\fsharp\TypedTree.fs(1,1): error EXCEPTION: The exception has been reported. This internal exception should now be caught at an error recovery point on the stack. Original message: Exception of type 'FSharp.Compiler.ConstraintSolver+ConstraintSolverError' was thrown.)
Compilation failed
```

Not sure if it rings a bell. It'd be nice to be able to merge this with fable-standalone working, but we can also wait until fcs-fable is updated to keep it in sync with Fable.Cli.

ncave:
@alfonsogarciacaro I was waiting for F# 6 and .NET 6 to officially drop before I start on fcs-fable, I guess I'll have to start now that it is released.

PR #2581, opened 2021-11-01
[Python] Add fixes to reflection

dbrattli:
- Add to reflection ` get_enum_underlying_type` , `get_generics`, `get_function_elements`, `get_tuple_elements`
- Fix module import path
- Add Lazy type
- Use Python arrays for F# arrays with numeric types instead of list


ncave:
@dbrattli Perhaps you can "un-fork" the Seq.fs if you implement `Array.ofSeq` to return proper arrays based on subtype, right? All those methods use `toArray` which uses `Array.ofSeq`. Or am I over-simplifying?

dbrattli:
@ncave The thing is that Python arrays are very limited in what you can do, so for `Seq` I want to not use arrays at all, but use plain Python lists (used for `ResizeArray`). Ref: https://stackoverflow.com/questions/22782879/api-why-dont-python-arrays-have-a-sort-method/22783330#22783330 

So for Array.fs I might have to use Python-lists as well as backing for sort, fold etc and then convert back to array of the right sub-type. In Python, arrays are mostly used for transmission of bytes, but in F# they are more often used like a list as well.

ncave:
@dbrattli That's fine, all I meant to say was that perhaps if you define `Array.ofSeq` in terms of ResizeArray you won't have to fork Seq.fs.

dbrattli:
@ncave Aha! Now I get it. Great suggestion. I'll try!

dbrattli:
It was actually doing it already in `Fable2Python`, so `Array.ofSeq` produces the (python) list we want. The problem was `toArray` itself passing through arrays, so if I comment the first line  then things works fine. So then the question is how do I fix this without generating an extra `Array.from($0)` for JS?

```fs
let toArray (xs: seq<'T>): 'T[] =
    match xs with
    //  | :? array<'T> as a -> a
    | :? list<'T> as a -> Array.ofList a
    | _ -> Array.ofSeq xs

let sortWith (comparer: 'T -> 'T -> int) (xs: seq<'T>) =
    delay (fun () ->
        let arr = toArray xs
        Array.sortInPlaceWith comparer arr // Note: In JS this sort is stable
        arr |> ofArray
    )

```

ncave:
@dbrattli Actually the commented version is correct, `Seq.toArray` needs to make a copy (that's the behavior in .NET):

```fs
    let a = [|1;2;3|]
    let b = a |> Seq.toArray
    a.[2] <- 4
    assert(b.[2] = 3)
```
Congratulations, you fixed a bug!
You can update it in your branch, I'll submit a fix in `main`.

dbrattli:
@ncave Awesome! Thank you for taking your time with this issue!

ncave:
See #2586.

PR #2583, opened 2021-11-04
Fix #2564: op_Implicit overloads

### LINKED ISSUE #2564: Return type overloading not supported by op_Implicit.
### Description

Multiple versions of `op_Implicit` are not supported.

### Repro code

```fs
type MyType = { Field : int}
        with
        static member op_Implicit(s:string):int = failwith "Unimplemented"
        static member op_Implicit(s:string):float = failwith "Unimplemented"

```

Native dotnet comiles while fable gives an error:
``` FABLE: Cannot have two module members with same name:```

### Related information

* Fable version: 3.4.2
* Operating system: MacOS

alfonsogarciacaro:
Fix #2564: op_Implicit overloads

Note: this is making the tests fails, but it works when I isolate the types declaring `op_Implicit` (one in ApplicativeTest and the other in CustomOperatorTests) in Quicktest. Not sure why.


MangelMaxime:
Replaced by #4712.

PR #2589, opened 2021-11-05
Merge main into beyond

ncave:
@alfonsogarciacaro I tried my hand at merging latest main into beyond, seems to work more or less, except two tests I had to [disable](https://github.com/fable-compiler/Fable/pull/2589/commits/4e009c7220520db32570e0450a552c942fe86260). Perhaps you can take a look, one has to do with uncurrying methods in object expr, the other is about generic decorators.

@dbrattli I haven't tried the Python tests, perhaps you can run them to see if everything still works.

P.S. We really need to merge more often, there was quite a churn in main in the last few months.


ncave:
@alfonsogarciacaro Also, there seems to be some problem with the React tests, perhaps the `Feliz` version is too old?
```
tests\React\fable_modules\Feliz.1.22.0\React.fs(1,1): error EXCEPTION: Method not found: 'Microsoft.FSharp.Core.FSharpOption`1<Fable.AST.SourceLocation> Import.get_Item3()'.
```

dbrattli:
@ncave There are 9 failing Python tests with this PR. I'm looking into what the problem is.

dbrattli:
@ncave Fixes for Python: https://github.com/ncave/Fable/pull/25

alfonsogarciacaro:
This is fantastic @ncave, thanks a lot for this work! I'm sorry things became so messy, I'm entirely to blame, I started to do some cleanup in `beyond`, then I tried to [split the languages](https://github.com/fable-compiler/Fable/tree/beyond-split) and then I went back to work on `main` :/ But I'll do my part from now on and will keep branches in sync when pushing to `main`.

I will also look into those failing tests. About the React ones, if I'm not mistaken they fail because Fable.AST has changed in `beyond` so Feliz `ReactComponent` plugin doesn't work any more. Once the AST is stable and we release Fable 4 (for JS) we should also update the plugin (hopefully it's the only one widely used).

ncave:
@alfonsogarciacaro Not a problem at all, thanks for merging! If the React tests can't be fixed by switching to a newer Feliz, then we'll have to turn off those tests until we can fix the plugin.

alfonsogarciacaro:
Yes, if I'm not mistaken they were switched off already for beyond, they probably became active again during the merge :)

Issue #2569, opened 2021-10-23
StringSplitOptions.TrimEntries is not supported


eagle-k:
### Description

.NET 5 has added [`TrimEntries`](https://docs.microsoft.com/en-us/dotnet/api/system.stringsplitoptions?view=net-5.0#System_StringSplitOptions_TrimEntries) to `StringSplitOptions`.
Is there any plan to support this in Fable?

### Expected and actual results

Expected: Each entry will be trimmed when `StringSplitOptions.TrimEntries`.
Actual: `StringSplitOptions.TrimEntries` is not supported and has the same behavior as `StringSplitOptions.RemoveEmptyEntries`.

### Related information

* Fable version: 3.4.5
* https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime/tests/System/String.SplitTests.cs

eagle-k:
Thanks for the reply!
I've added some tests based on the `dotnet/runtime` repo. You can see them here.
https://github.com/eagle-k/Fable/commit/b54a826101d8edac66043f37cd6dcc2887a75210

As it turns out, there are cases where not only `TrimEntries` but also the existing implementation behaves differently from dotnet.

```fs
testCase "case #1" <| fun () ->
    let M = System.Int32.MaxValue
    "a-b-c".Split("", M)
    |> equal [|"a-b-c"|]
    // actual
    // [|"a";"-";"b";"-";"c"|]

testCase "case #2" <| fun () ->
    let M = System.Int32.MaxValue
    "---o---o---".Split("--", M, StringSplitOptions.RemoveEmptyEntries)
    |> equal [|"-o"; "-o"; "-"|]
    // actual
    // TypeError: str.replace is not a function
```

I've tried to fix it, but it's proving difficult. I would appreciate it if others could contribute.

alfonsogarciacaro:
Sorry @eagle-k, I missed your last reply. Thanks a lot to @steveofficer for the fix 🙌 Hopefully this should be fixed now, I will push a new release shortly. Please feel free to reopen if there's still something missing.

PR #2590, opened 2021-11-07
Add support for StringSplitOptions.TrimEntries

steveofficer:
This covers both aspects of issue #2569 
1. It adds support for TrimEntries
2. It fixes the existing implementation to deal with separators like "" and finding the correct locations to perform the split


alfonsogarciacaro:
This is great @steveofficer, thanks a lot for this!

PR #2593, opened 2021-11-09
Pass only dirty files to FCS

alfonsogarciacaro:
So far, for recompilations we've just been calling `ParseAndCheckProject` on FCS, same as when compiling the whole project for the first time. This usually works fine when modifying files that are down the list of files because FCS caches the files on top, but for big projects, when editing a file on top of the list, FCS invalidates all the files below it, so the compilation can take a long time even if there are no other files depending on it.

In this PR, for watch compilations we call `ParseAndCheckFileInProject` instead so only "dirty" files are being recompiled which should help with the above case.

Note that I've touched the way files are recompiled to unify F# and Fable compilation, so we need to be sure watch compilations still work fine (for example, when changes cause errors in other files and then we fix the errors in a different iteration). It'd be great to have some integration tests about this.

### DIFF
--- a/src/Fable.AST/Plugins.fs
+++ b/src/Fable.AST/Plugins.fs
@@ -21,6 +21,7 @@ type CompilerOptions =
         Define: string list
         DebugMode: bool
         OptimizeFSharpAst: bool
+        // TODO: Rename this to TrimRootModule in beyond branch
         RootModule: bool
         Verbosity: Verbosity
         FileExtension: string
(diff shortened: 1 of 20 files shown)


ncave:
@alfonsogarciacaro I'm not sure how FCS works nowadays, hopefully that wouldn't trigger more work than just recompiling the project (i.e. would editing say the first 3 files in a project make FCS recompile the whole project 3 times?).

alfonsogarciacaro:
TBH, I don't think that Parallel helps performance a lot. For what I've seen in FCS code, `ParseAndCheckFileInProject` will recompile all files above if they've changed (i.e. calling in `ParseAndCheckFileInProject` will be almost the same as `ParseAndCheckProject`). The main gain is the files below won't have to be recompiled if they don't depend on the file changed (which happens often in Elmish projects).

Issue #2519, opened 2021-08-25
Support TimeOnly, DateOnly in .NET 6


kerams:
I know .NET 6 has not been released yet, but I'd just like to put this out there,

There are 2 new `DateTime`-related types that could see widespread use - [DateOnly](https://docs.microsoft.com/en-us/dotnet/api/system.dateonly?view=net-6.0) and [TimeOnly](https://docs.microsoft.com/en-us/dotnet/api/system.timeonly?view=net-6.0). Both come with loads of methods centered around arithmetic, but for start I think it would be enough to support their constructors and a way to get to the data (`Day`, `Month`, `Year` and `Hour`, `Minute`, `Second` properties). That alone would suffice in Fable.Remoting where the server does the arithmetic and the client just displays the data it received for example.

PR #2596, opened 2021-11-09
Support TimeOnly, DateOnly

kerams:
Implements https://github.com/fable-compiler/Fable/issues/2519.

`TimeOnly` is represented by a simple number, just like `TimeSpan`. `DateOnly` is represented by `Date` like `DateTime`.

Remaining work
- [x] Tests that comparison, hashing (dictionary lookup) works for both types
- [x] TimeOnly subtraction operator
- [x] TimeOnly.(Try)Parse
- [x] DateOnly.ToString
- [x] DateOnly.(Try)Parse
- [x] DateOnly.Add(Days|Months|Years)
- [x] DateOnly.ToDateTime

Not implementing
- `DateOnly` has a [constructor](https://docs.microsoft.com/en-us/dotnet/api/system.dateonly.-ctor?view=net-6.0#System_DateOnly__ctor_System_Int32_System_Int32_System_Int32_System_Globalization_Calendar_) which lets you specify the calendar used to intepret the other parameters. This would add another layer of complexity, and I don't think it's of any interest to the vast majority of users.
- DateOnly.ToLongDateString, DateOnly.ToShortDateString, TimeOnly.ToLongTimeString, TimeOnly.ToShortTimeString. You can just use ToString with the corresponding format specifier.
- DateOnly.ToString formats D, r, R, m, M, y, Y because they produce localized text.
- (Try)ParseExact on both types


tests/Main/DateOnlyTests.fs
```diff
@@ -0,0 +1,93 @@
...
+            let dateTime = DateTime (2000, 2, 28, 0, 30, 0, DateTimeKind.Local)
+            DateOnly.FromDateTime dateTime |> equal (DateOnly (2000, 2, 28))
+            let dateTime = DateTime (2000, 2, 28, 0, 30, 0, DateTimeKind.Utc)
+            DateOnly.FromDateTime dateTime |> equal (DateOnly (2000, 2, 28))
+            let dateTime = DateTime.Now
+            DateOnly.FromDateTime dateTime |> equal (DateOnly (DateTime.Now.Year, DateTime.Now.Month, DateTime.Now.Day))
+            //let dateTime = DateTime.MaxValue
+            //DateOnly.FromDateTime dateTime |> equal DateOnly.MaxValue
+            //let dateTime = DateTime.MinValue
+            //DateOnly.FromDateTime dateTime |> equal DateOnly.MinValue
```
kerams, line 92: These tests don't work because DateTime.MinValue/MaxValue is being created with `DateKind.Unspecified` and not UTC, so when you call `getTicks` on it, you get a different number back than what it was created with. Similarly, I also think there is a bug in `create` of `Date.ts` where `setFullYear` is called instead of `setUTCFullYear` when the requested kind is UTC.

alfonsogarciacaro:
Great work @kerams, thank you! Please just let me know when this is ready to merge 👍

tests/Main/DateOnlyTests.fs
kerams: It's not so simple and this scenario is a corner case with little practical use. Let's forget it.

kerams:
For some reason .NET SDK 5.0.403 was used for the last build, hence the failing tests.

kerams:
Ah, you changed the workflow file today. Well, without .NET 6 we're not going to get liftoff here :).

alfonsogarciacaro:
Sorry, we had some issues when updating to F# 6 (see #2605) but hopefully they're solved now, can you try syncing with `main`. Thanks!

kerams:
@alfonsogarciacaro, I'm not sure what's going on. The new tests pass the first time around, but then do not get compiled as .NET 6 following `/home/runner/work/Fable/Fable> node ./src/fable-compiler-js/src/app.fs.js tests/Main/Fable.Tests.fsproj build/tests-js`. I've switched `fable-compiler-js.fsproj` to .NET 6, but it didn't help.

alfonsogarciacaro:
@kerams Sorry, I should have explained this to you before. Fable standalone packs its own BCL DLLs because the point is dotnet/nuget doesn't need to be installed in the system (or the browser). These DLLs are in fable-metadata and are generated by custom build of FCS by (you probably guessed it) @ncave. We haven't updated the DLLs to net6 yet so that explains why fable-standalone doesn't find the new classes 

We can update the DLLs but maybe it's better to do it when @ncave updates fcs-fable... Yet another fork to make FCS compilable by Fable :) So for now we can disable them for fable-compiler-js. I can do that so you don't need to worry.

Thanks again for this work!

kerams:
Alright then, thanks.

ncave:
@alfonsogarciacaro I've updated the BCL to 6.0.100 in #2612, hopefully that works.

alfonsogarciacaro:
Thank you both! Released in v3.6.0-beta-001 👍

PR #2599, opened 2021-11-10
[Python] Add ParamObject to handle named arguments

dbrattli:
Fixes https://github.com/fable-compiler/Fable.Python/issues/18

We should perhaps give the `ParamObject` another name (or alias) for Python since it's not really a parameter object  when we expand the object into arguments before calling e.g:

```fs
[<ImportMember("....")>]
type MyPyClassWithOptionArgs
    [<ParamObject>] (?foo: int, ?bar: string, ?baz: float) =

    [<ParamObject(2)>] member _.method(foo: int, bar: string, baz: float, ?lol: int) = nativeOnly

let a = MyPyClassWithOptionArgs(bar="test")
a.method(10, "test", 1.0)
```

This generates the Python code:

```py
a = MyPyClassWithOptionArgs(bar = "test")

a.method(10, "test", baz = 1)
```

What would be the best way to do this for Python. E.g call it `NamedParams` instead? Is there a way we could e.g split attributes into separate language files @ncave @alfonsogarciacaro. PS: thanks for adding `ParamObject`, it fixes one of the last big problems with Python type bindings 🎉

### DIFF
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -98,6 +98,9 @@ type ParamObjectAttribute(fromIndex: int) =
     inherit Attribute()
     new () = ParamObjectAttribute(0)
 
+/// Alias for ParamObjectAttribute.
+type NamedParamsAttribute = ParamObjectAttribute
+
 /// Experimental: Currently only intended for some specific libraries
 [<AttributeUsage(AttributeTargets.Parameter)>]
 type InjectAttribute() =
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Nice to hear is working fine for Fable.Python! In the past I had a similar with attributes I wanted to rename, and this can be easily "fixed" with a type alias. `NamedParams` sounds good :)

https://github.com/fable-compiler/Fable/blob/535aa8a6775f461f9a8dc0c8956370c3cc941668/src/Fable.Core/Fable.Core.Types.fs#L87-L92

BTW, when using the index argument, I'd recommend to fully qualify it because this makes the intention clearer to readers:

```fsharp
type MyPyClassWithOptionArgs
    [<NamedParams(fromIndex=2)>] member _.method(foo: int, bar: string, baz: float, ?lol: int) = nativeOnly
```

PR #2600, opened 2021-11-10
Culture invariant tests

### LINKED ISSUE #2597: Fable .NET tests are dependent on the system locale
### Description

I'm seeing 8 tests that are failing in Expecto when running them locally. For instance:

```
[15:11:04 ERR] All/DateTime/Parsing doesn't succeed for invalid dates failed in 00:00:00.0040000.
. Actual value was true but had expected it to be false.
[15:11:04 ERR] All/Types/Mangled interfaces work with object expressions failed in 00:00:00.0060000.
.
          Expected string to equal:
          "Zar77.40rfalsefalsedbca"
                ↑
          The string differs at index 5.
          "Zar77,40rfalsefalsedbca"
                ↑
          String does not match at position 5. Expected char: '.', but got ','.
```

They are caused by the following:

```fsharp
let r, _date = DateTime.TryParse invalidAmericanDate
// ...
bar.DoSomething(multiplyTwice, 3.).ToString("F2")
```

The solution is to call the methods with a hard-coded `CultureInfo` outside of the JS tests.

kerams:
Fixes #2597.

I used `Replace` in some places because with `InvariantCulture` Mocha tests started failing with `ToString` producing unexpected different results.

### DIFF
--- a/tests/Main/DateTimeTests.fs
+++ b/tests/Main/DateTimeTests.fs
@@ -3,6 +3,7 @@ module Fable.Tests.DateTime
 open System
 open Util.Testing
 open Fable.Tests
+open System.Globalization
 
 let toSigFigs nSigFigs x =
     let absX = abs x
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro:
Thanks! We may need to fix `ToString` with CultureInfo in Fable but we can open a separate issue for that :+1:

Issue #2602, opened 2021-11-11
Cannot build to output directory named 'obj'


NickDarvey:
### Description

This is a low priority issue, though it did trip me up for a little while!

Building with an output directory of 'obj':
```
dotnet fable MyFableApp --outDir obj
```
leads to a referenced package not being available:
```
The value, constructor, namespace or type 'NamespaceInReferencedPackage' is not defined
```

### Repro code

https://github.com/NickDarvey/fable-outdir-package

### Expected and actual results

<details>
❯ dotnet fable Optml.Html.Fable.Tests --outDir obj/tests  
Fable: F# to JS compiler 3.2.9
Optml.Html.Fable.Tests> cmd /C dotnet restore Optml.Html.Fable.Tests.fsproj
  Determining projects to restore...
  All projects are up-to-date for restore.
Parsing Optml.Html.Fable.Tests\Optml.Html.Fable.Tests.fsproj...
Initializing F# compiler...
Compiling Optml.Html.Fable.Tests\Optml.Html.Fable.Tests.fsproj...
F# compilation finished in 2093ms
C:/repos/fable-outdir-package/Optml.Html.Tests/Testing.Fable.fs(3,22): (3,27) error FSHARP: The value, constructor, namespace or type 'Mocha' is not defined. (code 39)
C:/repos/fable-outdir-package/Optml.Html.Tests/Testing.Fable.fs(4,22): (4,27) error FSHARP: The value, constructor, namespace or type 'Mocha' is not defined. (code 39)
C:/repos/fable-outdir-package/Optml.Html.Tests/Testing.Fable.fs(5,23): (5,28) error FSHARP: The value, constructor, namespace or type 'Mocha' is not defined. (code 39)
C:/repos/fable-outdir-package/Optml.Html.Tests/Testing.Fable.fs(13,44): (13,49) error FSHARP: The value, constructor, namespace or type 'Mocha' is not defined. (code 39)
C:/repos/fable-outdir-package/Optml.Html.Tests/Testing.Fable.fs(14,36): (14,41) error FSHARP: The value, constructor, namespace or type 'Mocha' is not defined. (code 39)
C:/repos/fable-outdir-package/Optml.Html.Tests/Testing.Fable.fs(15,37): (15,42) error FSHARP: The value, constructor, namespace or type 'Mocha' is not defined. (code 39)
[...]

PR #2603, opened 2021-11-11
fix: #2602

NickDarvey:
See #2602.

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -183,6 +183,8 @@ type Runner =
     do!
         if outDirLast = Naming.fableHiddenDir then
             Error($"{Naming.fableHiddenDir} is a reserved directory, please use another output directory")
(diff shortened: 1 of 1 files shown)


src/Fable.Cli/Entry.fs
```diff
@@ -183,6 +183,8 @@ type Runner =
     do!
         if outDirLast = Naming.fableHiddenDir then
             Error($"{Naming.fableHiddenDir} is a reserved directory, please use another output directory")
+        elif outDirLast = "obj" then
```
NickDarvey, line 186: Would you prefer a link to the issue as a comment inline to explain to future visitors why this is the case, or do we rely on blame?

alfonsogarciacaro: It should be fine, `obj` is usually known as a special directory in .NET so it's understandable we reserve it. Now I just hope that people don't pick `.git` as out directory or similar :) Thanks!

PR #2604, opened 2021-11-11
Queue watch changes

alfonsogarciacaro:
Rigth now, Fable stops the watcher when it's compiling. Usually it's not a problem because watch compilations are fast, but if the compilation takes several seconds and users make some changes in the meanwhile these changes will go unnoticed. This PR adds a MailboxProcessor (actually, similar to Fable 2) so changes happening in the middle of a compilation are queued and not lost.

**UPDATE**: This PR now does a couple of things more:

- It removes the pseudo-cache of Fable compilation in Debug mode when there were already JS files and the F# sources hadn't been updated. I've realized the Fable compilation is needed to correctly build the watch dependency tree, so we cannot skip it.
- It goes back to @ncave's fork (actually [a fork of a fork](https://github.com/alfonsogarciacaro/fsharp/commits/service_slim_2021-06-08)) because the FCS Nuget package had performance issues ([discussion here](https://github.com/fable-compiler/Fable/issues/2605)) and also so we could keep fable-standalone compilation working.


MangelMaxime:
Oh this could explain some weird behaviour I had sometimes where I needed to manually trigger watch to some files manually.

PR #2621, opened 2021-11-23
Updated FCS-Fable

ncave:
- FCS-Fable rebased to 2021-11-15.


ncave:
I'm still evaluating the performance, seems a bit slower than before.

alfonsogarciacaro:
Thanks @ncave! Yes, unfortunately FCS seems to be getting slower in latest versions and this affects particularly the JS version. If you compile the RayTracer sample with fable.io/repl2 and fable.io/repl I get the following stats:

```
REPL2
FCS checker | 482.00
FCS parsing | 244.90
Fable transform | 138.50
Babel generation | 12.50

REPL
FCS checker | 0.10
FCS parsing | 574.70
Fable transform | 157.60
```

The terminology in the stats is wrong and somewhat misleading. The first value is the time to create the Interactive Checker and "FCS parsing" should include both parsing and type checking by FCS. It's interesting that the time to create the interactive checker is almost 0 in the current REPL, but it seems to be added to "FCS parsing" for every compilation. I assume that most of this time is spent reading the .dlls, may it happen that in repl2 we did it only once and now we're doing it for every compilation?

> On the plus side the running time of the generated JS code (here the time to render the image) seems to be improving

ncave:
> in repl2 we did it only once and now we're doing it for every compilation

No, I don't think that's the reason. Initial profiling doesn't reveal anything obvious unfortunately, so it could be some pervasive issue that is spread around, or could be a slow op that affects everything equally, or it could be an allocation issue.
It's very clear that the issue is there in FCS 41 on .NET too, even though .NET 6 is faster.

PR #2623, opened 2021-11-26
Follow-up for TS tagged union support (#2618)

cannorin:
- [x]  Erase the declaration of `TypeScriptTaggedUnion` 
- [ ] Check the tag when creating a `TypeScriptTaggedUnion` case

@alfonsogarciacaro Feel free to take over the second one If I'm too slow to implement it, since I'm still trying to figure out how to do that 😅  I think I will prioritize the docs over that.


cannorin:
↑ the above commit is actually not erasing the definition

cannorin:
Also: https://github.com/fable-compiler/fable-compiler.github.io/pull/116

alfonsogarciacaro:
Thanks again @cannorin! We probably can defer the tag checking to the next release 👍

PR #2630, opened 2021-11-30
[Python] Fixes 12

dbrattli:
- Better naming of arrow functions
- Fixes for dictionary item assignment
- Add more list tests
- Set project dir as outDir for Python if outDir is not set


alfonsogarciacaro:
@dbrattli I've been thinking about having a default outDir instead of putting the generated files next to the sources. It could be more consistent with dotnet to use `bin/Debug/fable` and `bin/Release/fable` by default. What do you think, would it work with Python?

dbrattli:
@alfonsogarciacaro No, it would not work with Python without adding an extra PYTHONPATH entry which in practice would require us to make a python wrapper. So either we install the modules right underneath (direct subfolder of) the project being compiled or we would need to pip install the modules as proper Python packages in site-packages. But the problem is that an F# project is not a proper Python package, so it makes things difficult. Thus the direct sub-folder is preferrable. Could we perhaps have it as a configurable subdir of the project being compiled? I.e `.`  vs `bin/Debug/fable` so Python could choose the first option.

alfonsogarciacaro:
> Thus the direct sub-folder is preferrable. Could we perhaps have it as a configurable subdir of the project being compiled?

Ah, ok. Then probably it's a good idea to have a default subdir but not use the .NET convention. I've noticed that also Rust requires a `cargo` file so I assume all the compiled files must be in the same dir too. @ncave Does Rust also have some conventions about how the sources should be structured? What `--outDir` default would work better for Rust?

> PS: looks like some JS test is failing. Is this fixed in some other branch?

Yes, it was [fixed here](https://github.com/fable-compiler/Fable/commit/c7aa75524489495bf9a00c7e2cddd6cd38e2c5d9#diff-90180378e9b28280f6876dd90204fceaf3da81fba2fac60462eee24738af5262) I need to make a new merge. Or you can just copy the fix... or wait until next Friday when the test will fix automatically 🤣

ncave:
@alfonsogarciacaro 
> Does Rust also have some conventions about how the sources should be structured?

Yes, it does, very much so. Rust has `src` sub-folder convention, so projects folders should be structured to mimic that, see `tests/Rust` and `fable-library-rust` for some examples.

>  What --outDir default would work better for Rust?

I have no opinion on the default value, --outDir should be whatever people want it to be.
So not sure what's best: `build`? `output`? Perhaps `build` is self-explanatory enough, if it doesn't clash with local folders.

dbrattli:
@ncave @alfonsogarciacaro Quick question about some JS interop in Native.fs I do not really understand so I'm having problems transpiling to Python:

```fs
let inline insertImpl (array: 'T[]) (index: int) (item: 'T): 'T[] = !! array?splice(index, 0, item)

let test () =
    let arr = [| "test" |]
    let _ = insertImpl arr 0 "ing"
    ()
```

This generates the following JS code:

```js
export function test() {
    const arr = ["test"];
    const tupledArg = [0, 0, "ing"];
    arr.splice([tupledArg[0], tupledArg[1], tupledArg[2]]);
}
```

But I don't understand why the arguments gets wrapped in `[]` i.e `[tupledArg[0], tupledArg[1], tupledArg[2]`. I get the same for Python which does not work, but I cannot really see how it should work for JS either. When I test it in the browser I cannot give `splice` an array of arguments like this. I'm a bit clueless here. Which file in the compiler is generating this code for `!!` and `?` ?

dbrattli:
Hmm, looks like it is compiled correctly when the library:

```js
export function insertRangeInPlace(index, range, array) {
    let index_1;
    let i = index;
    const enumerator = getEnumerator(range);
    try {
        while (enumerator["System.Collections.IEnumerator.MoveNext"]()) {
            const x = enumerator["System.Collections.Generic.IEnumerator`1.get_Current"]();
            (index_1 = (i | 0), array.splice(index_1, 0, x));
            i = ((i + 1) | 0);
        }
    }
    finally {
        enumerator.Dispose();
    }
}
```

But not in QuickTest:

```js
export function insertRangeInPlace(index, range, array) {
    let tupledArg;
    let i = index;
    const enumerator = getEnumerator(range);
    try {
        while (enumerator["System.Collections.IEnumerator.MoveNext"]()) {
            const x = enumerator["System.Collections.Generic.IEnumerator`1.get_Current"]();
            (tupledArg = [i, 0, x], array.splice([tupledArg[0], tupledArg[1], tupledArg[2]]));
            i = ((i + 1) | 0);
        }
    }
    finally {
        enumerator.Dispose();
    }
}
```

What is the difference?

dbrattli:
@alfonsogarciacaro @ncave Sorry, I'm referencing `Fable.Core.PyInterop` and not `Fable.Core.JsInterop` which seems to be the problem. Again, sorry for disturbing!

dbrattli:
Hmm, turned out that `Fable.Core.PyInterop` and `Fable.Core.JsInterop` where identical. But found this code which explain things:

```fs
        // When using Fable dynamic operator, we must untuple arguments
        // Note F# compiler wraps the value in a closure if it detects it's a lambda
        | FSharpExprPatterns.Let((_, FSharpExprPatterns.Call(None,m,_,_,[e1; e2])),_), args
                when m.FullName = "Fable.Core.JsInterop.(?)" ->
            let! e1 = transformExpr com ctx e1
            let! e2 = transformExpr com ctx e2
            let e = Fable.Get(e1, Fable.ExprGet e2, Fable.Any, e1.Range)
            let! args = transformExprList com ctx args
            let args = destructureTupleArgs args
            let typ = makeType ctx.GenericArgs fsExpr.Type
            let r = makeRangeFrom fsExpr
            // Convert this to emit so auto-uncurrying is applied
            return emitJsExpr r typ (e::args) "$0($1...)"

```

alfonsogarciacaro:
@dbrattli Yes, sorry, this operator is special because the F# compiler creates the binding and the replacements couldn't see the surrounding expression, so I dealt with it directly in FSharp2Fable. But... now replacements can see the let bindings in scope, so maybe it could be an opportunity to deal with `?` in Replacements and avoid confusions like this in the future: https://github.com/fable-compiler/Fable/blob/a7668ec36d5c9e87180762218eb26f0d7c650835/src/Fable.Transforms/Replacements.fs#L660-L668

dbrattli:
Great idea @alfonsogarciacaro . I can open a separate issue to track that fix. I think I will merge this now, so let's also fix the `ourDir` handling later when you decide how things should be in JS. It should be easy to align the code in Main.fs then and collapse if it becomes very similar for all the languages.

alfonsogarciacaro:
Hmm, I had a look at this and unfortunately it's going to be trickier than I thought .The pattern detects _applications_ to the result of accessing something with `?` operator. So this is different from checking the bound indents and cannot access it from the Replacements module :/

PR #2633, opened 2021-12-03
Merge main into beyond

ncave:
- Merge latest `main` into `beyond`

P.S. It's probably a good idea to merge `main` more often as the branches have diverged enough to warrant it.


alfonsogarciacaro:
Yes, I split the solutions yesterday because I was having issues with the IDE, sorry for that. Thanks @ncave!

dbrattli:
@alfonsogarciacaro @ncave One thing with these changes is that we now have get a circular dependency between Seq and List. Seq.js now depends on List.js  and List.js depends on Seq.js:

In List.js:

```js
import { foldBack as foldBack_2 } from "./Seq.js";
```

In Seq.js:

```js
import { length as length_1, tryItem as tryItem_2, isEmpty as isEmpty_1, tryHead as tryHead_2, ofSeq as ofSeq_1, ofArray as ofArray_1, toArray as toArray_1, FSharpList } from "./List.js";
```

This might work for JS, but fails when compiled to Python. I need to looks closer into what's going on here to see how we can fix this. Any suggestions?

```
/usr/local/Cellar/python@3.9/3.9.2_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_arithmetic.py:1: in <module>
    from util import Testing_equal
util.py:3: in <module>
    from fable_modules.fable_library.seq import (head as head_1, skip)
fable_modules/fable_library/seq.py:7: in <module>
    from .list import (FSharpList, to_array as to_array_1, of_array as of_array_1, of_seq as of_seq_1, try_head as try_head_2, is_empty as is_empty_1, try_item as try_item_2, length as length_1)
fable_modules/fable_library/list.py:8: in <module>
    from .seq import fold_back as fold_back_2
E   ImportError: cannot import name 'fold_back' from partially initialized module 'fable_modules.fable_library.seq' (most likely due to a circular import) (/Users/dbrattli/Developer/Github/Fable/build/tests/Python/fable_modules/fable_library/seq.py)
```

ncave:
@dbrattli This is wrong, `List` should not depend on `Seq`, I'll fix it.

ncave:
@dbrattli See #2637.

alfonsogarciacaro:
Ups, my fault! Thanks for the fix @ncave!

dbrattli:
But it's sort of nice that one language can find a bug in another 😄 Thanks for fixing @ncave !

PR #2643, opened 2021-12-07
Add Python tests to CI

dbrattli:
Add python tests to GitHub action

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,9 @@ jobs:
       with:
         toolchain: stable
 
+    - name: Install pytest runner
+      run: pip install pytest
+
     - name: Fable Tests
       run: dotnet fsi build.fsx test
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,3 +24,6 @@ jobs:
 
     - name: Test
       run: dotnet fsi build.fsx test
+
+    - name: Fable Tests - Python
+      run: dotnet fsi build.fsx test-py


dbrattli:
@alfonsogarciacaro @ncave There is btw a flaky (JS) test that I have seen quite a few times but haven't been able to fix. Any ideas on how to fix this?

```
25h[18:25:24 ERR] All.Async.Async.Sequential works errored in 00:00:01.0030000 <Expecto>
System.Exception: expected sequential operations to take longer than 1 second, but took 1
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1439.Invoke(String message) in D:\a\_work\1\s\src\fsharp\FSharp.Core\printf.fs:line 1439
   at Fable.Tests.Async.tests@264-375.Invoke(Int32[] _arg20) in /home/runner/work/Fable/Fable/tests/Main/AsyncTests.fs:line 267
   at Microsoft.FSharp.Control.AsyncPrimitives.CallThenInvokeNoHijackCheck[a,b](AsyncActivation`1 ctxt, b result1, FSharpFunc`2 userCode) in D:\a\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 464
   at Microsoft.FSharp.Control.Trampoline.Execute(FSharpFunc`2 firstAction) in D:\a\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 104
25h25h[18:25:36 INF] EXPECTO! 2,248 tests run in 00:00:15.9522527 for All – 2,247 passed, 0 ignored, 0 failed, 1 errored.  <Expecto>
System.Exception: Process exited with code 2
   at FSI_0001.PublishUtils.Platform.Process@163.FSI_0001.PublishUtils.IProcess.Run(String workingDir, String exePath, String[] args)
   at FSI_0002.test()
   at <StartupCode$FSI_0002>.$FSI_0002.main@()
Stopped due to error
```

alfonsogarciacaro:
Ah, I think I've seen this but only once in a while. Do you see it often? I guess we should extend the timeout in Expecto as we do it for Mocha (set to 10s with [-t 10000](https://github.com/fable-compiler/Fable/blob/c223454a3722ea342c773e347c43aefedbf1fa54/build.fsx#L79) argument).

ncave:
@alfonsogarciacaro @dbrattli

Or we change the condition to strictly less?

from: `if d <= TimeSpan.FromSeconds 1. then`
to: `if d < TimeSpan.FromSeconds 1. then`

alfonsogarciacaro:
@ncave Didn't check the actual test 😅 Good idea! 👍

dbrattli:
Hmm, I think it actually used to be like that. But I tried to make it less flaky and it became more flaky. Trying to fix to see if it helps: https://github.com/fable-compiler/Fable/pull/2645/files

dbrattli:
Reverted and ran it a few times before merging. Then it failed again when merging so there's something strange going on here: https://github.com/fable-compiler/Fable/runs/4472449184?check_suite_focus=true

ncave:
@dbrattli Reverted to what, strictly less or lessOrEqual?

dbrattli:
@ncave Strictly less

ncave:
@dbrattli Then it must be something else going on, not sure. But probably still makes sense to leave it as strictly less.

dbrattli:
Yes, I'm fixing it slightly to actually output the time with decimals since the current formatting only outputs `1` which is not helpful. The other thing is that the error message does not output what we are actually testing against. So fixing the test to test for `.TotalSeconds` instead and then output `.TotalSeconds` with decimals if it fails. https://github.com/fable-compiler/Fable/pull/2650

Issue #1296, opened 2017-12-18
`Seq.map String |> String.concat` fails at runtime


vasily-kirichenko:
```fsharp
open System
"fffff" |> Seq.chunkBySize 3 |> Seq.map String
```
According to fable.io, it compiles to
```js
import { chunkBySize, map } from "fable-core/Seq";
map(function (arg00) {
  return arg00.join('');
}, chunkBySize(3, "fffff"));
```
The "Run" button does not work on fable.io, so I cannot test it there, but such code crashes at runtime in my app with

![image](https://user-images.githubusercontent.com/873919/34094248-71ff14e8-e3dd-11e7-87ca-063e30832487.png)

![image](https://user-images.githubusercontent.com/873919/34094285-9e5a0278-e3dd-11e7-86bd-517c05da8e27.png)

`"fffff" |> Seq.chunkBySize 3 |> Seq.map String |> String.concat " "` - does not work
`"fffff".ToCharArray() |> Array.chunkBySize 3 |> Array.map String |> String.concat " "` - works

In F# REPL both ^^ return the same value `"fff ff"`.

PR #2646, opened 2021-12-08
Enable for Rust - Seq.map and Seq.map2

alexswan10k:
A little more progress on the seq module.

Not really sure what the best approach is for the fixed name ident problem I ran into here. It seems that "matchValue" is something generated by Fable. In some contexts (inner match) this was messing up the closure clone-on-capture mechanic as it did not exist. Happy to revisit if there are better ideas

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -3305,7 +3305,8 @@ module Util =
 
     let getIgnoredNames (name: string option) (args: Fable.Ident list) =
         let argNames = args |> List.map (fun arg -> arg.Name)
-        let allNames = name |> Option.fold (fun xs x -> x :: xs) argNames
+        let fixedNames = ["matchValue"]
+        let allNames = name |> Option.fold (fun xs x -> x :: xs) (argNames @ fixedNames)
         allNames |> Set.ofList
 
     let hasCapturedNames com ctx (name: string option) (args: Fable.Ident list) (body: Fable.Expr) =
--- a/src/fable-library-rust/src/Seq.fs
+++ b/src/fable-library-rust/src/Seq.fs
@@ -168,10 +168,8 @@ module Enumerable =
             | None -> None
         fromFunction next
 
-(*
     // Different implementation
-
-    [<CompiledName("Enumerator")>]
+    [<CompiledName("Enumerator2")>] //todo error if "Enumerator" - "An item with the same key has already been added"
     type FromFunctions<'T>(current, movenext, dispose) =
         interface IEnumerator<'T> with
             member _.Current = current()
@@ -185,6 +183,7 @@ module Enumerable =
     let fromFunctions current movenext dispose: IEnumerator<'T> =
         new FromFunctions<'T>(current, movenext, dispose) :> IEnumerator<'T>
 
+(*
     // implementation for languages where arrays are not IEnumerable
 
     let empty<'T>(): IEnumerator<'T> =
@@ -309,7 +308,7 @@ module Enumerable =
         let movenext() = e.MoveNext()
         let dispose() = try e.Dispose() finally f()
         fromFunctions current movenext dispose
-
+*)
     let generateWhileSome (openf: unit -> 'T) (compute: 'T -> 'U option) (closef: 'T -> unit): IEnumerator<'U> =
         let mutable started = false
         let mutable curr = None
@@ -337,7 +336,7 @@ module Enumerable =
                 | None -> finish(); false
                 | Some _ as x -> curr <- x; true
         fromFunctions current movenext dispose
-
+(*
     let unfold (f: 'State -> ('T * 'State) option) (state: 'State): IEnumerator<'T> =
         let mutable curr: ('T * 'State) option = None
         let mutable acc: 'State = state
@@ -408,8 +407,8 @@ let ofList (xs: 'T list) =
 //     // | :? list<'T> as a -> a
 //     | _ -> List.ofSeq xs
 
-// let generate create compute dispose =
-//     mkSeq (fun () -> Enumerable.generateWhileSome create compute dispose)
+let generate create compute dispose =
+    mkSeq (fun () -> Enumerable.generateWhileSome create compute dispose)
 
 // let generateIndexed create compute dispose =
 //     mkSeq (fun () ->
@@ -712,11 +711,11 @@ let length (xs: seq<'T>) =
             count <- count + 1
         count
 
-// let map (mapping: 'T -> 'U) (xs: seq<'T>) =
-//     generate
-//         (fun () -> ofSeq xs)
-//         (fun e -> if e.MoveNext() then Some (mapping e.Current) else None)
-//         (fun e -> e.Dispose())
+let map (mapping: 'T -> 'U) (xs: seq<'T>) =
+    generate
+        (fun () -> ofSeq xs)
+        (fun e -> if e.MoveNext() then Some (mapping e.Current) else None)
+        (fun e -> e.Dispose())
 
 // let mapIndexed (mapping: int -> 'T -> 'U) (xs: seq<'T>) =
 //     generateIndexed
@@ -727,14 +726,14 @@ let length (xs: seq<'T>) =
 // let indexed (xs: seq<'T>) =
 //     xs |> mapIndexed (fun i x -> (i, x))
 
-// let map2 (mapping: 'T1 -> 'T2 -> 'U) (xs: seq<'T1>) (ys: seq<'T2>) =
-//     generate
-//         (fun () -> (ofSeq xs, ofSeq ys))
-//         (fun (e1, e2) ->
-//             if e1.MoveNext() && e2.MoveNext()
-//             then Some (mapping e1.Current e2.Current)
-//             else None)
-//         (fun (e1, e2) -> try e1.Dispose() finally e2.Dispose())
+let map2 (mapping: 'T1 -> 'T2 -> 'U) (xs: seq<'T1>) (ys: seq<'T2>) =
+    generate
+        (fun () -> (ofSeq xs, ofSeq ys))
+        (fun (e1, e2) ->
+            if e1.MoveNext() && e2.MoveNext()
+            then Some (mapping e1.Current e2.Current)
+            else None)
+        (fun (e1, e2) -> try e1.Dispose() finally e2.Dispose())
 
 // let mapIndexed2 (mapping: int -> 'T1 -> 'T2 -> 'U) (xs: seq<'T1>) (ys: seq<'T2>) =
 //     generateIndexed
--- a/tests/Rust/tests/src/SeqTests.fs
+++ b/tests/Rust/tests/src/SeqTests.fs
@@ -396,20 +396,20 @@ let ``Seq.iteri works`` () =
     )
     total |> equal 20.
 
-// [<Fact>]
-// let ``Seq.map works`` () =
-//     let xs = [1.]
-//     let ys = xs |> Seq.map ((*) 2.)
-//     ys |> Seq.head
-//     |> equal 2.
+[<Fact>]
+let ``Seq.map works`` () =
+    let xs = [1.]
+    let ys = xs |> Seq.map ((*) 2.)
+    ys |> Seq.head
+    |> equal 2.
 
-// [<Fact>]
-// let ``Seq.map2 works`` () =
-//     let xs = [1.]
-//     let ys = [2.]
-//     let zs = Seq.map2 (*) xs ys
-//     zs |> Seq.head
-//     |> equal 2.
+[<Fact>]
+let ``Seq.map2 works`` () =
+    let xs = [1.]
+    let ys = [2.]
+    let zs = Seq.map2 (*) xs ys
+    zs |> Seq.head
+    |> equal 2.
 
 // [<Fact>]
 // let ``Seq.mapi works`` () =


ncave:
@alexswan10k Thanks, I may have to absorb this into another PR cause I opted for re-implementing the `Seq` from single-function, just so we have an easier way exposing as Rust iterators (and to simplify implementation).

alexswan10k:
No worries @ncave we can scrap it if you have a better implementation or the likes. Getting iterators etc right is a core feature anyway so it is worth taking the time to iron it out.

Let me know if there is something else on the periphery I can chip away at!

ncave:
@alexswan10k
Pretty much everything is up for grabs at this point, so whatever interests you. Any and all contributions are very welcome.

Core concepts like equality and comparison (and when to apply them to objects) still need some deep thinking, but they do need to be dealt with in order to enable all collection methods that need comparisons etc.

I guess we can start small and gradually work up the corner cases, we already have [some checks](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Transforms/Rust/Fable2Rust.fs#L266-L280) in place.

> Happy to revisit matchValue if there are better ideas

Perhaps we can add a restrict that it also has to be CompilerGenerated?

alfonsogarciacaro:
@alexswan10k @ncave What's `getIgnoredNames` used for? "matchValue" is one of the idents automatically generated by the F# compiler. But Fable should guarantee ident names are unique within the same declaration scope so you may get "matchValue_1", "matchValue_2", etc. There are automatically generated idents in other situations. Fable tries to eliminate them in binding beta reduction during FableTransforms, but it doesn't remove the binding if the value is captured by a closure: https://github.com/fable-compiler/Fable/blob/c223454a3722ea342c773e347c43aefedbf1fa54/src/Fable.Transforms/FableTransforms.fs#L88-L103

ncave:
@alfonsogarciacaro 

> What's getIgnoredNames used for?

For Rust closures, we need to collect all captured names so they can be cloned before the closure, but we need to exclude the closure arguments, that's what `getIgnoredNames` is for.

alfonsogarciacaro:
Thanks @ncave! Interesting, ideally we should store all variable declarations in the context so when transforming a lambda we can check if an ident reference is a free var or not, although your current method should also work. How is it with references to "root" scope, do they need to be cloned as well?

```fsharp
let foo = 5

let myMethod() =
   let myInnerLambda x =
      x + foo
   myInnerLambda 10
```

Maybe here you need to add also `TryCatch` expressions and nested lambdas/delegates which also declare arguments?

```fsharp
let myMethod() =
   let myInnerLambda x =
      let myInnerLambda2 y =
          x + y
      myInnerLambda 3
   myInnerLambda 10
```

https://github.com/fable-compiler/Fable/blob/c223454a3722ea342c773e347c43aefedbf1fa54/src/Fable.Transforms/Rust/Fable2Rust.fs#L3286-L3304

> Note: in principle there shouldn't be local name shadowing at this step. Fable guarantees that all idents in the same declaration scope are unique.

ncave:
@alfonsogarciacaro

> ideally we should store all variable declarations in the context so when transforming a lambda we can check if an ident reference is a free var or not

Are you suggesting storing all arguments and local declarations in context and checking each of them if captured in the closure, using `FableTransforms.isIdentCaptured` ? If so, would that not be inefficient, perhaps we need to update that method to take a set of idents to checks against, and return a subset of the captured ones. It would be great to have a general utility method for that.

ncave:
@alexswan10k 
Closing this as it was moved into #2647.

alexswan10k:
Thanks @ncave.

Without getting too into the weeds, the reason we need to clone everything over a boundary like this is to beat the borrow checker. All reference types are wrapped in a RefCounter https://doc.rust-lang.org/book/ch15-04-rc.html so a clone effectively gives us a new smart pointer to the same object on the heap. "Disposing" of one of these reduces the shared counter, unless it is the last one.. then it clears the memory. Poor mans GC :)

Because closures cannot reason if the captured thing will live longer than the thing outside, we clone to break the dependency (captured thing is otherwise borrowed). We only need to clone Rc wrapped references in practice, as rust mandates that there can only be 1 owner who is responsible for cleaning up the memory (although there may be some nuances I have missed).

alexswan10k:
Hi @ncave 

It seems that the low hanging fruits are mostly gone, I cannot seem to move anything forward any more :( Had a look at string.chunkBySize the other week and got hopelessly stuck.. then I found #1296. I guess this is blocked?

Anyway, I was experimenting to see if I could innocuously turn on the following seq test, which lead me down a rabbit hole. I was wondering if you are aware of this and perhaps have any ideas..
```
[<Fact>]
let ``Seq.sum with non numeric types works`` () =
    let p1 = {x=1; y=10}
    let p2 = {x=2; y=20}
    [p1; p2] |> Seq.sum |> (=) {x=3;y=30} |> equal true
```

Turns out there is a problem. The interface for this of course expects an Add trait.
image

I figured I could get the transpiler to just output a definition for each record. Maybe something like this
```
  impl Add for Point {
      type Output = Point;
      fn add(self, other: Point) -> Point {
          Point{x: self.x + other.x, y: self.y + other.y,}
      }
  }
```

Nope, turns out it is expecting a ```Rc<Point>``` and not a ```Point```. Ok, how about

image


Oh no! So I did some digging, which led me to [this](https://stackoverflow.com/questions/25413201/how-do-i-implement-a-trait-i-dont-own-for-a-type-i-dont-own). Turns out you cannot implement a trait on a thing if you are not in that crate because the compiler cannot guarantee it won't conflict with another definition elsewhere.

So next I tried this
```
            struct RcPoint(Rc<Point>);
            impl Add for RcPoint {
                type Output = Rc<Point>;
                fn add(self, other: RcPoint) -> Rc<Point> {
                    Rc::from(Point{x: self.0.x + other.0.x, y: self.0.y + other.0.y,})
                }
            }
```

Problem here is this is not picked up either, because a ```Rc<Point>``` is used, and not a ```RcPoint```! What is left to try, rewriting the call site for ```Record1 + Record2``` to expand the add operation inline? I feel like this has already been touched on when we did classes.

Nice work otherwise, excited to try a real project out when vnext is out.

ncave:
@alexswan10k

> It seems that the low hanging fruits are mostly gone...

Hi Alex,

I wouldn't worry about it, there are *plenty* of low hanging fruits left, from implementing any .NET BCL you want, to making the commented F# collection methods work, to adding support for .NET types like `decimal` and `bigint`, by porting, re-implementing or using external crates where appropriate for performance.

But that's all bike shedding at this point, until the two big elephants in the room are addressed, namely having proper imports without namespace clashing, and proper native Rust bindings (slightly lower priority only because simple bindings can already be done with some Emit and inline functions, but still badly needed at least as a concept).

As far as string handling goes, I made a few deliberate choices for performance and interoperability with native Rust. Since Rust strings are `UTF-8`, instead of `UTF-16` like JavaScript or .NET, we can either stick with Rust strings (which means accepting that things like `string.Length` are O(n)), or re-implement string and char as `UTF-16` and `uint16` (but that would make Rust interop harder).

For now I opted for sticking with native `UTF-8` and `uint32` for string and char. For that reason, all string functions are implemented in Rust, to use iterators to minimize conversions to char array and back where possible, as that would quadruple the string memory allocation for large strings.

For sequences, yes, there are still some missing default conversions (from string to seq of char, from seq of list/array to seq of seq, etc.). For now all those can be worked around with explicit conversions (e.g. from string to array of char to seq of char), so those are on the back-burner too.

For generic type constraints, the deliberate choice was to stick with [F# type constraints](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/generics/constraints) instead of attributes, and [convert](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Transforms/Rust/Fable2Rust.fs#L2884-L2918) them to traits, ideally ones that can be used with `#[derive]`.
If we can't make it all work, perhaps we can add our own traits, but then we'll have to implement them for all existing Rust primitive types, so perhaps as a last resort.

Thank you for your participation @alexswan10k, it's nice to have somebody to bounce ideas off.
I really appreciate your contributions and opinions, so please let me know if you spot an issue, or think we've made a wrong turn somewhere, or if you think we need to take a feature in an entirely new direction. All contributions are equally valuable, even just an opinion or a concept. Have a great day!

alexswan10k:
> But that's all bike shedding at this point, until the two big elephants in the room are addressed, namely having proper imports without namespace clashing, and proper native Rust bindings (slightly lower priority only because simple bindings can already be done with some Emit and inline functions, but still badly needed at least as a concept).

I have been thinking about the binding problem on and off over the weeks and have a few ideas. I am just aware this is quite a big (potentially disruptive) undertaking, but happy to maybe have a go here and perhaps lay some foundations. I guess this starts with a new Emit like attribute and probably a Fable AST representation of it.. one that has a lot more control of inbound and outbound stuff with conversions.

Have you decided on what path to go yet? Either
* Add strings to adulterate the output Rust AST
* Have some kind of a rust AST parser!
* Something else?
> 
> As far as string handling goes, I made a few deliberate choices for performance and interoperability with native Rust. Since Rust strings are `UTF-8`, instead of `UTF-16` like JavaScript or .NET, we can either stick with Rust strings (which means accepting that things like `string.Length` are O(n)), or re-implement string and char as `UTF-16` and `uint16` (but that would make Rust interop harder).

I think UTF-8 is entirely sensible given the constraints.

> For generic type constraints, the deliberate choice was to stick with [F# type constraints](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/generics/constraints) instead of attributes, and [convert](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Transforms/Rust/Fable2Rust.fs#L2884-L2918) them to traits, ideally ones that can be used with `#[derive]`. If we can't make it all work, perhaps we can add our own traits, but then we'll have to implement them for all existing Rust primitive types, so perhaps as a last resort.

Makes a lot of sense. I guess the only bit I am unsure of how to proceed with is my example above. It seems that there will be cases where you are comparing ```Rc<T>``` wrapped things. Is the intent to unwrap these where this happens?

> 
> Thank you for your participation @alexswan10k, it's nice to have somebody to bounce ideas off. I really appreciate your contributions and opinions, so please let me know if you spot an issue, or think we've made a wrong turn somewhere, or if you think we need to take a feature in an entirely new direction. All contributions are equally valuable, even just an opinion or a concept. Have a great day!

Appreciated! Same to you, super excited with how this is progressing. Just sorry I haven't been much use recently. Will keep an eye out anyway hopefully there are some things here and there I can help chip away at. Rome wasn't built in a day.

ncave:
@alexswan10k

> Have some kind of a rust AST parser!

That's probably taking it a bit too far, unless you mean it in the context of some `rust2fable` tooling.

For ad-hoc bindings, perhaps a more minimal approach would be to find a flexible enough representation to guide the transpiling to Rust native. F#/.NET has enough language features to be able to do that, IMO. The bindings can use a new attribute, or stick with `Import` (say on an interface) to distinguish themselves from normal F# code transpiling. Perhaps use  `inref<'T>` args in bindings to represent references, `byref<'T>` or `outref<'T>` for mutable references, etc.

I don't have all the details, I just know we need it, and it needs to make sense, in order to build a good community eco-system. See also #2779.

> comparing Rc<T> wrapped things. Is the intent to unwrap these where this happens?

Yes, or perhaps we can do what seems to be usually done in Rust for generic programming and references, and implement traits for both Rc and non-Rc wrapped generic types. Not sure yet if that would be sufficient to make it work.

alexswan10k:
Hmm I think I see where you are going with this... so off the top of my head you could define your input transforms IN f# rather than some proprietary string template.

```
[<Import("somefn")>]
let someFn (a: byref<int32>, b: Rust.Vec<Rust.Rc<int>>, c: inref<string>) =
  //input transforms
  let a = a.clone() 
  let b = b |> transformToArrayLikeThing
  let c = c.getValue()
  let res = call(a,b, c, d) //here you delegate down to the real rust function
  Rust.Rc(res.clone()) //some output transform
```

ncave:
@alexswan10k
Something like that, although I mostly meant using the "second" mode to have better fidelity to define imports using interfaces that map to e.g `std::` library or other Rust code, etc.

alexswan10k:
Interesting.

Ok @ncave  here is another idea I wanted to throw at the wall.

Maybe we can do this with types rather than a built in 'dual' mode.. although the effect would be largely the same. What if, we have a special low level type that can be used to work with real rust types. 

```
type Raw<'a> = ...
```

What if, when the compiler sees a Raw wrapped type, it completely omits any operations by the ref count "manager" (so no implicit wrapping or cloning). It might even be worth going further and having 1-1 types for anything that is raw rust, where conversion functions could be supplied to transition between "raw" and "managed". An example:

```
[<Import("std::awkward::awkwardFn")>]
let awkwardFn(a: Raw<string>, b: Raw<Ref<i64>>, c: Raw<SomeStruct>): Raw<SomeStructResult>

let useAwkwardApi a b c= 
  let a = mkRaw<string> s
  let b = mkRaw<Ref<64>> 42
  let c = mkRaw<SomeStructResult> {| X = 1... etc |}
  let res = awkwardFn(a, b, c)
  res |> mkManaged
```

The point here is the two conversion functions understand how to turn "dangerous" rust native types (things that can explode with the borrow checker but are invisible to the F# compiler) into implicit managed types that can be used without risk in F#. It may well be that there is also not a 1-1 mapping, and some degree of recursion required to unwrap etc. It could also perhaps take into account the state of global parallelism (are we using Rc or Arc etc).

By using types to transition between "managed" and "unmanaged"... (we need new terms or something!) you basically have the same effect of twin mode compilation but without having to define exact boundaries to where this may or may not happen, which I imagine is nigh on impossible. This way you have far more control.

Some random types off the top of my head

```
Rust --- F#
Vec<i32> --- Raw<Vec<i32>>
String --- Raw<Rust.String>
str --- Raw<Rust.str>
*str --- Raw<PtrTo<Rust.str>>
&str --- Raw<RefOf<Rust.str>>
&CustomStruct --- Raw<RefOf<CustomStruct>>
```
Maybe you don't need the extra raw wrapping, just trying to cover the scenario where anything can be ignored, simply by wrapping it in a Raw, and this allows this effect by composition. This allows you to model complex graphs of ignored structs etc.

alexswan10k:
> Some code, for example, doesn't need precise dual mode, as we already support low-fidelity "native" dual mode which avoids Rc-wrapping and doesn't pass by reference (it's all or nothing, so not able to model more complicated APIs):

It sounds like we might even be half way there then. Sorry my above post is a muddled consciousness dump! Let me go through your example but expand with this idea.

```
module Performance =
   [<Raw>]
    type Duration =
        abstract as_millis: unit -> Rust.uint128
        abstract as_secs_f64: unit -> Rust.float64
    [<Raw>]
    type Instant =
        abstract duration_since: Instant -> Duration
        abstract elapsed: unit -> Duration

    let now(): Instant = importMember "std::time::Instant"
    [<Import("std::time::Instant::now")>]
    let now2(): Instant = nativeOnly
    [<Emit("std::time::Instant::now")>]
    let inline internal now(): obj = nativeOnly

// sample usage
let measureTime (f: unit -> 'T): 'T * float =
    let t0 = Performance.now()
    let res = f ()
    let t1 = Performance.now()
    let elapsedRaw: Rust.Float64 = t1.duration_since(t0).as_secs_f64()
    let elapsed: float = Rust.float64.toManaged(elapsedRaw)
    res, elapsed * 1000.0
```

ncave:
@alexswan10k I don't know, custom types that have to be converted when used adds a lot of code on the usage side, instead of letting the transpiler do its job. I understand it adds safety, but I'm still leaning towards exploring first a fine-grained type conversion in the import declaration, to see how far we can push it.

ncave:
@alexswan10k Sorry to resurrect this one year old thread. Just picking your brain, as we still have this issue of not able to implement out-of-crate traits for LRC-wrapped objects (see [this comments](https://github.com/fable-compiler/Fable/pull/2646#issuecomment-1047563545) above). Currently we can implement out-of-crate traits only on non-wrapped (value-type) structs, but not on LRC-wrapped ones.

Any ideas how to overcome that? The usual work-around is to make either the trait or the type local to the crate. I don't think we can't use local smart ptr wrappers for each crate, as they would be incompatible. What about local supertraits that inherit from the out-of-crate ones (for, say, implementing `IDisposable` with `core::ops::Drop` trait), do you think that would work?

alexswan10k:
Hey @ncave , it’s not the most straight forward but do [extension traits ](http://xion.io/post/code/rust-extension-traits.html) help?

The wrapper is a good idea too. Any name works for me, although it might want to signify its local.

I think GATs might potentially also be able to simplify our Lrc story but not sure, needs some thought.

ncave:
@alexswan10k Thanks for the suggestions, I'll take a look.

alexswan10k:
I think you are right. Wrappers are probably the only option. Does implementing deref help at least on the consumption side? The good news is there is no runtime overhead. Maybe all that messy init code can be rolled up into the constructor.

ncave:
@alexswan10k Yes, wrapper will implement Deref and other ops. As mentioned, [CoerceUnsized](https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html) is unstable, so we can't hide the construction neatly, it's a bit verbose but needed for the implicit casting to dyn trait objects.

I'll take a look at the GAT too.

alexswan10k:
On the note of wrappers, what if every reference type came with its own locally defined wrapper? Anything needing it’s heap pointer would use this wrapper name instead of the lrc everywhere.

ncave:
@alexswan10k
> what if every reference type came with its own locally defined wrapper?
- Perhaps, if nothing else works. I'll keep it in mind, thanks!

ncave:
@alexswan10k On a side note, a [nice read](https://mmapped.blog/posts/15-when-rust-hurts.html) on some of the pains we're struggling with.

alexswan10k:
Good read

PR #2653, opened 2021-12-10
Language specific FABLE_COMPILER defines

dbrattli:
Add language specific defines for `FABLE_COMPILER`. In addition to `FABLE_COMPILER` and `FABLE_COMPILER_4` this PR will add one of the language specific below:

- `FABLE_COMPILER_PHP`
- `FABLE_COMPILER_RUST`
- `FABLE_COMPILER_DART`
- `FABLE_COMPILER_PYTHON`
- `FABLE_COMPILER_JAVASCRIPT`
- `FABLE_COMPILER_TYPESCRIPT`

it also adds

- `FABLE_COMPILER_4_OR_GREATER`

What do you think @alfonsogarciacaro @ncave. Would something like this be OK?

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -228,19 +228,14 @@ type Runner =
             "FABLE_COMPILER"
             "FABLE_COMPILER_3_OR_GREATER"
             "FABLE_COMPILER_4"
+            "FABLE_COMPILER_4_OR_GREATER"
             match language with
             | Php -> "FABLE_COMPILER_PHP"
             | Rust -> "FABLE_COMPILER_RUST"
             | Dart -> "FABLE_COMPILER_DART"
-            | Python -> yield! [
-                "FABLE_COMPILER_PY"
-                "FABLE_COMPILER_PYTHON" ]
-            | TypeScript -> yield! [
-                "FABLE_COMPILER_TS"
-                "FABLE_COMPILER_TYPESCRIPT" ]
-            | JavaScript -> yield! [
-                "FABLE_COMPILER_JS"
-                "FABLE_COMPILER_JAVASCRIPT" ]
+            | Python -> "FABLE_COMPILER_PYTHON"
+            | TypeScript -> "FABLE_COMPILER_TYPESCRIPT"
+            | JavaScript -> "FABLE_COMPILER_JAVASCRIPT"
         ]
         |> List.distinct
 
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -227,6 +227,12 @@ type Runner =
         |> List.append [
             "FABLE_COMPILER"
             "FABLE_COMPILER_3"
+            match language with
+            | Php -> "FABLE_COMPILER_PHP"
+            | Rust -> "FABLE_COMPILER_RUST"
+            | Dart -> "FABLE_COMPILER_DART"
+            | Python -> "FABLE_COMPILER_PY"
+            | JavaScript -> "FABLE_COMPILER_JS"
         ]
         |> List.distinct
 
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -226,13 +226,21 @@ type Runner =
         args.Values "--define"
         |> List.append [
             "FABLE_COMPILER"
-            "FABLE_COMPILER_3"
+            "FABLE_COMPILER_3_OR_GREATER"
+            "FABLE_COMPILER_4"
             match language with
             | Php -> "FABLE_COMPILER_PHP"
             | Rust -> "FABLE_COMPILER_RUST"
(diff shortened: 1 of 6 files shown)


MangelMaxime:
I wonder if it could make sense to include the version number in the directives, in case there are breaking changes in the future.

Example: `FABLE_COMPILER_4_JS`

*I put 4 because, I suppose all that will be in Fable 4 and not Fable 3*

alfonsogarciacaro (approved):
This looks good to me. Could also a good opportunity to change `FABLE_COMPILER_3` to `FABLE_COMPILER_4` in beyond. Although dotnet does something like `FABLE_COMPILER_3_OR_GREATER`, maybe we should do the same?

dbrattli:
Ok @alfonsogarciacaro adding the extra defines for `FABLE_COMPILER_4` and `FABLE_COMPILER_3_OR_GREATER`. I was also a bit unure about `FABLE_COMPILER_PY` since we spell out `PHP` and `RUST` so just adding both variants for js, py and ts. Is it too much?

- `FABLE_COMPILER_PY` and `FABLE_COMPILER_PYTHON`
- `FABLE_COMPILER_JS` and `FABLE_COMPILER_JAVASCRIPT`
- `FABLE_COMPILER_TS` and `FABLE_COMPILER_TYPESCRIPT`

MangelMaxime:
Code using compiler directives can already be confusing/hard to read without having duplicates of them.

I think, it is better to have a single version per language and to keep things consistent using the language full name seems better to me. If we use the short name/extension suffix there can be conflict/confusion between different languages.

alfonsogarciacaro:
I agree with @MangelMaxime, using just the fullname (e.g. `FABLE_COMPILER_PYTHON`) would likely be less confusing. I'm also more convinced now to use `FABLE_COMPILER_4_OR_GREATER` instead of `FABLE_COMPILER_4` :)

MangelMaxime:
I am not sure about `FABLE_COMPILER_4_OR_GREATER` you can't guarantee that in Fable 5 this will be compatible.

In Haxe another transpiler which exist since a long time they just use `haxe_3` and `haxe_4` making the supported version explicit.

Because if we do `FABLE_COMPILER_4_OR_GREATER` then how can we target a specific version?

src/Fable.Cli/Entry.fs
```diff
@@ -228,19 +228,14 @@ type Runner =
             "FABLE_COMPILER"
             "FABLE_COMPILER_3_OR_GREATER"
             "FABLE_COMPILER_4"
+            "FABLE_COMPILER_4_OR_GREATER"
```
alfonsogarciacaro, line 231: Hmm, I was thinking we should just use `FABLE_COMPILER_4_OR_GREATER` and remove `FABLE_COMPILER_4`, do you think it's better to use both? If we use FABLE_COMPILER_4_OR_GREATER only at the end we should also change the other instances below.

MangelMaxime: Personally, I would prefer `FABLE_COMPILER_3` and `FABLE_COMPILER_4` as we don't know if the code we are writing will really be supported with future version of Fable. 

We only know that we support the currently existing version of Fable.

dbrattli: I'm agnostic to this so that's why I added both options similar to .NET. But I will let you decide if I should remove one: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

MangelMaxime: I didn't know that you where basing your proposition on how .NET does it.

For me `FABLE_COMPILER_3` and `FABLE_COMPILER_4` seems enough from what I saw in Haxe worlds. 

But I understand if we want to stick to what .NET does if we feel it to be a good solution too.

dbrattli: Then I think we should probably stick with both `FABLE_COMPILER_4` and `FABLE_COMPILER_4_OR_GREATER`.  What do you think @alfonsogarciacaro Should we merge this as is?

alfonsogarciacaro: Ah, you're right @dbrattli! I thought .NET only used the X_OR_GREATER pattern 😅 I'd rather use only the X_OR_GREATER pattern for simplicity but it's true we likely should use both for consistency. In principle we should avoid breaking changes in future versions but we never know. As far as I know the only library using `FABLE_COMPILER_3` is Fable.SimpleJson but this is because before it was relying on the internal JS representation of some types like Map, and this can indeed change in future versions (even minor versions) so it's not a good practice.

dbrattli: Ok, great. Let's try this. Fable.SimpleJson is an obvious candidate since we now have Fable.SimpleJson.Python. Fable.JsonProvider also btw have a single line that needs to be handled differently with different languages. I'll merge this PR now. Thanks for reviewing!!

PR #2659, opened 2021-12-15
Precompile files

alfonsogarciacaro:
@ncave I'm trying to add a `precompile` command so users can precompile common files and packages that are changed less often and speed up the compilations. For this, I need to [generate a .dll assembly](https://github.com/fable-compiler/Fable/blob/9385f516ebbbb2ee8349b455139b62f759f47a70/src/Fable.Cli/Main.fs#L372-L389), so it can be referenced later from the main project. I'm using the `FSharpChecker.Compile` method from FCS and it works, only it takes  a long time and I cannot reuse the typed AST already generated (there's an overload to use the untyped AST but for this I don't have access to it currently and not sure if it will improve times a lot as parsing is much faster than type checking).

Do you know if it would be possible/easy to integrate the `exports` branch of your fork into `service_slim` to generate empty assemblies (only with metadata)? And whether these empty assemblies can be generated faster than assemblies with code? If it doesn't look feasible/easy to do we can forget about it though because precompiling shouldn't be done often anyways :)

> I'm planning to use this feature also to precompile libs for REPL so it would be useful there too.


ncave:
@alfonsogarciacaro Sorry for the late answer, didn't see your question. Ye, you can easily merge `export` and `service_slim`, but I don't think that will help, since `export` only exports already compiled and loaded/referenced assemblies, just without their code portion. So there is no shortcut there.

alfonsogarciacaro:
Thanks for the reply @ncave! You're right, I was having a look at this and realized my first assumption was wrong. The problem is actually I'm compiling twice: one for the normal F# - Fable compilation cycle and another to generate the .dll. Ideally we should have a way to tell `ParseAndCheckProject` to use the results to generate a .dll. I tried passing "--out:path/to/file.dll" in the options but it didn't work. Do you know if there's a way to generate the .dll assembly from the `ParseAndCheckProject` results?

ncave:
@alfonsogarciacaro 

> Do you know if there's a way to generate the .dll assembly from the ParseAndCheckProject results?

I'm sure there is a way, after all that's what the F# compiler does next, generating IL code from the type-checked AST.
But I don't know whether is available as an nice API, or do we need to expose it. Perhaps you can ask in the F# compiler repo, maybe one of the maintainers there can point us in the right direction and save us some digging through the compiler code.

alfonsogarciacaro:
@ncave I think [I got it](https://github.com/alfonsogarciacaro/fsharp/commit/b909c48cd75d9c8ae0f660545fdeba4f33cfb880) :) Seems to work, in my tests calling `compileOfTypedAst` seems to take half the time of `FSharpChecker.Compile`. Unfortunately it seems it cannot be parallelized, at one point I tried to generate the .dll in parallel to Fable compilation but the time was added at the end. It looks like there's a lock when accessing the `TypedImplFile`s.

ncave:
@alfonsogarciacaro This is great, looks very minimal, I like it.

ncave:
@alfonsogarciacaro 
> Unfortunately it seems it cannot be parallelized

Would it help if you detach the type-checking from the IL code generation (expose the `compileOfTypedAst` as separate API) and then stagger (or pub-sub) the type-checking and IL compiling in the client?

alfonsogarciacaro:
Thanks for the advice @ncave, I applied it and it [looks better now](https://github.com/alfonsogarciacaro/fsharp/commit/0d3b97dbf68524e346b5dfd3ce6c3350373caaa2) :)

After checking the project, I start the dll compilation [as a child](https://github.com/fable-compiler/Fable/pull/2659/commits/b774f072a2c40145a903642115208f8ef66a00c1#diff-c6e830abd1d7895047c70eaccf52015e9facdcd92ab5edaebe808369e7f409deR589-R591) but it still seems it doesn't do much work until the other operations are finished. I think now it's just because the Fable compilation and [the serialization of inline expressions](https://github.com/fable-compiler/Fable/pull/2659/commits/b774f072a2c40145a903642115208f8ef66a00c1#diff-00e6635d7e5369ae571fc801724f7fc98c470c7dcb93491fd616ea15be77dd75R719) just take up all the threads in the thread pool. If you can think of a better idea to start the work, please let me know :)

ncave:
@alfonsogarciacaro

Just to clarify what I mean by "stagger (or pub-sub) the type-checking and IL compiling in the client":
- Let's assume the type-checking cannot be parallelized. In this scenario we can also remove the parallelization of the parsing to keep the type-checking API single-threaded.
- This leaves some unused CPU while files are being type-checked, which we can use to generate IL code.
- So I was proposing that one (producer) thread is type-checking files *one by one*, and after *each* file is type-checked, creates work for other (consumer) thread(s) to generate IL code.
- Does that make sense? All this should happen *before* Fable starts transforming the AST, and, ideally, complete in the same time as before (if IL code generation is not much slower than type-checking).

alfonsogarciacaro:
That's a great idea @ncave! Indeed, type checking cannot be parallelized, [it uses a mapFold](https://github.com/ncave/fsharp/blob/9a1e74e5d14ef05f48285cd37ef18fc77d282c76/src/fsharp/FSharp.Compiler.Service/service_slim.fs#L235-L242) that takes the accumulated `tcState` before checking each file. It's always painful for me to see always that only 25% of my 4 CPU cores are being used during type checking :)

However, a changes as suggests will likely require many more changes in the FCS pipeline as it's basically implemented in a step-by-step sequential fashion. There are several points that expect the full list of `TypedImplFile`s, like [optimization](https://github.com/ncave/fsharp/blob/9a1e74e5d14ef05f48285cd37ef18fc77d282c76/src/fsharp/fsc.fs#L798-L800) (I tried skipping optimization but got runtime errors later in the pipeline) or [IL code gen](https://github.com/ncave/fsharp/blob/9a1e74e5d14ef05f48285cd37ef18fc77d282c76/src/fsharp/fsc.fs#L843).

Not sure if this can be changed to a pub-sub pattern, but if it's possible it should be beneficial for the F# compilation itself too.

In any case, you made me think we could try to apply the pub-sub pattern to F#-Fable compilation. Ideally this would kick-off Fable compilation before type checking finishes so Fable can use the free CPU.

BTW, I just remembered about the incoming support for reference assemblies in F#. Maybe we can try rebasing with [this branch](https://github.com/dotnet/fsharp/pull/12334) and see if it's faster to generate reference assemblies (these should be enough for fable precompilation).

ncave:
@alfonsogarciacaro +1 for Fable pushing the boundaries of the (b)leading edge of the F# compiler.
- `Fable`: Using "reference" metadata assemblies years before it became en vogue to do so :)

alfonsogarciacaro:
I'm merging this because I cannot find more opportunities to improve performance at the moment. I will try the sub-pub pattern with F#-Fable compilation in a different PR.

PR #2660, opened 2021-12-16
fable-library-rust update

ncave:
- Different `List` implementation to lower allocations
- More tests

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -97,8 +97,9 @@ type Entity =
     abstract IsFSharpModule: bool
     abstract IsFSharpUnion: bool
     abstract IsFSharpRecord: bool
-    abstract IsValueType: bool
+    abstract IsFSharpAbbreviation: bool
     abstract IsFSharpExceptionDeclaration: bool
+    abstract IsValueType: bool
     abstract IsInterface: bool
     abstract IsMeasure: bool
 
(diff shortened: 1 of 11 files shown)


alexswan10k:
This is really starting to come together @ncave, great work. I see you have most of seq, list, and array working, and somehow you have got around the class file problem for IEnumerable which is awesome.

It is a shame lists have to be wrapped in a MutCell, but I guess some of these problems are a consequence of closely mirroring the behaviour on the dotnet runtime.

Can I help by tackling the Map module? I see nobody has started on it yet, and it seems like you have enough there now to do this, and of course maps are rather handy!

ncave:
@alexswan10k

> It is a shame list elements have to be wrapped in a MutCell, but I guess some of these problems are a consequence of closely mirroring the behaviour on the dotnet runtime.

No, that's not it, actually it's a feature ;)
Mutable list tails allow for forward list building, avoiding the list reverse.
Also, `Cell` is a pretty cheap abstraction for mutable refcounted pointers anyway.

> Can I help by tackling the Map module?

Absolutely, whatever you like, the `fable-library/Map.fs` is a good starting point.

alexswan10k:
Oh, that is clever, I never thought of that. Pragmatism. 

It is a shame the Rust world will see it though, but I suppose we could easily add some extra conversion functions at the boundaries to make the F# types more rust like - to_vec(), to_slice() etc.

Have you had any more thoughts on the dreaded Arc switch? Can we swap all the rust Rc coded files out with arc equivalents? I still think there might be an argument to have both sets side by side, so you can (as a rust consumer) work with either simultaneously. Not sure how you distribute otherwise also as package is precompiled presumably. I suppose you could have two packages.

Cool anyway

ncave:
@alexswan10k Yes, I suppose the build can publish two packages, `fable_library_rust` and `fable_library_rust_sync`.

PR #2675, opened 2021-12-27
Include interface declarations in AST

### LINKED ISSUE #2673: Represent Interfaces in the Fable AST
### Description

Interfaces are erased and not part of the Fable AST. For type annotations in languages such as Python and TypeScript it would be nice to be able to represent the interfaces as code. What is the current thinking about interfaces and e.g TypeScript? Would it be possible to add interfaces to the AST and let the language decide if it want's to erase them or not? I would be interested in contributing this, but want to check if there are any gotchas I should be aware of?

alfonsogarciacaro:
Fixes #2673

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -1597,10 +1597,10 @@ module Util =
         let error msg =
             $"%s{msg}: %s{ent.FullName}"
             |> addErrorAndReturnNull com [] None
-        match ent.IsInterface, ent.Ref.SourcePath with
-        | true, _ -> error "Cannot reference an interface"
-        | _, None -> error "Cannot reference entity from .dll reference, Fable packages must include F# sources"
-        | _, Some file ->
+        match com.Options.Language, ent.IsInterface, ent.Ref.SourcePath with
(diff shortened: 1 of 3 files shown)


dbrattli:
This is awesome! What's the right way to deal with reflection declarations. Should we output reflection declaration for interfaces (if the language outputs them). If I do I get an error: `error FABLE: Cannot reference an interface: QuickTest.MyInterface` https://github.com/fable-compiler/Fable/blob/main/src/Fable.Transforms/FSharp2Fable.Util.fs#L1601

dbrattli:
Should we perhaps also add an `isAbstract` attribute to `MemberInfo` so we can translate `MembersFunctionsAndValues` entries into `Fable.MemberDecl`?

alfonsogarciacaro:
@dbrattli Ah, sorry. I forgot to remove the interface check also here (or maybe we should include a language check as interfaces still cannot be referenced in JS): https://github.com/fable-compiler/Fable/blob/7db476d7897ae97fafffcb543f7a63b0b6125cba/src/Fable.Transforms/FSharp2Fable.Util.fs#L1600-L1601

About turning the interface members into `MemberDecl`, if we can already get the info from `Entity` type I think it shouldn't be necessary They're not actually declarations (they don't have a body) and member declarations are already tricky in Fable (attached vs detached members).

PR #2676, opened 2021-12-28
Pub-sub pattern for F#-Fable compilation

alfonsogarciacaro:
See this comment: https://github.com/fable-compiler/Fable/pull/2659#issuecomment-1000889052

This is fantastic, thanks a lot for the idea @ncave! In practicality it means Fable compilation times disappear and get diluted in F# type checking times. In most cases, Fable compilation is much faster than F# type checking so we're still constrained by that. But when compiling standalone (where F# and Fable compilation took more or less the same) we're cutting the compilation time almost in half!

Changes in FCS necessary for this are here: https://github.com/alfonsogarciacaro/fsharp/commit/2a497180fd9366531f89cf6d7dcf3447637c9d28#diff-115ce16832f5a44cba9c8b3796afe6b06c8f41a6502466034dd5dae950d53ea1R237


dbrattli:
Very nice!! Great work! 🎉

PR #2678, opened 2021-12-28
[Python] interface handling

dbrattli:


src/Fable.Transforms/Python/Fable2Python.fs
```diff
@@ -775,7 +774,12 @@ module Annotation =
             | _ ->
                 let ent = com.GetEntity(ent)
                 if ent.IsInterface then
-                    typingModuleTypeHint "Any" []
+                    let name =
+                        ent.FullName.Split('.')
+                        |> Array.last
+                        |> (fun name -> name.Replace("`", "_"))
+                        |> Helpers.clean
+                    makeGenericTypeAnnotation com ctx name genArgs repeatedGenerics, []
```
dbrattli, line 782: @alfonsogarciacaro This sort of works but probably not the easiest way to do things. How would you generate imports for the interfaces used? I also need to decide if generic interfaces should be named `MyInterface_1` ($ in js) or just `MyInterface`. The problem with `_1` is with e.g `IEnumerable_1` which is not valid. Any suggestions?

dbrattli: Unblocked by https://github.com/fable-compiler/Fable/pull/2680

PR #2679, opened 2021-12-28
[Python] Fix increment / decrement

### LINKED ISSUE #2677: [Python] Invalid codegen for `incr`
### Description

Using `incr(a)` causes `void(a.contents++)` which is invalid python syntax.

### Repro code

```fsharp

let f x =
    incr x
    !x
```

-> 

```python
from typing import Any

def f(x: Any) -> int:
    void(x.contents++)
    return x.contents
```

```python
from typing import Any

def f(x: Any) -> int:
    x.contents += 1
    return x.contents
```

### Related information

* Fable version: `dotnet fable --version`:
   ```
     ➜ fable-py --version
     4.0.0-alpha-020
   ```
* Operating system: Windows 10(19043.1415)

dbrattli:
Fixes #2677


dbrattli:
Waiting for #2678

PR #2685, opened 2021-12-29
Optimize Map keys and values

### LINKED ISSUE #2684: Fable 3.7 generated javascript performance
### Description

Generated javascript collection performance comments (under electron 16/Chromium recent version)

Fable when compared with F# 6 .Net runs 2X slower. Not bad. Some rough edges:

1. Sequences are a lot slower than under .Net
2. As a result of 1., `Map.values` - which generates a sequence - is about 5X slower than an optimal `Map.toList >> List.map snd` implementation. It is possibly more space efficient.
3. I don't really understand how much optimisation of sequences is done by the F# .Net compiler. You'd think in many cases they could be transformed to non-sequence equivalents before the Javascript generation but obviously not.

From: 
https://github.com/tomcl/issie/blob/a17356f2927f8b1fd663120cf1eac14edfae5a53/src/Renderer/Common/TimeTests.fs

### Electron16 (Chromium) javascript performance

```
Comparing speed filtering an input list, array, seq of size 90000.
The output is the same type as the input except for iter and for which change a mutable output value.
 In those cases the output must be a list: arrays cannot incrementally add one element quickly.
 Iter and For result in a reversed list as output: for non-reversed add on time of List.rev.
        Code Fragment           Time (output size=1)  Time (output size=89999) 
        sequence (seq) time:    12.82                  16.83
        list time:              5.11                  12.32
        array time:             3.40                  5.92
        array collect time:     8.30                  50.40
        list collect time:      0.56                  4.23
        list iter->list time:   0.64                  2.00
        array iter->list time:  0.23                  1.20
        array for loop time:    0.61                  5.80
 
Testing time in ns to read the values of a Map of given size (per value)
 Size  in items:   100       1000       10000
 viaArrays:       51.30     47.21      48.56
 viaArraysSumBy:  21.73     19.95      23.36
 viaLists:        31.19     28.71      44.45
 viaListsSumBy:   16.76     16.34      20.50
 viaValues:       214.51     183.44      190.54
 viaRamdomLookup: 92.45     164.95      215.52
```

### .Net 6 performance

```
Comparing speed filtering an input list, array, seq of size 90000.
The output is the same type as the input except for iter and for which change a mutable output value.
In those cases the output must be a list: arrays cannot incrementally add one element quickly.
Iter and For result in a reversed list as output: for non-reversed add on time of List.rev.

       Code Fragment           Time (output size=1)  Time (output size=89999)
       sequence (seq) time:    11.04                  19.33
       list time:              9.50                  25.66
       array time:             7.09                  17.83
       array collect time:     1.86                  2.69
       list collect time:      0.84                  8.54
       list iter->list time:   0.64                  2.57
       array iter->list time:  0.29                  2.12
       array for loop time:    0.38                  4.97

Testing time in ns to read the values of a Map of given size (per value)
Size in items:   100       1000       10000
viaArrays:       31.91     31.30      39.89
viaArraysSumBy:  22.93     25.59      31.91
viaLists:        26.80     28.83      39.71
viaListsSumBy:   21.79     21.28      24.25
viaValues:       35.49     35.18      36.25
viaRamdomLookup: 56.28     94.32      117.29
```

ncave:
- Fixes #2684

### DIFF
--- a/src/fable-library/Map.fs
+++ b/src/fable-library/Map.fs
@@ -654,10 +654,10 @@ type Map<[<EqualityConditionalOn>]'Key, [<EqualityConditionalOn; ComparisonCondi
         | None -> false
 
     member __.Keys: ICollection<'Key> =
-        MapTree.toSeq tree |> Seq.map (fun kvp -> kvp.Key) |> ResizeArray :> _
+        MapTree.toArray tree |> Array.map (fun kvp -> kvp.Key) :> _
 
     member __.Values: ICollection<'Value> =
-        MapTree.toSeq tree |> Seq.map (fun kvp -> kvp.Value) |> ResizeArray :> _
+        MapTree.toArray tree |> Array.map (fun kvp -> kvp.Value) :> _
 
     member m.TryFind key =
 // #if TRACE_SETS_AND_MAPS


alfonsogarciacaro:
Thanks!

PR #2683, opened 2021-12-29
[Python] inlining incorrectly postpones variable mutations

thautwarm:
So far there is a bug described at fable-compiler/Fable.Python#32
This seems not resolved, and I add a dedicated test case to the `beyond` branch.
Thanks to @dbrattli for the instructions.

PR #2687, opened 2021-12-30
Inlining incorrectly postpones variable mutations

dbrattli:
Adding unit-tests from #2683 to JavaScript. One of the tests added by @thautwarm for Python seems to be failing for JS as well.

The problem here seems to be that the right side `TestInliningMutation__call(__)` is inlined instead of being evaluated before the addition.

Failing code:
```js
export function TestInliningMutation__Run(__) {
    return __.left + TestInliningMutation__call(__);
}
```

Expected code:
```js
export function TestInliningMutation__Run(__) {
    const right = TestInliningMutation__call(__) | 0;
    return (__.left + right) | 0;
}
```

### DIFF
--- a/tests/Main/MiscTests.fs
+++ b/tests/Main/MiscTests.fs
@@ -278,11 +278,9 @@ type TestInliningMutation(l: int, r: int) =
             left <- l
             r
 
-        member __.Run() =
+        member _.Run() =
             let right = call()
             left + right
-
-let ``inlineData PR #2683`` =  [3, 2, 5; 5, 10, 15; 10, 20, 30]
 
 module Extensions =
     type IDisposable with
@@ -1198,6 +1196,8 @@ let tests =
         value |> equal 4
 
     testCase "Mutating variables is not postponed (functions)" <| fun () ->
+        let ``inlineData PR #2683`` =  [3, 2, 5; 5, 10, 15; 10, 20, 30]
+
         let runCase (l: int) (r: int) (expect: int) =
             let mutable left = 0
             let call() =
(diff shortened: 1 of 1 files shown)


dbrattli:
Somehow quicktest produces the right non-inlined code:

```fs
type TestInliningMutation(l: int, r: int) =
    let mutable left = 0

    let call() =
        left <- l
        r

    member __.Run() =
        let right = call()
        left + right
```

```js
mport { class_type } from "./fable_modules/fable-library.4.0.0/Reflection.js";

export class TestInliningMutation {
    constructor(l, r) {
        this.l = (l | 0);
        this.r = (r | 0);
        this.left = 0;
    }
}

export function TestInliningMutation$reflection() {
    return class_type("QuickTest.TestInliningMutation", void 0, TestInliningMutation);
}

export function TestInliningMutation_$ctor_Z37302880(l, r) {
    return new TestInliningMutation(l, r);
}

export function TestInliningMutation__Run(__) {
    const right = TestInliningMutation__call(__) | 0;
    return (__.left + right) | 0;
}

function TestInliningMutation__call(this$) {
    this$.left = (this$.l | 0);
    return this$.r | 0;
}
```

dbrattli:
Using `dotnet run . --project ../Fable.Cli` on the QuickTest code above produces the wrong result:

```js
export function TestInliningMutation__Run(__) {
    return __.left + TestInliningMutation__call(__);
}```

ncave:
@dbrattli Seems to work ok in the repl (which is still running 3.4.10), so it must be a relatively recent change. This is perhaps best left for @alfonsogarciacaro to take a look and see if it is related to #2634 (or not).

alfonsogarciacaro:
Actually this seems to be an "old" bug. It probably works in the REPL because by default it compiles in debug mode which does less inlining. Hopefully it [should be fixed](https://github.com/fable-compiler/Fable/commit/d666769f99c2bfff9211edd4a2361382d3fac6c2) in `main` now. I need to finish #2681 to port the fix to `beyond`.

alfonsogarciacaro:
The test is included in #2699 so we can close this, but we keep open #2683 that should work after merging #2699.

dbrattli:
Thanks for fixing @alfonsogarciacaro

Issue #2489, opened 2021-07-21
What are the capabilities / limitations of the plugins?


MangelMaxime:
### Description

Hello,
Fable 3 re-introduced a plugin system but I think it still doesn't have documentation on it.

In current form what are the capabilities and limitation of the plugins?

Here are some questions I have in mind but if you have others things to add to the list please do :)

Can they allow to rewrite a type definition? For example, if I attach a `[<Pojo>]`attribute to a record can I control how it is defined or used?

Can it allow modification of the import paths? For example, could it intercept all the `./.fable/fable-library.3.2.9/XXX` calls and rewrite them to use a `fable-library` which could be an npm package?

Can it prevent generation of some of the code like removing reflection informations?

Can it only rewrite "function call" like Feliz does?

MangelMaxime:
Can the plugins rewrite the module generated code?

For example, for this code

```fs
module TestB =

    let getName (v : TestB) =
        if v.Type = 1 then
            "Class"
        else
            "Function"
```

can it generates something like

```js
export let Config = {}

Config.getName = function (v) {
    if (v.Type === 1) {
        return "Class";
    }
    else {
        return "Function";
    }
}

// or

export const Config = {
    getName: function (v) {
        if (v.Type === 1) {
            return "Class";
        }
        else {
            return "Function";
        }
    }
}
```

instead of 

```fs
export function TestBModule_getName(v) {
    if (v.Type === 1) {
        return "Class";
    }
    else {
        return "Function";
    }
}
```

alfonsogarciacaro:
I should write some documentation about plugins, shouldn't I? 😅 Only thing is plugins have access to the internal AST and this is the only thing that's going to change for JS in Fable 4 so maybe it's better to keep them "a secret" a bit longer.

Anyways, I'll try to quickly summarize the mechanism of plugins: Fable 3 allows for libraries to tell Fable it should load and scan the .dll for "hooks" to different steps of the compilation. This is similar to Type Providers in the sense the user doesn't need to pass explicitly the plugins to the compiler (as it happened with Fable 1). At the moment there're only two such hooks (intercepting function declarations and function calls), we can add more but as said above maybe we should wait until Fable 4 for expanding plugin capabilities. And also like TPs, plugins introduce complexity to the compilation process so it's better to explore alternatives before implementing one.

> Can they allow to rewrite a type definition?

[...]

PR #2710, opened 2022-01-10
Add flag to disable reflection

alfonsogarciacaro:
See https://github.com/fable-compiler/Fable/issues/2489#issuecomment-1007081946

### DIFF
--- a/src/Fable.AST/Plugins.fs
+++ b/src/Fable.AST/Plugins.fs
@@ -26,6 +26,7 @@ type CompilerOptions =
         Verbosity: Verbosity
         FileExtension: string
         TriggeredByDependency: bool
+        NoReflection: bool
     }
 
 type PluginHelper =
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -6,9 +6,11 @@ open Fable
 
 type CliArgs(args: string list) =
     let argsMap =
-        let even = List.length args % 2 = 0
-        // If arguments are odd, assume last has true value in case it's a flag
-        let args = if even then args else args @ ["true"]
+        let args =
+            // Assume last arg has true value in case it's a flag
+            match List.tryLast args with
+            | Some key when key.StartsWith("-") -> args @ ["true"]
+            | _ -> args
         (Map.empty, List.windowed 2 args) ||> List.fold (fun map pair ->
             match pair with
             | [key; value] when key.StartsWith("-") ->
@@ -73,6 +75,7 @@ let knownCliArgs() = [
 
   // Hidden args
   ["--precompiledLib"], []
+  ["--noReflection"], []
   ["--typescript"], []
   ["--trimRootModule"], []
   ["--fableLib"], []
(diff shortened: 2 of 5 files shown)


kerams:
@alfonsogarciacaro, would it be possible to also have an attribute that we could selectively apply on types? Sort of use at your own risk thing? Very often you do need reflection in a part of the application (serializing data contract types in Remoting for instance) but not in another (Elmish models, messages).

For

```fsharp
type Msg =
    | TabClicked
    | UpdatePasswordClicked
    | UpdateProfileClicked
    | UpdateAccountClicked
```

I get

```js
import { Union } from "fable-library/Types.js";
import { union_type } from "fable-library/Reflection.js";

export class Msg extends Union {
    constructor(tag, ...fields) {
        super();
        this.tag = (tag | 0);
        this.fields = fields;
    }
    cases() {
        return ["TabClicked", "UpdatePasswordClicked", "UpdateProfileClicked", "UpdateAccountClicked"];
    }
}

export function Msg$reflection() {
    return union_type("Test.Msg", [], Msg, () => [[], [], [], []]);
}
```

and none of that is treeshaken by Terser. What I would like is `[<Fable.Core.NoReflection>]` that would produce:

```js
import { Union } from "fable-library/Types.js";

export class Msg extends Union {
    constructor(tag, ...fields) {
        super();
        this.tag = (tag | 0);
        this.fields = fields;
    }
}
```

(When no case has any fields, perhaps `fields` could be removed from the ctor? But that's a whole different request.)

kerams:
Actually, looking at that desired output, you could even reuse that definition for all unions with no reflection as there's nothing specific to the original F# DU. The size savings would thus be even larger.

MangelMaxime:
I add completely missed that you added this flag @alfonsogarciacaro. 😅

@kerams I just want to point that in case you are using `Elmish.Debugger` you are going to need the reflection information because it is using Thoth.Json to transform into JSON for passing the information to Redux dev tools.

kerams:
@MangelMaxime, sure but in that case you can conditionally compile the attribute just for production :).

MangelMaxime:
@kerams Yes sure, just wanted to point it out in case you encounter this bug in the future. 

Because, I don't know what error / warning would be emitted by Fable in such a case and if it would be user friendly or not.

kerams:
If you use it in your code and something breaks, it's your problem. It might be more of an issue when it's baked in a third party library.

alfonsogarciacaro:
Thanks @kerams. I think it makes sense to have an attribute to regulate this behavior on a per-type basis. Although we need to be careful because adding more specific behavior increases complexity and sometimes it doesn't actually bring the benefits we expect. Some notes:

- I added this flag because it was a minor thing and in principle it was intended for JS libraries as @MangelMaxime originally asks. I'm not very concerned about the effect of reflection on the bundle size because bundlers should be able to remove the `$reflection` methods if they're not used.
- Your proposal focuses on unions and goes beyond discarding the `$reflection` method, it also discards separated classes for each union. This has many effects on type testing, printing, equality, the ability to implement interfaces, etc. This is tricky because it can be abused quickly, and I know from experience users always want the benefits but none of the drawbacks :wink:
- Because of this, your proposal reminds me more of the previous work by @ncave in #2279 to optimize records and unions by reducing them to their minimal shape. Maybe we could revisit that PR (adapting it so it can be activate per-type with an attribute instead of/as well as with a compiler switch) and give users the option to optimize union/records that are only used as light data-transports.

PR #2711, opened 2022-01-10
[Python] avoid `dict[_, _]` to support py38

thautwarm:
This is just a tiny fix, but this issue suggests that it could be better for us to set up a test that specifies python versions.


dbrattli:
The main target is Python 3.9, but since 3.8 is relatively easy to handle then let's support that as well. I'm making a fix for the build/test action to have a separate Python job with matrix handling (so we don't do JS, Rust testing as well for the different Python versions). Python 3.11 will also be available later this year (2022-10-03), so we don't want to have too long version tail to maintain before we make release ourselves.

PR #2712, opened 2022-01-10
Python version matrix handling

dbrattli:
- Split python tests into separate job that will run concurrently
- Concurrently testing should speed up the test process (we might want to consider the same for Rust)
- Test against multiple python versions 3.8, 3.9, and 3.10

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,14 +8,10 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
-      with:
-        python-version: '3.9'
 
     - name: Setup .NET
       uses: actions/setup-dotnet@v1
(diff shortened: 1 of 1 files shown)


dbrattli:
Looks like all three Python checks actually finished before Fable (JS) itself, so this will speed up the overall test time. Should we do the same for Rust as well? @ncave @alfonsogarciacaro 

Screenshot 2022-01-10 at 17 07 21

ncave:
@dbrattli Perhaps, we could potentially split off other heavy tests in Fable too, if they can be run in parallel.
cc @alfonsogarciacaro

dbrattli:
Yes, that is a great suggestion @ncave Most of the heavy work is with Fable. E.g "Compile itself to JS, then compile again the test suite and run it" should be a good candidate. Every job in the actions file will run concurrently. For a good overview see https://stackabuse.com/getting-started-with-github-actions-software-automation/

alfonsogarciacaro:
This makes a lot of sense @dbrattli, thank you! It will improve the pipeline to separate each language in a different job. About JS, it's true we could parallelize the first test suite compilation/run and Fable self-compilation, though running the test suite now is more or less fast and catches most of the errors, so we don't need to start self-compilation if CI is going to fail anyways.

There're actually other tests that take a long time: `tests/Compiler`. This is because it [references Fable.Standalone](https://github.com/fable-compiler/Fable/blob/b8971666bcc3e464072db23f776baf515d6c8126/tests/Compiler/Compiler.fsproj#L15) which needs to be compiled. It'd be nice to remove this reference as Fable.Standalone is not meant to be used in .NET. For that we should enable [the APIs the Compiler tests need](https://github.com/fable-compiler/Fable/blob/b8971666bcc3e464072db23f776baf515d6c8126/tests/Compiler/Util/Compiler.fs#L44-L61) in Fable.Cli and maybe also integrate those tests with `tests/Integration` and `tests/ProjectConfigs`. Then we could set the integration tests also as a indepentent job from unit tests for each language.

alfonsogarciacaro (approved):

PR #2717, opened 2022-01-11
File lock

alfonsogarciacaro:
Fable precompilation is causing issues when two processes try to trigger precompilation of a common library at the same time. This tries to avoid that.

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -268,10 +268,15 @@ type Runner =
             |> Some
         else None
 
-    return!
+    let startCompilation() =
         State.Create(cliArgs, ?watchDelay=watchDelay)
         |> startCompilation
         |> Async.RunSynchronously
+
+    return!
+        match outDir with
+        | None -> startCompilation()
+        | Some outDir -> File.withLock outDir startCompilation
 }
 
 let clean (args: CliArgs) rootDir =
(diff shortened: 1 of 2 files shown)


ncave:
@alfonsogarciacaro Probably unrelated, but what is the mechanism by which Fable decides if file was already compiled?
I'm just asking to see if there is a way to turn it off for repeated debugging, without modifying the source file every time.

alfonsogarciacaro:
@ncave I've updated the description with a rationale for this PR. Also you're right the way to decide when to skip compilation should be better documented. In Fable 3.7 is a 3-step process:

1. Check if the cached info for project cracker is up-to-date: https://github.com/fable-compiler/Fable/blob/31e1d46ffa80c567ef58d920b6a178173cb90d6b/src/Fable.Cli/ProjectCracker.fs#L606-L619

2. If it is, check if the most recent compilation corresponded to the same config (debug/release) and used the same compiler options: https://github.com/fable-compiler/Fable/blob/31e1d46ffa80c567ef58d920b6a178173cb90d6b/src/Fable.Cli/ProjectCracker.fs#L625-L629

3. Then check if there are already compiled files with a more recent timestamp (not sure if there's a better method): https://github.com/fable-compiler/Fable/blob/31e1d46ffa80c567ef58d920b6a178173cb90d6b/src/Fable.Cli/Main.fs#L615-L621

To disable this, you can pass the `--noCache` option.

PR #2725, opened 2022-01-13
Add --delimiter option to keep contents of target files

alfonsogarciacaro:
### UPDATE

This should be ready to merge @dbrattli. As discussed below, the PR has been repurposed to add a `--delimiter` option so you can create "regions" in the target file. This way Fable won't overwrite the whole contents of the file but will put the generated code in the specific region and keep the rest. The main use case is when using Python interactive so you can easily resend the F# functions after updating them and have the immediately available in Python scope.

For now this is only available in Python but it's easy to implement it in other languages if needed. I've also added a `region` helper to Fable.Core so you can declare regions in F# and Fable will try to match them with regions surrounded by the delimiters in the target file. If it can't it will just print the code at the end of the file (should we error/warn the user in that case?). I guess most of the times a single region will be used but it can be useful if you need to separate the imports and the rest of F# code.

Besides this, the PR does a couple of other things:

- Refactor the `Printer` interfaces so all languages use the same one. For this I've also touched of the Rust code, please let me know if that's ok @ncave
- Add support for compiling .fsx scripts including Nuget references. Though it probably needs more testing. For example, it seems the F# logs get lost. Not sure why this is, maybe FCS/service_slim has trouble because of the extension?

### ORIGINAL POST

This is my first PR for Python!

I'm testing Fable Python with VS Code interpreter. It works nicely and I find it easier to use than Jupyter notebooks (I need to learn more about how to write bindings though).

I can write a helper to add the the `# %%` comment VS Code uses to delimit the cells. However I cannot print anything before the imports which is cumbersome because I need to rerun the imports when adding a new package.

screencast

The comment is just hard-coded for now and I hope it doesn't break anything else. Not sure if there's a better way to make it configurable somehow, what do you think @dbrattli?


alfonsogarciacaro:
BTW, this is the code I used in the example (note I'm using the bang trick in `Emit` so the string is printed as is):

```fsharp
open System
open Fable.Core
open Fable.Core.PyInterop

[<Emit("# %%")>]
let cell<'T> = ()

cell
printfn "Hello from Fable Python!"

module Bindings =
    [<Erase>]
    type pyplot =
        abstract show: unit -> unit

    [<Erase>]
    type matplotlib =
        abstract pyplot: pyplot
        [<Emit("$0._enums.$1!.demo()")>]
        abstract demo: key: string -> unit

open Bindings

cell
let matplotlib: matplotlib = importAll "matplotlib"
matplotlib.demo("CapStyle")
matplotlib.pyplot.show()

// Add a cell at the end to prevent VS Code
// from doing it automatically
cell
```

dbrattli:
This is super cool @alfonsogarciacaro 🎉  

One idea would be to make it easier in general to generate comments in Fable generated code. Emit is one possibility. Are there others? Any previous work on this? Any possibility of F# comments `//` being transpiled? They are not part of the typed AST so perhaps not.

Another way we could do this is to add an extra argument to `Emit` to generate a file header. Thus if `header=true` or  `front=true` then we will/should emit it before anything else such as open statements.

We could also detect the use of emitted cells in `transformEmit` and conditionally generate the first one `# %% Fable Python` if cells are being used. Perhaps add the `let ``# %%``<'T> = ()` to `Fable.Core.PyInterop` and detect if it's being used or not.

alfonsogarciacaro:
Thanks for the reply @dbrattli. Yes, the comments are not in the AST so this is going to be difficult (I think you can access the XML doc comments though).

I need to think a bit about this. After playing with this more it has become clear to me we need to fully integrate F# and Python.  When I was learning about Machine Learning a couple of years ago I realized with Python you copy a lot of code from documentation, tutorials. Trying to do as we've done with JS (interop techniques, bindings) is going to be a lot of work and may not be enough. I tried using the F# Template Highlighting extension, but I could only make the syntax highlighting work, not autocompletion or tooltips.

Another idea that just came to my mind is to add a "delimiter" option, so instead of fully overwriting files, Fable can decide which regions should be overwritten and which kept as is so it's possible to mix hand-written and auto-generated python in the same file. This shouldn't be difficult to implement in Fable and will give us a similar experience as with Jupyter notebooks but with the advantages of working in a proper IDE :)

I'm starting to get really excited about the possibilities of the F#/Python integration. Thanks a lot for making it possible @dbrattli!

dbrattli:
@alfonsogarciacaro You should also try Fable.Jupyter. It works within vscode just by opening .ipynb files. You can select F# (Fable Python) kernel. I have just fixed it so it works with Windows again (after choking on the new same line output). Found the CI env option to get simpler output: https://pypi.org/project/fable-py/

dbrattli:
FYI: you can actually use the Fable Python kernel from a python (.py) file and send F# snippets instead. In the kernel we can decide if we want to compile as Python or as F#. So if we could embed F# some way within Python files then it could work some way. If you don't understand what I'm talking about then I could show you on e.g discord when you have time

alfonsogarciacaro:
Yes, probably what I'm dreaming of can already be done with notebooks. I think I had Fable.Jupyter working at some point but I tried yesterday (in Linux) and I couldn't make it work. And because I couldn't see any logs or anything I was very confused. I'm likely missing something very obvious, so I'll try to ping you in Discord later.

Besides making it work, there are a couple of things I'd like to have with the notebooks. Do you think they're feasible?

- Having the same development experience when using a notebook as when using an IDE (both for Python and F#). I think it's possible to open .ipynb files in VS Code, but not sure if you'll get the F# tooling work in that case.
- For Fable development, have the same experience I have with Quicktest: that is, have Fable run in watch mode, but if I make a change in the compiler it will restart automatically

dbrattli:
Yes, we need to think of both. Normal Python (vscode) development and Jupyter. For Jupyter the F# tooling does not work. Syntax highlighting works, but I haven't integrated with LSP yet. [Here's](https://github.com/fable-compiler/Fable.Jupyter/blob/main/fable_py/kernel.py#L326) where it needs to be added. So we need to start it the same way I start Fable, and then use some LSP library to talk with it.

For Jupyter you can get quck-test functionality by not PIP installing it the normal way (and a quick fix to the kernel):

1. pip uninstall fable_py # uninstall the current kernel
2. pip setup.py develop # local installation (link)
3. python -m fable_py install # registers the fable_py kernel with jupyter

Then we change [this line](https://github.com/fable-compiler/Fable.Jupyter/blob/main/fable_py/kernel.py#L144) to do ` dotnet run -p ../../../Fable/src/Fable.Cli --lang Python -outDir ...` instead. Then Fable complier changes will be visible directly while changes to the Jupyter kernel (kernel.py) needs a kernel restart (menu or `%reset` magic command). For Fable compiler we can probably have an environment setting so it's easy to change

alfonsogarciacaro:
Thanks for the explanations @dbrattli! It'd be nice to have F# working as a first-class citizen in Jupyter notebooks. For now, I managed to get a proof of concept of compilation with delimiters (and using an .fsx with Nuget references too!).

screencast

dbrattli:
That's amazing!! 😱 I need to take a closer look at this!!

dbrattli (approved):
This is a fantastic contribution. Very nice!! PS: you might want to split quicktest and add a quicktest-py directory, so we don't overwrite the original quicktest code

alfonsogarciacaro:
@dbrattli I've updated the first post and the title :)

dbrattli:
Looks very nice! It add an alternative way for interoperability with other languages so looking forward to try it out. Thanks for fixing this 🎉

dbrattli:
@alfonsogarciacaro When trying this feature (quicktest-py) I get an exception:

```
Unhandled exception. System.Exception: No common base dir
   at Fable.Cli.Main.FileWatcherUtil.getCommonDir@181(FSharpList`1 restDirs, String dir) in /Users/dbrattli/Developer/Github/Fable/src/Fable.Cli/Main.fs:line 189
   at Fable.Cli.Main.FileWatcherUtil.getCommonBaseDir(FSharpList`1 files) in /Users/dbrattli/Developer/Github/Fable/src/Fable.Cli/Main.fs:line 191
   at Fable.Cli.Main.FsWatcher.Observe(FSharpList`1 filesToWatch) in /Users/dbrattli/Developer/Github/Fable/src/Fable.Cli/Main.fs:line 229
   at Fable.Cli.Main.Watcher.Watch(ProjectCracked projCracked) in /Users/dbrattli/Developer/Github/Fable/src/Fable.Cli/Main.fs:line 532
   at Fable.Cli.Main.state@662.Invoke(Watcher w) in /Users/dbrattli/Developer/Github/Fable/src/Fable.Cli/Main.fs:line 662
```

The argument to `getCommonBaseDir` is:
```
["/Users/dbrattli/Developer/Github/Fable/src/quicktest-py/quicktest.fsx";
 "/var/folders/40/6klqtrv17txdp1x8zkv40bg40000gn/T/12476--3c9f3d43-2ae0-45db-8a8a-db45055145da/Project.fsproj.fsx";
 "/Users/dbrattli/Developer/Github/Fable/src/quicktest-py/quicktest.fsx"]
```

So the problem is the `/var/folders/...` entry. Any ideas?

alfonsogarciacaro:
Ah, interesting. It looks as if the F# compiler has created a  temporary `Project.fsproj.fsx` file. Didn't know this was the case, my guess this is done to resolve the Nuget references. If this is something more or less "standard" I guess we can just filter out files ending with `.fsproj.fsx`.

alfonsogarciacaro:
It looks to be this: https://github.com/dotnet/fsharp/blob/5819395514b334aecf268f9c13d2d19965d9690d/src/fsharp/FSharp.DependencyManager.Nuget/FSharp.DependencyManager.fs#L288

dbrattli:
Thanks, I will try this later and see how it works

PR #2729, opened 2022-01-14
Pragmatically fix test with culture specific decimal separator

andagr:
This PR fixes the last two tests that failed in sv-se (and others?) culture.

The fix is a bit, ehm, crude but:
In many cultures, e.g. sv-se, commas are used for decimal separator.
In an interpolated string this can normally be controlled by using
the class `FormattableString.Invariant()`.
However, we also want to test %-specifiers, but that cannot be
combined with `FormattableString.Invariant()`, only general
.NET-style specifiers work. Therefore, a simple replace of commas to
the invariant style dots is used.

### DIFF
--- a/tests/Main/StringTests.fs
+++ b/tests/Main/StringTests.fs
@@ -227,7 +227,7 @@ let tests =
           |> equal "Name: Phillip, Age: 29"
 
       testCase "string interpolation works with inline expressions" <| fun () ->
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Actually we have the same "hack" in a couple of other places 😅 so it should be fine, thanks!

PR #2730, opened 2022-01-15
[Python] Records should be hashable

thautwarm:
fix fable-compiler/Fable.Python#44


dbrattli:
One thing is that I'm considering transpiling Records as Python dataclasses i.e `@dataclass(frozen=True)`. I need to investigate how this will work with `__slots__`.  Ref: https://github.com/ericvsmith/dataclasses/issues/28

thautwarm:
I'm not sure if this is a good idea. Dataclasses are pretty useful in pure Python, but as the generated code, I doubt if there are advantages over classes that are specifically designed for F#.

There are many incompatible parts, for instance, we don't really use the same equality as dataclasses, which means that `frozen=True` might not be that useful to us.

In fact, `__eq__`, `__hash__`,  `__repr__`(for F# compatibility), comparisons(`__gt__` stuffs), many of them are not applicable to F# records...

dbrattli:
The idea is that it's common to use data classes for e.g Python serialization libraries and I would expect they introspect the class to know what fields to serialize. Thus there should be some way to generate such classes from Fable in order to be interoperable with such libraries. Not sure I understand the problem with the methods you mention. They will not be generated if they are already present (i.e generated by Fable). 

For example Faust [models](https://faust.readthedocs.io/en/latest/userguide/models.html), but I now see they have another problem that records cannot inherit from Faust `Record` anyways. Perhaps we instead should have a `[<DataClass>]` attribute to say that normal classes (not records) we generate should become a Python dataclass!?

thautwarm:
Okay, I didn't consider serialization as I've implemented seriaization using reflection in F#(that's why I ask features for faster serializations)... 

> Not sure I understand the problem with the methods you mention. They will not be generated if they are already present (i.e generated by Fable).

My initial feeling is that if we need to generate most of the record methods ourselves, it might not be a good idea to generate dataclasses. 

> Perhaps we instead should have a [<DataClass>] attribute to say that normal classes (not records) we generate should become a Python dataclass!?

It looks good to me, but as you've mentioned serialization support for dataclasses, I need to consider more about this issue.

dbrattli (approved):
Very nice, thanks for fixing!

PR #2731, opened 2022-01-15
[Python] Fix accessing DU tags

thautwarm:
fix fable-compiler/Fable.Python#41

The way to define abstract static methods comes from [docs.python.org/3/library/abc](https://docs.python.org/3/library/abc.html):

```python
@staticmethod
@abstractmethod
def cases():
      ...
```


dbrattli (approved):
Looks good. Thanks for fixing!

thautwarm:
Thanks for the review!

PR #2734, opened 2022-01-15
Use a local link for FCS.

philderbeast:
On the main branch, the readme has a broken link to [FSharp Compiler Services](https://fsharp.github.io/FSharp.Compiler.Service/). I found https://github.com/dotnet/fsharp and https://github.com/fable-compiler/Fable/tree/main/lib/fcs. Since we're using the later anyway and it has its own readme then why not link there?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 [Follow us on Twitter!](https://twitter.com/FableCompiler)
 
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Hmm, not sure if a link to the local folder is going to be very informative, we should probably fix it so it points to the correct FCS documentation: https://fsharp.github.io/fsharp-compiler-docs/

In principle the fact that we're using a fork with a couple of tweaks should be transparent to the user. But you're right we should disclose it for the sake of transparency, we could add a note for that and point to the branch used to generate the binary: https://github.com/ncave/fsharp/pull/2

alfonsogarciacaro:
Thanks!

PR #2737, opened 2022-01-16
Fix typeof<obj>.IsInstanceOfType

chkn:
`typeof<obj>.IsInstanceOfType(...)` should always return true. Currently it always returns false.


alfonsogarciacaro:
Thanks!

PR #2738, opened 2022-01-16
[Python] Use Fantomas formatting and check CI

dbrattli:
So this PR might be a bit controversial with Fable 😬 but I think it's mostly a good thing with a common code style when we have many contributors. So starting with Python related code we will now enforce the formatting based on the latest Fantomas 4.6 (set as a dotnet tool) and the rules set in .editorconfig. I'm pretty agnostic to what the rules should be as long as we have a set of common rules. The current .editorconfig tries to match the current code style, but please make PR's to modify if to the Fable style.

This PR will only enforce the formatting of `src/Fable.Transforms/Python` but we should consider adding this to the rest of the code base as well. 

To format the Python code, you will need to run: 

```sh
> dotnet tool restore
> dotnet fantomas src/Fable.Transforms/Python
```

Fantomas is added to GitHub actions so if the Python related code is not formatted according to the rules, the build will fail. Please chime in if you do not want this kind of restrictions in Fable (even if only applied to the Python specific code) @alfonsogarciacaro @ncave


ncave:
@dbrattli Some of the code base in Fable comes from `dotnet/fsharp` where the isn't *any* formatting, not even EOL white-space removal. So yes, IMO it would be preferable if we keep the original formatting and not enforce Fantomas on everything.

dbrattli:
@ncave It's possible to ignore parts of the code-base using `.fantomasignore`, so copied / auto-generated code can be ignored from formatting. https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#ignore-files-fantomasignore

src/Fable.Transforms/Python/Fable2Python.fs
```diff
@@ -548,17 +653,22 @@ module Helpers =
...
+            (lower, Naming.NoMemberPart)
+            ||> Naming.sanitizeIdent (fun _ -> false)
 
         let path =
             match relative, normalizedPath with
             | _, "" -> ""
             | true, "." -> "."
             | true, path ->
-                path  // translate path to Python relative syntax
+                path
```
alfonsogarciacaro: Fantomas has removed the comment here.

alfonsogarciacaro:
Yes, I've been thinking about adding Fantomas many times, but at the end I discarded by issues like the one pointed out by @ncave. Particularly right now it can make keeping `main` in `beyond` in sync more difficult, so I would postpone the decision at least until `beyond` replaces `main`. It should be fine to apply Fantomas to `Fable.Transforms/Python` though.

src/Fable.Transforms/Python/Fable2Python.fs
dbrattli: Yes, I've seen this happen a couple of times. Added issue at: https://github.com/fsprojects/fantomas/issues/2027

dbrattli:
Ok, thanks for the comments. It makes sense to wait with this until after beyond merges with main. I'll still add this for Python to test if it can work, but will remove the CI hook if it becomes to annoying for the rest of you.

.editorconfig
```diff
@@ -7,3 +7,12 @@ indent_style = space
 # TS files
 [*.ts]
 indent_size = 2
+
+# Fantomas (see https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md)
+[*.fs]
+indent_size=4
+max_line_length=140
+fsharp_space_before_uppercase_invocation=false
```
nojaf, line 15: Out of curiosity, why are you mentioning these settings here?
They have the same values as the default ones, so this won't do anything.

dbrattli: @nojaf Yes, you are right. It seems I had a .editorconfig in a parent dir as defaults and forgot about it (scary). I'll remove it and simplify this file. Thanks for letting me know. I have wondered about that a couple of time.

PR #2744, opened 2022-01-18
Fix build for .fsx files

dbrattli:
- Remove auto-generated project file


alfonsogarciacaro (approved):
Maybe we can add a comment like: F# compiler may include files with this extension in temporary folders for resolution of Nuget references in script files.

PR #2747, opened 2022-01-19
Put integration/js tests in different folders

alfonsogarciacaro:
~~I wanted to put the integration tests in a different job, but some of them still depend on the fable-library build so for now I kept them within the JS tests to avoid having to build fable-library multiple times, but we may consider to do that later.~~


alfonsogarciacaro:
I wanted to put the integration tests in a different job, but some of them still depend on the fable-library build so for now I kept them within the JS tests to avoid having to build fable-library multiple times, but we may consider to do that later.

dbrattli:
@alfonsogarciacaro To improve on this I think you would have to use e.g the `@cache` action:

- https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
- https://levelup.gitconnected.com/github-actions-how-to-share-data-between-jobs-fc1547defc3e

And then use the `need` keyword to serialize the jobs dependent on the build:

- https://docs.github.com/en/actions/using-workflows/advanced-workflow-features

dbrattli:
But building fable-library multiple times should not really be a problem since it takes less than a minute of the total build time, so the overall build time should become a lot better even if building fable library multiple times.

alfonsogarciacaro:
Thanks for the input @dbrattli! I tried to do it in #2750 although I'm missing the cache to keep the `build/fable-library` dir between the jobs.

PR #2757, opened 2022-01-20
Fix Match.Groups

alfonsogarciacaro:
@dbrattli Fix for `Match.Groups`, please note this PR targets your python-regexp branch.


dbrattli (approved):
@alfonsogarciacaro Thanks for the help, Regex can be tricky, but it wasn't as bad as I had feared.

PR #2759, opened 2022-01-21
[Python] Revert TypeInfo caching

dbrattli:
- Use tuple for `FieldInfo`, not a list
- Format code


dbrattli:
FYI: @thautwarm The caching is reverted for now. Is it perhaps possible for you to do any caching in your library instead?

PR #2764, opened 2022-01-24
Add --noParallel option

alfonsogarciacaro:
In some big projects watch compilations can take more than 10 seconds. Because now there's more gap between each file compilation, Webpack triggers multiple times which can be annoying. This PR adds the `--noParallel` to have a "classic" compilation where FCS ends type checking all files before Fable kicks in. Note the name is a bit misleading because Fable compilation itself still keeps running in parallel, but I couldn't come up with something better.

~~There's also a fix to "watch" .proj files, as some projects use them. Though it's not ideal because .proj files are not included in the project references returned by the cracker so they remain invisible to Fable.~~

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -77,6 +77,7 @@ let knownCliArgs() = [
   // Hidden args
   ["--precompiledLib"], []
   ["--noReflection"], []
+  ["--noParallel"], []
   ["--typescript"], []
   ["--trimRootModule"], []
   ["--fableLib"], []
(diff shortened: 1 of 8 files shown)


adelarsq:
> Note the name is a bit misleading because Fable compilation itself still keeps running in parallel

@alfonsogarciacaro  What about `--runCompilerServiceTypeCheckFirst` or `--runCompilerServiceTypeCheckBefore`?

alfonsogarciacaro:
Maybe a bit long, but `--noParallelTypeCheck` should work, thanks a lot for the suggestion @adelarsq!

adelarsq:
@alfonsogarciacaro I agree. I think that `--noParallelTypeCheck` is better.

PR #2765, opened 2022-01-24
Dart test

alfonsogarciacaro:
Just because I want another green tick in CI builds 😸 

Also, given that there's been tremendous progress in the Python and Rust targets, I'm giving up on the idea for now to create a extended set of the Fable AST that we could use to perform some of the transformations that are happening now at the last step of the compilation (tail-call optimization, variable hoisting...). So Dart will have its own AST and Fable2Dart step (and later its own Replacements module) as the other targets.


ncave:
@alfonsogarciacaro Sounds good. We do need to extract some common functionality from Replacements (i.e. all Replacement methods that FSharp2Fable calls), with a default implementation, so they can be implemented differently by each language.

alfonsogarciacaro:
You're definitely right @ncave, I had something similar in mind. The Replacements module is the part I'm less proud of in Fable because it's a big mess... and now we have this mess multiplied by 3 :)

When we created the `beyond` branch I wanted to give Replacements an interface and pass it to the compiler so we could have different implementations depending on the language target. Actually I think I started some work for it a few months ago in a branch but it was a bit more refactor than I anticipated. In any case, we should probably try to clean it up little by little by giving it a diamond structure as we're doing with the compiler pipeline:

```
                 Replacements.Util (common parts)
Replacements.Rust    Replacements.Python     Replacements.JS
                 Replacements (API exposed to FSharp2Fable and others)
```

The Replacements API should have the language switch so FSharp2Fable can just make single calls. When the API is stable we can consider if turning it into an interface.

ncave:
@alfonsogarciacaro Yes, something like that. It's very nice to be able to update individual Replacements without worrying about impact to other languages, but the cost is increased maintenance when AST is updated or features are added, so we have to strike a balance.

PR #2770, opened 2022-01-25
[Beyond] Replacements cleanup

alfonsogarciacaro:
This is a first step to do the Replacements cleanup discussed [here](https://github.com/fable-compiler/Fable/pull/2765#issuecomment-1020209453). Basically trying to avoid (too much) duplicated code and expose the same surface to the rest of modules (in particular FSharp2Fable) so we don't need to check the language at every call site. Some notes:

- Wanted to rename (JS) Replacements but I was afraid this would make syncing with master more difficult so for now the "common" surface is named `Replacements.Api`.
- The helpers for (un)currying runtime helpers (`curryExprAtRuntime` and friends) are JS-only atm, not sure how they beahve for other languages.
- @dbrattli @ncave If you see more opportunities to refactor code please go ahead. For example, it'd be nice to also move `applyOp` to Replacements.Util. There are other language-dependent operations we could abstract somehow, like sanitizing identifiers.

@ncave It looks I've broken the Rust compilation. If you've a moment could you please take a look? I'll try to review tomorrow nonetheless.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -22,7 +22,7 @@ let private transformBaseConsCall com ctx r (baseEnt: FSharpEntity) (baseCons: F
     let argTypes = lazy getArgTypes com baseCons
     let baseArgs = transformExprList com ctx baseArgs |> run
     let genArgs = genArgs |> Seq.map (makeType ctx.GenericArgs)
-    match Replacements.tryBaseConstructor com ctx baseEnt argTypes genArgs baseArgs with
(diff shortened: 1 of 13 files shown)


dbrattli:
This looks very promising, thanks for doing this @alfonsogarciacaro. I must say that I didn't expect things to go so well with multiple languages with separate tests for each language, but it looks like we might pull this off after all 😅

alfonsogarciacaro:
@dbrattli Indeed, I'm also surprised of how smoothly targeting multiple languages is going 😸 Since having different jobs in the build it's become much easier to pinpoint which problem affects which language, thanks a lot for that!

PR #2774, opened 2022-01-27
[Dart] Imports

alfonsogarciacaro:
First support for imports in Dart and add a few more binary expression tests.

@dbrattli @ncave In this PR I'm also trying to unify the code to decide the extension of the generated files based on the language. I realized we had two methods to change extensions, the `ChangeExtension` polyfill and `replaceExtension` with slightly different behavior so I've removed the latter for simplicity. I've also made more explicit (`changeExtensionButUseDefaultExtensionInFableModules`) a mild hack I added when transitioning to Fable 3: files compiled within `fable_modules` would always use `.fs.js` extension, because in some packages they can conflict with native .js files. I've restricted the hack now to JS as we should avoid this practice in other languages and always use different names for the files regardless of the extension.

### DIFF
--- a/tests/Dart/Tests.Util.fs
+++ b/tests/Dart/Tests.Util.fs
@@ -1,0 +1,15 @@
+module Fable.Tests.Dart.Util
+
+open Fable.Core
+
+type Assertion<'T> =
+    interface end
+
+[<ImportAll("package:test/test.dart")>]
+type Test =
+    static member test(msg: string, f: unit -> unit): unit = nativeOnly
+    static member expect(actual: 'T, assertion: Assertion<'T>): unit = nativeOnly
+    static member equals(value: 'T): Assertion<'T> = nativeOnly
+
+let testCase msg f = Test.test(msg, f)
+let equal expected actual = Test.expect(actual, Test.equals(expected))
--- a/build.fsx
+++ b/build.fsx
@@ -505,7 +505,7 @@ let testRust() =
     copyFile (projectDir </> "Cargo.toml") buildDir
     runInDir buildDir "cargo test"
 
-let testDart() =
+let testDart(isWatch) =
     let projectDir = "tests/Dart"
     let runDir = "tests/Dart/run"
 
@@ -513,6 +513,8 @@ let testDart() =
         "--exclude Fable.Core"
         "--lang Dart"
         "--noCache"
+        if isWatch then
+            "--watch"
     ]
     runInDir runDir "dart test main.dart"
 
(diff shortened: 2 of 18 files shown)


tests/Dart/Tests.Util.fs
```diff
@@ -0,0 +1,15 @@
+module Fable.Tests.Dart.Util
+
+open Fable.Core
+
+type Assertion<'T> =
+    interface end
+
+[<ImportAll("package:test/test.dart")>]
```
Nhowka, line 8: That's pretty cool! I'm wondering if we can fork the [dartdoc](https://github.com/dart-lang/dartdoc/) package to get the public api and generate bindings for any other package or even read the `pubspec.yaml` and generate a file with bindings for all the current package dependencies.

alfonsogarciacaro: Thanks for checking @Nhowka! I think that'd be great idea. I'm still learning Dart so if someone with more experience could have a look at forking dartdoc that'd be awesome.

Note that bindings for Dart will likely be slightly different than in JS. For JS we usually define an interface and then assign it to a value. In Dart it's not possible to assign a module to a value, so I assume we'll write the bindings usually as a static class with dummy member bodies (`nativeOnly`) as in here. We've also discussed about doing the same in JS but there're already many bindings written with the "interface style".

ncave: > we'll write the bindings usually as a static class with dummy member bodies (nativeOnly) as in here. We've also discussed about doing the same in JS but there're already many bindings written with the "interface style".

@alfonsogarciacaro
Can you give an example or point to the discussion, we're looking for the best representation of ad-hoc bindings for Rust too.

alfonsogarciacaro: As usual the discussion is scattered in different places, but I think there was a summary here: https://github.com/fable-compiler/fable-browser/issues/77

Nhowka: I'll take a shot on dartdoc, maybe generating a JSON representation with all the surface might be reusable on that side and we can consume it to generate our bindings.

Would it be feasible to attribute an entire module with an `Import`? I'm wondering that because as Dart has top-level methods and fields besides the instance ones, so maybe we could leverage that to make a better 1:1 representation of the api. Also as Dart has no inner classes nor overloads, top level functions and values would be the static top-level methods and fields, and types and/or interfaces would represent the classes. For the generated file, the import could also have the same name as the module, like

```fs
[<Import("package:some_package/some_file.dart")>]
module SomeFile = ...
```
turning into
```dart
import "package:some_package/some_file.dart" as SomeFile;
```

Dart is kinda quirky as every class is also an interface that can be implemented, it can be somewhat useful but I'm not sure how we could or even should represent that... Named constructors and factories could probably be encoded as static methods returning the class without any issues.

alfonsogarciacaro: That's a good appreciation @Nhowka! Though I think it may be a bit difficult. Let's move the discussion to a separate issue so we can share the experience with other languages 👍 #2779

PR #2773, opened 2022-01-25
[Python] Fix delimiter issue

dbrattli:
There is an issue if you have the black formatter installed (you should) that it will automatically rewrite `#fsharp` to `# fsharp` since the style guide [PEP 8 states about comments](https://www.python.org/dev/peps/pep-0008/#inline-comments): "They should start with a # and a single space.". So black will rewrite the delimiter, and then you suddenly have a double set of delimiters in the file, both `#fsharp` and `# fsharp`, and then black will again rewrite `#fsharp` ...

PS: It looks a bit scary to have a space ` ` in the build command, but using quotes or `\ ` did not work so I suppose it's the correct way!?

PR #2777, opened 2022-01-27
[Beyond] Building standalone in beyond

dbrattli:
I was looking at what it would take to compile the Fable compiler to Python by looking at fable-compiler-js and Fable.Standalone. In order to have `dotnet fsi build.fsx standalone` work I had to make these fixes to handle the removal of `ITypeResolver` and also have a 4.0.0 version of Fable.Core with e.g `nativeOnly` defined. So this means that the worker cannot be part of the current CI. What is the btw role of the worker? I guess I just need to build `dotnet fsi build.fsx compiler-js` instead and look at making e.g `fable-compiler-py`?

### DIFF
--- a/src/fable-standalone/src/Worker/Worker.fsproj
+++ b/src/fable-standalone/src/Worker/Worker.fsproj
@@ -11,7 +11,7 @@
     <Content Include="util.js" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Fable.Core" Version="3.1.6" />
-    <PackageReference Include="Thoth.Json" Version="4.0.0" />
+    <PackageReference Include="Fable.Core.Experimental" Version="4.0.0-alpha-030" />
+    <PackageReference Include="Thoth.Json" Version="7.0.0" />
   </ItemGroup>
 </Project>


alfonsogarciacaro:
Wow, are you already thinking of compiling the F# compiler to Python? Remember that you may face some issues with Stack Overflow exception for very long functions (as in [here](https://github.com/fable-compiler/Fable/pull/2773#issuecomment-1022861676)), you will need to debug Fable to know which functions should be broke into parts.

Yes, in principle you should try to adapt compiler-js. Fable standalone is designed to be used from either Node.js (compiler-js) or from the REPL. In the case of the REPL the Worker is used to interact with the main app, as Fable can freeze the UI if running in the same thread. If I'm not mistaken, the Worker is not used in the CI of this repo but we should keep it compilable for when we need to update the REPL.

dbrattli:
Thanks for the info. I see that my Mac consistently gives stack overflow on standalone, but it compiles on Windows. I was silly enough to google "stack overflow fable ...", but then just got Fable results at stackoverflow.com 🤣

ncave:
@dbrattli You can also take a look at the `fable-standalone`'s [test compiler](https://github.com/fable-compiler/Fable/tree/beyond/src/fable-standalone/test/bench-compiler) project, basically the same as `compiler-js`, just easier to test with (uses project references to `fable-standalone`, doesn't use the Worker).

Issue #2781, opened 2022-01-28
[Beyond] Numeric types


alfonsogarciacaro:
Right now, Fable only includes as "primitive" numeric types those representable in JS. Not sure how are you solving for Python and Rust but I'm realizing this is very limiting. We should likely at least include Int64/UInt64, if a target language doesn't have support for it they should deal with that in `Replacements.makeTypeConst` and then in the Fable2XXX step just fail with "no supported" if a numeric literal of this type remains in the AST (shouldn't happen).

> BTW, it looks that JS BigInt has [good support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#browser_compatibility) so this is also good timing to support longs natively in JS.

I'm less sure about BigInt and decimal because they're classes in disguise. Though it seems JS BigInt is not only for longs and be arbitrarily large (except in TypedArrays where it's constrained to 64bits). Thoughts? @ncave @dbrattli

ncave:
@alfonsogarciacaro Makes sense. For Rust I get around the lack of 64-bit ints in the Fable AST by serializing (encoding) them as [floats wrapped in an enum](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Transforms/Rust/Replacements.fs#L56-L63). It works, but having proper 64-bit ints in the AST is of course much clearer.

dbrattli:
Not really sure why things work, just focusing on getting the unit-tests to pass 😄  For Python I compile everything to Python `int` since they are variable in length. BigInt is also compiled to `int`. Long uses the low / high bits stuff. For Decimal I compile them to the Python Decimal type. But it would be really great if the AST would  have some proper value types (float / int) instead of "number".

alfonsogarciacaro:
Fixed by #2789

PR #2789, opened 2022-01-29
[Beyond] Add Int64/UInt64 to Fable AST

alfonsogarciacaro:
Edit: See #2781 and comments below.


ncave:
@alfonsogarciacaro I would recommend if you're changing the AST anyway, you should also add the F# `nativeint` and `unativeint` for full fidelity of the primitive numeric types.

alfonsogarciacaro:
Thanks for the suggestion @ncave. At the end I've also added Decimal and BigInt to NumberKind. Decimal, because the documentation lists it as one of [the basic types](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/basic-types) and BigInt to prevent it from remaining an outlier, as we had a couple of patterns to include it with other numeric types, and also Python and JS now can support bigints natively.

It's been considerable work but I hope it's for the best. I've also touched a lot of Rust/Python code to make things compile, I hope it's ok but feel free to change whatever you think it's necessary. One thing to note is I moved `makeTypeConst` from Replacements to Transforms.Util so each Fable2XXX module is responsible to convert the `NumberConstant`s to whatever works for their languages.

However, there's still work to do as CI is not passing. I need to fix the tests for JS (13 failing), but could you please take a look at Rust/Python? @ncave @dbrattli

@ncave It seems the Rust build is failing because in some number constants the `NumberKind` and actual type don't match. Not sure what's the best way to solve this, so I'm leaving it up to you 😸 Either ignoring the kind and only checking the boxed value or trying to find the places where the NumberConstant is built with the wrong kind and fix them. https://github.com/fable-compiler/Fable/blob/349bca7cca8cb861e6b32803f417f77126bc362c/src/Fable.Transforms/Rust/Fable2Rust.fs#L1617-L1626

@dbrattli I had to disable the Fantomas check in test.yml. This is because Fantomas complained Python/Replacements.fs was not valid F# code even if it compiles correctly. Can you please have a look? Also, please note that I've removed `System.Web.HttpUtility.JavaScriptStringEncode` from PythonPrinter. The `EscapeStringLiteral` method was added to the IWriter interface to make BabelPrinter compilable to JS, as we don't have replacements for `System.Web.HttpUtility` (in JS we use `JSON.stringify`). However I think JS strings supports unicode now so maybe we can just use some Regex substitutions directly in the Printer modules and remove `IWriter.EscapeStringLiteral`. https://github.com/fable-compiler/Fable/blob/349bca7cca8cb861e6b32803f417f77126bc362c/src/Fable.Cli/Pipeline.fs#L83-L91

For the future we can try to make the Replacements module leaner by doing the same we 've done with `makeTypeConst`: abstract the basic operations that we already have in the AST (add, multiply... and probably also equality and comparison) as well as type conversions and resolve them in the Fable2XXX step.

dbrattli:
@alfonsogarciacaro Looks good, Python supports unicode strings but not control chars such as null bytes:

e.g code such as:

```py
def test_char_is_control_works() -> None:
    Testing_equal(False, is_control("a"))
    Testing_equal(True, is_control("\u0000"))
    Testing_equal(True, is_control("\u001f"))
    Testing_equal(True, is_control(""))
    Testing_equal(True, is_control(""))
```

will now be written as
```py
def test_char_is_control_works() -> None:
    Testing_equal(False, is_control("a"))
    Testing_equal(True, is_control("")). # <--- null byte
    Testing_equal(True, is_control(""))  # <--- some control byte
    Testing_equal(True, is_control(""))
    Testing_equal(True, is_control(""))
```

So we also need to escape 0x0000 - 0x001f. ([ref](https://www.utf8-chartable.de/unicode-utf8-table.pl?number=1024&unicodeinhtml=hex))I can help later today after work.

dbrattli:
@alfonsogarciacaro As for fantomas you need to run fantomas before checking in the code i.e `dotnet fantomas src/Fable.Transforms/Python`. The CI step will not rewrite, but just check the formatting. But it will be a hassle for you so we should probably just disable it for now.

alfonsogarciacaro:
@dbrattli Ah, ok. Then we need to find a way to replace the control characters or just use `HttpUtility.JavaScriptStringEncode` as the `IWriter.EscapeStringLiteral` implementation, or leave `JavaScriptStringEncode` in PythonPrinter as before but add a replacement for it in case we are compiling Fable.Transforms to Python (or we want to show Python compilation in the REPL). 

About formatting, I did run `dotnet fantomas src/Fable.Transforms/Python` before pushing, but that's when I got the error.

dbrattli:
@alfonsogarciacaro Aha, then there's a bug with Fantomas itself. Let's just disable for now. I'm working on the escape function to see if I can make the tests run again. There are a couple of bugs for arrays and regexp as well.

src/Fable.Cli/Pipeline.fs
```diff
@@ -68,6 +70,62 @@ module Js =
     }
 
 module Python =
+
+    let encodeString charRequiresEncoding (value: string) : string =
```
dbrattli, line 74: @alfonsogarciacaro I've reimpl `JavaScriptStringEncode` from HttpEncoder.cs. We could probably reuse this function for the other languages as well and just have separate `charRequiresEncoding` for each. What do you think?

alfonsogarciacaro: That's a great idea! :+1: We can also add a note to say where it comes from. Maybe move it to Prelude.fs (Naming module) where we have the other polyfills?

ncave:
@alfonsogarciacaro Rust tests fixed in #2794.

dbrattli:
@alfonsogarciacaro I have now fixed Python for this PR 😄

PR #2794, opened 2022-02-02
Fixed Int consts

ncave:
@alfonsogarciacaro 

This fixes the Rust tests. Only minor fixes were needed, great job with the AST change!

On an unrelated note, do we want to add the missing unary operators too (e.g. `UnaryAddressOf`)?


alfonsogarciacaro:
Thanks a lot @ncave! Yes, originally `NumberKind` and operators came from the Babel AST. Now that we've made the numbers closer to F# representation we can do the same with operators too. I was thinking of also adding `compare` to the operators so we can abstract most of the common operators out of `Replacements` and resolve them with each language particularities in the `Fable2XXX` step. Please feel free to send a PR to add `UnaryAddressOf` for now :+1:

PR #2811, opened 2022-02-07
Moar AST/Replacements cleanup

alfonsogarciacaro:
See comments below.


dbrattli:
@alfonsogarciacaro The reason for Python failing is because strict equal needs to be used for `ReferenceEquals`. If I understand things correctly then it will be transformed as `Fable.Binary`? So to make this work we need to use `==` i.e `Eq`/`NotEq` (in Python AST) with literals and `is` i.e `Is` (in Python AST) with any other object. I can have a look at this later tonight (after work).

alfonsogarciacaro:
This is yet another round to simplify the AST and try to polish a few things. It's the last batch in a while I promise! Summary:

- Remove JS-only operators that weren't actually used in the Fable AST.
- Consolidate Binary(Un)Equal and Binary(Un)EqualStrict to avoid confusion. Fable only used non-strict equality in JS for null comparison and option testing so it worked either for `null` and `undefined`. I've just moved this check to the Fable2Babel so we don't need to include this language-specific feature in the Fable AST. Unfortunately, I broke the Python tests when doing this (I also broke Rust but @ncave went ahead and fixed it). Can you please check @dbrattli? It's [this commit](https://github.com/fable-compiler/Fable/pull/2811/commits/c6ee7fa30f29360ff3c74571f160d3112a0aeead#diff-b509f47efb3f6e67625a40a85e41c7c18a9dd5dd8fba664a3e3863959369af20).
- Consolidate `Number` and `Enum` types. This is because number can already be "plain" or "IsMeasure" so we can make "IsEnum" and extra particular case and simplify the matchings (as most of the times the number optimizations apply to enums as well). As we already have `NumberKind` I named the new field `NumberDetails`, though it's a union instead of a record. Please let me know if you can think of a better name ([commit](https://github.com/fable-compiler/Fable/pull/2811/commits/faaee0b1635fe46613c4b2e6d8fd933ad156da27)).
- I also added `DelegateDetails` because I need to indicate in JS when a function shouldn't be compiled as arrow for the decorators. I like the -Info, -Details field because it allows us to add some extra info when needed without having to change a lot of matchings. Maybe we could extend the pattern to other AST elements (`Lambda` for example still has a `name: string option` which will become a bit of an inconsistency). The main issue is when constructing the elements. Ideally we should have static members allowing optional arguments as @dbrattli did with Babel/Python ASTs, but unfortunately from the beginning I used curried helpers in Transforms.Util (makeBinOp, makeCall, etc...) and it may be a lot of work to rewrite that now.

BTW, my intention was to move operator resolution (including equality and comparison) to the Fable2XXX step and in the Fable AST leave only the operators to help with constant folding and pattern matching. I [started doing that](https://github.com/fable-compiler/Fable/pull/2811/commits/faaee0b1635fe46613c4b2e6d8fd933ad156da27) but it seems to be trickier than I hoped. This probably as much as we can trim the Replacements module for now.

alfonsogarciacaro:
@dbrattli I didn't see you reply before finishing my comment, sorry! Yes, Python tests are broken because of the `Binary(Un)Equal` and `Binary(Un)EqualStrict` consolidation. I tried to fix that in Fable2Python but couldn't make it work 😅 It'd be great if you could have a look. Ideally you just use `BinaryEqual` for everything and then make the distinction in the Fable2Python step as JS does now, but please let me know if this doesn't work for you and you need to resolve this already in Replacements.

src/Fable.Transforms/Python/Replacements.fs
```diff
@@ -2969,7 +2910,9 @@ let objects (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr opt
     match i.CompiledName, thisArg, args with
     | ".ctor", _, _ -> typedObjExpr t [] |> Some
     | "ToString", Some arg, _ -> toString com ctx r [ arg ] |> Some
-    | "ReferenceEquals", _, [ left; right ] -> makeEqOp r left right BinaryEqualStrict |> Some
+    | "ReferenceEquals", _, [ left; right ] ->
+        //makeEqOp r left right BinaryEqual |> Some
+        emitExpr None t [ left; right ] "$0 is $1" |> Some
```
dbrattli, line 2915: @alfonsogarciacaro Not sure what to do here. We cannot use `BinaryEqual` since that will compile to `==` and not `is`. And there is no way we can decide in Fable2Python if `BinaryEqual` should be `is` for any names. It must be `==` except for `ReferenceEquals`. Any ideas? I could make a `referenceEquals` function in `utils.py` to avoid the emit, but then the code is not as clean Python as it could have been. I'm unsure ...

ncave: @dbrattli In theory you could do the same check to determine if it's a reference type in `Fable2Python` that was done before in `Replacements.Python`. The other option is to use some work-around by wrapping that `BinaryEqual` op in something (e.g. typecast to something like obj) so you can distinguish it from the regular equals.

alfonsogarciacaro: Yes, this is a pattern I use for some optimizations that happen after beta reduction. As @ncave says in Fable 3 I added a "tag" to TypeCast to mark this potential optimizations during the Replacements. However, later I felt this was stretching too much the purpose of TypeCast and in beyond I added an "OptimizableInto" optional tag to CallInfo. So you can emit a dummy call with OptimizableInto and catch it in Fable2Python. You can see how it's done in Fable2Babel/transformCall.

dbrattli: @alfonsogarciacaro  Ok, I'll put it on the backlog, so please feel free to merge this when you feel for it.

dbrattli (approved):
Great work!

PR #2818, opened 2022-02-14
[Rust] Support simple interpolated strings

ncave:


alfonsogarciacaro:
Nice one @ncave! Actually the last (for now) change I have in mind for the Fable AST is to add interpolated strings as most languages support them already, so we don't need to do hacks with emit as it's now.

PR #2821, opened 2022-02-16
Fix off-by-one for array allocation in Array.windowed

### LINKED ISSUE #2815: [Python] Array.windowed raises IndexError
### Description

Array.windowed raises an IndexError with fable-py, the error has no effect with fable js

### Repro code

```f#
let arr = [|1;2;3|]
let w = arr |> Array.windowed 2
printfn "%A %A" arr w
```

expected
```
array('i', [1, 2, 3]) [array('i', [1, 2]), array('i', [2, 3])]
```
actual 
```
IndexError: list assignment index out of range
```

i think the expression
```f#
source.Length - windowSize
```
in
https://github.com/fable-compiler/Fable/blob/3a06765ee30cc1289d785bddfd49bb432f53143c/src/fable-library/Array.fs#L901

should be changed to
```f#
source.Length - windowSize + 1
```

like [here](https://github.com/dotnet/fsharp/blob/2bdae3516274907a064d02641c75740653694a76/src/fsharp/FSharp.Core/array.fs#L783)

### Related information

* Fable version: 4.0.0-alpha-032
* Operating system; Ubuntu 20.04

dbrattli:
The array allocated in `Array.windowed` is not large enough for the generated windows (off-by-one). @alfonsogarciacaro This should probably be back-ported to main, but JavaScript is immune to the error so not critical.

Fixes #2815

### DIFF
--- a/src/fable-library/Array.fs
+++ b/src/fable-library/Array.fs
@@ -791,7 +791,7 @@ let averageBy (projection: 'T -> 'T2) (array: 'T[]) ([<Inject>] averager: IGener
 let windowed (windowSize: int) (source: 'T[]): 'T[][] =
     if windowSize <= 0 then
         failwith "windowSize must be positive"
-    let res = FSharp.Core.Operators.max 0 (source.Length - windowSize) |> allocateArray
+    let res = FSharp.Core.Operators.max 0 (source.Length - windowSize + 1) |> allocateArray
     for i = windowSize to source.Length do
         res.[i - windowSize] <- source.[i-windowSize..i-1]
     res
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro (approved):
Ah, I see. Yes, it's probably a good idea to port this to JS and maybe other implementations (Rust?)

dbrattli:
@alfonsogarciacaro No, it's in the common F# code, this this will fix for all the languages. I was just thinking about the `main` branch.

ncave:
@dbrattli Rust has its own Array implementation in `fable-library-rust`, so it's already correct.

PR #2831, opened 2022-02-22
Add StringTemplate to Fable AST

alfonsogarciacaro:
This PR adds `StringTemplate` in the AST so we can have "formal" representation of string interpolation, as it's already present in most languages, instead of masking it with `Emit`. This is the last change I wanted to do in the AST for now so if we don't find anything else in the upcoming weeks we can start to consider it stable enough for a beta release and plugin updates. Some notes:

- The StringTemplate has an optional tag field to represent [JS tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) though this may not be needed in other languages.
- Likewise, I'm also using string templates to represent [F# formattable string](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/interpolated-strings#interpolated-strings-and-formattablestring-formatting) in JS. This was mainly needed for Fable.Lit as lit caches the templates in a WeakMap which doesn't work with strings. But in other languages maybe you cannot use the interpolated string as FormattableString should give access to the argument array and the formatted template in runtime.
- For now, I'm just failing in other languages when trying to convert the StringTemplate. Please check out the code in this PR to see how it's done in JS Replacements for reference if you want to implement it in Rust/Python @ncave @dbrattli

### DIFF
--- a/src/Fable.Transforms/Replacements.Util.fs
+++ b/src/Fable.Transforms/Replacements.Util.fs
@@ -2,6 +2,8 @@ module Fable.Transforms.Replacements.Util
 
 #nowarn "1182"
 
+open System.Text.RegularExpressions
+
 open Fable
 open Fable.AST
 open Fable.AST.Fable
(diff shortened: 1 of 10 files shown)


src/Fable.Transforms/Replacements.Util.fs
```diff
@@ -254,6 +256,41 @@ let makeRuntimeType genArgs fullName =
...
+        let str = str.Replace("%%" , "%")
+        (Some [], Regex.Matches(str, @"((?<!%)%(?:[0+\- ]*)(?:\d+)?(?:\.\d+)?\w)?%P\(\)") |> Seq.cast<Match>)
+        ||> Seq.fold (fun acc m ->
+            match acc with
+            | None -> None
+            | Some acc ->
+                // TODO: If arguments need format, format them individually
+                let doesNotNeedFormat =
+                    not m.Groups.[1].Success
+                    || m.Groups.[1].Value = "%s"
```
ncave, line 284: @alfonsogarciacaro Can we parameterize these so they can be passed in as an array, e.g. see [here](https://github.com/fable-compiler/Fable/blob/beyond/src/Fable.Transforms/Rust/Replacements.fs#L955).

PR #2842, opened 2022-03-12
Type.IsInstanceOfType works for types decorated with Global/Import

chkn:
Same idea as #2631, but extends it to work with `Type.IsInstanceOfType` as well. As a side effect, `Activator.CreateInstance` should also work with imported/global types now, but I haven't tested that.


alfonsogarciacaro:
Good one, thank you!

PR #2859, opened 2022-04-18
Dart array tests

alfonsogarciacaro:
As usual, this PR includes a bunch of mixed stuff besides adding some array tests for Dart compilation, but it's another step forward in any case.

@ncave @dbrattli How are you resolving when statements get nested in expressions? JS supports [sequence expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator) so most of the times I could resolve nested lets by hoisting the var declaration and creating a sequence expression with the assignment. Not sure about Python/Rust but Dart doesn't support sequence expressions . At the beginning I added a sequence expression helper that turns assignments into arguments of a function (later ignored), but now I'm trying to [extract the statements](a0d9d0cf25548e4d595a08b918fc8ad90f9dc8b9) to generate better code. How do you do it for Python/Rust?


dbrattli:
@alfonsogarciacaro For Python most of the functions in `Fable2Python` returns a tuple `Expression * Statement list`. Thus you can always produce any number of statements within expressions. These will be "lifted" out to the closest statement scope and be written there. For Python a sequence expression will be transformed to a function where all expressions will be tranformed to statements. Then it returns the result of the last expression. This function will be defined as a statement and then called as an expression like below:

```
let transformSequenceExpr' (com: IPythonCompiler) ctx (exprs: Expression list) (stmts: Statement list) : Expression * Statement list =
    let ctx = { ctx with BoundVars = ctx.BoundVars.EnterScope() }

    let body =
        exprs
        |> List.collecti (fun i expr ->
            // Return the last expression
            if i = exprs.Length - 1 then
                stmts @ [ Statement.return' (expr) ]
            else
                exprAsStatement ctx expr)

    let name = Helpers.getUniqueIdentifier "expr"

    let func = Statement.functionDef (name = name, args = Arguments.arguments [], body = body)

    let name = Expression.name (name)
    Expression.call (name), [ func ]
```
It makes the code a bit messy and I was afraid of painting myself into a corner with this, but so far it works well 😅

ncave:
@alfonsogarciacaro Rust is [primarily](https://doc.rust-lang.org/reference/statements-and-expressions.html) an expression language, so I was able to ditch the statement/expression separation and concentrate on expressions, although that required re-implementing some things like tail calls elimination which could not be reused from the JS implementation.

alfonsogarciacaro:
Ah, thanks @ncave, didn't remember that Rust is expression-based 😅 That's very helpful (and also the pattern matching support). And thanks a lot @dbrattli too for the explanation! This gave me the idea to add a new case to the `ReturnStrategy` passed to `transformAsStatements`. Thanks to this I've been able to [remove the duplicity of transforming as expression or statements](https://github.com/fable-compiler/Fable/pull/2859/files#diff-e865800cded862e6a7b41e472ddd8812d0af3f1d3341da2d7f1b29a6c4c3d2ea). Now all transform functions return a statement list (can be empty) and an optional captured expression (available when the return strategy is `Capture`). This requires some care when combining expressions but it simplifies the code (e.g. it's not necessary to hoist vars anymore) and helps generate better code too 😄

ncave:
@alfonsogarciacaro

> and also the pattern matching support

Yes, well, unfortunately as you well know F# (FCS) decomposes the pattern matching a bit too much, so we have to re-assemble it back from the pieces, which is hard to do in more complicated cases, as it's hard to reconstitute the original pattern matching expression. It's unfortunate the original decision tree representation is not exposed by FCS. It's there internally, but it gets transformed to a representation that is more useful for code generation by languages without pattern matching support, but this way we're missing the opportunity to have an exact translation for languages that do support pattern matching (like Rust).

That said, the partial conversion from decision tree back to pattern matching works fine for most simple cases, even though it may not be always pretty, but that's ok.

alfonsogarciacaro:
Oh, you're right, didn't even think of that. I don't have much experience with FCS Syntax Tree but, could it be possible/helpful to use the ranges to match the `DecisionTree` (from the typed-checked expressions) with the corresponding `SynExpr.Match`? This could help in reconstructing the original representation 🤔

ncave:
@alfonsogarciacaro

> use the ranges to match the DecisionTree (from the typed-checked expressions) with the corresponding SynExpr.Match

Not sure, sounds like an advanced magic, you may have to elaborate a bit ;)

Don't know if it's possible to reverse [this transformation](https://github.com/fable-compiler/Fable/blob/snake_island/src/fcs-fable/src/fsharp/symbols/Exprs.fs#L1249), maybe it's lossy. It would be so much easier to just expose the underlying `TDSwitch`/`TDSuccess`/`TDBind`.

PR #2861, opened 2022-04-24
Add NoReflectionAttribute

kerams:
Implements https://github.com/fable-compiler/Fable/pull/2710#issuecomment-1107456133.

Given

```fsharp
[<Fable.Core.NoReflectionAttribute>]
type U =
    | A
    | B of int

type R = {
    W: U
}
```

I get

```js
export class U extends Union {
    constructor(tag, ...fields) {
        super();
        this.tag = (tag | 0);
        this.fields = fields;
    }
    //cases() {
    //    return ["A", "B"];
    //}
}

//export function U$reflection() {
//    return union_type("QuickTest.U", [], U, () => [[], [["Item", int32_type]]]);
//}

export class R extends Record {
    constructor(W) {
        super();
        this.W = W;
    }
}

export function R$reflection() {
    //return record_type("QuickTest.R", [], R, () => [["W", U$reflection()]]);
    return record_type("QuickTest.R", [], R, () => [["W", null]]);
}
```

with comments denoting changes with the attribute. There's also a warning `warning FABLE: QuickTest.U is annotated with NoReflectionAttribute, but is being used in types which support reflection. This may lead to runtime errors.` because `R` uses `U`, but does not itself have the attribute.

@alfonsogarciacaro, if you think this is a good addition, I can add tests and try to figure out https://github.com/fable-compiler/Fable/pull/2710#issuecomment-1107456736.

One thing that worries me is that `Union.name()` refers to the case names, and it's used both in `toString` and `toJSON`. ~~Perhaps `NoReflection` could also imply that default stringification/serialization is for all intents and purposes not supported?~~ For these cases we could have a different base union class in TS that would just use the tag instead of case name. No specialized classes for unions with no reflection would then be needed to be emitted at all.

### DIFF
--- a/src/fable-library/Types.ts
+++ b/src/fable-library/Types.ts
@@ -54,6 +54,49 @@ export function unionToString(name: string, fields: any[]) {
   }
 }
 
+export class CommonUnion implements IEquatable<CommonUnion>, IComparable<CommonUnion> {
+  public tag!: number;
+  public fields!: any[];
+
+  public toJSON() {
+    return this.fields.length === 0 ? String(this.tag) : [String(this.tag)].concat(this.fields);
+  }
+
+  public toString() {
+    return unionToString(String(this.tag), this.fields);
+  }
+
+  public GetHashCode() {
+    const hashes = this.fields.map((x: any) => structuralHash(x));
+    hashes.splice(0, 0, numberHash(this.tag));
+    return combineHashCodes(hashes);
+  }
+
+  public Equals(other: CommonUnion) {
+    if (this === other) {
+      return true;
+    } else if (!sameConstructor(this, other)) {
+      return false;
+    } else if (this.tag === other.tag) {
+      return equalArrays(this.fields, other.fields);
+    } else {
+      return false;
+    }
+  }
+
+  public CompareTo(other: CommonUnion) {
+    if (this === other) {
+      return 0;
+    } else if (!sameConstructor(this, other)) {
+      return -1;
+    } else if (this.tag === other.tag) {
+      return compareArrays(this.fields, other.fields);
+    } else {
+      return this.tag < other.tag ? -1 : 1;
+    }
+  }
+}
+
 export abstract class Union implements IEquatable<Union>, IComparable<Union> {
   public tag!: number;
   public fields!: any[];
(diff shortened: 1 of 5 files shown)


kerams:
Ok, unions are now "erased" in terms of the type system but not to the degree achieved with `EraseAttribute`.

Given

```fsharp
[<Fable.Core.NoReflectionAttribute>]
type U =
    | A
    | B of int

type R = {
    W: U
}

let a = { W = B 1 }

match a.W with
| A -> ()
| B 1 -> ()
```

I get

```js
//export class U extends Union {
//    constructor(tag, ...fields) {
//        super();
//        this.tag = (tag | 0);
//        this.fields = fields;
//    }
//    cases() {
//        return ["A", "B"];
//    }
//}

//export function U$reflection() {
//    return union_type("QuickTest.U", [], U, () => [[], [["Item", int32_type]]]);
//}

export class R extends Record {
    constructor(W) {
        super();
        this.W = W;
    }
}

export function R$reflection() {
    //return record_type("QuickTest.R", [], R, () => [["W", U$reflection()]]);
    return record_type("QuickTest.R", [], R, () => [["W", null]]);
}

//export const a = new R(new U(1, 1));
export const a = new R(new CommonUnion(1, 1));

(function () {
    const matchValue = a.W;
    if (matchValue.tag === 1) {
        if (matchValue.fields[0] === 1) {
        }
        else {
            throw (new Error("Match failure: QuickTest.U"));
        }
    }
})();
```

src/fable-library/Types.ts
```diff
@@ -54,6 +54,49 @@ export function unionToString(name: string, fields: any[]) {
...
+  public GetHashCode() {
+    const hashes = this.fields.map((x: any) => structuralHash(x));
+    hashes.splice(0, 0, numberHash(this.tag));
+    return combineHashCodes(hashes);
+  }
+
+  public Equals(other: CommonUnion) {
+    if (this === other) {
+      return true;
+    } else if (!sameConstructor(this, other)) {
```
kerams, line 78: I'm not exactly sure what the idea behind this check is, but it probably does not make sense in the context of this shared union class.

alfonsogarciacaro: This was to prevent a case where you compare two unions with the same shape but different type (after casting them to obj). As you said, it doesn't make sense if you're not generating different classes for each union.

kerams: Thanks. Yeah, calling equals on 2 different union types may return true here. It's one of the trade-offs for size savings.

PR #2866, opened 2022-04-28
Dart: add Map/Set implementations

alfonsogarciacaro:
This adds Map/Set implementations to Dart library. I'm a bit slow on tests, but Dart compiler (and linter) are quite good so I'm focusing in generating nice code at the moment.

@ncave Upon reflection, I realized you're right and we shouldn't go too crazy with List optimizations. ResizeList works well when lists are used as arrays. It's true this happens many times because List is the default collection in F#, but we should avoid giving a penalty to devs using lists "properly" as linked lists/stacks.

Also, because Dart gives warnings when using non-null assertions unnecessarily (`!` postfix operator in Dart) I need to keep track of the variables in the current (local) function scope. Maybe you can do something similar to detect which values are being captured by local functions? https://github.com/fable-compiler/Fable/pull/2866/commits/eb668412e6b465184fefac6742fd8ac6dc587458


ncave:
@alfonsogarciacaro No, what I had to fix was slightly different. For Rust, we want non-capturing inner functions to be generated as functions, not lambdas, as it's much cleaner and supports recursion. But that only works for non-capturing functions, and because they have different signatures than closures, they have to be boxed when passed as parameters (but not boxed when called), so they can match the closure signature. Closures don't support recursion in Rust, so we add fixed-point combinators where needed (if tail calls cannot be eliminated). Anyway, TMI.

PR #2872, opened 2022-05-04
updated to a newer version of FCS/Core

voronoipotato:
It's probably not something that should actually ever get merged but may theoretically? be useful as an exploration of parts that would need to get done to update FSC/Core to latest main branch in fcs. The extra underscores are debug points like "debugAt/debugFrom" etc which I typed out more fully in Fsharp2Fable.fs . I don't know how useful they would actually be, but they don't appear to get in the way.

Accompanying changes for the F# compiler are in my version of the service_slim. I mostly was able to use NCave's service_slim with a couple small tweaks. It compiles and things seem to run okay.

https://github.com/voronoipotato/fsharp/tree/service_slim

I did this because I saw the note about how you can build NCave's service_slim, and then immediately went down a rabbit hole of curiosity. If this isn't useful in any sense to anyone that's okay :). 

Thanks!


alfonsogarciacaro:
Thanks a lot for investigating this @voronoipotato! It's true that the way we're handling the FCS dependency is super-messy right now 😅 We should try to automate it more and at least indicate which FCS version/commit the custom build is based on.

As an example of the mess, the current custom build contains [a change](https://github.com/ncave/fsharp/pull/10/files) that's still not merged into @ncave's `service_slim` fork and I was using for the experimental precompilation feature.

In any case, as we're still using @ncave's `service_slim` as the source of truth, the preferred approach ([as you actually already did](https://github.com/ncave/fsharp/pull/11)) is to sync ncave/fsharp/service_slim with dotnet/fsharp/main and then build the custom .dll and update Fable. Maybe this could be an opportunity to add the information of the FCS version/commit somewhere (maybe an assembly attribute so we can access it from Fable?).

ncave:
@alfonsogarciacaro I don't mind updating FCS-Fable more often, but we probably want to align it to some F# or FCS version release or specific commit that adds a F# language feature.

voronoipotato:
I think a release makes a lot of sense since it creates a bit of structure and regularity for updating FCS/Fable. I went for main as a proof of concept, and it's very possible that I missed a step that needs to be done so it would be wise to not take my work as is. It was more exploratory than a proper unit of work.

ncave:
@voronoipotato Fable 4 already has latest FCS, feel free to close this PR if not needed.

voronoipotato:
Closing as no longer needed :) 

Thanks!

PR #2886, opened 2022-05-17
Enable more F# features when targeting Dart

alfonsogarciacaro:
This PR:
1. Enables `ref` support for Dart
2. Fixes measure types by deleting them from type annotations in Dart
3. Uncurries classes and (anonymous) records' fields

@dbrattli @ncave I think point 3. will also bite you. During the uncurry transformation, it's assumed that record/union/class function fields are uncurried. But unfortunately we cannot uncurry them in FableTransforms because `FSharpFields` belong to the entities which are out of the AST, so in the case of Dart I have to uncurry the types in the Fable2Dart step. It's probable you need to do it for Rust and Python too so please take the changes in this file as reference: https://github.com/fable-compiler/Fable/pull/2886/commits/3bd6eeac5d00eea29473344e5296a6269edb265b#diff-e865800cded862e6a7b41e472ddd8812d0af3f1d3341da2d7f1b29a6c4c3d2ea

> Note: If you happen to have type annotations for union fields in your language (in Dart atm I'm just using a dynamic array) you will have to uncurry them too.

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -59,6 +59,7 @@ type Constraint =
 
 type GenericParam =
     { Name: string
+      IsMeasure: bool
       Constraints: Constraint list }
 
 type Parameter =
@@ -139,7 +140,7 @@ type Type =
     | List of genericArg: Type
     | LambdaType of argType: Type * returnType: Type
     | DelegateType of argTypes: Type list * returnType: Type
-    | GenericParam of name: string * constraints: Constraint list
+    | GenericParam of name: string * isMeasure: bool * constraints: Constraint list
     | DeclaredType of ref: EntityRef * genericArgs: Type list
     | AnonymousRecordType of fieldNames: string [] * genericArgs: Type list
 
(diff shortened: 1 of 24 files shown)


alfonsogarciacaro:
@ncave Ah, ok. Thanks for checking! So Python/Rust do the "uncurrying" when emitting the type annotation. Interesting, I had the impression this wouldn't work as there are situations where lambdas are not uncurried (e.g. when lambdas are returned from methods) but it seems to be working for Python/Rust so let's see :)

Thanks also for tidying up the uncurry code, putting the helpers together in FableTransforms makes a lot of sense. `uncurryLambdaType` is a bit different to `uncurryType` as it doesn't deal with options and the returned values are different so I have changed its name to try to make the semantics clearer. I hope that's ok: https://github.com/fable-compiler/Fable/pull/2913/commits/8bad324a503e93f5023802b84eacfbe5e9c8015d

Nhowka:
Not sure if relevant, but Dart has crazy good extensions that can target functions in a generic way while keeping type-safety, enabling code like [this](https://dartpad.dev/?id=8f6925dda5f9ba4d082882757316de90). Could this be useful if they were inside the core library?

alfonsogarciacaro:
That's nice! I'm currently just wrapping the function where needed, but an extension like this will make the output nicer. Thanks for sharing!

alfonsogarciacaro:
@ncave I was trying to convey that the two functions behave differently, but you're right that they're still close enough to bear similarity in the name, so I renamed it again to `uncurryLambdaType` 👍

PR #2888, opened 2022-05-18
[Rust] Emit now works with string templates

ncave:


alfonsogarciacaro:
@ncave Seems this PR triggers a build warning about `EmitExpresion` not being handled in Rust.AST.Impl.fs

image

PR #2908, opened 2022-05-31
[Rust] Bindings with Emit on interfaces

ncave:
@alexswan10k
- Support modelling Rust interop bindings with Emit on interfaces.
- Calling convention is Rust native by default (not pass-by-ref).

Note: Unlike `Enum` on `module let` bindings which gives you more control, in interface member signatures we can only use F# language features, so we have to model passing references using something like F# `byref/inref/outref`. This is still TBD.

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -605,7 +605,7 @@ module TypeInfo =
         | HasEmitAttribute value ->
             let genArgs = genArgs |> List.map (transformType com ctx)
             mkEmitTy value genArgs
-        | ent when ent.IsInterface = true ->
+        | ent when ent.IsInterface ->
             transformInterfaceType com ctx entRef genArgs
         | ent ->
             let genArgs = transformGenArgs com ctx genArgs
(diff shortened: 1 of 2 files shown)


alexswan10k:
Nice. 


Ah yes, the ref keyword has got me thinking now. We currently pass everything by reference for function parameters but this is inconsistent with dotnet which only passes by ref if the ref keyword is present, and value by default. Should we be trying to model this do you think? You would get clones at every boundary which is not ideal, but otherwise you are often passing a pointer to a pointer to an object which is kinda weird anyway such as ``` &Rc```.

ncave:
@alexswan10k To clarify, modelling pass by reference using `byref/inref/outref` can be tricky as they are already implemented using mutable cells. Perhaps we can keep `byref/outref` as is using mutable cells, and only use `inref` for pass by reference?

alfonsogarciacaro:
@ncave @alexswan10k I've added a `NEXT` section to release notes so when sending a PR if you want you can add a brief description of the changes that will be included in the notes of the next release 👍 https://github.com/fable-compiler/Fable/blob/72f82094ba6106a115e0c94cefadc215c8c62426/src/Fable.Cli/RELEASE_NOTES.md?plain=1#L1-L3

alexswan10k:
No need to apologise @alfonsogarciacaro 😆 when there are release notes we will be sure to write them up. Thanks for mentioning. My comments are too often an incoherent stream of consciousness so I should apologise 😀

We are still somewhat battling with the lack of a garbage collector, which means everything has to be abstracted with smart pointers. The problem is there are many incompatible smart pointers to choose from, all with their own trade offs.

I will write up tickets when it is obvious what needs to be done, but a lot of this is just mentally working through the options.

ncave:
@alexswan10k Re: passing by ref, see also the original discussion [here](https://github.com/ncave/Fable/pull/16).

alexswan10k:
Good points, a good refresher. The benchmark we did was a particularly useful exercise. It clearly shows that passing by reference has significant performance advantages, and we know that reference counting is generally slow anyway.

That being said, I did not fully consider giving a user control over this might give you the best of both worlds. I am definitely coming round to the idea that control is preferred here, which means picking a sensible default. The problem is that our best representation of references (inref) is opt-in, which means by-value should probably be default. Gahh! I am happy to look at this if we go that way, since it is arguably my mess. I think it is worth thinking this over carefully before we commit to anything though.

We will still need reference tracking though for inref, I imagine the big change here is switching to leaveContextByValue as default. 

I think the generic argument problem may well go away because we no longer represent closures and interfaces by ```&impl something```, but ```Rc<impl/dyn Something>```. This means it can be passed by value. I think the caveat with generics is that whatever you choose (value or reference), it must apply for all types, so you cannot have value for ints, and ref for structs.


I was going to create a generic issue for memory management stuff to consolidate all these ideas and track everything. That ok?

ncave:
@alexswan10k
> That ok?

Very much so, please do, thanks!

Also, micro-benchmarks are obviously not representative of real performance difference in an average code base, I'm sure we can create micro-benchmarks that show that passing by value is faster in certain cases, but that's not the point. I don't thing there is that much actual difference compared to other heavier costs. I fully agree that the sensible defaults is what we should be doing, so whatever makes most sense.

alexswan10k:
Totally agree. Ticket is [here ](https://github.com/fable-compiler/Fable/issues/2916) btw, not sure if you can edit but feel free to. I will try keep it up to date with discoveries/ideas etc.

What are the other key outstanding issues that were total dealbreakers? Are you happy with where we are on Emit now for bindings? Import was another big one, wasn't it? Not sure how to move that forward heh.

ncave:
@alexswan10k Yes, we will use aliases to avoid name clashing, the question is what is the level of granularity we want to go to with the use clauses, i.e. do we want to keep using wildcards at some level, or do we want to go all the way down and declare a use clause for each imported name.

alexswan10k:
I would suggest we go all the way down and spit out one for each function/type imported, it is probably the easiest to implement for MVP. We can always come back and rework it to be prettier/more lint friendly in the future. That feels pretty solvable though.

Beyond that is there anything else that is fundamentally wrong or needs to be working with imports? You can sort of call stuff directly by namespace using emit too if it is an external lib you are importing.

ncave:
@alexswan10k No, that's it, just a matter of implementing it properly and make sure it works properly across with a complex multi-file project with many modules.

PR #2918, opened 2022-06-02
[Rust] - Switch back to pass by value as default

alexswan10k:
Explore consequences of switching to pass by value.

* Pass by value as default. Clone when multiple usages
* Ironed out a bunch of usage tracking issues
* Ironed out some ref tracking issues
* Ironed out closure plurality problem - a closure becomes fnOnce if it assumes it is the only consumer of an externally captured thing. We now assume all closures may be called more than once, so their captured clones must be assumed to be plural.
* Allow native libs to be called with value OR ref
* Precursor to inref -> & work https://github.com/alexswan10k/Fable/pull/2
* Self is still pass by reference. Passing self by value leads to transferred ownership which leads to problems.


alexswan10k:
@ncave  can I bother you with some assistance. I have 1 damn call site left that won't compile and i can't figure it out. After that I think this will work. It is the ```map.tryGetValue```. Something clever is going on to turn it from a ```fn(param)``` to a ```fn(param, outParam, _)```, and I cannot work it out! I presume you did this bit. Any ideas?

Quite a lot to uncover here if this works, I will write it up later. The bottom line is - references via any ref type thing, far cleaner code, more cloning, but with the benefit that single usage can just pass transparently pass through. Native libs can be controlled to pass individual params by ref or by value, although the method is rough at best. Also uncovered a bug with closures and for loops around FnOnce.

I need to do a cleanup pass too...

ncave:
@alexswan10k I'll take a look.

alexswan10k:
@ncave figured it out. The new reference rules have some subtle edge cases around when to adhere to the dereference & operator. I made a mistaken assumption. Fixed.. Looks like a bunch of new errors have materialized now....

ncave:
@alexswan10k Great job, it works now. I sent you just a minor cosmetic pr.

Tested it on a slightly bigger code base (than a micro benchmark): [fable-raytracer](https://github.com/ncave/fable-raytracer)
Numbers below are specific to this benchmark and machine, so may not be indicative of performance in general.
(smaller/faster is better):
- By ref:  *1030 ms*
- By val:  *1230 ms*

For native Rust binary (`test-rust`), passing by value runs 20% slower than passing by reference, but I wouldn't be too quick to jump to conclusions yet (except the obvious one that passing big objects by value is slower than passing by reference). This benchmark is very sensitive to that exact difference, and in that sense is a good exercise.

For example, just changing the `Intersection` object to a struct (by adding `[<Struct>]`) makes the current fastest performer .NET native (`test-native`) almost 50% *slower* (goes from *800 ms* to *1200 ms*), but has much smaller impact (~5%) on the speed of the managed .NET code (`test-dotnet`), cause I believe there is an internal optimization that .NET does to pass large value objects by reference.

alexswan10k:
Ah nice one. It is good to have some numbers. So do you think this is progress? I know we are losing a little bit of performance but the generated api surface does appear generally cleaner. There were also a lot of subtle bugs around the ident usage tracking that I think I have managed to iron out too. I think we should be able to claw back the performance where it is required by letting a user control when stuff is passed by ref.

I might need some help for the next phase. I had a go at trying to convert inref to references, but there were some tricky problems around the fact there is a transform already turning all references into native refCell's. I think we need to do something like pass through the AddrOf operators, but this obviously breaks the mutref transforms. https://github.com/alexswan10k/Fable/pull/2

I will do a quick cleanup pass if we choose to merge. Can we squash? I have a sprawling mess of commits heh

ncave:
@alexswan10k In theory we don't need to represent `inref` as mutable cell, only `byref/outref`. But a byref<'T> can be passed where an inref<'T> is expected, so to avoid that we have to get the cell value at call site and pass by reference.

I can squash and merge if you want.

alexswan10k:
I agree, I think it should probably not be a mutable cell, only byref/outref. The problem I had was that the inboud AST did not have a representation that distinguished the difference. It eagerly transforms onto the native::refCell, which made it difficult to use the context of the Fable2Rust transform to help guide the process. I tried to unravel this but it obviously breaks refCell's as they then have to do this transform further down.

Anyway this is tidied up now, I will open.

ncave:
@alexswan10k Ok merged. The underlying .NET type is different, so there is a way to distinguish between `inref` and `byref`, but we may need to update the Fable AST to represent that fact. I'll take a look.

alexswan10k:
Excellent. Maybe I can take a look at the attribute approach in the meantime. I am  torn if this is even a good idea, although having a non-invasive way to achieve this optimization is quite appealing.

ncave:
@alexswan10k Great, it could even turn out that controlling passing by reference with an attribute may be preferable, as you said is may be less intrusive.

PR #2923, opened 2022-06-06
Rust - Attribute to make a function pass by ref

alexswan10k:
A few things added and tweaked here

* Removed concept of TakingOwnership as everything is now taking ownership by default
* Added ByRef attribute - I am surprised there is nothing built in to do this, please correct me if there is a better way. There is a CompilerServices one but it is not available by default. I assume the assembly would need to be included or something.
* Added IsParamByRefPreferred which is set to true when 
* Added a bunch of tests, half which are commented out because their mechanics do not yet work.
* Added placeholder switch to emit & on an inref. This requires callers to have visibility over this which is not yet implemented.

Currently, only a root function tag works for ALL parameters. Crude, but it is a start. 
```
[<Fable.Core.ByRef>]
let byrefAttrRootObjDecFn (x: Obj) =
    x.X + 1
```

I will look at param level control next, although this seems to be problematic as attributes are not on the AST. Will do some digging.


ncave:
@alexswan10k This is great, let me try to adapt my code on top of this, as there is some overlap (as expected).
I like the flexibility to have either function-level `ByRef` attribute or parameter-level `inref<'T>` type for fine control.

alexswan10k:
Damm, hope i didnt make too much work for you @ncave. Let me know if i can help in any way.

Yeah i think both methods have merits. Its probably just something that we need to experiment with and see what works best.

alexswan10k:
I have been investigating if it is possible to do the following, but I seem to be hitting similar shortcomings with both the Fable AST and more importantly the F# AST.

```
let byrefAttrObjFn ([<Fable.Core.ByRef>] x: Obj) =
    x.X + 1
```

First and most obvious - Fable params do not have this metadata. That lead me to https://github.com/fable-compiler/Fable/blob/7c6b52fd03c7f6d92e1434032350ec70f59bdf36/src/Fable.Transforms/FSharp2Fable.fs#L1417
 and see what was going on with the debugger. To my surprise, the Attributes field for each parameter does not seem to have any content at all. I have tried different fable attributes and nothing seems to be coming through. I have also tried to traverse the args model in the debugger to see if anything else fits, but nothing seems likely to be an Attribute. @alfonsogarciacaro I don't suppose you are aware of what I might be missing here? I presume you have done something to get this working for other Fable parameter attributes such as ParamListAttribute and InjectAttribute.
 
 
image

If I get anywhere with this, will probably push out to another PR. Just investigating at the moment.

alfonsogarciacaro:
@alexswan10k If I understand it correctly, the attributes are not in the arguments (the expressions passed to the call) but in the parameters of the method signature. That's why I had to add to add `CallMemberInfo` to the calls so you can easily access the parameter info from the call:

https://github.com/fable-compiler/Fable/blob/5b14f7c3efcf10496d923f20e0315eb0acb3ce4e/src/Fable.AST/Fable.fs#L332-L353

BUT... I didn't include the attributes in the ParamInfo within CallMemberInfo so it's still not useful for you 😅 We can either add the attributes there or have an add-hoc property like I did for `IsNamed`. Probably we'll need the attributes anyways for maximum flexibility so I'll add them. This should also be a good chance to transform CallMemberInfo in a reference instead of duplicating them in each call so we don't bloat the AST.

alfonsogarciacaro:
@alexswan10k I'm making the changes. At the end I'm doing some refactoring to remove some duplication in the Fable AST, but if you can wait a bit, hopefully I should have it ready tomorrow so you can have access to the parameter attributes 👍

alexswan10k:
Of course. Sorry I did not mean to make more work for you @alfonsogarciacaro I appreciate you have enough going on. Many thanks though, this will definitely help.

ncave:
@alexswan10k Now that I think of it, perhaps parameter-level attributes are not such a hot idea, not sure if they can be supported on return types, etc. So perhaps no need to go down this rabbit hole. I'm looking at `inref` instead, it's a bit more involving to cover all corner cases than I anticipated, but at least in theory there are no blockers.

alexswan10k:
I am happy to hold off if you have reservations. Attributes are supported on return types, and you can constrain what is actually allowed where. I don't know if the compiler understands this though.

My initial thought was that inref is more typesafe because it is explicit, so the F# domain maps to the Rust domain 1-1. However, because call sites and idents are tracked, and because of all of the clever auto uncurrying etc, I believe that in principle an implicit attribute syntax is still actually typesafe, and references simply become a compiler optimization. My main concerns were around function composition, where the inputs and outputs are glued together without any control over dereferencing and unwrapping but so far this has not yet been an issue in practice. I guess we need some more testing to really confirm this.

ncave:
@alexswan10k Here is a PR for the `inref`: https://github.com/alexswan10k/Fable/pull/3

Things that work:
- Somehow passing `byref` into an `inref` slot works :)

Things that don't work yet:
```fs
// TODO - ByRef as return body not working yet
let byrefIdentity (x: int byref): int = x

// TODO - ByRef as return type not working yet
let byrefReturns (x: int byref): int byref = &x
```

alexswan10k:
This is awesome progress @ncave, great work. I really think this change is a big improvement because it gives users the ability to choose rather than dictating one way or another (with all its compromises). Having param level granularity is also a must.

I have merged into this branch. Feel free to merge if you are happy :)

PR #2927, opened 2022-06-12
[Rust] Better imports

ncave:
- Better imports, resolving name clashing.

### DIFF
--- a/src/Fable.Transforms/Rust/AST/Rust.AST.Helpers.fs
+++ b/src/Fable.Transforms/Rust/AST/Rust.AST.Helpers.fs
@@ -1215,7 +1215,7 @@ module Items =
         { attrs = mkVec attrs
           id = DUMMY_NODE_ID
           span = DUMMY_SP
-          vis = PUBLIC_VIS
+          vis = INHERITED_VIS
           ident = ident
           kind = kind
           tokens = None }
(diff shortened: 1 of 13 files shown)


ncave:
@alexswan10k This should unblock the progress for Rust support. Try it, let me know what you think.

@alfonsogarciacaro Given that files are compiled in random order (and not always, some may be skipped), we need access to the full list of all project file names (from the project file). This is mainly for libraries or test projects that need to export all their modules. Is there a way to access `Project` from `Compiler`?

Currently we're getting around this by manually re-importing all files in the main file, but it will be nice to avoid that.

alexswan10k:
This is great @ncave. Will do some testing. We could probably easily simulate the unit test collision scenario too I imagine - name a function the same thing in two different files and import both into the same test?

alfonsogarciacaro:
@ncave `Project` is defined at the very end of Fable.Transforms while `Compiler` interface is declared at the very beginning, so exposing the full `Project` class is going to be difficult (besides, it has a dependency on on `FSharpAssembly` from FCS). But if you need to access the source files it should be ok to add a property to [`Compiler` interface](https://github.com/fable-compiler/Fable/blob/473c1cfc95d5cef6c24e6c985e2408b57bd9a614/src/Fable.Transforms/Global/Compiler.fs#L55) and then implement it in [`CompilerImpl`](https://github.com/fable-compiler/Fable/blob/473c1cfc95d5cef6c24e6c985e2408b57bd9a614/src/Fable.Transforms/State.fs#L191).

But... it happens we don't even store the files in order in `Project` 😅 (they're stored as a map currently). So if you need the ordered list we'll have to add an extra argument to [the constructor](https://github.com/fable-compiler/Fable/blob/473c1cfc95d5cef6c24e6c985e2408b57bd9a614/src/Fable.Transforms/State.fs#L103-L106) and fill from the [static `From` method](https://github.com/fable-compiler/Fable/blob/473c1cfc95d5cef6c24e6c985e2408b57bd9a614/src/Fable.Transforms/State.fs#L121-L138) which is how the project is initialized from Main.fs (maybe we should make the constructor private to be sure).

ncave:
@alfonsogarciacaro Added `SourceFiles` to `Project` and `Compiler` in #2929.

PR #2928, opened 2022-06-13
Rust - Async - first (really rough) pass to support builder + tasks + Configurable pointer types

alexswan10k:
This is a bit of an experiment and introduces the idea of being able to tag constructs like records and unions with a pointer type, so that all instances become switched out with a different pointer type. I am not convinced this is even the right approach so feel free to discuss or persuade me otherwise!

Currently this supports Rc and Arc, for Records and Unions.

--- 
I have been doing some thinking and reading around with respect to threading, I was always under the impression that the Cell concept would need to be substituted for a RwLock or Mutex, but today the penny dropped that dotnet does not give you this threadsafe guarantee anyway, and you would have to implement your own locks to create threadsafe collections arbitrarily etc. I believe it is basically equivalent to us making MutCell Send and Sync and throwing out the compile-time guarantee completely. This vastly simplifies the problem, as the issue then is only around choosing Rc or Arc. I guess we would additionally need a way to simulate a higher level object lock, but this can probably be done with a Mutex.

I like the idea of having control over pointers at the type level, but the big elephant in the room is obviously the standard library, which would probably have to support Arc for everything, or nothing. Perhaps we could make some collections such as List support Arc, but not Array (as it is not threadsafe in either domain).


alexswan10k:
@ncave  particularly curious on your thoughts to if this has legs, or if we should bin it

ncave:
@alexswan10k I thought we may use AtomicCell or [GcCell](https://github.com/Manishearth/rust-gc/blob/master/gc/src/lib.rs#L378) and Rc/Arc/[Gc](https://github.com/Manishearth/rust-gc/blob/master/gc/src/lib.rs#L48) on a compiler switch.
But if that doesn't work, we can try to make a custom Sync MutCell.

I guess this PR is in preparation for it, can you please elaborate what are your plans for it?
We probably need a small multi-threaded sample to work on the design.

alexswan10k:
In principle, I am still on board with a compiler switch. The problem though is this kind of goes against the Rust mantra of "choose your guarantees", and you effectively force the whole codebase down to the lowest common denominator. If 1 call site needs an Arc, the whole codebase must become Arc, and pay the price. I am torn about what this means for the std library too. Should there be two, or even three libraries? The library can be compiled multiple times, but the .rs files must be explicitly duplicated. I feel like this could be confusing too, having to know what lib to install vs what compiler switch is set, although perhaps there is a way to streamline this by abstracting it from a user (In Arc mode you get the Arc std library, etc).


My thoughts on the attribute approach are that you would be able to explicitly control what wrapper you might use (gc/arc/rc) at the type level, so that you can only enable types that are necessary to cross thread boundaries, and "pay for what you use". This could be expanded to encompass whatever wrappers are required going forward. It could also allow a single std library, although it would end up gravitating to the lowest common denominator again. Interestingly you could also have both mechanisms working side by side, using an attribute to opt out of code definitely not part of a parallel workflow for performance reasons. There are interesting questions here such as collections that are not safe in dotnet are provably not safe when compiling to rust. By choosing to model this as Rc for example, we can make those programs illegal but is this even a good idea as this means some valid F# programs will not be valid Fable2Rust programs.

Finally, the point I was trying to make about MutCell is that in dotnet, there are basically no guards. An array for example is "not thread safe", meaning it is vulnerable to exactly the sort of contention Sync is supposed to prevent. From what I understand forcing MutCell to Send and Sync is not a million miles from dotnet behaviour, even though it is not thread safe. As you know, dotnet uses higher level mechanisms such as locks to allow a user to "make it safe". I do accept that there are probably gnarly edge cases that could lead to undefined behaviour, but perhaps there are additional internal tweaks around atomicity etc that may alleviate this.

_I have been keeping an eye on the rust-gc crate too, and this definitely an interesting approach, but it looks like their thread-safe implementation is not going anywhere any time soon. I did however envisage that the attribute approach could be extended to support "custom", thus allowing you to plug in any wrapper by name going forward. In practice, this might be harder than I thought considering Gc requires auto-derived traits._

I will have a look at perhaps putting together some tests or something.. agree we need some real examples

alexswan10k:
Another possibility with respect to conditional compilation are [features](https://doc.rust-lang.org/cargo/reference/features.html). The benefit is they apply to both rust and any generated F# code (assuming it is correctly tagged of course). You would probably have to output both paths and tag them up with different features.

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -240,3 +240,11 @@ type [<Erase>] U9<'a, 'b, 'c, 'd, 'e, 'f, 'g, 'h, 'i> =
     static member op_ErasedCast(x:'i) = Case9 x
     static member inline op_ErasedCast(x:'t) : U9<_, _, _, _, _, _, _, _, ^U> =
         Case9 (^U: (static member op_ErasedCast: 't -> ^U) x)
+
+
+type PtrType =
```
alexswan10k, line 245: Move to Fable.Core.Rust

ncave: @alexswan10k Looks great, very minimal.

Very minor suggestion, do we want to rename `PointerType`/`PtrType` to `ReferenceType`/`RefType`, just to keep it aligned with .NET lingo? Also, we may need the name for `PointerAttribute` (or `PtrAttribute`) to represent actual non-rc pointers (for webasm/js interop etc.).

PR #2929, opened 2022-06-13
[Rust] Import project files

ncave:
- Simply uses F# project files to generate module imports, in addition to the imports already defined in files.
This alleviates somewhat the problem with files being compiled out of order, as now all F# project files will be imported.

Note: There is still some gap when collecting native (non-F#) import modules, in theory a native import can be left out if files are compiled out of order, i.e. if the last file in the project (where the modules are) gets compiled before the file with the native import.

Technically native imports are not properly supported yet, so it's still something to consider what the best solution is. One possible way to solve this is to make sure Fable compiling is done in order (or at least done before the last file in the project).


ncave:
@alexswan10k
I'm not sure if there is a good solution when caching is enabled and not all files are being compiled (only the ones that have changed). The only solution I see right now is to disable cache and always compile everything (and do it in proper project order). That would require adding a new compiler flag to serialize compilation, I believe.

Anyway, what this PR does should work most of the time, but it's not foolproof.

PR #2933, opened 2022-06-21
AST clean up: deduplicate member info

alfonsogarciacaro:
After the discussion about accessing the attributes of the parameters from the call site, I got embroiled in a refactor/cleanup of the AST. I still need to fix some issues and sync with latest changes in Rust, but I'm opening the PR so you can have a look if you have time @ncave @dbrattli and give your opinion. This got much messier than I hoped but the good news is I've no intentions of making more (big) changes in the AST for the Fable 4 cycle 😅 Please see the main points of the PR below:

- Remove duplication of member info in AST. 
- Remove entirely member info from AST. The info is retrieved instead through references (as with Entity info)
- ~~Now `MemberFunctionsAndValues` in entities are a map (lazily initialized on first access) so a member ref is an entity ref plus a string key for this map.~~
- We can generate member refs (this is usually done when generating object expressions in Replacements). `GeneratedMember` is added for this purpose to the AST and mocks most of the properties of `MemberFunctionsAndValue` interface. It could be made more complex, but for now is just intended to cover the existing use cases in the code base.
- Member declarations that implement an abstract signature have an extra `ImplementedSignatureRef`. This is important because unfortunately `IsGetter/IsSetter` doesn't work well for implementing members so we need to check the abstract signature in this case.
- Remove fields in the AST that were specific for a language (like `IsConstructor`/`ExportDefault` for JS or `IsConst` for Dart). In its place I've extended the `Tag` mechanism to other items so we can have flexibility to add extra information as plain strings. But I didn't do it for all items, maybe we should be consistent but this would break many patterns.


ncave:
@alfonsogarciacaro Would be interesting to see, once you make it work, if the new AST has any impact on performance (i.e. if wall clock compile time is less or more with the new AST).

src/Fable.AST/Fable.fs
```diff
@@ -58,40 +68,43 @@ type Constraint =
...
     abstract HasSpread: bool
-    abstract IsMangled: bool
     abstract IsInline: bool
     abstract IsPublic: bool
     abstract IsInstance: bool
     abstract IsValue: bool
     abstract IsMutable: bool
     abstract IsGetter: bool
     abstract IsSetter: bool
-    abstract IsProperty: bool
```
ncave, line 81: @alfonsogarciacaro Can we keep `IsProperty`?

ncave: @alfonsogarciacaro Added in #2934.

ncave:
@alfonsogarciacaro Fixed Rust tests in #2934.

alfonsogarciacaro:
@ncave Thanks a lot for reviewing the PR and sorry for not replying to your comments! 🙇 I got a bit obsessed with this and I wanted to make it work. At the end, I had problems with storing the members as maps in the entities, because it's not guaranteed that the Entity is always the same actual object so the members map was initialized multiple times. For now, I've tried to unify the different ways to find a member in [a single method](https://github.com/fable-compiler/Fable/blob/a1568f737b983d6c21f5cad32fa61639018e74ab/src/Fable.Transforms/FSharp2Fable.Util.fs#L43-L67) which is working well and fortunately it doesn't seem to affect performance too much. But maybe we can consider later to create some caches on the fly so we don't need to retrieve them all the time (apparently FCS recreates them for every `TryGetMemberFunctionAndValues` call) and we can also create some maps for constant-time access.

Thanks a lot for #2934! Sorry I couldn't merge because I made more changes locally in between, but I included the bits needed to fix the Rusts tests in 629cd015d9a075441173ce0c0c6a0dfd75473974

About the `IsProperty` members, I'm a bit confused by them. When debugging, apparently they don't appear for all entities but only those with extension members (?) and doesn't seem to add much value because the getters/setters are listed separately. If I'm not wrong, `IsProperty` members are mainly useful as a way to access the getter and setter methods (and when debugging I also found this to be non-consistent), but we don't have these properties at the moment in Fable `MemberFunctionOrValue` interface. Also, conceptually are a bit problematic because Fable doesn't considered indexed properties as actual getters/setters so we can get some contradictory cases. So for now I thought it'd be simpler to ignore them in Fable, but we can discuss bringing them back if necessary.

In any case, PR seems finally ready so if you want to have a last look I promise I'll respond to your comments this time 😉

ncave:
@alfonsogarciacaro Looks good to me, ready to merge.

> we can discuss bringing them back if necessary

I still think preserving full fidelity of F# AST (all entities and methods) is better, as some plugin might need that in the future, especially since in this case it doesn't cost anything to preserve it, so I might submit a PR adding `IsProperty` back, and we can discuss.

alfonsogarciacaro:
Ok, you convinced me, I added back `IsProperty` 😸 I was concerned it could interfere when finding getters/setters but it seems to work fine. Also the inconsistency with indexed properties (they are getters/setters in F# AST but not in Fable AST) but we can see how this works and resolve it in the future (indexed properties are rare in F# anyways).

One "problem" we have with getter/setter resolution is that for declarations the resolution is done in the last step (Fable2Babel, Fable2Python, etc) while for calls, it's done in the [FSharp2Fable step](https://github.com/fable-compiler/Fable/blob/09103fc01d6a1a03734c72de1b8df95c4f427cc4/src/Fable.Transforms/FSharp2Fable.Util.fs#L1993-L2019). Also, currently we're compiling getters/setters as actual getters/setters in the target languages only if the members are not mangled.

PR #2935, opened 2022-06-23
Attempt to fix Python imports

alfonsogarciacaro:


dbrattli:
@alfonsogarciacaro I'll start looking into this. This seems to have broken compilation of NuGet references w/Fable. I was planning to handle this a different way, so perhaps now is the right time since stuff needs to be done anyways.

alfonsogarciacaro:
@dbrattli Thanks for looking into this. Do you have an example of what's not working? I did the changes because Fable.Python was not working for the Python sample. After this PR it seemed to go fine: https://github.com/alfonsogarciacaro/fable-py-sample

dbrattli:
@alfonsogarciacaro It broke https://github.com/fable-compiler/Fable.Python and in particular NuGet dependencies like with https://github.com/fable-compiler/Fable.Python/tree/main/examples/timeflies 

I'm experimenting with translating F# projects to proper Python packages and using [Poetry](https://python-poetry.org/) (think Paket) local references instead in pyproject.toml files which are the standard Python package files. This could enable us to easily decide to use local (static) Fable.Core or pre-built from PyPI (dynamic). With poetry the we would have super-simple dependencies since we could just reference the packages without having to know where they are located.

I tried this yesterday and it looks to be working. This is how the `pyproject toml` file would look for timeflies:

```toml
[tool.poetry]
name = "timeflies"
version = "0.1.0"
description = "Fable Time Flies Demo"
authors = ["Dag Brattli <dag@brattli.net>"]
license = "MIT"

[tool.poetry.dependencies]
python = "^3.9"
fable-library = ""
fable-python = ""
fsharp-control-async-rx = ""

[tool.poetry.dev-dependencies]
fable-library = {path = "./fable_modules/fable-library", develop = true}
fable-python = {path = "./fable_modules/fable-python", develop = true}
fsharp-control-async-rx = {path = "./fable_modules/fsharp-control-async-rx", develop = true}

[tool.fable]
package-dir = "."

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

Then the files reference the dependencies in a super simple direct way like this:

```py
from queue import Queue
import sys
from tkinter import Tk, Frame, Label
from typing import Any, List, Tuple, TypeVar, Callable
from fable_library.async_ import run_synchronously
from fable_library.async_builder import singleton, Async
from fable_library.mailbox_processor import receive, start, post
from fable_library.reflection import TypeInfo, class_type, int32_type, union_type
from fable_library.seq import map_indexed, to_list
from fable_library.string import to_console, interpolate, printf
from fable_library.types import Array, Union
from fable_library.util import ignore
from fable_python.stdlib.tk_inter import Event
from fsharp_control_async_rx.async_observable import (
    AsyncRx_subject,
    AsyncRx_flatMap,
    AsyncRx_map,
    AsyncRx_delay,
    AsyncRx_ofSeq,
    FSharp_Control_IAsyncObservable_1__IAsyncObservable_1_SubscribeAsync_307B9986,
)
```

I need to experiment a bit more this weekend, but it looks really promising so I don't think we should try to fix it since it might all be simplified if I can make this work.

dbrattli:
@alfonsogarciacaro It will require us to generate proper Python packages so F# projects needs to go in a two-level folder structure e.g: `fable_modules/fable-library/fable_library` where `fable-library` contains the `pyproject.toml` file:
Screenshot 2022-07-23 at 08 34 33

dbrattli:
@alfonsogarciacaro See here for an example of how Python packages needs to be structured: https://python-poetry.org/docs/basic-usage/

dbrattli:
@alfonsogarciacaro PS: your example may work, but for more complex NuGet's that needs to reference Fable.Python (or other libraries) the code would need to find Fable.Library going up into `..fable_library` or `...fable_library` which is currently not working.

dbrattli:
@alfonsogarciacaro Using packages and `pyproject.toml` solves the problem for complex project references, but is overkill for simple fsx scripts like your example. But they could then simply just pip install the packages they depend on if we also deploy Fable.Library (fable-library) to PyPI

alfonsogarciacaro:
@dbrattli Something I may be misunderstanding: when you talk about Python packages, it's not about packages that are distributed through a Python register. Fable packages will still be distributed through Nuget but Fable will generate local folders that will **look like** packages to Poetry. Is that right?

dbrattli:
Yes @alfonsogarciacaro , they will look like packages to Poetry, and then we are finally free to decide if we want to use that generated package (or a pre-built from the Python PyPI register). If Poetry finds a combination of local and remote versions of the same package in the package resolution (i.e when handling a dependency chain of Fable NuGets) it will pick the best fit (according to version rules). 

See example at: https://github.com/fable-compiler/Fable.Python/pull/59/files#diff-09ebea1cdb8693ef913f1475adef7cfa878ff57fa03c66d3a527c5ac1df98fe6

For timefiles it references the generated packages, but it could have referenced https://pypi.org/project/fable-library/ instead

dbrattli:
@alfonsogarciacaro Then F# libraries like AsyncRx can bundle a `pyproject.toml` file so dependency handling will be handled all the way without having to use a tool like e.g Femto. 

- https://github.com/dbrattli/AsyncRx/blob/main/src/FSharp.Control.AsyncRx.fsproj#L30
- https://github.com/dbrattli/AsyncRx/blob/main/pyproject.toml

TODO: is to have fable-compiler update that `pyproject.toml` file when compiling so you don't have to add the entries manually yourself.

PR #2941, opened 2022-06-28
[WIP] Fix #2869: Avoid unnecessary updates

### LINKED ISSUE #2869: Avoid unnecessary updates
**UPDATE:** This is a good first issue for contributors as it doesn't require specific knowledge about Fable code. See the steps necessary to fix in [this comment](https://github.com/fable-compiler/Fable/issues/2869#issuecomment-1153577474) below.

This is a behaviour I detected after reporting #2868 

### Description

Let say that in my project I have the following files:

- `CssModules.fs`
- `App.fs`
- `Login.fs`
- `Navbar.fs`

and `CssModules.fs` is referenced in all the other files.

When `CssModules.fs` change Fable is recompiling all the files that depends on it and rewrite them on the disk even if nothing changed in the output of each of these files.

This cause a lot unnecessary "hot update" to be triggered by the bundlers. I think Fable should only write a file on the disk if the output of the compilation for that file changed since the last compilation.

### Related information

* Fable version: `3.7.5`
* Operating system: Windows

hensou:
- Keep output in a memory stream
- Only write to file if there is a change or if the file is new

Fix #2869

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -144,11 +144,56 @@ module private Util =
             if fileExt.EndsWith(".ts") then Path.replaceExtension ".js" fileExt else fileExt
         let sourceDir = Path.GetDirectoryName(sourcePath)
         let targetDir = Path.GetDirectoryName(targetPath)
-        let stream = new IO.StreamWriter(targetPath)
+        let memoryStream = new IO.MemoryStream()
+        let stream = new IO.StreamWriter(memoryStream)
         let mapGenerator = lazy (SourceMapSharp.SourceMapGenerator(?sourceRoot = cliArgs.SourceMapsRoot))
+        let WriteStreamToFileAsync(stream : IO.Stream, filePath : string) = async {
+            stream.Seek(0, IO.SeekOrigin.Begin) |> ignore
+            use fileStream = new IO.StreamWriter(filePath)
+            do! stream.CopyToAsync(fileStream.BaseStream) |> Async.AwaitTask
+            do! fileStream.FlushAsync() |> Async.AwaitTask
+        }
+        let IsMemoryStreamEqualToFileAsync() = async {
+            memoryStream.Seek(0, IO.SeekOrigin.Begin) |> ignore
+            use fileStream = IO.File.OpenRead(targetPath)
+
+            let bufferSize = 1024
+
+            let readBytesAsync (stream: IO.Stream) size = async {
+                let buffer = Array.zeroCreate size
+                let rec nextBytesAsync() = async {
+                    let! count = stream.AsyncRead(buffer, 0, size)
+                    if count = 0 then return None
+                    else 
+                        let res = 
+                            if count = size then buffer
+                            else buffer |> Seq.take count |> Array.ofSeq
+                        return Some(res)
+                }
+                return! nextBytesAsync()
+            }
+
+            let rec compareBytesAsync bytes1 bytes2 = async {
+                match bytes1, bytes2 with
+                | Some(b1), Some(b2) when b1 <> b2 -> return false
+                | Some(_), Some(_) -> 
+                    let! nextBytes1 = readBytesAsync memoryStream bufferSize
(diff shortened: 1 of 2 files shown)


src/Fable.Cli/Main.fs
```diff
@@ -144,11 +144,56 @@ module private Util =
...
+                    let! nextBytes1 = readBytesAsync memoryStream bufferSize
+                    let! nextBytes2 = readBytesAsync fileStream bufferSize
+                    return! compareBytesAsync nextBytes1 nextBytes2
+                | None, None -> return true
+                | _ -> return false
+            }
+
+            let! memoryBytes = readBytesAsync memoryStream bufferSize
+            let! fileBytes = readBytesAsync fileStream bufferSize
+            return! compareBytesAsync memoryBytes fileBytes
```
kerams, line 189: You could first compare `memoryBytes.Lenght` and the file size before starting to read the latter from disk and comparing byte by byte.

src/Fable.Cli/Main.fs
```diff
@@ -144,11 +144,56 @@ module private Util =
...
+            do! fileStream.FlushAsync() |> Async.AwaitTask
+        }
+        let IsMemoryStreamEqualToFileAsync() = async {
+            memoryStream.Seek(0, IO.SeekOrigin.Begin) |> ignore
+            use fileStream = IO.File.OpenRead(targetPath)
+
+            let bufferSize = 1024
+
+            let readBytesAsync (stream: IO.Stream) size = async {
+                let buffer = Array.zeroCreate size
```
kerams, line 163: A little nitpick, maybe pull this buffer up so that we don't allocate additional 1KB on every chunk read?

hensou: Sure, it makes sense, I'll do that.

src/Fable.Cli/Main.fs
```diff
@@ -144,11 +144,55 @@ module private Util =
...
+
+            let buffer = Array.zeroCreate 1024
+
+            let readBytesAsync (stream: IO.Stream) = async {
+                let rec nextBytesAsync() = async {
+                    let! count = stream.AsyncRead(buffer, 0, buffer.Length)
+                    if count = 0 then return None
+                    else 
+                        let res = 
+                            if count = buffer.Length then buffer
```
alfonsogarciacaro, line 168: As we are using a single buffer, if `count = 0` we just return a reference to the buffer. But in the lines below:

```fsharp
                    let! nextBytes1 = readBytesAsync memoryStream
                    let! nextBytes2 = readBytesAsync fileStream 
```

`nextBytes1` is going to be same as `nextBytes2` (if count is 0 in both cases), because `readBytesAsync` returns the same array reference in both cases and the second `readBytesAsync` will overwrite the buffer.

alfonsogarciacaro (approved):

alfonsogarciacaro:
This is fantastic @hensou, thanks a lot for this! I've rewritten a bit the `IsMemoryStreamEqualToFileAsync` function to fix the issue with the shared buffer and simplify a bit the code. I hope that's ok with you.

```fsharp
        let IsMemoryStreamEqualToFileAsync() = async {
            let areStreamsEqualAsync (stream1: IO.Stream) (stream2: IO.Stream) =
                let buffer1 = Array.zeroCreate<byte> 1024
                let buffer2 = Array.zeroCreate<byte> 1024

                let rec areStreamsEqualAsync() = async {
                    let! count1 = stream1.AsyncRead(buffer1, 0, buffer1.Length)
                    let! count2 = stream2.AsyncRead(buffer2, 0, buffer2.Length)
                    match count1, count2 with
                    | 0, 0 -> return true
                    | count1, count2 when count1 = count2 && buffer1 = buffer2 ->
                        return! areStreamsEqualAsync()
                    | _ -> return false
                }

                areStreamsEqualAsync()

            memoryStream.Seek(0, IO.SeekOrigin.Begin) |> ignore
            use fileStream = IO.File.OpenRead(targetPath)

            return! areStreamsEqualAsync memoryStream fileStream
        }
```

PR #2945, opened 2022-06-30
[WIP] Writes sourceMapping information to the memory stream when needed

### LINKED ISSUE #2869: Avoid unnecessary updates

hensou:
SourceMapping information was not being written to the memory stream, making it always write to disk on save. This commit writes the respective information to the memory stream.

fixes #2869

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -153,7 +153,12 @@ module private Util =
             use fileStream = new IO.StreamWriter(filePath)
             do! stream.CopyToAsync(fileStream.BaseStream) |> Async.AwaitTask
             do! fileStream.FlushAsync() |> Async.AwaitTask
-            return true
+
+            if cliArgs.SourceMaps then
+                let mapPath = filePath + ".map"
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Sorry @hensou, the title still includes [WIP] in the title, but it seems to be working. So I will merge so we can push a new release to test it out. Thanks again!

PR #2949, opened 2022-07-05
Main fcs

ncave:
- Updated .NET binaries to latest .NET 6.0.6
- Updated FCS-Fable to latest.
- Updated FCS binaries to latest.

Note: Unfortunately I had to redo the entire `FCS-Fable` PR from scratch :(
There was just too much file moving, reformatting and code churn for git to handle, so all PR files had to be redone manually. Hopefully that doesn't happen too often.

@alfonsogarciacaro There seems to be a new `FSharpGlyph.TypeParameter` union case, I'll leave it up to you to add the mapping for it in [Interfaces/Glyph](https://github.com/fable-compiler/Fable/blob/main/src/fable-standalone/src/Interfaces.fs#L6).


alfonsogarciacaro:
Wow, that was quite a lot of work. Thanks a lot for this @ncave! Yes, let's hope they don't reshuffle the files to often 🤞

ncave:
@alfonsogarciacaro Next step will be to figure out where/why did FCS lose some performance along the way.

alfonsogarciacaro:
@ncave Yes, we should log somewhere the compilation times of fable-standalone so we can see the trend. Unfortunately, it seems Github removes the full log of old builds so I cannot see exactly when it started to take longer. I'll try to run some builds with past versions locally. My feeling, though I may be wrong, is this because some changes in Fable and not in FCS as we didn't update it for a while. It may be related to [resolveInlineExpr](https://github.com/fable-compiler/Fable/blob/e12a9af0a18daa5c7a6b2fdb1f0f1bd5d867405a/src/Fable.Transforms/FSharp2Fable.fs#L1484) as I had to cover more cases to fix issues, so it does quite a bit of work, but I'm not 100% sure.

PR #2956, opened 2022-07-14
Updated fable-standalone

ncave:
- Updated `fable-standalone` with language support.
- Removed `EscapeStringLiteral`, as language printers know best how to print their strings.

### DIFF
--- a/src/Fable.Transforms/Python/Fable2Python.fs
+++ b/src/Fable.Transforms/Python/Fable2Python.fs
@@ -546,7 +546,7 @@ module Helpers =
             | ""
             | "." -> fileDir
             | _ ->
-                IO.Path.GetFullPath(Path.Combine(Path.GetDirectoryName(com.CurrentFile), fileDir))
+                Path.GetFullPath(Path.Combine(Path.GetDirectoryName(com.CurrentFile), fileDir))
                 |> Path.normalizePath
 
         let outDir = com.OutputDir |> Option.defaultValue projDir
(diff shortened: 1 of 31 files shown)


ncave:
@alfonsogarciacaro This should move us closer to a new REPL for Fable 4. Perhaps all that's needed is to expose the target language as a drop-down, and add a condition to only run the generated code if target is `JavaScript`.

alfonsogarciacaro:
This is great, thanks a lot @ncave! Sorry I couldn't look into this, but I think I can add the dropdown now that you've made the hard part :) About `EscapeStringLiteral` originally it was to aovid the WebUtility dependency, but if we can do it with custom code, that's much better :+1:

src/Fable.Transforms/Python/Fable2Python.fs
```diff
@@ -546,7 +546,7 @@ module Helpers =
             | ""
             | "." -> fileDir
             | _ ->
-                IO.Path.GetFullPath(Path.Combine(Path.GetDirectoryName(com.CurrentFile), fileDir))
+                Path.GetFullPath(Path.Combine(Path.GetDirectoryName(com.CurrentFile), fileDir))
```
alfonsogarciacaro, line 549: I usually like to prefix with `IO.` to make sure we're not using the internal Fable polyfills but probably we should resolve the potential confusion in a better way (if Fable's polyfill doesn't have the exact same behavior we should change the name).

ncave: @alfonsogarciacaro In this case we do want the Fable prelude version of it, which compile to .NET `Sytem.IO.Path.GetFullPath` when not compiled with Fable.

alfonsogarciacaro:
Buf, there're so many moving pieces in the REPL that every time I touch something I've to spend a lot of time investigating if all the necessary files have been copied properly 😅 It seems that some of the working JS samples for current repl are not working now so I need to look into this. But for now I've published the experimental repl4 in this url: https://fable.io/repl4/

screencast

Hopefully this will help people that want to have a quick feeling of how Fable 4 is working. Thanks again for your work @ncave!

PR #2957, opened 2022-07-17
[Python] fixes python test wrt culture info

dbrattli:
Fixes python tests to use invariant culture to not output floats with commas instead of dots for decimal separator. Please check it looks ok since it touches non-python code as well @alfonsogarciacaro


alfonsogarciacaro (approved):
Looks good to me!

PR #2966, opened 2022-07-21
[Rust] Made Async optional

ncave:
- Made `Async` module optional to speed up compilation when not needed.
- It can be enabled by adding `features = ["futures"]` in `Cargo.toml` where needed, like this:
```toml
[dependencies]
fable_library_rust = { path = "../../fable-library-rust", features = ["futures"] }
```

### DIFF
--- a/src/fable-library-rust/Cargo.toml
+++ b/src/fable-library-rust/Cargo.toml
@@ -4,4 +4,4 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-futures = { version = "0.3.21", features = ["executor", "thread-pool" ] }
+futures = { version = "0.3.21", features = ["executor", "thread-pool" ], optional = true }
--- a/src/fable-library-rust/src/Async.rs
+++ b/src/fable-library-rust/src/Async.rs
@@ -1,5 +1,6 @@
 #![allow(non_snake_case)]
 
+#[cfg(feature = "futures")]
 pub mod Async_ {
     use std::future::{self, Future, ready};
     use std::pin::Pin;
@@ -61,6 +62,7 @@ pub mod Async_ {
     }
 }
 
+#[cfg(feature = "futures")]
 pub mod AsyncBuilder_ {
     use std::{sync::Arc, pin::Pin, future::{Future, ready}};
 
@@ -98,6 +100,7 @@ pub mod AsyncBuilder_ {
     }
 }
 
+#[cfg(feature = "futures")]
 pub mod ThreadPool {
     use std::sync::RwLock;
 
@@ -116,6 +119,7 @@ pub mod ThreadPool {
     }
 }
 
+#[cfg(feature = "futures")]
 pub mod Task_ {
     use std::{sync::{Arc, RwLock}, thread::{self, JoinHandle}, time::Duration, task::Poll, pin::Pin};
 
@@ -292,6 +296,7 @@ pub mod Task_ {
     }
 }
 
+#[cfg(feature = "futures")]
 pub mod TaskBuilder_ {
     use std::{sync::Arc, rc::Rc};
 
--- a/tests/Rust/Cargo.toml
+++ b/tests/Rust/Cargo.toml
@@ -4,4 +4,4 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-fable_library_rust = { path = "../../fable-library-rust" }
+fable_library_rust = { path = "../../fable-library-rust", features = ["futures"] }


ncave:
@alexswan10k

This naming is likely temporary, this feature flag will be most probably be renamed or broken down to better represent the optional modules. Let me know what you think, should we have one feature for each sub-module in `Async`?

alexswan10k:
All in one is probably the easiest place to start, and leads to the simplest user experience. The two modules are relatively dependent right now too, so I think this makes sense. Can always increase granularity down the road if needed.

I was going to start looking at the whole Rc redirection thing next so this ties in nicely with being able to switch between them using this switch. You happy with that?

ncave:
@alexswan10k
> By the way - do we have a way to output something like this from F#?

Yes, we already have the aptly named `Attr` attribute in the Rust namespace. I was debating myself whether to change it to `RustAttr`, even if you have to write `Rust.RustAttr` sometimes. Let me know what you think.

Currently supported on functions, but we have to add support for it on modules and types and other top-level items.

See `MiscTests` for an example, but could be something like this:

```fs
[<Attr("cfg", [|"feature = \"futures\""|])>]       // #[cfg(feature = "futures")]

[<Attr("cfg", [|"not(feature = \"futures\")"|])>]  // #[cfg(not(feature = "futures"))]
```

Only `outer` attributes are supported for now (attached to an item). If we need `inner` attributes, we'll have to figure out how to map them, the easiest thing will be to just declare them on some item as well. Perhaps if the attribute name starts with a `!` to mean `inner`, or do you prefer an explicit attribute propert like `, IsInner = true`?

ncave:
@alexswan10k 
Then again, perhaps it's best to be explicit and rename the `Attr` to `OuterAttr`. I can make this change, if you agree.

alexswan10k:
Happy with all of the above. There seems to be an unexpected problem. The attribute works great, but the imports sit outside the definition!

ncave:
@alexswan10k
> the imports sit outside the definition!

Yeah, could be an issue.
Ideally the imports should be harmless if everything inside is properly wrapped in a feature. For example all modules in `Async` are wrapped in feature, so it's fine to import `Async`.

alexswan10k:
Hmm interesting. We can do this but we would effectively need to create empty stubs for every function.

I have done this for example (so that an empty module is always present), and am getting problems still because the functions do not exist
```
#[cfg(not(feature = "futures"))]
pub mod TaskBuilder_ {

}
```

```
24 | use fable_library_rust::TaskBuilder_::new;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `new` in `module_12b3243d::TaskBuilder_`
```

ncave:
@alexswan10k Yes, that's what I meant above.
Perhaps we can check if the declaring entity of an import has an outer or inner attribute, and apply it to the import as well.
I'll see if that works.

ncave:
@alexswan10k Do we want to copy all attributes from the declaring entity to each import, or do we want to somehow restrict that to only some attributes (if it has a the word "feature" in it, perhaps)? Or do we need a separate attribute just for features (ideally not, if we can avoid it)?

alexswan10k:
I think the way they get around it is simply by nesting the imports as part of the module which is optionally turned on.

I wonder if it would be possible to do something like that - have a F# module defined in the module, which is tagged at root with this attr. The problem currently is that the imports are outside of the root module. Could they be moved inside?

ncave:
@alexswan10k Another option achieving the same is perhaps an inner feature attribute in the module? Would that work?

The client importing from that module has to wrap in features either way, regardless of how it's done.

alexswan10k:
Happy to try. Forgive me, I am not 100% sure how InnerAttr's work relative to Attr's. Would you be able to sketch out an example?

ncave:
@alexswan10k I'll try both and see which works better, thanks for brainstorming!

alexswan10k:
NP :) Do shout if there is anything I can help with.

ncave:
@alexswan10k Here is how it looks right now after I added attributes for modules and types (already merged):

```rust
#![allow(dead_code)]
#![allow(non_snake_case)]
#![allow(non_camel_case_types)]
#![allow(non_upper_case_globals)]
#![allow(unused_parens)]
#![allow(unused_imports)]
#![allow(unused_variables)]
#![allow(unused_attributes)]
use fable_library_rust::Async_::Async_1;
use fable_library_rust::Async_::awaitTask;
use fable_library_rust::Async_::runSynchronously;
use fable_library_rust::Async_::startAsTask;
use fable_library_rust::AsyncBuilder_::bind;
use fable_library_rust::AsyncBuilder_::delay;
use fable_library_rust::AsyncBuilder_::r_return;
use fable_library_rust::Native_::Arc;
use fable_library_rust::Native_::Lrc;
use fable_library_rust::Task_::Task_1;
use fable_library_rust::Task_::bind as bind_1;
use fable_library_rust::Task_::delay as delay_1;
use fable_library_rust::Task_::from_result;
use fable_library_rust::Task_::r_return as r_return_1;
use fable_library_rust::TaskBuilder_::TaskBuilder;
use fable_library_rust::TaskBuilder_::new;
pub mod Fable {
    use super::*;
    pub mod Tests {
        use super::*;
        pub mod AsyncTests {
            #![cfg(feature="futures")]   // <--- attr added on top module ends up here, modules above are namespaces
            use super::*;
            #[test]
            pub fn shouldExecPrim() {
            //...
            }
  ```
  
 Still thinking about what's the best approach, feel free to shout other ideas if you think of something.

alexswan10k:
Hmm interesting. So outer attr looks like it will work if the imports can be moved within the root module.

I am just wondering if this presents problems for namespaces at root rather than modules. I guess these are also just modules in Rust so probably fine.

PR #2968, opened 2022-07-21
[Rust] Lrc - Abstracting over Rc and Arc using feature

alexswan10k:
* Create an alias Lrc which is able to represent a Rc or an Arc depending on the feature switch "futures"
* Add Lrc as a explicit concept in transforms
* Tweak libs to redirect onto Lrc where manually hard-coded

### Testing:
* Run tests on futures (this is the default in CI) - Call sites correctly represent Arc all the way down
* Run tests on no futures + comment out async tests - Call sites correctly represent Rc

### Further things that still need looking at
* Is 1 feature switch sufficient granularity? (simplest user experience but less control)
* Add a new test-rust-something and push feature on-off to cargo run call. Conditionally enable "futures" feature switch for lib. Adjust CI so that it can be run for both modes
* MutCell

### DIFF
--- a/src/Fable.Core/Fable.Core.Rust.fs
+++ b/src/Fable.Core/Fable.Core.Rust.fs
@@ -22,8 +22,9 @@ type InnerAttrAttribute private (name: string, value: string option, items: stri
 
 //Rc/Arc control
 type PointerType =
-    | Rc = 0
-    | Arc = 1
+    | Lrc = 0
+    | Rc = 1
+    | Arc = 2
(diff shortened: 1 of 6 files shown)


ncave:
@alexswan10k Looks great, is it ready to merge?

alexswan10k:
I think so yes, we can handle the details of the CI integration over here https://github.com/alexswan10k/Fable/pull/5
I can rebaseline this once merged

PR #2969, opened 2022-07-21
WIP - Add futures switch to CI and tests

alexswan10k:


alexswan10k:
@ncave here is the rebased PR for the CI integration, but with the problem that the tests cannot correctly be deactivated due to imports attempting to fetch items not present.

I made a few changes to the build script since https://github.com/alexswan10k/Fable/pull/5 to allow CI to skip the F# double build.

build.fsx
```diff
@@ -234,7 +234,7 @@ let buildLibraryRust() =
     copyFiles libraryDir "*.toml" buildDir
     copyFiles sourceDir "*.rs" outDir
 
-    runInDir buildDir ("cargo build")
+    // runInDir buildDir ("cargo build")
```
alexswan10k, line 237: @ncave here is an interesting question - Do we even need to do this here? Currently, you build anyway when you run the tests (everything builds from source in Rust), so this makes getting feedback harder when using the rust lang server to view tests when altering the AST.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -3688,13 +3688,19 @@ module Util =
...
                 // TODO: perhaps collect other use decls from usage in body
-                let useItem = mkGlobUseItem [] ["super"]
-                let useDecls = [useItem]
+                let useDecls =
+                    if isRoot && not isPath then
+                        let importItems = com.GetAllImports() |> transformImports com ctx
+                        importItems
+                    else
+                        let useItem = mkGlobUseItem [] ["super"]
+                        [useItem]
```
alexswan10k, line 3703: @ncave I am struggling to get this to work as intended. It seems that we can move the logic here so that imports are rendered inside the module.

My biggest problem here is that I cannot tell the difference between a "namespace" and the "root module", relative to arbitrarily defined submodules defined within the file. Have you any thoughts on some logic that might get us to this?

This PR in its current state correctly renders everything as we discussed, but it breaks certain library modules that have exactly 1 defined submodule etc.

alexswan10k: I think I have cracked it. 

This was way harder than I anticipated - I had to track the depth and surgically remove stuff from the context root scope as it left the module. Everything should be able to be run in CI now with the feature on and off, although I am a little nervous there may still be edge cases. It is good to have lots of tests though as nearly everything that wasn't quite right broke catastrophically, giving me some confidence.

Let me know what you think. I believe this is principally feature complete and working now but happy to make changes if there are obvious problems.

build.fsx
ncave: @alexswan10k Yes and no, technically no, we don't have to have it, but it is a useful early quick check if the `fable-library-rust` is broken by itself. If that's what's throwing the Rust lang server in the VS Code extension, then we can keep it commented. I tend to use the older Rust extension as it works better for me, not sure why, which one do you use?

alexswan10k: I use rust-analyzer, which I believe is the community favourite. I believe the official one has less features, and has some shortcomings. I have found rust-analyzer to work best for me.

I am more than happy to add this back in if you like, I was just trying to trim some fat :)

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -3834,6 +3845,14 @@ module Compiler =
...
 
             member _.GetAllImports() = imports.Values |> Seq.toList
+            member _.ClearAllImports ctx =
+                for importKv in imports do
+                    let depthsLeft = importKv.Value.Depths |> List.filter (fun dp -> dp > ctx.ModuleDepth)
+                    importKv.Value.Depths <- depthsLeft
+                    if importKv.Value.Depths.Length = 0 then
+                        imports.Remove(importKv.Key) |> ignore
+                        ctx.UsedNames.RootScope.Remove(importKv.Value.LocalIdent) |> ignore
+
```
ncave, line 3855: @alexswan10k Do you mind explaining what the import cleaning is supposed to do, what's the intent?

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: @alexswan10k Looks very minimal, I like it.

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: Sure. So this is necessary because when you exit a scope, the used names were previously memorized. This means that the parent scope will try and declare new aliases which do not line up with the useages in the code (it thinks there are collisions). By clearing out any idents which have gone out of scope (declared in a submodule), this fixes the problem.

PR #2972, opened 2022-07-22
[Rust] Added namespace entities

ncave:


ncave:
@alexswan10k
I tried to improve a bit on your solution, by further deduping some of the repeating imports across module levels.
Seems to work, although fundamentally is not perfect since we're transforming modules depth-first, so if a module adds an import that *later* appears also in it's parent, there is no way to take it back as it is already included.

alexswan10k:
> @alexswan10k
> 
> I tried to improve a bit on your solution, by further deduping some of the repeating imports across module levels.
> 
> Seems to work, although fundamentally is not perfect since we're transforming modules depth-first, so if a module adds an import that *later* appears also in its parent, there is no way to take it back as it is already included.

Indeed. Looks good. Let's see how we get on for now, can perhaps revisit if this does not cut it.

ncave:
@alexswan10k
We can probably improve on that if process module member declarations out of order (i.e. process inner modules last), then assemble them back in the correct order.

ncave:
@alexswan10k Also the new Fable 4 [REPL](https://fable.io/repl4/) is up (with language support), if you haven't seen it (thanks @alfonsogarciacaro!)

alexswan10k:
This is really cool tbh, I will have a bit more of a play tomorrow. In the context of non js languages, this is great for quickly testing output and exploring behaviour. 

Maybe one day it can be joined up with https://play.rust-lang.org/

Great work.

alexswan10k:
Just been playing around with the REPL. Pretty exciting, I already think there is enough here to be able to use it for sketching out high-level domain models + logic rapidly, even though right now you probably are going to struggle to write a full program in it.

A few observations:
* Do expressions are not yet implemented! Should be an easy one to add.
* for some reserved names I get a r#name. Is this valid Rust? I should check this..
* thread_local was a great start but it probably isn't going to cut it for futures mode. Needs some thought this. I have a dirty unsafe way of doing this in the Async module (for the thread pool), but there may be better ways.

On a side note that has given me a random thought, I wanted to quickly scribble down. I think in the Js version, outputted files default to a convention of modulename.fable.js so they are easy to .gitignore, and thus easy to interleave with other js/ts files. Perhaps we can do the same in Rust? I presume this convention might not play well with the module system though so we may need to use a dash instead.

I do still feel like most benefit will come from being able to combine F# generated Rust code directly with a bigger Rust program (where the core harness/api integration happens in Rust), and this is mostly how I plan to use it. For this reason, I also wonder if it makes sense to augment the existing libraries with some interop helpers too. Is there any way we can do this currently as they are F# generated? Specifically, I imagine it would be useful to implement [From](https://doc.rust-lang.org/std/convert/trait.From.html) and [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) traits for things like List/Set/Array/Map for things like vec etc.

ncave:
@alexswan10k 
> Modude do bindings are not yet implemented
   
   - Yes, those are normally done with IIFE, which works just fine on closures in function scope, but I haven't looked if it will works on static variables (which is what module let bindings are).
   
   - Another feature not yet implemented is object expressions (anonimous objects).  I was thinking maybe just generate unique local temp class names for them, perhaps that will work.

> for some reserved names I get a r#name. Is this valid Rust?
    
 Yes, this is intentional, that's how Rust deals with reserved symbols. See `rawIdent` and `sanitizeIdent` in `Rust.AST.Helpers.fs` 
    
> thread_local was a great start but it probably isn't going to cut it for futures mode.

Yes, this  is for single-threaded. For futures, we need to synchronize the access to mutable module let bindings, which people shouldn't be using anyway in a functional language, but alas, they're part of the language, so we need something there.

> in the Js version, outputted files default to a convention of modulename.fable.js so they are easy to .gitignore, and thus easy to interleave with other js/ts files. Perhaps we can do the same in Rust?

That is already how it works in Fable for all languages including Rust. If you don't specify an output folder, it will put the `.rs` output file next to the `.fs` source file with an extension `.fs.rs` (or `.fs.js`, or `.fs.py`, etc). You can also use custom file extensions with the `-e` (`--extension`) parameter.

> I do still feel like most benefit will come from being able to combine F# generated Rust code directly with a bigger Rust program

I agree, we can perhaps add some native interop helpers to `fable-library-rust` (written in Rust) that can be used by client code to interop more easily with generated code.

> Following from that, I imagine the next elephant in the room will be how to serialize built-in types

Fable usually avoids including serialization directly, since it uually means taking on a big dependency. Normally it's done in a separate package (e.g. [Fable.SimpleJson](https://github.com/Zaid-Ajaj/Fable.SimpleJson)). Perhaps we can do something like that for Rust too.

Or, since we have a special way of dealing with dependencies in Rust using features, perhaps we can get away with adding a dependency on Serde. We'll have to add more dependencies anyway on packages for some core F# functionality like [decimal](https://docs.rs/rust_decimal/latest/rust_decimal/), [bigint](https://docs.rs/num-bigint/latest/num_bigint/) etc.

ncave:
@alexswan10k Another question is, can we support [no-std](https://docs.rust-embedded.org/book/intro/no-std.html) for some core version of the F# language.

ncave:
@alexswan10k Since we're talking about serialization, there are also all the binary serializers (e.g. [speedy](https://github.com/koute/speedy), [borsh](https://github.com/near/borsh) etc.). There are a lot of Rust packages that are useful, but bindings for them probably belong in their own projects (i.e. `Fable-Serde`, or something like that).

alexswan10k:
Agree good points.  I think what might be tricky here is a lot of them encourage decorating your types with some kind of auto implemented trait, which is great when you have total control over the code, but less great when it is a library (such as fable-core) which can/should not take on a dependency to that crate. I do think serde's [remove derive](https://serde.rs/remote-derive.html) approach is an interesting workaround though. Perhaps other libraries provide something similar.

The other thing I wanted to check was structs being wrapped in pointers. Was original behaviour not to keep value types unwrapped? This seems closest to how .Net works.

ncave:
@alexswan10k 
> I do think serde's [remove derive](https://serde.rs/remote-derive.html) approach is an interesting workaround

Yes, perhaps we can that approach for common interfaces like IEnumerable etc?

> structs being wrapped in pointers. Was original behaviour not to keep value types unwrapped? This seems closest to how .Net works

Yes, I think we tried to make this work consistently before the switch to pass by value, perhaps it's time to revisit. It will be nice to be able to reason about it the same way .NET works.

> All parameters are *passed by value* (unless you use the *byref/inref/outref* keywords to pass them *by reference*). Passing *a reference [ptr]* is not the same thing as *passing by* reference.

I clearly remember reading somewhere about a CLR JIT optimization for passing the `this` by reference for struct value type members, but I think we already do this for `self`. Also the struct return values can be sometimes optimized by RVO in some cases, and the JIT has a lot of flexibility to move value types to the heap when needed, unless restricted from doing so.

In any way, these are internal optimizations we can ignore for now and stick to the basic notion of *passing by value* [unless explicitly modified with a keyword](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#passing-structure-type-variables-by-reference).

So yes, we should try to make it work as expected, so people can follow the standard [.NET guidance on value types](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct):

> ✔️ CONSIDER defining a struct instead of a class if instances of the type are small and commonly short-lived or are commonly embedded in other objects.

> ❌ AVOID defining a struct unless the type has all of the following characteristics:
    - It logically represents a single value, similar to primitive types (int, double, etc.).
    - It has an instance size under 16 bytes.
    - It is immutable.
    - It will not have to be boxed frequently.
    
> In all other cases, you should define your types as classes.

ncave:
@alexswan10k Nevermind ;)

As usually happens when I go on a tangent without checking first, looks like we already pass structs by value as expected, so nothing to do, it just works ;)

alexswan10k:
Sounds good.

Thats a fair point actually, self could be problematic. It looks like it is passed by reference though if you start putting methods on a struct, which is actually correct. Looking at the output though for a non struct reference type it seems like we are doing some slightly weird stuff when returning self (cloning and then wrapping in a new Rc, which is not entirely correct). I cannot find the original conversation now where we hammered class methods out but there was definitely a way to implement with self represented as an optional wrapper rather than the unwrapped thing. I will dig around.

 I agree though. this probably needs another pass. I will perhaps try and take a look at this sometime next week.

ncave:
@alexswan10k 
It works just as you would expect for `[<Struct>]` classes, records and unions. And pure structs too, only there is some issue with pure struct implicit constructors that needs to be fixed, but it's unrelated.

ncave:
@alexswan10k Yes, return types probably need to be looked at if there is unnecesary wrapping/unwrapping, and we eventually want to support [byref returns](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/byrefs#byref-returns) too.

alexswan10k:
I think this is the example I was talking about https://github.com/ncave/Fable/pull/17#issuecomment-938847177

Now I am not how confident my past self was in testing this, but I do seem to remember checking this against the language server. In short, it looks like you can implement your own trait against a ```Rc<T>``` rather than T, thus ensuring self is a ```Lrc<T>``` for a class and ```T``` for struct.

Another side note is tuples are always by value. I imagine they should probably also be Lrc wrap by default, unless these are explicitly struct tuples. Just having a quick look through the code I am not sure why this is not working exactly. It seems to check this in isCopyableType. 

Oh well, one for next week.

ncave:
@alexswan10k 

> tuples are always by value. I imagine they should probably also be Lrc wrap by default, unless these are explicitly struct tuples

We can, but I believe we intentionally made all of them work as struct tuples, for simplicity. It's an implementation detail anyway, I don't think anybody actually relies on the fact that tuples are reference type in F#. I know it's a slippery slope, why not say the same for Records etc., but tuples are usually small, more similar to options and result/choice types.

In fact, perhaps in F# the default should be for tuples to be value types, until they become bigger than certain size.
Same thing for options and result/choice types.

I've probably linked this article before: https://bartoszsypytkowski.com/writing-high-performance-f-code/
We should strive to make sure most if not all the techniques from that article work as expected.

alfonsogarciacaro:
Sorry, already quite on top of the thread, but a quick note on IIFE: in Dart I managed to remove them by removing the duplicity of `TransformAsExpr` and `TransformAsStatements` in Fable2Babel. In Fable2Dart the transform functions return `Statement list * Expr`, so in case there's a statement (usually a variable declaration) in a place where only expressions are allowed, I can push it to the top until there's a place where I don't need the IIFE. Some situations require special care though, like `while` loops to make sure the statements are re-evaluated for every iteration: https://github.com/fable-compiler/Fable/blob/417808fd9c886c8d1bd2de183ea31e2f537611ff/src/Fable.Transforms/Dart/Fable2Dart.fs#L1594-L1604

Not sure if this can be applied to Rust, as its AST is expression-based, but I just wanted to mention it in case it's helpful.

ncave:
@alfonsogarciacaro In this case this is about IIFE for module do bindings.

alexswan10k:
@ncave good article, will definitely come back to this and have a read.

I have vaguely followed the[RFC discussion](https://github.com/fsharp/fslang-suggestions/issues/813) around trying to make tuples struct by default, and fully accept that this should probably be the default. What worries me a little is performance characteristics may be unexpectedly different between Rust and .Net, and there is no way to give the user control over which tuple type they are using if there is no distinction.

In theory tuples should be small, but even a tuple of 4 Rc pointers on a 64 bit system is min 8 bytes * 4 = 32. In practice it is almost definitely more than that because you also have the count, and alignment (my guess is thus 16 bytes*4 = 64 but cannot find the exact figure right now). If you started nesting actual structs this could quickly warrant using a pointer. The article implied 16 bytes was a good maximum to aim for.

ncave:
@alexswan10k Makes sense, I agree we should try not deviate (much) from the .NET F# implementation, so people can reason about the generated code paradigms using the F# documentation. So now by default tuples/records/unions are reference types, unless marked as `struct`. I left options/results/choices as value types always, as these are small. See #2979.

Also added struct anonymous records, and module do bindings (which should work on most platforms, except webasm).

alexswan10k:
This is great work as always @ncave, I do hope I have not forced your hand on this one though :) I am fully aware of how subjective some of these decisions are and am always happy to go with consensus.

Going to have a look at the unwrapped self problem next.

PR #2974, opened 2022-07-23
[Python] Use poetry for Python

dbrattli:
Will now generate proper Python packages within fable-modules and uses poetry with pyproject.toml to reference local (and PyPI) packages.


Sajadahf:
Can I use v3.6 for poetry python?

dbrattli:
@Sajadahf we will target Python 3.9+ after https://github.com/fable-compiler/Fable/pull/2974

PR #2980, opened 2022-07-25
fix: standalone import path for Python

### LINKED ISSUE #2977: [Python] Fable repl 4 imports not formatted correctly
### Description

In the Fable repl 4, the Python imports are not formatted correctly. Not sure what's wrong @ncave and what makes it different from the cli, but I would love to help out if you can point me in the right direction.

### Repro code

Default repl example

Expected result:

```py
from abc import abstractmethod
from typing import (Protocol, Optional, Any)
from fable_library.string import (to_console, printf)
from fable_library.types import Array
```

Actual result:

```py
from abc import abstractmethod
from typing import (Protocol, Optional, Any)
from fable-library/string.py import (to_console, printf)
from fable-library/types.py import Array
```

dbrattli:
Fixes #2977 

@alfonsogarciacaro I decided to go with @ncave suggestion since changing `getLibPath` to use dots (`.`) broke PythonWriter in the CLI that uses stuff like `Imports.getImportPath` etc.

### DIFF
--- a/src/fable-standalone/src/Worker/Worker.fs
+++ b/src/fable-standalone/src/Worker/Worker.fs
@@ -44,11 +44,14 @@ type State =
       Worker: ObservableWorker<WorkerRequest>
       CurrentResults: Map<string, IParseAndCheckResults> }
 
-type SourceWriter(sourceMaps: bool) =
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Cool, thanks!

PR #2981, opened 2022-07-25
Rust - Records and Unions now are correctly unwrapped when Struct attribute is used to tag as value

alexswan10k:
* Added isValue to Fable AST for Records and Unions in keeping with @ncave 's work here https://github.com/fable-compiler/Fable/pull/2978
* Fixed call sites
* Added some missing tests to codegen (although cannot actually prove more than both sides are symmetrical). Struct unions already had a test.
* Added placeholder for Lrc wrapped implementations - todo (see tagged comment below...)
* Bonus - Added explicit Box pointer type. If you are sure you are only going to use 1 copy without cloning, why pay the price of an Rc or an Arc?

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -409,9 +409,9 @@ type ValueKind =
     | NewArray of newKind: NewArrayKind * typ: Type * kind: ArrayKind
     | NewList of headAndTail: (Expr * Expr) option * typ: Type
     | NewTuple of values: Expr list * isStruct: bool
-    | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list * isStruct: bool
+    | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list
     | NewAnonymousRecord of values: Expr list * fieldNames: string [] * genArgs: Type list * isStruct: bool
-    | NewUnion of values: Expr list * tag: int * ref: EntityRef * genArgs: Type list * isStruct: bool
+    | NewUnion of values: Expr list * tag: int * ref: EntityRef * genArgs: Type list
     member this.Type =
         match this with
         | ThisValue t
@@ -428,9 +428,9 @@ type ValueKind =
         | NewArray (_, t, k) -> Array(t, k)
         | NewList (_, t) -> List t
         | NewTuple (exprs, isStruct) -> Tuple(exprs |> List.map (fun e -> e.Type), isStruct)
-        | NewRecord (_, ent, genArgs, isStruct) -> DeclaredType(ent, genArgs)
+        | NewRecord (_, ent, genArgs) -> DeclaredType(ent, genArgs)
         | NewAnonymousRecord (_, fieldNames, genArgs, isStruct) -> AnonymousRecordType(fieldNames, genArgs, isStruct)
-        | NewUnion (_, _, ent, genArgs, isStruct) -> DeclaredType(ent, genArgs)
+        | NewUnion (_, _, ent, genArgs) -> DeclaredType(ent, genArgs)
 
 type CallInfo =
     { ThisArg: Expr option
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -409,9 +409,9 @@ type ValueKind =
     | NewArray of newKind: NewArrayKind * typ: Type * kind: ArrayKind
     | NewList of headAndTail: (Expr * Expr) option * typ: Type
     | NewTuple of values: Expr list * isStruct: bool
-    | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list
+    | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list * isStruct: bool
     | NewAnonymousRecord of values: Expr list * fieldNames: string [] * genArgs: Type list * isStruct: bool
-    | NewUnion of values: Expr list * tag: int * ref: EntityRef * genArgs: Type list
+    | NewUnion of values: Expr list * tag: int * ref: EntityRef * genArgs: Type list * isStruct: bool
     member this.Type =
         match this with
         | ThisValue t
@@ -428,9 +428,9 @@ type ValueKind =
         | NewArray (_, t, k) -> Array(t, k)
         | NewList (_, t) -> List t
         | NewTuple (exprs, isStruct) -> Tuple(exprs |> List.map (fun e -> e.Type), isStruct)
-        | NewRecord (_, ent, genArgs) -> DeclaredType(ent, genArgs)
+        | NewRecord (_, ent, genArgs, isStruct) -> DeclaredType(ent, genArgs)
         | NewAnonymousRecord (_, fieldNames, genArgs, isStruct) -> AnonymousRecordType(fieldNames, genArgs, isStruct)
-        | NewUnion (_, _, ent, genArgs) -> DeclaredType(ent, genArgs)
+        | NewUnion (_, _, ent, genArgs, isStruct) -> DeclaredType(ent, genArgs)
 
 type CallInfo =
     { ThisArg: Expr option
--- a/src/Fable.Transforms/Rust/AST/Rust.AST.Symbols.fs
+++ b/src/Fable.Transforms/Rust/AST/Rust.AST.Symbols.fs
@@ -167,7 +167,7 @@ module kw =
     let RustPrelude = [
         "Copy"; "Send"; "Sized"; "Sync"; "Unpin";
         "drop"; "Drop"; "Fn"; "FnMut"; "FnOnce";
-        "Box"; "ToOwned"; "Clone";
+        "ToOwned"; "Clone";
         "PartialEq"; "PartialOrd"; "Eq"; "Ord";
         "AsRef"; "AsMut"; "Into"; "From"; "Default"
         "Iterator"; "Extend"; "IntoIterator";
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -421,11 +421,10 @@ module TypeInfo =
             | HasEmitAttribute _ -> None
             // do not make custom types Rc-wrapped by default. This prevents inconsistency between type and implementation emit
             | HasReferenceTypeAttribute ptrType ->
-                if not (isCopyableType com Set.empty typ) then
-                    Some ptrType
-                else None
-            | _ ->
-                if not (isCopyableType com Set.empty typ) then Some Lrc else None
+                Some ptrType
+            | ent ->
+                if ent.IsValueType then None
+                elif not (isCopyableType com Set.empty typ) then Some Lrc else None
 
     let isCloneableType (com: IRustCompiler) typ =
         match typ with
@@ -824,12 +823,14 @@ module TypeInfo =
                     //     let callee = com.TransformAsExpr(ctx, reflectionMethodExpr)
                     //     Expression.callExpression(callee, generics)
 
+        maybeWrapInPtrTy com ctx typ ty
+
+    let maybeWrapInPtrTy com ctx typ ty =
         match shouldBeRefCountWrapped com typ with
         | Some Lrc -> makeLrcTy com ctx ty
         | Some Rc -> makeRcTy com ctx ty
         | Some Arc -> makeArcTy com ctx ty
         | _ -> ty
-
 (*
     let transformReflectionInfo com ctx r (ent: Fable.Entity) generics =
         if ent.IsFSharpRecord then
@@ -1537,7 +1538,7 @@ module Util =
         then expr
         else expr |> makeLrcValue
 
-    let makeRecord (com: IRustCompiler) ctx r values entRef genArgs =
+    let makeRecord (com: IRustCompiler) ctx r isStruct values entRef genArgs =
         let ent = com.GetEntity(entRef)
         let fields =
             Seq.zip ent.FSharpFields values
@@ -1552,7 +1553,7 @@ module Util =
         let genArgs = transformGenArgs com ctx genArgs
         let path = makeFullNamePath ent.FullName genArgs
         let expr = mkStructExpr path fields // TODO: range
-        if isCopyableEntity com Set.empty ent
+        if isStruct || isCopyableEntity com Set.empty ent
         then expr
         else expr |> maybeWrapSmartPtr ent
 
@@ -1566,7 +1567,7 @@ module Util =
             else
                 fullName
 
-    let makeUnion (com: IRustCompiler) ctx r values tag entRef genArgs =
+    let makeUnion (com: IRustCompiler) ctx r isStruct values tag entRef genArgs =
         let ent = com.GetEntity(entRef)
         // let genArgs = transformGenArgs com ctx genArgs
         let unionCase = ent.UnionCases |> List.item tag
@@ -1576,7 +1577,7 @@ module Util =
             if List.isEmpty values
             then callee
             else callFunction com ctx None callee values
-        if isCopyableEntity com Set.empty ent || ent.FullName = Types.result
+        if isStruct || isCopyableEntity com Set.empty ent || ent.FullName = Types.result
         then expr
         else expr |> maybeWrapSmartPtr ent
 
@@ -1630,9 +1631,10 @@ module Util =
         | Fable.NewTuple (values, isStruct) -> makeTuple com ctx r isStruct values
         | Fable.NewList (headAndTail, typ) -> makeList com ctx r typ headAndTail
         | Fable.NewOption (value, typ, isStruct) -> makeOption com ctx r typ value isStruct
-        | Fable.NewRecord (values, entRef, genArgs) -> makeRecord com ctx r values entRef genArgs
+        | Fable.NewRecord (values, entRef, genArgs, isStruct) -> makeRecord com ctx r isStruct values entRef genArgs
         | Fable.NewAnonymousRecord (values, fieldNames, genArgs, isStruct) -> makeTuple com ctx r isStruct values
-        | Fable.NewUnion (values, tag, entRef, genArgs) -> makeUnion com ctx r values tag entRef genArgs
+        | Fable.NewUnion (values, tag, entRef, genArgs, isStruct) ->
+            makeUnion com ctx r isStruct values tag entRef genArgs
 
     let calcVarAttrsAndOnlyRef com ctx (e: Fable.Expr) =
         let t = e.Type
@@ -3331,7 +3333,7 @@ module Util =
         let idents = getEntityFieldsAsIdents com ent
         let fields = idents |> List.map Fable.IdentExpr
         let genArgs = getEntityGenArgs ent
-        let body = Fable.Value(Fable.NewRecord(fields, ent.Ref, genArgs), None)
+        let body = Fable.Value(Fable.NewRecord(fields, ent.Ref, genArgs, ent.IsValueType), None)
         let name = declName //TODO: is this always correct?
         let paramTypes = idents |> List.map (fun id -> id.Type)
         let memberRef = Fable.GeneratedMember.Function(name, paramTypes, body.Type, entRef = ent.Ref)
@@ -3350,7 +3352,7 @@ module Util =
                 let fieldIdents = idents |> List.map (fun id -> Map.find id.Name identMap)
                 let fields = fieldIdents |> List.map Fable.IdentExpr
                 let genArgs = getEntityGenArgs ent
-                let body = Fable.Value(Fable.NewRecord(fields, ent.Ref, genArgs), None)
+                let body = Fable.Value(Fable.NewRecord(fields, ent.Ref, genArgs, ent.IsValueType), None)
 
                 // add return value after the body
                 let body = Fable.Sequential (exprs @ [body])
@@ -3530,9 +3532,14 @@ module Util =
                     []
                 else
                     let ty =
-                        let genArgs = getEntityGenArgs ent |> transformGenArgs com ctx
+                        let genArgs = getEntityGenArgs ent
+                        let entRef = Fable.DeclaredType(entRef, getEntityGenArgs ent)
+                        let genArgs = genArgs |> transformGenArgs com ctx
                         let bounds = mkTypeTraitGenericBound [entName] genArgs
-                        mkTraitTy [bounds]
+                        let ty = mkTraitTy [bounds]
+                        // TODO : Wrapping the impl block in a Lrc breaks casting (see IEnumerable), because it no longer implements the interface for T but for Lrc<T>
+                        //maybeWrapInPtrTy com ctx entRef ty
+                        ty
                     let genArgs = getEntityGenArgs tEnt
                     let nameParts =
                         if isNonInterface then tEntRef.FullName + "Methods"
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -189,6 +189,9 @@ module TypeInfo =
     let makeArcTy com ctx (ty: Rust.Ty): Rust.Ty =
         [ty] |> makeImportType com ctx "Native" "Arc"
 
+    let makeBoxTy com ctx (ty: Rust.Ty): Rust.Ty =
+        [ty] |> makeImportType com ctx "Native" "Box"
+
     // TODO: emit Lazy or SyncLazy depending on threading.
     let makeLazyTy com ctx (ty: Rust.Ty): Rust.Ty =
         [ty] |> makeImportType com ctx "Native" "Lazy"
@@ -652,6 +655,7 @@ module TypeInfo =
         | Lrc
         | Rc
         | Arc
+        | Box
 
     let (|HasReferenceTypeAttribute|_|) (ent: Fable.Entity) =
         ent.Attributes |> Seq.tryPick (fun att ->
@@ -662,6 +666,7 @@ module TypeInfo =
                     | 0 -> Some Lrc
                     | 1 -> Some Rc
                     | 2 -> Some Arc
+                    | 3 -> Some Box
                     | _ -> None
                 | _ -> None
             else None)
@@ -830,6 +835,7 @@ module TypeInfo =
         | Some Lrc -> makeLrcTy com ctx ty
         | Some Rc -> makeRcTy com ctx ty
         | Some Arc -> makeArcTy com ctx ty
+        | Some Box -> makeBoxTy com ctx ty
         | _ -> ty
 (*
     let transformReflectionInfo com ctx r (ent: Fable.Entity) generics =
@@ -1308,6 +1314,9 @@ module Util =
     let makeArcValue (value: Rust.Expr) =
         makeCall ["Arc";"from"] None [value]
 
+    let makeBoxValue (value: Rust.Expr) =
+        makeCall ["Box";"from"] None [value]
+
     let maybeWrapSmartPtr ent expr =
         match ent with
         | HasReferenceTypeAttribute a ->
@@ -1315,6 +1324,7 @@ module Util =
             | Lrc -> expr |> makeLrcValue
             | Rc -> expr |> makeRcValue
             | Arc -> expr |> makeArcValue
+            | Box -> expr |> makeBoxValue
         | _ ->
             match ent.FullName with
             | Types.fsharpAsyncGeneric
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -1641,9 +1641,12 @@ module Util =
         | Fable.NewTuple (values, isStruct) -> makeTuple com ctx r isStruct values
         | Fable.NewList (headAndTail, typ) -> makeList com ctx r typ headAndTail
         | Fable.NewOption (value, typ, isStruct) -> makeOption com ctx r typ value isStruct
-        | Fable.NewRecord (values, entRef, genArgs, isStruct) -> makeRecord com ctx r isStruct values entRef genArgs
+        | Fable.NewRecord (values, entRef, genArgs) ->
+            let isStruct = (com.GetEntity entRef).IsValueType
+            makeRecord com ctx r isStruct values entRef genArgs
         | Fable.NewAnonymousRecord (values, fieldNames, genArgs, isStruct) -> makeTuple com ctx r isStruct values
-        | Fable.NewUnion (values, tag, entRef, genArgs, isStruct) ->
+        | Fable.NewUnion (values, tag, entRef, genArgs) ->
+            let isStruct = (com.GetEntity entRef).IsValueType
             makeUnion com ctx r isStruct values tag entRef genArgs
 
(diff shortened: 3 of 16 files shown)


src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -3530,9 +3532,14 @@ module Util =
...
                 else
                     let ty =
-                        let genArgs = getEntityGenArgs ent |> transformGenArgs com ctx
+                        let genArgs = getEntityGenArgs ent
+                        let entRef = Fable.DeclaredType(entRef, getEntityGenArgs ent)
+                        let genArgs = genArgs |> transformGenArgs com ctx
                         let bounds = mkTypeTraitGenericBound [entName] genArgs
-                        mkTraitTy [bounds]
+                        let ty = mkTraitTy [bounds]
+                        // TODO : Wrapping the impl block in a Lrc breaks casting (see IEnumerable), because it no longer implements the interface for T but for Lrc<T>
```
alexswan10k, line 3540: @ncave I think I have hit a brick wall on this, and not sure where to go!

In short, this is the change that makes ```self``` go from a ```T``` to a ```Lrc<T>```. The problem here is that the whole way we do casting depends on the underlying thing implementing the trait for ```T```, so changing it to ```Lrc<T>``` means it is no longer visible to the compiler!

```
error[E0277]: the trait bound `Enumerator<T>: IEnumerator_1<T>` is not satisfied
   --> src\.\Seq.rs:200:13
    |
200 |             Seq_::Enumerable::Enumerator::_ctor__5FE5556A(next, Lrc::from(dispose).clone()).clone()
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IEnumerator_1<T>` is not implemented for `Enumerator<T>`
    |
    = help: the trait `IEnumerator_1<T>` is implemented for `Rc<Enumerator<T>>`
    = note: required for the cast to the object type `dyn IEnumerator_1<T>`

error[E0277]: the trait bound `Enumerator<T>: IEnumerator_1<T>` is not satisfied
   --> src\.\Seq.rs:207:13
    |
207 |             Seq_::Enumerable::Enumerator::_ctor__5FE5556A(next, dispose).clone()
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IEnumerator_1<T>` is not implemented for `Enumerator<T>`
    |
    = help: the trait `IEnumerator_1<T>` is implemented for `Rc<Enumerator<T>>`
    = note: required for the cast to the object type `dyn IEnumerator_1<T>`

error[E0277]: the trait bound `Enumerable_1<T>: IEnumerable_1<T>` is not satisfied
   --> src\.\Seq.rs:487:9
    |
487 |         Seq_::Enumerable::Enumerable_1::_ctor__1538198A(f).clone() as Lrc<dyn IEnumerable_1<T>>
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IEnumerable_1<T>` is not implemented for `Enumerable_1<T>`
    |
    = help: the trait `IEnumerable_1<T>` is implemented for `Rc<Enumerable_1<T>>`
    = note: required for the cast to the object type `dyn IEnumerable_1<T>`
```

This feels like a lose-lose situation because, on the one hand, I cannot see how you can cast into a wrapped implementation off the top of my head (needed for IEnumerable), but on the other hand returning self on a fluent api will actually create new copies at the moment, which is subtly wrong. If I built a fluent api with internal mutable state for example (the evil OO way), the results would be different between Rust and .Net because Rust makes new copies when returning this. What do we do? :)

alexswan10k: Some possible ideas:
* Make the implementors of IEnumerable Structs (almost definitely a bad idea)
* Make the option to implement wrapped or unwrapped self as a controllable attribute (also pretty awful)
* Implement twice for unwrapped and wrapped, although this probably means fishing out methods which do not line up. This probably also breaks on edge cases where you return self through the interface (also pretty awful)
* Double wrap anything castable (not great either, but probably the most realistic solution. See below)

image

src/Fable.AST/Fable.fs
```diff
@@ -409,9 +409,9 @@ type ValueKind =
     | NewArray of newKind: NewArrayKind * typ: Type * kind: ArrayKind
     | NewList of headAndTail: (Expr * Expr) option * typ: Type
     | NewTuple of values: Expr list * isStruct: bool
-    | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list
+    | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list * isStruct: bool
```
ncave, line 412: @alexswan10k This is unnecessary, Records and Unions already have entity reference, so you can just check if the entity is `IsValueType` where needed. It needed to be added to anonymous records only because they don't have an entity attached  to them in the F# AST. Please undo this portion of the PR (the change to the AST).

ncave, line 412: @alexswan10k This is unnecessary, Records and Unions already have entity reference, so you can just check if the entity is `IsValueType` where needed. Please undo this portion of the PR (the change to the AST).

ncave, line 412: @alexswan10k This is unnecessary, Records and Unions already have entity reference, so you can just check if the entity is `IsValueType` where needed. It needed to be added to anonymous records only because they don't have an entity attached  to them in the F# AST. Please undo this portion of the PR (the change to the AST).

alexswan10k: Ah yes, and so do unions. My bad. Let me revisit this.

alexswan10k: Fixed.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1537,7 +1548,7 @@ module Util =
         then expr
         else expr |> makeLrcValue
 
-    let makeRecord (com: IRustCompiler) ctx r values entRef genArgs =
+    let makeRecord (com: IRustCompiler) ctx r isStruct values entRef genArgs =
```
ncave, line 1551: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1552,7 +1563,7 @@ module Util =
         let genArgs = transformGenArgs com ctx genArgs
         let path = makeFullNamePath ent.FullName genArgs
         let expr = mkStructExpr path fields // TODO: range
-        if isCopyableEntity com Set.empty ent
+        if isStruct || isCopyableEntity com Set.empty ent
```
ncave, line 1566: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1566,7 +1577,7 @@ module Util =
             else
                 fullName
 
-    let makeUnion (com: IRustCompiler) ctx r values tag entRef genArgs =
+    let makeUnion (com: IRustCompiler) ctx r isStruct values tag entRef genArgs =
```
ncave, line 1580: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1576,7 +1587,7 @@ module Util =
             if List.isEmpty values
             then callee
             else callFunction com ctx None callee values
-        if isCopyableEntity com Set.empty ent || ent.FullName = Types.result
+        if isStruct || isCopyableEntity com Set.empty ent || ent.FullName = Types.result
```
ncave, line 1590: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1630,9 +1641,13 @@ module Util =
         | Fable.NewTuple (values, isStruct) -> makeTuple com ctx r isStruct values
         | Fable.NewList (headAndTail, typ) -> makeList com ctx r typ headAndTail
         | Fable.NewOption (value, typ, isStruct) -> makeOption com ctx r typ value isStruct
-        | Fable.NewRecord (values, entRef, genArgs) -> makeRecord com ctx r values entRef genArgs
```
ncave, line 1633: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1630,9 +1641,13 @@ module Util =
...
+        | Fable.NewRecord (values, entRef, genArgs) ->
+            let isStruct = (com.GetEntity entRef).IsValueType
+            makeRecord com ctx r isStruct values entRef genArgs
         | Fable.NewAnonymousRecord (values, fieldNames, genArgs, isStruct) -> makeTuple com ctx r isStruct values
-        | Fable.NewUnion (values, tag, entRef, genArgs) -> makeUnion com ctx r values tag entRef genArgs
```
ncave, line 1635: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -1630,9 +1641,13 @@ module Util =
...
+        | Fable.NewUnion (values, tag, entRef, genArgs) ->
+            let isStruct = (com.GetEntity entRef).IsValueType
+            makeUnion com ctx r isStruct values tag entRef genArgs
```
ncave, line 1650: @alexswan10k Please remove change as `isCopyableEntity` already handles value types correctly.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -189,6 +189,9 @@ module TypeInfo =
     let makeArcTy com ctx (ty: Rust.Ty): Rust.Ty =
         [ty] |> makeImportType com ctx "Native" "Arc"
 
+    let makeBoxTy com ctx (ty: Rust.Ty): Rust.Ty =
+        [ty] |> makeImportType com ctx "Native" "Box"
```
ncave, line 193: Please change to:
```fs
    let makeBoxTy com ctx (ty: Rust.Ty): Rust.Ty =
        [ty] |> makeImportType com ctx "Native" (rawIdent "Box")
```

src/Fable.Transforms/Rust/AST/Rust.AST.Symbols.fs
```diff
@@ -167,7 +167,7 @@ module kw =
     let RustPrelude = [
         "Copy"; "Send"; "Sized"; "Sync"; "Unpin";
         "drop"; "Drop"; "Fn"; "FnMut"; "FnOnce";
-        "Box"; "ToOwned"; "Clone";
+        "ToOwned"; "Clone";
```
ncave, line 170: @alexswan10k Please undo this change, we can't change the fact what's in the Rust prelude.
If you need to avoid it when creating a symbol, you can use `rawIdent` (see below).

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -421,11 +424,10 @@ module TypeInfo =
...
             // do not make custom types Rc-wrapped by default. This prevents inconsistency between type and implementation emit
             | HasReferenceTypeAttribute ptrType ->
-                if not (isCopyableType com Set.empty typ) then
-                    Some ptrType
-                else None
-            | _ ->
-                if not (isCopyableType com Set.empty typ) then Some Lrc else None
+                Some ptrType
+            | ent ->
+                if ent.IsValueType then None
```
ncave, line 429: @alexswan10k If `isCopyableType` already returns true for value types, is this needed?

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: It works because the mutated thing was nested in a Lrc.

I have written a new test which demonstrates this problem more accurately. The line commented out is where it breaks.

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: I don't think this is accurate. If you remove the value type check, structs are always rendered as Lrc. I accept the fix might be in the wrong place and need to be in isCopyableEntity, but I am not sure this logically follows. 

What has wrapped status got to do with if something is copyable or not? The logic is - value types should be unwrapped, surely?

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: Agree for unions. Not sure how I ended up here. Unions appear to work! Fixed.

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: fixed

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -3530,9 +3545,14 @@ module Util =
                     []
                 else
                     let ty =
-                        let genArgs = getEntityGenArgs ent |> transformGenArgs com ctx
+                        let genArgs = getEntityGenArgs ent
+                        let entRef = Fable.DeclaredType(entRef, getEntityGenArgs ent)
+                        let genArgs = genArgs |> transformGenArgs com ctx
```
ncave, line 3549: This change from one to three lines of code seems unnecessary, as the end result looks the same. Is it needed?

alexswan10k: See line 3553, which requires entRef to be pulled out.

I cannot activate this line because it breaks IEnumerable casting, but I believe this must be turned on as soon as possible to enable self references to be of the Lrc wrapper

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: I have just tried this, but it does not work

```
error[E0433]: failed to resolve: use of undeclared type `Box_`
   --> tests\src\.\RecordTests.rs:387:17
    |
387 |                 Box_::from(Fable::Tests::RecordTests::BoxRecord { a: a.a + 1i32 })
    |                 ^^^^ use of undeclared type `Box_`

error[E0433]: failed to resolve: use of undeclared type `Box_`
   --> tests\src\.\RecordTests.rs:391:70
    |
391 |                 let actual: i32 = Fable::Tests::RecordTests::add1Box(Box_::from(
    |                                                                      ^^^^ use of undeclared type `Box_`
```

The issue seems to be that the filtering is happening further down within makeImportType

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: @alexswan10k 
> structs are always rendered as Lrc

We already have a lot of tests for struct classes, struct records, struct unions, etc., and they don't render as Lrc.
Do you have an example?

alexswan10k: Sure. Check out StructRecord in RecordTests and associated tests. It is wrapped in a Lrc before this change.

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: @alexswan10k Then can you try this change?

```fs
    let getLibraryImportName (com: IRustCompiler) ctx moduleName typeName =
        let selector = moduleName + "_::" + (rawIdent typeName)
```

alexswan10k: Unfortunately this seems to create bigger problems downstream

```
error[E0252]: the name `arrayEmpty` is defined multiple times
  --> src\.\Array.rs:17:9
   |
16 |     use crate::module_5c9aa1da::Native_::arrayEmpty;
   |         ------------------------------------------- previous import of the value `arrayEmpty` here
17 |     use crate::module_5c9aa1da::Native_::arrayEmpty;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `arrayEmpty` reimported here
   |
   = note: `arrayEmpty` must be defined only once in the value namespace of this module

error[E0252]: the name `arrayEmpty` is defined multiple times
  --> src\.\List.rs:14:9
   |
13 |     use crate::module_5c9aa1da::Native_::arrayEmpty;
   |         ------------------------------------------- previous import of the value `arrayEmpty` here
14 |     use crate::module_5c9aa1da::Native_::arrayEmpty;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `arrayEmpty` reimported here
   |
   = note: `arrayEmpty` must be defined only once in the value namespace of this module
```

I don't really understand exactly how Rc and Arc are different. Are they not in the prelude?

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: @alexswan10k I see, can you then rewrite it to:
```fs
let entTyp = Fable.DeclaredType(entRef, genArgs)
...
//maybeWrapInPtrTy com ctx entTyp ty
```

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: @alexswan10k The REPL hasn't been updated yet with latest, but you're right. What isCopyableType/isCopyableEntity does is to check if we can derive from "Copy". In this case we cannot (since one of the fields is not Copy).

So your change is correct and I was wrong.

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: This is fine.

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: No worries :)

I believe it just tells us if something implements the Clone trait or not. It is a pretty close approximation most of the time I agree, but both a ```Lrc<RawStruct>``` and a ```RawStruct``` are both cloneable so I am not sure this actually tells us the whole story.

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: This is fine, I can remove it later.

src/Fable.Transforms/Rust/Fable2Rust.fs
ncave: This is fine, I can remove it later.

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: Oh I can move this logic inside. Let me fix..

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: will fix, give me a bit

alexswan10k:
Ah no you beat me to it, I had some more fixes to your comments, along with a test to prove unwrappedness (found  a way to do with emit). gimme a sec and I will open another PR if that is ok?

ncave:
@alexswan10k No, lemme check in mine first ;)

ncave:
@alexswan10k You can rebase on #2983 if you don't want to wait 5 min until the tests go through ;)

src/Fable.Transforms/Rust/Fable2Rust.fs
alexswan10k: @ncave just been experimenting with this... I think I might have a pretty dreadful but workable solution that means we can make this work without changing the implementation of a trait to be wrapped.

What if we do this (specifically the "wrappedSelf" magically generated parameter):
```
        pub trait IHasAdd {
            fn Add(&self, wrappedSelf: fable_library_rust::Native_::Lrc<dyn IHasAdd>, x: i32, y: i32) -> i32;
        }
```

And then implementors would do this
```
            impl IHasAdd for Adder {
                fn Add(&self, casted: Lrc<dyn IHasAdd>, x: i32, y: i32) -> i32 {
                    x + y + self.m
                }
            }
```

And consumers would have to do something horrible like this
```
            pub fn Class_interface_impl_works_trivial() {
                let a: Lrc<Fable::Tests::InterfaceTests::Adder> =
                    Fable::Tests::InterfaceTests::Adder::_ctor__Z524259A4(1i32);
                let casted = (a.clone() as Lrc<dyn IHasAdd>);
                let actual: i32 = casted.Add(casted.clone(), 2i32, 1i32);
                assert_eq!(4i32, actual)
            }
```

Additionally, it crossed my mind that we only need to do this hack when a user writes some code that returns "this", or consumes "this" directly as part of a returned object. It is kind of an edge case (although pretty common OO pattern).

We might be able to get this working too for abstract classes, but it seems pretty apparent now that this will not allow you to cast between concrete instances in a class hierarchy under any circumstances. Is this a dealbreaker do you think - do we need to go back to the drawing board on this?

ncave: @alexswan10k I don't know, another option is to change the representation of the mutable fields to be Lrc-wrapped, so copies of the object can see the other copy's mutations. I know, a functional nightmare, it's bad for so many reason :)

How is self returned in Rust usually, isn't there the same issue?

alexswan10k: Yeah Lrc wrapped mutable fields may well work, although I am somewhat worried we could get into a pickle for the odd cases when you really need referential comparison, never mind the performance overhead.

Well in normal Rust, the user has total control so this is not really an issue. If they want to write a fluent api, i imagine they just have to deal with the borrow checker, or implement it using ```impl for Rc<T>```. The problem is we have to accommodate for that AND dyn casting as we have to abstractly solve for all cases. A Rust user would probably just be forced to choose either or and deal with the consequences elsewhere.

ncave: @alexswan10k Our interface story is not great anyway. I'm almost tempted to go back and replace Lrc<dyn traits> with enums. But that has the giant issue that the client can't implement new types for that interface, so interop between code bases is limited. Sill, what we have right now *almost* works. How about returning that self by reference, if that's even possible?

alexswan10k: I am not sure what you mean. So self can be:
* by value (consume, so borrow checker will break on multiple uses as move)
* by reference (use as many times as you want, but ```&T``` will be self and not ```Lrc<T>```
* Change impl to use ```Lrc<T>```, which makes self a ```Lrc<T>```. Really this is exactly what we want, but it breaks casting!

ncave: @alexswan10k I've explored a little bit returning by ref for fluent returns in #2986, not sure yet if we can make it fully work. It's [turned off](https://github.com/fable-compiler/Fable/blob/snake_island/src/Fable.Transforms/Rust/Fable2Rust.fs#L2899) for now and unfinished, just so it doesn't break anything, just wanted to share and get your thoughts before we go too far with it.

Returning `self` works just fine in Rust, and we can fix the return type on the member function impl, and perhaps even in the interface, but we need to somehow make the type `ref` instead of `Lrc` in let bindings too. Here is what it ideally should look like if it all works (and maybe I'm just not thinking through all the reasons why it wouldn't):

```rust
#![allow(dead_code)]
#![allow(non_snake_case)]
#![allow(non_camel_case_types)]
#![allow(non_upper_case_globals)]
#![allow(unused_parens)]
#![allow(unused_imports)]
#![allow(unused_variables)]
#![allow(unused_attributes)]
use fable_library_rust::Native_::Lrc;
use fable_library_rust::Native_::array;
use fable_library_rust::String_::string;
pub mod TestApp {
    use super::*;
    use fable_library_rust::Native_::MutCell;
    #[derive(Clone, Debug, Default, Hash)]
    pub struct FluentA {
        a: MutCell<i32>,
    }
    impl TestApp::FluentA {
        pub fn _ctor__Z524259A4(x: i32) -> Lrc<TestApp::FluentA> {
            let a: i32;
            ();
            a = x;
            ();
            Lrc::from(TestApp::FluentA{a: MutCell::from(a),})
        }
    }
    pub trait FluentAMethods {
        fn get_X(&self)
        -> i32;
        fn Add1(&self)
        -> &TestApp::FluentA;  // <--- ref return type, works (unsound, only checks if return type is the same as the enclosing entity)
    }
    impl TestApp::FluentAMethods for FluentA {
        fn get_X(&self) -> i32 { self.a.get() }
        fn Add1(&self) -> &TestApp::FluentA {  // <--- ref return type, works (sound, inspects the body if it returns `self`)
            self.a.set(self.a.get() + 1i32);
            self
        }
    }
    pub fn main(_args: Lrc<MutCell<Vec<Lrc<str>>>>) -> i32 {
        let a: Lrc<TestApp::FluentA> =
            TestApp::FluentA::_ctor__Z524259A4(1i32);
        let b: &TestApp::FluentA = a.Add1().Add1();   // <--- ref local type, not done yet. TODO: not sure how to do that yet
        println!("a = {0}", a.get_X());
        println!("b = {0}", b.get_X());
        0i32
    }
}
pub fn main() {
    let args: Vec<String> = std::env::args().collect();
    let args: Vec<Lrc<str>> = args[1..].iter().map(|s| string(s)).collect();
    TestApp::main(array(args));
}

```

alexswan10k: This is a pretty creative solution @ncave , I do like it.

Although this solves the consuming fluent api problem as defined in the test, it unfortunately will fall down when trying to do other things. The problem is there is now asymmetry between ```&T``` and ```Lrc<T>``` which is fine, till the returned object at the end of the chain is thrown over a boundary where it needs to be cloned. I don't see how you can do this without actually having an Lrc. 

The problem is that with this pattern you really do need the reference you start with to be the same as the reference you end with, as a user may choose to use either to pass over to a framework or similar. 

_For the record I hate this pattern and have spent way too much time trying to teach my peers to just create new instances (functional style) rather than mutating this, as the whole thing is far more predictable. The reason people do it this way is that it is just easier, at the cost of many bugs of course. The problem is even asp.net core does this in its setup api._

alexswan10k: Right now I see two ways forward that work for both chaining and casting to abstract interfaces
* The double self compiler generated parameter hack
* Putting anything declared as an interface behind a double pointer ```Lrc<dyn Lrc<ISomeInterface>>```

ncave: @alexswan10k I agree fully. I didn't like the double parameter option at all, so I'm glad you went with double wrapping for interfaces instead.

alexswan10k: Well the biggest issue with it was that there was no obvious way to decide if it should or shouldn't be there, so you would probably end up having it there all the time, which would be a mess.

Double indirection is annoying because of the performance overhead, but at least the experience is kind of predictable, also for the generated traits etc. There is some discussion about having sized dyn (so you don't need a pointer), but I imagine this is way off. One to keep an eye on - I imagine this can be revisited.

alexswan10k: Oh, I had a look at constructors while I was in that space, in an attempt to clean up the unnecessary cloning, and got into a bit of a pickle. I think the logic that decides cloning is just fundamentally limited and probably needs a total rewrite at some point.

I was wondering - would it be possible to output some kind of lazily evaluated thing into the output AST, so that I can make my mind up if to clone or not based on some future thing given to the compiler further down the AST? The problem is you do not know till you have visited all the idents if the first usage needs to clone or not. Alternatively I guess you could do an extra pass over the AST, although the context would have to be completely recreated, which may not be straight forward.

No rush with this one anyway as what we have works, it is just noisy.

ncave: @alexswan10k If you want to do this just inside constructors, we already modify the constructors AST body [here](https://github.com/fable-compiler/Fable/blob/snake_island/src/Fable.Transforms/Rust/Fable2Rust.fs#L3420-L3428), so it's just a matter of adding another case to modify whatever you need as you want it.

ncave: @alexswan10k But I'm not actually sure why you want to remove the cloning in the constructors. Since `self` is Lrc ptr, closing is really cheap. I think it's fine the way it is right now, I would just leave it.

alexswan10k: Sure. Have a look at generated seq.rs ctor's for Enumerable line 164 for example. You can see it gets confused and tries to clone next and dispose even though you can easily see these are only used once. This is because we currently just treat any reference to an ident as a usage, rather than explicitly work out if the usage is leaving the context (a move). A set is being counted as a usage here. I did look at extending the reference counting to cover this, but it created other problems when the first usage was a move (but count was 1 because all subsequent usages were set).

I think the bottom line is the whole approach is just limited. You really need to be able to have evaluated all usages first to be able to understand the usage you are looking at, and thus know if that should be cloning or not.

I agree though this is not a big problem, it's just annoying me :) 

What is probably more of an issue is the generated constructors have cryptic compiler-generated names such as _ctor__5FE5556A. They should probably stick to the "new" convention, maybe "new_1", "new_2". Oh well one step at a time.

ncave: @alexswan10k Actually that's a feature, not a bug. Not just for constructors, any member name (except names from interfaces) is mangled to contain a hash of the parameter types, so you can have member overloading with different parameters. It can be forced on with interface methods too with the `Mangle` attribute to allow overloading.

alexswan10k: Ah yes, the typescript output does this too, it completely makes sense for collisions.

The main argument I have against this is that it will just not be a pretty interop experience (in either language). Would it not be possible to generate prettier names such as how ident collisions are avoided?

ncave: @alexswan10k If you have any ideas, feel free to open an issue and discuss. But yes, that's why there is no mangling by default on interfaces, usually bindings and interop is done with interfaces.

alexswan10k: Makes sense. Will do. Cheers.

alexswan10k: Been mulling this over.. Considering you can just expose a ctor via a public static function (which is pretty much what you do in rust anyway), I am not sure this is even necessary. I will save ticket spam for now.

PR #2989, opened 2022-07-26
Rust - represent self as a Lrc<T> for method calls using double pointer indirection

alexswan10k:
Following on from https://github.com/fable-compiler/Fable/pull/2981#discussion_r928690837 this implements class method traits as Lrc<T>, thus allowing self to be of type ```&Lrc<T>``` and ```not &T```. This is done by double pointer indirection ```Lrc<dyn Lrc<YourStruct>>```, which is not ideal, but currently the only seemingly practical way to achieve this dynamic dispatch approach in Rust. This is important because it allows self/this to be correctly cloned, preserving underlying referential integrity. This is required for fluent/builder patterns when internal mutability is present, and it can not be reasoned which reference a user will potentially pass over to another area of the program after configuration.

image

Changes
* Add concept of dynamic - Currently this is just anything expressed through its interface, but in theory you could use this to cast between instances of concrete classes too, but it is kind of an all or nothing approach. For flat hierarchies perhaps this could just be optimized out.
* Split out static methods so they no longer get rendered into the Methods trait. They now get rendered straight into the root type, so they can correctly be referenced by static name.
* Something weird with dictionaries, they are not a single struct that implements a methods trait, so casting causes problems. Will likely need an explicit workaround? For now I have deduced that it is safe to do absolutely nothing, so this should hopefully suffice for now.

The trait gen code is copy pasted, so there are some obvious refactor opportunities here! Apologies, I will come back to this.

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -3599,7 +3599,6 @@ module Util =
                         let genArgTys = genArgs |> transformGenArgs com ctx
                         let bounds = mkTypeTraitGenericBound [entName] genArgTys
                         let ty = mkTraitTy [bounds]
-                        let ety = com.GetEntity(tEntRef)
                         ty
                     let genArgs = getEntityGenArgs tEnt
                     let nameParts =
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -428,6 +428,13 @@ module TypeInfo =
             | ent ->
                 if ent.IsValueType then None else Some Lrc
 
+    let shouldBeDoubleWrapped (com: IRustCompiler) typ=
+        match typ with
+        | Fable.DeclaredType(entRef, _) ->
+            let ent = com.GetEntity(entRef)
+            ent.IsInterface && not ent.IsValueType
+        | _ -> false
+
     let isCloneableType (com: IRustCompiler) typ =
         match typ with
         | Fable.String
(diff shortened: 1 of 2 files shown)


src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -3564,23 +3580,48 @@ module Util =
...
+                let implItemStatic =
                     let ty =
                         let genArgs = getEntityGenArgs ent
                         let genArgTys = genArgs |> transformGenArgs com ctx
                         let bounds = mkTypeTraitGenericBound [entName] genArgTys
                         let ty = mkTraitTy [bounds]
-                        // TODO: Wrapping the impl block in a Lrc breaks casting (see IEnumerable), because it no longer implements the interface for T but for Lrc<T>
-                        // let entTyp = Fable.DeclaredType(entRef, genArgs)
-                        // maybeWrapInPtrTy com ctx entTyp ty
+                        let ety = com.GetEntity(tEntRef)
```
ncave, line 3602: @alexswan10k This doesn't seem to be used.

alexswan10k: Good spot, fixed

ncave:
Very minimalistic solution, I like it.

alexswan10k:
> Very minimalistic solution, I like it.

Ta. I imagine we are probably not out of the woods yet with all the class/interface stuff, but hopefully this gets us a little closer. I think the full blown inheritance stuff can probably be kicked down the road for now, so long as we have a way forward with it.

ncave:
@alexswan10k Btw, if you're revisiting to refactor, perhaps we can merge the `impl` for constructors with the `impl` for static methods.

PR #2999, opened 2022-07-27
Rust - Add interfaces to generic constraints

alexswan10k:
It was not previously possible to constrain a generic parameter by an interface, this fixes this.

Additionally did a little bit of refactoring around class trait generation, so less code is repeated.


alexswan10k:
Tests run, but just noticed output is really weird, and probably wrong if using interior mutability
```
        member this.AddThroughCaptured x y = adder.Add x y
```
```
impl<a: IHasAdd + Clone + 'static> Fable::Tests::InterfaceTests::IConstrained_1<a>
                for Lrc<AdderWrapper_1<a>>
            {
                fn AddThroughCaptured(&self, x: i32, y: i32) -> i32 {
                    let copyOfStruct: Lrc<MutCell<a>> =
                        Lrc::from(MutCell::from(self.adder.clone()));
                    copyOfStruct.get().Add(x, y)
                }
            }
```

Looks like this is coming from fcs-fable.

ncave:
@alexswan10k 
> Looks like copyOfStruct is coming from fcs-fable. Some kind of optimization?

Yes, The F# compiler sometimes takes defensive copies when copying a struct. In theory you should be able to control that behavior to some extent with the [IsReadOnly](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/structs#readonly-structs) attribute on the struct, but not sure when it's a generic type. I tried adding it to the `adder` but it doesn't seem to do much.

```fs
type AdderWrapper<'a when 'a :> IHasAdd> ([<IsReadOnly>] adder: 'a) = 
```

alfonsogarciacaro:
@alexswan10k Yes, I've seen this in JS too when using structs (which don't have an effect in JS) and it's a bit annoying. I haven't investigated much but I assume it's a way F# uses to keep the semantics of records when using structs.

PR #3002, opened 2022-07-27
Rust - Removing more clones

alexswan10k:
Progress. I went round on a few solutions actually (another far more comprehensive/complicated but ultimately flawed solution). The Expr recursive function might be useful though so I will put it on a branch. I also dreamt up a kind of cool way to dump out debug data into the output as a text comment so more on that later. It is on the other branch too.

I am not proud of this, but mutation seems like the only reasonable way here to get accurate ident usage count tracking, without tearing the whole AST walker apart. It is still not perfect but this has definitely cleared up some noise.

Take Seq.rs for example, I recon there are perhaps 20-30 usages cleaned up:

image
image

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -460,7 +460,10 @@ module TypeInfo =
             | Fable.FieldGet _
             | Fable.ListHead _
             | Fable.ListTail _ -> true
-            | _ -> false
+            | Fable.TupleIndex _
+            | Fable.ExprGet _
+            | Fable.UnionTag _ -> false
+            // TODO: Add isForced flag to distinguish between value accessed in pattern matching or not
         | Fable.IdentExpr _
             -> true
         // | Fable.TypeCast(e, t) -> isCloneableExpr com t e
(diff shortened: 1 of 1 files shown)


src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -449,7 +453,14 @@ module TypeInfo =
...
         match e with
-        | Fable.Get _
+        | Fable.Get(_, kind, _, _) ->
+            match kind with
+            | Fable.OptionValue _
+            | Fable.UnionField _
+            | Fable.FieldGet _
+            | Fable.ListHead _
+            | Fable.ListTail _ -> true
+            | _ -> false
```
ncave, line 463: Can you list the other kinds in the `false` category instead of a `_`, just so it's a complete pattern match?

alexswan10k: Of course. All done.

alexswan10k:
By the way @ncave, thought you might find this hack interesting/useful to know about:

https://github.com/alexswan10k/Fable/blob/2d3c5d50d6a252040a6bf26ae737e028d3bb11a8/src/Fable.Transforms/Rust/AST/Rust.AST.Helpers.fs#L888

Basically when you are debugging an offensive expression, you can just spit out some data into the output as an inline comment, so code can still be executed as normal. This was super useful for me when trying to debug ident usage counts.

ncave:
@alexswan10k By all means check it in so you don't have to carry it forward from branch to branch, seems useful.

PR #3006, opened 2022-07-28
Rust - First pass at getting the interop experience with built in collections a little better

alexswan10k:
So within 5 minutes of trying to use generated Rust in the real world, one quickly gets into a lot of frustration around trying to interop between F# collections and built-in rust collections. This is a first step to try and bridge the gap between these worlds to ultimately provide a better developer experience.

* Implement ExtInterop module in Rust.
* Add from_vec and into_vec for List and Array currently
* Add tests to cover trivial roundtripping scenario


alexswan10k:
##### Caveats to interop that need further discussion and thought

There were immediately some confusing things which threw me off, which I think warrants its own discussion. Specifically, I quickly jumped to the conclusion that I would just be able to use the iter trait on F# collections, to which I was unpleasantly surprised by finding out this was calling onto Option.iter! I was also not able to implement and thus predictably use For and Into because we are using aliases.

I think this points to a deeper problem, in that none of the built-in F# collections are able to implement standard traits currently such as Iter, From, and To because they are **aliases** and not **structs**. I wonder if (I know this is a huge pain) it makes more sense to use real thin wrapper structs rather than aliases for built-in collections. This way those structs can correctly implement traits, giving Rust users a far more predictable experience, and likely allowing them to work out of the box with less conversions. I believe it will be the same size as the underlying struct if it is a thin wrapper, as there is no metadata overhead in Rust as far as I am aware.

Perhaps it could look like this:

```
//old world
pub type List_1<T> = Option<Lrc<crate::List_::Node_1<T>>>;

//new world
struct List_1<T: Clone + 'static>(Option<Lrc<crate::List_::Node_1<T>>>);

//edit - actually perhaps it needs to be more like this, where it then becomes treated like a standard ref type and wrapped in Lrc
struct List_1<T: Clone + 'static>(Option<crate::List_::Node_1<T>>);

impl <T> From<Vec<T>> for List_1<T>{
    fn from(vec: Vec<T>) -> Self {
        todo!()
    }
}

//impl iterator for List etc

```

It looks like this comes from the F# representation, so perhaps this can be fixed by swapping out the alias of Some/None to a real concrete type (or struct if performance would be better)
```
// old world
type List<'T> = Node<'T> option
type 'T list = List<'T>

//new world
// [<Struct>] //?
type List<T> = 
    | Some of Node<'t>
    | None
```

Additionally, by implementing the Deref trait, I imagine we can also reduce the amount of churn, as you do not need to pull the underlying field explicitly.

_Edit_ 
I have just been pondering over the definitions and the following aliases basically wrap a useable Rust type anyway
* Array -> Vec
* HashSet -> HashSet
* HashMap -> HashMap

The following are problematic however (which are commonly used, and also some of my personal favourites)
* List
* Set
* Map

I do think there may well be an argument for wrapping ALL of them, for the simple reason that it gives an external user a predicable api surface, and allows us to internally change our mind on the implementation details without breaking everything for everyone

ncave:
@alexswan10k `List/Map/Set` use `option<node>` to represent an empty node, which makes the implementation more memory-efficient and reduces allocation.Technically it is possible to convert that to a wrapper type at the cost of more allocations, but since this is only for interop, can this wrapper type be in `ExtInterop.rs` instead?

alexswan10k:
I am not sure the allocation story is true. My understanding is these are the same size:

```
struct Thing { ...fields };
struct Wrapped(Thing);
```

You certainly could create a wrapper in the interop layer, but a user would have to know to wrap the built-in type for every usage, which seems highly error prone. Perhaps there is a way to do this that I haven't considered though.

ncave:
Yes, if the wrapper is a struct (in F#) it should be ok.

On a side note, we may eventually have to move to Options for everything that is Box/Lrc if we want to represent nulls.
Options on pointer types as memory efficient (same size as the pointer), so that's a plus.

Perhaps we need a generic wrapper for nullable types?

alexswan10k:
Agree, ```Nullable<T>``` sounds like a good idea.

ncave:
@alexswan10k Can we perhaps start with those collection wrappers in `ExtInterop.rs`, and then decide which of them make sense to move to actual types?

ncave:
@alexswan10k For sequences, I have some [helpers](https://github.com/fable-compiler/Fable/blob/snake_island/src/fable-library-rust/src/Native.rs#L111-L136) to and from `iter`, but they only work for clonable types. I haven't been able to make them work in the general case, Rust keeps complaining.

alexswan10k:
We can probably do that, although the representation may actually have to change as there is currently no way to represent the following in F# at the moment. 
```
struct Wrapped(Thing);
```
Perhaps we can rewrite single case DU's as this.

With respect to sizing, I wrote a quick test that proves the following are the same, and the enum is 4 bytes bigger because it has to store the extra enum pos:

```
mod Test {
                struct Thing { a: i32 }
                struct Wrapped(Thing);
                struct WrappedThingWithNamedField { thing: Thing }
                enum EnumedThing {
                    SomeVal(Thing),
                    Nothing,
                }

                #[test]
                pub fn are_all_same_size() {
                    let aSize = std::mem::size_of::<Thing>();
                    let bSize = std::mem::size_of::<Wrapped>();
                    let cSize = std::mem::size_of::<WrappedThingWithNamedField>();
                    let dSize = std::mem::size_of::<EnumedThing>();
                    assert_eq!(aSize, 4);
                    assert_eq!(aSize, bSize);
                    assert_eq!(bSize, cSize); 
                    assert_eq!(dSize, cSize + 4);//4 bytes for enum value + underlying max size of 4
                }
            }
```

As for seq's and cloning, I think that is expected. Basically everything we pass around with shared references needs to be cloneable. into_iter() is supposed to consume the underlying collection, so that is how you generally get around the cloning problem. There is also iter() which returns references and does not consume.

ncave:
@alexswan10k 
> wrapping is a zero cost abstraction.

Yes, also when wrapping pointers with an option.

My concern was about wrappers that are reference types.
But struct wrappers should be ok.

alexswan10k:
Agree, let me fix the file name and do a quick iteration on some struct wrappers within interop.rs. I can hopefully replace these proprietary implementations with ```From``` and ```To``` trait impls.

alexswan10k:
OK I have pushed an update which now uses thin wrappers for Array and List as we discussed (so they are the same size). I am not so sure about array since it is actually wrapped in a Lrc already. I think this may potentially need to be the other way around.

Anyway, important things of note
* In conversions implement with From (only vec atm)
* Out conversions implemented with To (only vec atm)

I also implemented deref for transparent access to underlying types, again I think we can use this to reduce friction.

Knowing you need to wrap your list alias in a List before using it is not intuitive though, so I do still think longer term these types should be wrapped always. Perhaps we can use  the From trait though to roll this out gradually, where going from self to self eventually becomes a no-op.

tests/Rust/tests/src/ExtInteropTests.rs
```diff
@@ -0,0 +1,32 @@
+pub mod ExtInteropTests {
```
ncave, line 1: Perhaps change this to `InteropTests` too.

ncave: OK, makes sense. But then, why not move those tests there, in InteropTests?

alexswan10k: The interop tests are written in F#, they are intended to test interop from the point of view of an F# developer.

The rust interop tests are written in Rust, because they are testing the rust experience of consuming F# generated types (currently just collections). I am not sure it makes sense to try and force these through with F#, even though it might be possible.

With that being said, I am not sure there is an easy way to merge both into 1 file?

build.fsx
```diff
@@ -515,6 +515,10 @@ let testRust testMode =
     cleanDirs [buildDir </> "tests"]
     cleanDirs [buildDir </> ".fable"]
 
+    // copy rust only tests files (these must be present when running dotnet test as import expr tests for file presence)
+    makeDirRecursive (buildDir </> "tests" </> "src")
+    copyFile (projectDir </> "tests" </> "src" </> "ExtInteropTests.rs") (buildDir </> "tests" </> "src" </> "ExtInteropTests.rs")
```
ncave, line 520: Perhaps copy all `*.rs` tests, so you don't have to change it every time.

alexswan10k: This would be far better. Can I just use a wildcard here?

ncave:
@alexswan10k I can merge if ready, please let me know.

build.fsx
alexswan10k: This doesn't seem to work.
```
    copyFile (projectDir </> "tests" </> "src" </> "ExtInteropTests.rs") (buildDir </> "tests" </> "src" </> "*.rs")
```

alexswan10k:
I am happy to merge if you are. I cannot fix your wildcard thing though.

build.fsx
alexswan10k: fixed

PR #3010, opened 2022-07-29
Rust - More work on collection interop (Set, List, Map)

alexswan10k:
@ncave sorry I really am keeping you busy this week :) Following on from yesterday's discussion on zero cost wrappers, I wanted to get your input and be sure we are on the same page etc. I know we were going to defer the decision to add wrappers as first class citizens but it was a lot easier than I anticipated...

This implements List, Set, and Map as first class struct wrappers (zero cost as we covered yesterday). I have consequentially also adjusted the interop layer so you can now directly ```From``` and ```Into``` these types, which is a far more intuitive interop experience as everything is just available on the raw type. At the moment we obviously only support Vec, but you can see how this pattern can be extended to cover all sorts of useful things.

Additionally I also got carried away and supported iter, following on from your seq_as_iter helper. This is really handy because it means you can just collect into a vec after using built in Rust combinators, or even theoretically collect INTO a F# collection if we also implemented fromiterator..

These From and Into implementations are not supposed to be fast, they are just supposed to define the required interfaces and work. Optimization can come later.

**Following on from this**
* Array I am not so sure of, but I think the same pattern should work in theory.
* Who knows what we do with seq
* HashSet and Hashmap I guess need this treatment too. Meh, no rush :)
* Names are a little weird because they are ```List_```. Can we make them without the underscore - ```List```?
* Since our tuple Rc by default change, it is possible that some library types may want to use struct tuples under the hood for performance reasons, such as Map uses these in a few places. I didn't want to go changing too much at once so have left this for now.

Another interesting possibility for way down the road is that you could perhaps even "auto implement" from and into traits for any Fable generated types. The great thing is that tree shaking (when turned on) will throw out anything you are not using anyway, so the bloat is never paid at runtime in production (assuming everything is correctly configured).

### DIFF
--- a/src/fable-library-rust/src/Interop.rs
+++ b/src/fable-library-rust/src/Interop.rs
@@ -5,7 +5,7 @@ pub mod ListExt {
 
     use super::super::List_;
     use crate::{
-        List_::{cons, iterate, singleton, mkList},
+        List_::{cons, iterate, mkList, singleton},
         Native_::{List_1, Lrc, MutCell},
     };
 
@@ -47,7 +47,7 @@ pub mod ListExt {
     }
 }
 
-pub mod Array {
+pub mod ArrayExt {
     use std::ops::Deref;
 
     use crate::{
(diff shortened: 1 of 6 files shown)


ncave:
@alexswan10k 
- Yes, adding struct wrappers to every type where needed should be trivial (including Array, HashMap, HashSet).
- Seq is an interface, so perhaps for now leave it with the helpers from and to iter.
- Names can be changed to be without the _1, _2 etc.
- Map as sequence already exposes `KeyValuePair<K, V> seq` which are struct tuples.

ncave:
@alexswan10k About

```fs
[<Fable.Core.Emit("$0.root")>] // otherwise Fable assumes conservatively the root must be cloned as owned.
let inline private getRoot s = s.root
```
I'm thinking of removing the emit, if you're ok with that, cause:
1) those clones are really cheap,
2) imo at some places where it's happening it's arguably a deficiency in our cloning evaluation.

alexswan10k:
Happy either way. My reasoning was that fixing the algorithm here is non trivial because it is an input parameter, so you cannot know how it is used elsewhere without checking the whole program.

alexswan10k:
Random thought... does inlining do anything useful here? If it was inlined you could easily evaluate it was the only usage, although currently we do not eval field usages and assume they are shared

ncave:
@alexswan10k
> currently we do not eval field usages and assume they are shared

I guess that's what I meant, perhaps there is a room for improvement there in the future.
Anyway, see #3013 for slight renaming and cleanup.

I decided against making the `Map.toSeq` return struct tuples, as that would be a deviation from the [documented](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-mapmodule.html#toSeq) API.

PR #3019, opened 2022-07-30
Rust - more collection interop improvements

alexswan10k:
Couple of little improvements in an attempt to better standardize the Rust interface.

* Added implementations of ```FromIterator```, so a user can directly collect an iterator into an F# collection.
* Renamed iter to from_iter so that it more accurately reflects Rust conventions. A rust user expects ```iter``` to emit references, and ```into_iter``` to emit values (and consume the underlying object). Currently we can only emit values, so ```into_iter``` is implemented, and ```iter``` is a commented out placeholder with the correct interface. This is still not entirely accurate because into_iter will also consume the underlying collection, which is not feasible when you are working with Rc shared types. I personally feel this combination is the closest match to Rust expectations but happy to discuss/iterate where required.

On a side note I noticed equality does not work for Map and Set, which was a little unexpected since this works in .Net. I am worried this will break equality on model hierarchies where Sets/Maps/Lists are used as properties. I think this is to do with the SetTree and MapTree type being marked as such. Is it possible to do anything about this? You can see this in the tests, which have had to convert back to vec to check the result.

### DIFF
--- a/src/fable-library-rust/src/Interop.rs
+++ b/src/fable-library-rust/src/Interop.rs
@@ -1,11 +1,17 @@
 pub mod ListExt {
     // use std::ops::Deref;
-    use crate::List_::{List, cons, mkList};
-    use crate::Native_::{seq_as_iter};
-    use crate::Seq_::{ofList};
+    use crate::List_::{cons, mkList, reverse, List};
+    use crate::Native_::seq_as_iter;
+    use crate::Seq_::ofList;
 
     impl<T: Clone> List<T> {
-        pub fn iter(&self) -> impl Iterator<Item = T> {
+        //todo - non-consuming iter by ref
+        // pub fn iter<'a>(&self) -> impl Iterator<Item = & 'a T> {
+        //     let s = ofList(self.clone());
+        //     seq_as_iter(&s)
+        // }
+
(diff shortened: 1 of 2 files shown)


ncave:
@alexswan10k 
> Currently we can only emit values, so `into_iter` is implemented, and `iter` is a commented

Can't we emit reference, i.e. ptr.as_ref()?

> equality does not work for Map and Set

It should work, I'll take a look.

alexswan10k:
ptr.as_ref sounds good in principle but I am not sure you are constrained to have a ```Rc<T>``` at the point you dereference (its just a ```T```). Happy to add in though if you have any idea how to do this? I figured it was not super important now since the value version ```into_iter``` works fine with cloning anyway. I just think it is important to have them the right way round though, so a Rust user gets what they expect.

ncave:
@alexswan10k 
> I noticed equality does not work for Map and Set
Equality gets translated to `Map.equals` and `Set.equals` which basically convert map/set to sequences and compare them.

alexswan10k:
That makes sense. I guess it would be nice though if we could also support built in equality? To be fair there is no reason why we could not just manually implement PartialEq (or whatever is needed) in the interop layer and just call onto ```Map.Equals``` or ```Set.equals``` as the generated code does

PR #3027, opened 2022-08-01
Enable type names for erased types

ncave:
Allows for proper propagation of `Emit`-ted type names (even if they're also `Erase`-d) when used as generic arguments in other types.


alfonsogarciacaro:
Interesting. Actually if we want to improve TypeScript compilation maybe we need to have actual erased types (or at least erased unions) in Fable AST 🤔

ncave:
@alfonsogarciacaro 
The erased types are already in the AST, they just get converted to `Fable.Any`.
First I tried just commenting [this line](https://github.com/fable-compiler/Fable/blob/snake_island/src/Fable.Transforms/FSharp2Fable.Util.fs#L1151), but that breaks the `!!` operator.

PR #3029, opened 2022-08-01
[Rust] Minor cloning update

ncave:


ncave:
@alexswan10k

I wish we could simplify [transformLeaveContext](https://github.com/fable-compiler/Fable/blob/snake_island/src/Fable.Transforms/Rust/Fable2Rust.fs#L1623-L1642), feels like it has extra code for the exceptions to the rule and it just patches to make tests pass, instead of operating on clear rules that can be reused elsewhere as well.

alexswan10k:
I am all ears if you have ideas on how to improve this, but we both knew this was going to be a hard problem, as the only real way to be able to correctly solve this comprehensively is to track ident ownership as the borrow checker in Rust does. It is actually even harder because we can't just tell a user it is wrong like the borrow checker does, we have to implicitly figure out how to make it correct.

In a nutshell, it seems to me the rules could perhaps be boiled down further to the following concepts
* If the target expects a ref, then we are borrowing so no clone
* Beyond this, is it cloneable? If not you are screwed anyway so maybe we do not even need to check this. If you are cloning a struct (copying) or cloning an Rc, it is kind of irrelevant. However, we do not want to clone things that are ```Copy``` though (i32 etc) - in practice, it is probably a no-op, but it creates noise we definitely should not add if possible.
* Tracking - If either the only usage or the last usage, do not clone as move (understanding borrows vs moves is important here) I actually think this part is pretty much working now we track usages.

To the last point, when I did the last tidy up on usage tracking, I actually pushed it a bit further to try and model tracking every usage type. If you had this (and could line them up - very hard), you could know more, because you would have context of other usages in scope. The counter has a similar effect in practice, but it does not let you know if previous usages were borrows or move/clones. Maybe this doesn't matter though.

https://github.com/fable-compiler/Fable/pull/3030

Probably not good in its current form, but perhaps a seed for an idea. I am not convinced you actually need this granularity though, as usage count tracking as it is today should probably suffice.

ncave:
@alexswan10k

Yeah, I don't know if we want to spend much effort on this, since we still have other other language features to work on.
I was just expressing my frustration that I spent quite some time trying to improve it without anything to show for it.

Anyway, I guess your more comprehensive approach with a truth table might be warranted, once we get to it.

Issue #3032, opened 2022-08-02
Rust - Creating a record from another module correctly imports, but incorrectly adds module prefix


alexswan10k:
Shapes.fs
```
namespace Shapes

type Square = {
    x: int
    y: int
}
```

lib.fs
```
open Shapes
let shape () = { x = 1; y = 2 }
```

##### Expected Rust
```
    use crate::module_42b5373b::Shapes::Square;
    pub fn shape() -> Lrc<Square> {
        Lrc::from(Square{x: 1i32, y: 2i32,})
    }
```

##### Actual Rust
```
    use crate::module_42b5373b::Shapes::Square;
    pub fn shape() -> Lrc<Square> {
        Lrc::from(Shapes::Square{x: 1i32, y: 2i32,})
    }
```

##### Tested with
4.0.0-snake-island-alpha-019

##### Workaround
Make a create function and use that from another fs file/module
```
module Square = 
    let create (x, y) = { x = x; y = y}
```

ncave:
@alexswan10k Thanks, I'll take a look.

alexswan10k:
No need, I think I have just fixed it, give me 2 secs to push :)

alexswan10k:
Fixed

PR #3034, opened 2022-08-02
Rust - Double prefix fix for NewRecord #3032

alexswan10k:
Failing scenario + fix for #3032

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -1495,7 +1495,8 @@ module Util =
                     else expr
                 mkExprField attrs fi.Name expr false false)
         let genArgs = transformGenArgs com ctx genArgs
-        let path = makeFullNamePath ent.FullName genArgs
+        let entNs, entName = splitNameSpace ent.FullName
+        let path = makeFullNamePath entName genArgs
         let expr = mkStructExpr path fields // TODO: range
         if ent.IsValueType
         then expr
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -1495,7 +1495,8 @@ module Util =
                     else expr
                 mkExprField attrs fi.Name expr false false)
         let genArgs = transformGenArgs com ctx genArgs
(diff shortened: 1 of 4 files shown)


alexswan10k:
Actually this will work for a single usage I am not sure this is entirely correct. What we really want to be able to do is get the scoped alias that has been created (which may not be the same as the name).

I think there are a few places this pattern prevails so may well be worth checking all of them

alexswan10k:
@ncave I may need some help. I have deliberately broken this again by adding a collision (local record + aliased record). We have this going on in a few places where we do this sort of thing, and assume the imported name is just fine.
```
let entNs, entName = splitNameSpace ent.FullName
```

What we need to do in all these cases is get the generated locally aliased name from the import. How do we do that?

The import correctly creates the alias by the way, so that bit works. It is just the transform of the actual NewRecord.

ncave:
@alexswan10k See `getEntityFullName` it should give you a proper name when the entity is in another file.

ncave:
@alexswan10k There may be other paces too, where we use `ent.FullName` directly, instead of passing it through `getEntityFullName`.

alexswan10k:
I have just gone through them and I believe some of the usages are actually correct, as they are transforming the Class or Interface etc, so you really do want the raw value as there is no import yet.

I found 2 usages that were definitely wrong though and this seems to have fixed the problem.

Nice one. Just gotta wait for CI.

src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -3005,7 +3007,7 @@ module Util =
                 | Fable.DeclaredType(entRef, _) ->
                     let ent = com.GetEntity(entRef)
                     if ent.IsInterface then
-                        let entNs, entName = splitNameSpace ent.FullName
+                        let entName = getEntityFullName com ctx entRef
```
ncave, line 3010: @alexswan10k Do we need to split the full name here? Not saying we do, just asking.

alexswan10k: I am not sure. Doesn't this just give you the imported alias name (or type name if no alias present)? We do have some test coverage for this area and everything passes. There could well be holes though.

Issue #2916, opened 2022-06-01
[Rust] - Managing memory without a garbage collector


alexswan10k:
Probably the biggest impedance mismatch between F# and Rust is the managed/unmanaged memory divide. Rust however has smart pointers, which can be used to achieve pretty near the same result. The philosophy in Rust however is "pick your guarantees", which is somewhat problematic when we really would like a one size fits all approach. 

The point of this ticket is to track progress and finalize how these representations translate to Rust, and how we can ensure sensible defaults where possible, but allow enough control where it makes sense.

So far we have the following big open questions:

##### Currently implemented by:
* All functions currently pass by ref and return owned (value)
  * Most efficient - least amount of cloning, benchmarked [here](https://github.com/ncave/Fable/pull/16#issuecomment-933419602)
  * Leads to most Rust-like api's where generally pass by reference is preferred.
* All constructors and Emit calls pass by value and can take ownership. An emit template can additionally pull references to pass down into a native function, thus allowing it to do both pass by value or by reference.
##### Problems:
* Passing by reference is not consistent with dotnet default (passing by value)
* There is currently no way to express byValue. This leads to weird things like double dereferencing pointers ```&Rc<T>```, or primitives being passed by reference ```&i32```.
  * It is not possible to pass some by value and some by reference due to the fact generics require these to be consistent. This was covered [here](https://github.com/ncave/Fable/pull/16#issuecomment-932938696)
##### Ideas:  
* Could use [inref ](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/byrefs#inref-semantics)to represent a reference.
    * Most similar to dotnet behaviour
    * Typesafe on the F# side - You cannot pass a ref to a function expecting a value, etc
    * Must bind value out before passing it down by reference making pipelining awkward.
    * Module scope values must be mutable
    * Major refactor/breaking change as the default would have to be flipped back to byValue
      * This means passing by value again everywhere which means lots of cloning, which is slow.
 * Could use a function and/or parameter level attribute
   * Function level attribute could flip a whole function to byRef with little friction, allowing opt-in speed benefits
   * Non standard & inconsistent with dotnet
   * Not typesafe on F# side
   * May cause problems when currying, applying functions to functions, etc
##### Direction:
* [x] Switch back to pass by value as default
* [x] ByRef attribute to switch a function over to passing by ref in its entirity
* [x] ByRef attribute on parameters to allow finer grained implicit control
[...]

PR #3040, opened 2022-08-03
Rust - implement unsafe-cells feature switch + Monitor

alexswan10k:
Folowing on from https://github.com/fable-compiler/Fable/issues/2916 and https://github.com/fable-compiler/Fable/issues/2916#issuecomment-1196939666 this moves threading forward a bit:

* unsafe-cells feature switch - turns on Send and Sync for MutCell
* Add test for mutating declared on another thread
* Implement Monitor
* Added some tests for above
* Added placeholders for the next steps - full locks + Thread etc

### DIFF
--- a/tests/Rust/tests/src/AsyncTests.fs
+++ b/tests/Rust/tests/src/AsyncTests.fs
@@ -93,30 +93,53 @@ let shouldExecuteMutationOnTask () =
     do comp.Result
     x |> equal 1
 
+// open System.Threading
 // [<Fact>]
 // let ``should execute mutation on thread unsafe`` () =
 //     let mutable x = 1
-//     let t = new System.Threading.Thread(fun () -> x <- x + 1)
+//     let t = Thread(fun () -> x <- x + 1)
 //     t.Start()
 //     t.Join()
 //     x |> equal 2
 
-// module Monitor =
-//     [<Fact>]
-//     let monitorShouldWorkWithSystemObj () =
-//         let o = new System.Object()
-//         System.Threading.Monitor.Enter(o)
-//         System.Threading.Monitor.Exit(o)
+module Monitor =
+    // [<Fact>]
+    // let monitorShouldWorkWithSystemObj () =
+    //     let o = new System.Object() // todo - this doesn't work, and outputs unit
+    //     System.Threading.Monitor.Enter(o)
+    //     System.Threading.Monitor.Exit(o)
 
-//     type Data = {
-//         x: string //deliberately use reference type to confirm nested Lrc
-//     }
+    type Data = {
+        x: string //deliberately use reference type to confirm nested Lrc
+    }
 
-//     [<Fact>]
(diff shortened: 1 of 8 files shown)


tests/Rust/tests/src/AsyncTests.fs
```diff
@@ -93,30 +93,53 @@ let shouldExecuteMutationOnTask () =
...
-// module Monitor =
-//     [<Fact>]
-//     let monitorShouldWorkWithSystemObj () =
-//         let o = new System.Object()
-//         System.Threading.Monitor.Enter(o)
-//         System.Threading.Monitor.Exit(o)
+module Monitor =
+    // [<Fact>]
+    // let monitorShouldWorkWithSystemObj () =
+    //     let o = new System.Object() // todo - this doesn't work, and outputs unit
```
ncave, line 108: @alexswan10k Yeah, looks like in `Rust/Replacements` (and elsewhere) this gets converted to an object expression, but it doesn't have to be like that, it can be whatever we want it to be. Please also keep in mind that currently the `obj` type (i.e. `System.Object`) is erased to `Fable.Any`, but that may not be ideal either, see discussion in #3027.

alexswan10k:
Thanks @ncave. On another point do you have any idea why system.theading.thread blows up? Looks like it was coming from the F# compiler

ncave:
@alexswan10k 
> why system.theading.thread blows up

Fixed by #3041. To put it simply, Fable tries to only use and implement the bare minimum of .NET assemblies.
Technically that's not a requirement for the Fable CLI, it's for Fable REPL, but I guess for consistency we're restricting ourselves to a fixed set of core .NET assemblies.

alexswan10k:
Oh nice one. Is this even the right call do you think? I am a little torn if to even implement thread, it's just it maps quite nicely to the rust thread construct and makes testing some basic stuff like locks a little less noisy.

ncave:
@alexswan10k 
> I am a little torn if to even implement thread

It's up to you, I don't think currently the other languages implement any of it, but perhaps they might.

I think the implementation criteria for Fable is "Most or all that is in `FSharp.Core`" and a bare minimum of .NET BCL types that are expected to be be there for quality of life out of the box (e.g. dates, times, random, etc.). Again, this criteria was based mostly on compiling to JavaScript for use in a browser, so it had a natural boundary (no IO, etc.).

But now that we have languages that target wider platforms, perhaps the criteria can be updated with the times. But probably not much, we still don't want (nor do we have Microsoft's resources) to re-implement and maintain the whole .NET BCL library. So most of it (where needed) can and should be done in separate project packages that can be maintained on their own.

alexswan10k:
Agree.

Kind of an obvious in retrospect, but this implementation is not thread-safe across different CPU caches. I believe this needs to actually use a OS level mutex but calling ```sys::MovableMutex```.

I will give this another pass at some point.

PR #3044, opened 2022-08-04
Rust -- implement Thread.Sleep()

alexswan10k:


src/Fable.Transforms/Rust/Replacements.fs
```diff
@@ -2430,6 +2430,8 @@ let threads com (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: E
     match thisArg, i.CompiledName, i.GenericArgs with
     | _, "Thread", [tType] ->
         Helper.LibCall(com, "Thread", "new", tType, args, ?loc=r) |> Some
+    | _, "Sleep", _ ->
```
ncave, line 2433: We want to be specific about which overload is being implemented, something like this:
```fs
| None, "Sleep", [Number(Int32,_)] ->
```

ncave (changes requested):

src/Fable.Transforms/Rust/Replacements.fs
ncave: @alexswan10k Or, in case you are matching on arguments, you can do something like this:
```fs
    match i.CompiledName, thisArg, args with
    | "Sleep", None, [ExprType(Number(Int32,_))] ->
```

ncave: It's fixed, nothing to do.

alexswan10k: Sorry would have been happy to fix this but you beat me to it :)

I will be in this area again soon as I wanted to add Start, Join, etc.., will clamp it down then.

Issue #3051, opened 2022-08-05
Unexpected results for nested options with Option.orElse


IanManske:
### Description

For nested options (i.e. `Option<Option<_>>`) where the innermost option is `None`, `Option.orElse` does not behave as expected.
Instead of returning the second arguement `option` as is, it is unwrapped and the inner `option` is returned. This occurs because `Option.orElse` is compiled into JS using `defaultArg`.
This same issue exists for `Option.orElseWith`.

### Repo code
[REPL](https://fable.io/repl/#?code=DYUwLgBAhhC8EGUD2BbEEBySB2IBQA9ARAGIDEAXNHIqulrgNwQBSCVM8yaEA3hADcowAK4gqI7ABMQAMwCWuKRAC+EPAAcATorCzsEAERQqAUgCCAGgjyAzt3TxTAI2t2GjiC8PUoAOgBJezpoQNsPPDxQSGcaGAAfAD4IAHkNMHkcPyQtAFFgW3QACgcIEpDDXAEQLUMASjrCYnIykAAPDRAAYzAQKTqqWK4QiKJSMjKoHpFhAYghzBwQZjZBmkkZBSVNHWw9A0NnMysbYJ4nV1OPGm95+bDS5zCIoA&html=Q&css=Q)

### Expected and actual results
Expected:
```fsharp
(Some None) |> Option.orElse (Some (Some "never")) // = Some None
// the expected JS representation is: Some { value: undefined }
```
Actual:
```fsharp
(Some None) |> Option.orElse (Some (Some "never")) // = None
// the actual JS representation: undefined
```
### Related information

* Fable version: 3.7.10 (REPL), also tested on 3.7.17

alfonsogarciacaro:
Closed by #3052, will release a version with the fix soon.

PR #3052, opened 2022-08-05
Fix using nested options with Option.orElse

IanManske:
This address issue #3051.
I've made changes to the obvious areas, but I'm not sure if anything else needs to be changed.
The (new) tests pass and I also got the expected results as described in the issue with a local package build.


alfonsogarciacaro:
Sorry for the late reply @IanManske, I was unavailable last week. Thanks a lot for the fix! This is me again trying to be too clever to reduce the bundle size. These dirty tricks always bite me at the end ;)

Thanks a lot for the fix and the tests!

PR #3061, opened 2022-08-08
Renamed Implicit to Primary constructors

### LINKED ISSUE #3058: Question: Implicit vs compiler generated constructors
I'm a bit confused about the naming of some functions, i.e `transformClassWithImplicitConstructor` vs `transformClassWithCompilerGeneratedConstructor`. It seems to me that `transformClassWithImplicitConstructor` has a very explicit `cons` argument since the type that is being translated contains one (or more) constructors. 

While `transformClassWithCompilerGeneratedConstructor` do not have any explicit constructor and will generate one for you. Thus the constructor is implicit here.

Shouldn't the names instead then be `transformClassWithExplicitConstructor` and `transformClassWithCompilerGeneratedConstructor`, or have I misunderstood something?

- Explicit: fully revealed or expressed without vagueness
- Implicit: capable of being understood from something else though unexpressed

ncave:
Fixes #3058.


dbrattli (approved):
Looks good!

ncave:
Merging for now, we can rollback if there are other opinions later.

PR #3067, opened 2022-08-12
Rust - preliminary pass at GUID's

alexswan10k:
So this is only partly there. Implemented:

* Basic GUID struct
* Parse
* ToString
* Compare
* Donkey work plumbing of adding type, tests, redirects etc

Fable transpiling problems
* Could not intercept ```new Guid()``` for some reason

This was mainly to get the ball rolling on how we go about this, and it has raised some questions:

* Everyone in the rust community seems to use the [rand](https://docs.rs/rand/latest/rand/) crate for doing any kind of rand generation (required for ```NewGuid()```). We are going to need this for other things too. Do we take on a dependency to rand?
* A lot of the complicated stuff here is copied from the [uuid](https://docs.rs/uuid/0.5.1/uuid/index.html) crate. This crate is obviously far more comprehensive, but it also supports other UUID formats that .NET does not need anyway out of the box. We could shim this and call down onto this library too but it would again require taking on a dependency. Do we want to do this?

Assuming that we take on dependencies, should they be gated behind a feature? 

In principle taking on dependencies is fine, but I do worry a little about transitive dependencies forcing decisions on users about what version they use. Now Rust does support multiple different transitive dep versions in a tree (as node does too), but these will obviously be compiled twice and may present conversion challenges etc.


ncave:
@alexswan10k Looks good, just a few random thoughts:
> A lot of the complicated stuff here is copied from the [uuid](https://docs.rs/uuid/0.5.1/uuid/index.html) crate
- Yes, we should probably just take dependency on `uuid` instead of copying the source and having to maintain it. Looks like it has a feature just for `v4` that [uses](https://docs.rs/uuid/1.1.2/uuid/struct.Uuid.html#method.new_v4) the `getrandom` crate which should be [compatible](https://docs.rs/getrandom/latest/getrandom/#webassembly-support) with `wasm` targets:
```
[dependencies]
uuid = { version = "1.1.2", features = ["v4"] }
```
> Do we take on a dependency to rand?
- `rand` is nice but isn't it big? Also I'm not sure it compiles for `wasm32-unknown-unknown` targets.
If we need simple random generators for implementing `System.Random` we can probably use `getrandom` for non-deterministic or something simpler like `fastrand` or `oorandom` for deterministic.

> Assuming that we take on dependencies, should they be gated behind a feature?
- Perhaps the criteria whether it should be feature-gated can be if it can or cannot compile to `no_std` (or the limited `std` that is supported by the `wasm` targets).

- Perhaps we can strive to make the core Fable-generated Rust code `no_std` compatible, and add features for everything else. That may require switching to [alloc](https://doc.rust-lang.org/alloc/) for `Vec/String/Box/Rc/Arc` and [hashbrown](https://github.com/rust-lang/hashbrown) for `HashMap/HashSet`.

- Talking about features, what do you think of splitting the `futures` feature into two features, `atomic` and `threading`?
This way we can still provide atomic pointers where needed, makes it a bit more flexible and a bit clearer what's included.

alexswan10k:
> @alexswan10k Looks good, just a few random thoughts:
> 
> > A lot of the complicated stuff here is copied from the [uuid](https://docs.rs/uuid/0.5.1/uuid/index.html) crate
> 
> * Yes, we should probably just take dependency on `uuid` instead of copying the source and having to maintain it. Looks like it has a feature just for `v4` that [uses](https://docs.rs/uuid/1.1.2/uuid/struct.Uuid.html#method.new_v4) the `getrandom` crate which in theory [should be compatible](https://docs.rs/getrandom/latest/getrandom/#webassembly-support) with `wasm` targets:
> 
> ```
> [dependencies]
> uuid = { version = "1.1.2", features = ["v4"] }
> ```
I think I agree with this, I was just worried about the bloat. It is absolutely true though that this crate has pretty granular feature control so happy to go with this.

> > Do we take on a dependency to rand?
> 
> * `rand` is nice but isn't it big? Also I'm not sure it compiles for `wasm32-unknown-unknown` targets.
>   If we need random generators for implementing `System.Random` we can probably use `getrandom` for OS-provided non-deterministic or something simpler like `fastrand` or `oorandom` for deterministic.
> 

Happy with this too. On a side note if we take on a dependency to uuid we do not need to worry about rand right now, as it internally does some stuff to randomly generate guids for the v4 spec

> > Assuming that we take on dependencies, should they be gated behind a feature?
> 
> * Perhaps the criteria whether it should be feature-gated can be whether it can (or cannot) compile to `no_std` (or the limited `std` that is supported by the `wasm` targets). What is your opinion?

This makes sense. I guess anything that adds bloat to an assembly feels like something that a user should have control over, although you obviously want the out of the box experience to just work. Maybe we can solve this with sensible defaults.

> * Perhaps we can strive to make the core Fable-generated Rust code `no_std` compatible, and add features for everything else. That may require switching to [alloc](https://doc.rust-lang.org/alloc/) for `Vec/String/Box/Rc/Arc` and [hashbrown](https://github.com/rust-lang/hashbrown) for `HashMap/HashSet`.
> * Talking about features, what do you think of splitting the `futures` feature into two features, `atomic` and `threading`?
>   This way we can provide atomic pointers (with or without threading) as needed, makes it a bit more flexible and a bit clearer what's included.

I agree completely with this. More granularity is better here. I think it is highly likely our threading constructs are going to suck for a while, and a user might have higher confidence writing threaded code in Rust, but writing the constituent functions in F#. This would allow one without the other. This is also partly why I insisted on the "unsafe-cell" feature as explicit opt-in, because the Rust approach to threading is superior, and provably correct - it seems a shame to throw this out just because .NET does not have those guarantees.

no-std would be a great goal to aim for. I don't know how feasible this is, but its certainly a good idea to actively try and position everything to work this way if possible. Even if it doesn't work for MVP it will reduce pain down the road when we do eventually get to it.

ncave:
@alexswan10k 
> On this note, what do we do about Guid. Feature to enable the dependency/lib + default of on?

Perhaps just take on the dependency (with `v4` only) without a feature for now, let's see how much it adds.
I need to disable the default features on `rust_decimal` too, it pulls `serde` right now by default.

src/fable-library-rust/src/Guid.rs
```diff
@@ -0,0 +1,32 @@
+// [Y combinator](https://en.wikipedia.org/wiki/Fixed-point_combinator#Fixed-point_combinators_in_lambda_calculus)
```
ncave, line 1: This comment perhaps not needed.

alexswan10k: Agree! Copy paste... fixed

ncave:
@alexswan10k Looks good, I'll take a look at why new and empty guilds don't work.

PR #3080, opened 2022-08-16
Rust - First bit of progress on DateTime

alexswan10k:
Pushing DateTime forward a tiny bit. 2 tests working.

Added:
* DateTime.rs with modules for DateTime, DateTimeOffset, and TimeSpan (mostly placeholders at the moment).
* Added appropriate replacements & type replacements for above types (WIP)
* Added apis for
  * new_ymd
  * new_ymdhms
  * new_ymdhmsms
  * to_string
* Added date feature switch - compiles modules & pulls in Chrono
* Added guid feature switch - compiles module & pulls in uuid
* Turned on above 2 by default, but they can be turned off with ```default-features = false``` if a user wants to not use these, and cut down on some bloat.
* Added a test for ```Box``` multi owner -  behaviour is clone and fork (already working) which is a nice alternative option when you do not want the reference counting overhead.

This is far from working in any useful sense, but it hopefully lays some groundwork for wrapping the chrono types etc. I assume we are happy to go with chrono? There are other options but this seems like the preferred option to most.


src/fable-library-rust/src/DateTime.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // impl core::fmt::Display for DateTime {
+    //     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+    //         write!(f, "{}", self.0.to_string())
+    //     }
+    // }
+
+    pub fn new_ymd(y: i32, m: i32, d: i32) -> DateTime {
+        let l = Local.ymd(y, m as u32, d as u32).and_hms(0, 0, 0);
+        DateTime(LocalUtcWrap::CLocal(l))
```
alexswan10k, line 24: Is this even correct? What does .NET do out of the box, local or UTC?

ncave: @alexswan10k I think it's [unspecified](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.-ctor?view=net-6.0#system-datetime-ctor(system-int32-system-int32-system-int32)).

alexswan10k: Hmm, that could be tricky to model. Will have to have a think. Perhaps NativeDate can be used as a 3rd option or something.

ncave: @alexswan10k I agree it's confusing, I guess that's why the `DateTimeOffset` struct was introduced.
I guess we can consider switching the representation of `DateTime` to `chrono::naive::NaiveDateTime` + DateTimeKind.

PR #3083, opened 2022-08-17
[Rust] Initial no_std support

ncave:
- Added object reference equality.
- Changed string concatenation.
- Changed module let bindings to not use `thread_local`.
- Added initial `no_std` support.


alexswan10k:
Awesome

PR #3097, opened 2022-08-20
Upgrade to Ionide.ProjInfo

### LINKED ISSUE #3047: Differentiating development and production builds from F# code
### Description

I'm looking for a way to alter my Fable application depending on whether it is being used in a development or production context. In my case, I want to use a different service implementation during development, but another example might be including a debug page only in development builds.

I cannot find any documentation on how to achieve this, perhaps using compilation symbols (e.g. `FABLE_COMPILER_DEVELOPMENT`) or the likes. Any help appreciated.

IanManske:
This attempts to fix #3047.
At a surface level, this seems to work (e.g., all tests pass and `DEBUG` operates as intended).

However I have not tested:
- relative vs absolute file paths (I believe `Ionide.ProjInfo` provides absolute but unnormalized paths).
- cross-targeting
  - seems to 'work', in the sense that the first listed framework is chosen.
  - this is despite the fact that I haven't been able to get the `IsCrossTargetingBuild` prop to appear.
- other edge cases I am unaware of

Note: I used the latest version of `Ionide.ProjInfo` which requires `net6.0`. Upgrading *seems* to not have broken anything. Still, this could be held off and shipped with Fable 4 (if that's upgrading to `net6.0`), otherwise we can use an earlier `Ionide.ProjInfo` version compatible with `net5.0`.

~Also, the changes removed the existing error reporting, as this is now internal to `Ionide.ProjInfo`. I believe functions from it (marked in the code) can still throw. This may or may not interact weirdly with `retryGetCrackedProjects`.~


IanManske:
I'm having trouble figuring out how to fix the following runtime error:
```
Could not load file or assembly 'System.Configuration.ConfigurationManager, Version=6.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. Could not find or load a specific file. (0x80131621)
```
This is encountered by Ionide.ProjInfo while trying to load any of our `Fable.Library.fsproj` projects.

It only happens while running on dotnet `6.0.400` (and maybe some prior versions). A [test branch](https://github.com/IanManske/Fable/tree/test-branch) I made running dotnet `6.0.168` does not have this error and [all tests there pass](https://github.com/IanManske/Fable/actions/runs/2907834684).

Additionally, this error did not happen before I merged this branch into snake_island (commit [8953f25](https://github.com/fable-compiler/Fable/pull/3097/commits/8953f25e06da7469a2125dac6d15d32f58f99fe3)). Maybe it was something I did instead?

I would appreciate any help or insight into this, as I don't have much experience with errors like these. Thanks!

alfonsogarciacaro:
Thanks for retargeting the PR @IanManske. Oh, that's a weird error indeed. It does ring a bell but I don't remember where exactly I saw it, I'll check. (There've been some regressions in the F# compiler so we need to update it but not sure if it's related.)

IanManske:
I tested all dotnet runtime versions `6.0.{1,2,3,4}00` and only `6.0.400` fails.
It seems like [this](https://github.com/microsoft/MSBuildLocator/issues/159) related issue explains the problem.
We are using the lastest version `17.2` for `Microsoft.Build.Framework` which shipped for dotnet `6.0.300`.
So, this will hopefully be solved once `Microsoft.Build.Framework` version `17.3` releases and we can update it [here](https://github.com/IanManske/Fable/blob/cd204e2f64fed2eb98e5d1af55150ed8e4f9a505/src/Fable.Cli/Fable.Cli.fsproj#L50).
I'll continue to work on this, but I'll be doing testing using dotnet version `<6.0.400` for the time being.

alfonsogarciacaro:
@IanManske @dbrattli @ncave @alexswan10k I've renamed "old" `main` to `fable3` and `snake_island` to `main`. The open PRs have been retargeted. Locally you should only need to check out main and pull, please work on this branch from now on. Thanks and let me know if you've any trouble!

IanManske:
# Progress Update
`Microsoft.Build.Framework` 17.3 released, but it has not fixed the issue.
So, this PR is still broken for dotnet `6.0.400` for now but should work for all other `6.0.x` versions.
Just to demonstrate that the CI tests pass, I've temporarily set the dotnet version to `6.0.300` in the github actions workflow.

# Changes
Upgraded to `Ionide.ProjInfo` (duh):
- This uses `Microsoft.Build.Graph` under the hood and may be faster for projects with lots of project references.
- Added logging for project loading (`logProjectLoad`).

Tests:
- Added multi-targeting project configuration case inspired by #3047.
- Added simple project order preservation test.
  - I put it under `tests/Integration/Compiler` -- not sure if this is the correct place.
- Upgraded projects, etc. to `net6.0`.
- Updated actions on `test.yml` (but again, using dotnet `6.0.300` for now).

Most changes are in `Fable.Cli/ProjectCracker.fs`.
I made sure to check that I did not break:
  - compilation order (i.e., project references including fable packages).
  - paths are full and normalized.
  - tests pass of course.

Please double check if necessary.

# Left to do / Skipped
- Retrying project loads when encountering file lock.
  - `retryGetCrackedProjects` only catches IO execeptions. I'm guessing `Dotnet.ProjInfo` used to throw IO exceptions, but I don't think `Ionide.ProjInfo` does.
- Update ReadMe's (e.g., list `net6.0` requirement)
- Do we need to update fcs or `lib/` as well to be in line with `net6.0`?
- Anything else I may have missed.

So yeah, I think the main part of this PR is finished.
Let me know if anything needs to be changed or fixed!

alfonsogarciacaro:
This is great @IanManske, thanks a lot! My only concern is that this will fail when 6.0.400 is used, is that right? In that case, should we wait until there's a fix to prevent Fable from breaking or should we try to restrict the SDK version somehow?

IanManske:
It's always the smallest of things...
I believe I've got it working for `6.0.400` @alfonsogarciacaro . I just had to remove an unused package. 🤦‍♂️

As a side note, I think we still need `FSharp.SystemTextJson` for `Fable.Cli`? There was a todo note regarding upgrading to `net6.0` that I removed.

alfonsogarciacaro:
Awesome @IanManske, thanks again!

Sorry for the late reply, I couldn't check this during the weekend. I will test your branch and if there's no problem I will merge 👍

About  FSharp.SystemTextJson I need to check the end. I think at the beginning I was just planning to use very simple serialization and net6 should support F# unions and records by default. But IIRC at the end I used some specific features from FSharp.SystemTextJson so probably we cannot remove the library anyways.

IanManske:
So actually, this PR breaks loading scripts which I forgot to check (we don't have any tests for this).
I'm working on a solution right now, but I may need some help with it. Part of this is actually identifying what the problem is.

The problem may be that `Ionide.ProjInfo` needs to load `Microsoft.Build.Framework` specifically? at [runtime](https://github.com/ionide/proj-info#getting-started). However, we also need to reference/include `Microsoft.Build.Utilities.Core` in order to load scripts (this is the package that I removed, since I thought we didn't need it). `Microsoft.Build.Utilities.Core`, in turn, also needs `Microsoft.Build.Framework` which conflicts? with `Ionide.ProjInfo`'s runtime requirement.

Otherwise, the [previously mentioned issues](https://github.com/fable-compiler/Fable/pull/3097#issuecomment-1224168203) might also be relevant.

alfonsogarciacaro:
I see, thank for pointing that out. TBH, script loading has been working intermittently, but it'd be nice to have it working for Fable 4, particularly for quickly testing compilation of Python scripts.

IanManske:
I've tried to get script loading working, but I've only been able to get it to work for specific versions at a time without breaking project loading. I think I'm missing something, because I believe the current `Fable.Cli.fsproj` (or something similar) should work. Both [FsAutoComplete](https://github.com/fsharp/FsAutoComplete/blob/main/paket.dependencies) and [FSharpLint](https://github.com/fsprojects/FSharpLint/blob/master/paket.dependencies) (although the latter uses `net5.0`) have something like
```
nuget FSharp.Compiler.Service
nuget Ionide.ProjInfo

nuget Microsoft.Build copy_local: false
nuget Microsoft.Build.Framework copy_local: false
nuget Microsoft.Build.Tasks.Core copy_local: false
nuget Microsoft.Build.Utilities.Core copy_local: false
```
in their `paket.dependencies`. The current proposed `Fable.Cli.fsproj` I think follows this. With this configuration, project loading works, but script loading fails due to not having `Microsoft.Build.Utilities.Core`. Adding this package breaks project loading but fixes script loading. I believe this is true for all `6.0.x` sdk versions.

[FSharp.Formatting](https://github.com/fsprojects/FSharp.Formatting/blob/main/paket.dependencies) does not exclude the runtime assets:
```
nuget FSharp.Compiler.Service
nuget Ionide.ProjInfo

nuget Microsoft.Build.Framework
nuget Microsoft.Build.Tasks.Core
nuget Microsoft.Build.Utilities.Core
```
I have tried this as well, and this does work (project loading and script loading).
However, I could only seem to get it working for sdk `6.0.400` (fails for `6.0.300`, etc.).

alfonsogarciacaro:
If using sdk 6.0.400 makes both project and script loading work, we probably can make it a requirement for the release of Fable 4, as it is a new major version and I believe most users keep the SDK frequently up-to-date. Would that work?

IanManske:
@alfonsogarciacaro I guess that will work for now. I'll continue to try and find a solution, although university is starting and I may be busy.
I also noticed that loading other scripts from a script doesn't seem to work. That's a separate problem, so I might open a different PR for that.
Otherwise, one final thing is that missing dependencies are currently being silently ignored. I'm not sure if I should add a `Log.warning` or `Log.verbose` (see [cf10c79](https://github.com/fable-compiler/Fable/pull/3097/commits/cf10c793011cf621194fb7595a4dc8edd3c390da) and the [culprit test](https://github.com/fable-compiler/Fable/runs/8194006472?check_suite_focus=true)).

kentcb:
Hi @IanManske @alfonsogarciacaro . Is there anything I can do to help get this one over the line? It's quite a pain point for us. Also, can I clarify whether this will be Fable 4 only, or will Fable 3 be patched as well?

alfonsogarciacaro:
TBH I'm a bit concerned because there are many moving pieces right now and changing the project cracker brings some risks given the current way, even if old, it's working fine except in very punctual cases. Also, there are many reports of issues with latest versions of dotnet sdk and Ionide, so I was hoping to wait until things stabilize both in Fable and Ionide/dotnet sdk before merging this.

In principle, we're not updating Fable 3 anymore unless there's a critical bug. @kentcb, the problem you mention is the one reported in #3047 or a new one? Could you solve the issue by not targeting multiple frameworks?

kentcb:
@alfonsogarciacaro  #3047 is indeed the bug I'm trying to get around. Unfortunately, we cannot drop multi-targeting as we have projects that are built on both .NET Core and .NET FX.

alfonsogarciacaro:
@kentcb There's one addition I've been considering: passing `FABLE_COMPILER` also when parsing the .fsproj as property. This would allow you to have different frameworks depending on whether you want to compile for Fable or not. Example:

```
    <TargetFrameworks Condition="'$(FABLE_COMPILER)' == 'true'">net6.0</TargetFrameworks>
    <TargetFrameworks Condition="'$(FABLE_COMPILER)' != 'true'">net6.0;netstandard2.0</TargetFrameworks>
```

Would that work for you? That's a simple non-breaking change and we could publish a patch version of Fable 3. Or, you can also set `FABLE_COMPILER` as env var and you don't even need to upgrade Fable.

kentcb:
@alfonsogarciacaro I think that would work as a workaround, yes.

alfonsogarciacaro:
@kentcb I've published v3.7.19 with the change, can you please give it a try? Thanks!

alfonsogarciacaro:
@IanManske I've merged with `main` and fixed the merge conflicts in `Ionide-ProjInfo-sync` branch. But Github doesn't let me send a PR to your fork (or I'm doing something wrong). Can you please rebase with my branch instead?

IanManske:
@alfonsogarciacaro Sorry I haven't been following this too closely. I've rebased my fork, let me know if anything looks off. I know that dotnet `6.0.401` came out, but I haven't gotten around to testing it. I'm not sure if this and/or other updates have broken anything (although, it seems like the CI passes).

IanManske:
Since `net7.0` is out and ProjInfo is still wonky, I'll remark this as draft.
If we end up coming back to ProjInfo just lmk, and I can start up again from here.

alfonsogarciacaro:
Thanks a lot @IanManske. Yes, things are getting even more complicated with net7. I'm currently trying a hacky approach of "disguising" the .fsproj as .csproj and using Buildalyzer to invoke a design build.

I agree it's a good idea to keep the PR on hold for now and revisit it later. Thanks a lot for your work!

MangelMaxime:
I am closing because in the end using `csproj` in disguise was the choosen approach.

Alse with the release .NET 8, we can ask MSBuild directly to resolve the dependencies for us now. It will be explored in the future.

PR #3100, opened 2022-08-22
[Python] type var cleanup. Use Any for types starting with $$

dbrattli:
Btw: What are these inferred types starting with `$$` and then a number @alfonsogarciacaro ?


alfonsogarciacaro:
If I remember correctly compiler-generated generic param names. There's a note about it here: https://github.com/fable-compiler/Fable/blob/2fab92763b6a26bd5a8666a3020782f9ed4a7964/src/Fable.Transforms/FSharp2Fable.Util.fs#L804-L806

The F# compiler prefixes compiler-generated generic params with `?` and I'm replacing it with `$`. Additionally generic params are prefixed with $ to avoid clashes with other identifiers (e.g. Dart doesn't allow a generic to be named the same as the function). But we can use language-specific solutions in `genParamName` as we do in other cases.

PR #3101, opened 2022-08-22
Fix local package version generation to satisfy test.

IanManske:
Makes sure the local package build version number will always match the version expected by tests.


alfonsogarciacaro:
Nice catch, thank you!

PR #3103, opened 2022-08-23
Rust - fix byref nested context not correctly propagating

alexswan10k:
Found another couple more edge cases while experimenting with v21...

Looks like there were two things going on
* Nested context propagating things that should have been reset (fixed)
* Compiler optimizing structs leading to unexpected output in the AST (no idea where to start)

I have written a failing test that covers these (uncomment marked ```[<Struct>]``` line). The first issue I have resolved. The second one I am at a loss as where to start. I don't suppose you might have any ideas @ncave?

It seems to generate the following sort of problematic output:
```
                    let inputRecord: Lrc<MutCell<Lrc<MutCell<Entities::Spatial>>>> =
                        Lrc::new(MutCell::new(&ent.Spatial.get()));
                    Entities::Spatial {
                        Rotation: (ent.Spatial).Rotation + r,
                        Position: inputRecord.Position.clone(),
                    }
```
Where inputRecord is not my code. It is generated.

I have pretty much ruled out
* Leave context + context propagation problems
* Any notion of ident mutability in the Fable2Rust side
* Any notion of ident tracking
* Passing ByRef (both fail in exactly the same way)
* Type LRC double wrapping bugs

I believe what is happening is the F# compiler is generating the inputRecord implicit mutable local var, which is leading to some kind of unexpected input AST. I know there was some work done to reduce this before. Could this be related, or perhaps provide a way out?

Ideally if we could just turn off the optimization in the F# phase that would be helpful, but I imagine this is non trivial.


tests/Rust/tests/src/RecordTests.fs
```diff
@@ -351,3 +351,28 @@ let ``Mutating records work`` () =
...
     equal -10 x''.uniqueA
     equal -20 x''.uniqueB
+
+module ComplexEdgeCases =
+    open Fable.Core.Rust
+    [<Measure>] type m
+    [<Measure>] type Rad
+
+    //TODO - When this is turned into a struct, the F# compiler seems to do an optimization which seems to break everything
+    //[<Struct>]
```
alexswan10k, line 361: here
```suggestion
    [<Struct>]
```

ncave:
@alexswan10k I'll take a look, thanks.

ncave:
@alexswan10k Sorry, I was busy with something else. Fixed in #3114.

Credit goes to @alfonsogarciacaro for nicely optimizing away the extra copies from the body.

alexswan10k:
Many thanks @ncave and @alfonsogarciacaro this is great.

Something is still not quite right as it is trying to unwrap the record as a mutable get or something... I will come back to this.

ncave:
@alexswan10k Fixed in #3116. I checked in your tests (without the `ByRef` attr as the outer record is already a reference type (Lrc)), but you can update it as you see fit.

alexswan10k:
Thanks @ncave you are way ahead of me as usual :)

I put the byref back in because it actually found another edge case when dereferencing a nested expression, due to byref ctx propagation (see transformLeaveContext change). I agree that for this use case it is not really necessary though.

PR #3105, opened 2022-08-25
[Rust] Fixed static, member and interface imports

ncave:
- Fixed static, member and interface calls when imported from another file.

### DIFF
--- a/src/Fable.Transforms/Rust/AST/Rust.AST.Helpers.fs
+++ b/src/Fable.Transforms/Rust/AST/Rust.AST.Helpers.fs
@@ -36,7 +36,7 @@ module Naming =
         elif rustPrelude.Contains(ident) then ident + "_"
         else stripRaw ident // no need to keep it raw here
 
-    let splitFullName (name: string) =
+    let splitNameParts (name: string) =
         name.Split([|"."; "::"|], System.StringSplitOptions.RemoveEmptyEntries)
         |> List.ofArray
 
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -30,16 +30,11 @@ type UsedNames = {
     CurrentDeclarationScope: HashSet<string>
 }
 
-type TypegenContext = {
-    IsParamAnyType: bool
-    IsParamByRefPreferred: bool
-}
-
 type ScopedVarAttrs = {
     IsArm: bool
     IsRef: bool
     IsBox: bool
-    HasMultipleUses: bool
+    // HasMultipleUses: bool
     mutable UsageCount: int
 }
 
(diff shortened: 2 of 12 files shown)


ncave:
@alexswan10k Please evaluate when you have a chance and tell me what you think.

The class member calls (static, member and interface calls) were completely broken, when imported from another file. To fix that, I had to move quite a few things around:
- changed imports and calling convention for static methods (only import the class).
- changed calling convention for interfaces (from method to static call).
- moved member methods out of their Methods trait into the class impl (except overrides).
- fixed fluent self types from `self: &Self` to `self: &Lrc<Self>`, when applicable.
- commented some unused code.

If you have the time, I would appreciate your opinion on those changes. None of that is set in stone, we can still move things back and forth (and we'll probably have to, say to make type extensions in a different file than the class/struct work). Still, I think this is a step forward, as it makes more things work.

alexswan10k:
Hi @ncave ,

Definitely some good improvements here. Some thoughts:

> * changed imports and calling convention for static methods (only import the class).
This makes sense
> * changed calling convention for interfaces (from method to static call).
I think this is fine, it should not matter either way
> * moved member methods out of their Methods trait into the class impl (except overrides).
So this is an interesting one. The whole idea of the Methods trait is it allows dynamic dispatch and eventually inheritance. It does come with a major caveat though, which is that all instances need to be ```Lrc<dyn Lrc<Class>>```, so you pay the price everywhere. I don't think this was previously implemented yet. It also does not look like we have any tests for any inheritance stuff yet, so I guess the big question. Do we even bother?
> * fixed fluent self types from `self: &Self` to `self: &Lrc<Self>`, where applicable.
Again, I think this makes sense. It sees correct to always accept ```&Lrc<Self>``` in the context of something which is a pointer-wrapped type. I guess the detail here is that previously self referred to the type in the impl block, so the context of self would subtly change depending on the calling convention.
> * commented some unused code, minor cleanup.
Great. All looks good. Some old iterations that were laying around still that needed a purge.

All in all, I am happy. I guess the Methods trait thing deserves discussion though, as this really is our only current way out for inheritance and dynamic dispatch. As I have mentioned before I would rather inheritance did not exist because of the endless bugs and complexity it introduces. Always-on dynamic dispatch also comes with a performance penalty as we know.

At the same time, I could see this being unexpected for an unknowing user, and thus problematic when expecting to compile generic F# (packages etc) which happens to use these patterns. I wonder even if this could be delegated to the feature switch, but I am also patently aware that every feature switch multiples the output space of possible Rust programs by 2 thus leading to combination explosion. I can't see how we can feasibly test this all these combinations in CI without needing 100+  agents!

As for the Methods approach, I have still not figured out all the details anyway. For example, I am struggling to wrap my head around how that approach would help to implement the two most evil concepts of OO - the virtual method, and the overridden method. I imagine this is solvable though, perhaps with some additional codegen. Anyway, I am hinging on parking all this completely, at least untill a keen OO programmer who cares about this stuff comes along and volunteers to work all this out! I think we are in a good place with interface dynamic dispatch already.

ncave:
@alexswan10k 
>
   >     changed calling convention for interfaces (from method to static call).
> I think this is fine, it should not matter either way

To clarify, it doesn't work the other way, for some reason Rust cannot resolve what to call and complains when interface methods are called as an instance method call (when multiple classes implement the same interface). Luckily the other syntax works fine (that's probably why it's there).

> It seems correct to always accept `&Lrc<Self>` in the context of something which is a pointer-wrapped type. I guess the detail here is that previously self referred to the type in the impl block, so the context of self would subtly change depending on the calling convention.

Yes, previously fluent methods were implemented in trait impl for `Lrc<Self>`, so their self type was still `&Lrc<Self>`. Since we can no longer implement for `Lrc<Self>`, the type had to be adjusted to match.

> The whole idea of the Methods trait is it allows dynamic dispatch and eventually inheritance.
It does come with a major caveat though

Yes, I agree we can keep a big feature like inheritance on the back burner, until all the other more important features are completed.

> I am also patently aware that every feature switch multiples the output space of possible Rust programs by 2 thus leading to combination explosion.

Yes, we may have to eventually combine some optional features to simplify testing.

PR #3109, opened 2022-08-26
Updated FCS-Fable

ncave:
- Updated `FCS-Fable` to latest.


ncave:
@alfonsogarciacaro 

Just FYI, there seems to be a change in the intellisense `DeclarationListItem`:
```fs
type public DeclarationListItem =

    /// Get the text to display in the declaration list for the declaration.
    /// This is a display name without backticks.
    [<Obsolete("Renamed to NameInList - this is the text to appear in the declaration list, while NameInCode is the text inserted on completion.")>]
    member Name: string

    /// This is a display name without backticks.
    member NameInList: string

    /// This is a display name with backticks if necessary.
    member NameInCode: string
```

PR #3113, opened 2022-08-27
[Rust] Output string type

ncave:
- Output string type alias in the generated code.

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -392,6 +392,7 @@ module TypeInfo =
         | Fable.Boolean
         | Fable.Char
         | Fable.Number _
+        | Fable.String
         | Fable.GenericParam _
         // struct containers, no need to Rc-wrap
         | Fable.List _
@@ -407,7 +408,6 @@ module TypeInfo =
             -> if ctx.RequiresSendSync then Some Arc else Some Lrc
 
         // should be Rc-wrapped
-        | Fable.String
         | Fable.Regex
         | Fable.Array _
         | Replacements.Util.Builtin (Replacements.Util.BclHashSet _)
@@ -617,7 +617,7 @@ module TypeInfo =
         ]
         mkDynTraitTy bounds
 
-    let numberType com ctx kind: Rust.Ty =
+    let transformNumberType com ctx kind: Rust.Ty =
         match kind with
         | Int8 -> "i8" |> primitiveType
         | UInt8 -> "u8" |> primitiveType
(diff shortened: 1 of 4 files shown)


ncave:
@alexswan10k On a somewhat related topic, do we need to revisit the implementation decision on strings?

I know we have discussed this already in the past, but we might want to re-evaluate what is the best string type:

- Current string type, always heap allocated. Simple, easy to use, to pass around, and to reason about.
```rs
pub type string = Lrc<str>;
```
- Or, perhaps we can just use `String`. Sounds enticing as it will be very interop-friendly, but cloning it makes a copy so we'll have to pass it by reference everywhere to avoid the extra allocation, and it's going to be cloned when it's returned from a function.
```rs
pub type string = String;
```
- Or, perhaps we can do better with an enum that differentiates between static and allocated strings, to avoid allocating strings that are already baked in the binary:
```rs
pub enum CowRcStr {
    Borrowed(&'static str),
    Shared(Lrc<String>),
}

pub type string = CowRcStr;
```
Those are just hypothetical questions, perhaps all we need for now is a good interop story for strings, and we can postpone the memory optimization for later.

alexswan10k:
Some good ideas here. I agree, there is are quite a few potential edge cases we are trying to capture here, whilst trying to expose a coherent API surface for Rust consumption.

This to me looks like another text book case for creating our own wrapper (just like Array/List/Set).

I propose we start with something like the following, but we could easily extend to your enum variant, which is a great idea by the way.

```
pub struct FString {
  pub item: str
}
```
Then we can carefully expose a stable api surface, which allows us to freely change the underlying types without breaking consuming Rust code.

```
impl FString {
  pub fn new(s: &str) {
...
}

   pub fn as_str() -> &str {
...
}

pub fn to_owned() -> String {
...etc
}
}
```
Not sure which of these are covered by traits but obviously those that are can be managed by traits.

As we covered before, wrapper structs are the same size, so this is zero cost (unless of course you have enums with more than 1 value, which require extra space for the enum variant).

My gut feel on this is that the Lrc wrapper should stay outside of the struct, and be treated just like any other Class or Record. I am not sure if this is feasible though since I think str's might need to be behind a pointer. If internal though perhaps we could even consider using a Box. 

Lrc is a fine stopgap though. The great thing about wrapping it is we do not have to decide now.

ncave:
If we're going to do a wrapper, we might as well do the enum, like it's done [here](https://github.com/jac3km4/dart-kernel/blob/master/src/prim.rs#L106) (and a similar one is used in Servo).
I stumbled upon those while thinking if we should use `alloc::borrow::Cow` to minimize cloning, same concept.

alexswan10k:
Agree. Sounds like a good plan.

PR #3114, opened 2022-08-28
[Rust] Fixed struct record copy

ncave:


alfonsogarciacaro:
Nice @ncave! So this and [this comment](https://github.com/fable-compiler/Fable/pull/3103#issuecomment-1229348247) means that by marking these values as immutable we active the binding beta reduction optimization? In that case, this change will indeed benefit JS as well. I find the extra copies a bit annoying. It doesn't happen so often when compiling to JS but I've seen it sometimes when using types from a shared domain which may be decorated with `Struct`.

In any case I'm surprised the copy can be optimized, I thought this was one of the operations that prevented the erasing 😅

PR #3124, opened 2022-08-31
Fix #3110: More information for imports

### LINKED ISSUE #3110: AST - not enough information in LibraryImport
In some cases I want to know if a `LibraryImport` is a module member, or a static member import.
Can this information be added before we lock down the AST?
I know it's not important in the JS case, cause everyting gets rendered as a module function, but it is important for other languages where calling conventions may be are different.
```fs
type ImportKind =
   | UserImport of isInline: bool
   | LibraryImport
   | MemberImport of isInstance: bool * fullPath: string
   | ClassImport of fullPath: string
```

alfonsogarciacaro:
This is what I came up with to fix #3110, please have a look when you have a moment and let me know what you think @ncave. As discussed, I added `MemberRef` and `EntityRef` to member and class imports but the problem with LibraryImport is `makeImportLib` is used in many places, so passing the information required to generate the member or class information would require too many changes. I've added instead a `LibraryImportInfo` and a `ImportKind.Library` helper as in other places. This way if in the future you need to add extra info to LibraryImport it, hopefully, won't break other code.

Right now what you can do is add an extra `makeImportLib` helper for Rust accepting an `isInstance` argument. Would that work for you?

### DIFF
--- a/src/Fable.Transforms/Python/Replacements.fs
+++ b/src/Fable.Transforms/Python/Replacements.fs
@@ -2231,7 +2231,7 @@ let sets (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: Exp
 
         let args = injectArg com ctx r "Set" mangledName i.GenericArgs args
 
-        Helper.LibCall(com, "set", mangledName, t, args, i.SignatureArgTypes, isInstance=not isStatic, ?thisArg = thisArg, ?loc = r)
+        Helper.LibCall(com, "set", mangledName, t, args, i.SignatureArgTypes, ?thisArg = thisArg, ?loc = r)
         |> Some
 
 let setModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Expr option) (args: Expr list) =
@@ -2255,7 +2255,7 @@ let maps (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: Exp
 
         let args = injectArg com ctx r "Map" mangledName i.GenericArgs args
 
-        Helper.LibCall(com, "map", mangledName, t, args, i.SignatureArgTypes, isInstance=not isStatic, ?thisArg = thisArg, ?loc = r)
+        Helper.LibCall(com, "map", mangledName, t, args, i.SignatureArgTypes, ?thisArg = thisArg, ?loc = r)
         |> Some
 
 let mapModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Expr option) (args: Expr list) =
--- a/src/Fable.Transforms/Python/Replacements.fs
+++ b/src/Fable.Transforms/Python/Replacements.fs
@@ -2231,7 +2231,7 @@ let sets (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: Exp
 
         let args = injectArg com ctx r "Set" mangledName i.GenericArgs args
 
-        Helper.LibCall(com, "set", mangledName, t, args, i.SignatureArgTypes, ?thisArg = thisArg, ?loc = r)
+        Helper.LibCall(com, "set", mangledName, t, args, i.SignatureArgTypes, isInstance=not isStatic, ?thisArg = thisArg, ?loc = r)
         |> Some
 
 let setModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Expr option) (args: Expr list) =
@@ -2255,7 +2255,7 @@ let maps (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: Exp
 
         let args = injectArg com ctx r "Map" mangledName i.GenericArgs args
 
-        Helper.LibCall(com, "map", mangledName, t, args, i.SignatureArgTypes, ?thisArg = thisArg, ?loc = r)
+        Helper.LibCall(com, "map", mangledName, t, args, i.SignatureArgTypes, isInstance=not isStatic, ?thisArg = thisArg, ?loc = r)
         |> Some
 
 let mapModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Expr option) (args: Expr list) =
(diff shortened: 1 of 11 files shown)


ncave:
@alfonsogarciacaro I guess what I really need is to be able to distinguish between static, instance and module members.
Do you think adding one more property can help with that, or do we need an (optional) MemberRef?

alfonsogarciacaro:
If you don't have the member ref you need a [lot of info to generate it]() and the module vs type info is not in the member itself but in the entity. It's also possible to have class imports from the library so we would have to add also `EntityRef option` for completeness. So I'd say that if you only need to distinguish between static, instance and module members, adding this to `LibraryImportInfo` will minimize the necessary code changes. And if you do need MemberRef or EntityRef, you can just create a member/class import. Would it be important in this case to know they're coming from the library (you can already set the library path in the import path)?

ncave:
@alfonsogarciacaro I guess you're right, instance imports are already created as `MemberImport`, so perhaps it's just the name of the property, it should be named either `IsStatic` or `IsModuleMember`, so it can determine whether it's a class member or module member.

alfonsogarciacaro:
@ncave Do you mean `LibraryImportInfo`? Would this work for you?

```fsharp
type LibraryImportInfo = {
    // For class members
    IsInstance: bool
    // For module members (invalidates IsInstance property
    IsModuleMember: bool
    // We can add an extra field for type references if necessary (invalidates other properties)
    IsEntity: bool
}
```

We can make it a union instead but maybe it's too much nesting: Import(ImportKind(LibraryImportKind)).

ncave:
@alfonsogarciacaro What I meant was, `LibraryImport`s are never instance members, they are either module members or static class members. So perhaps just one property should suffice, say `IsModuleMember`

alfonsogarciacaro:
It's actually possible to import instance members (unless we're understanding here instance members as _attached_ instance members which cannot be imported). This is mostly done automatically now. For example, with `Map` replacements we can import both instance and static members of `Map` type: https://github.com/fable-compiler/Fable/blob/51bf16a9d9a190b25991bdf7ed7859832ee994b3/src/Fable.Transforms/Replacements.fs#L1692-L1699

Right now, this information [is lost](https://github.com/fable-compiler/Fable/blob/51bf16a9d9a190b25991bdf7ed7859832ee994b3/src/Fable.Transforms/Replacements.Util.fs#L38) but if we add it to the AST, we can pass it when needed. An alternative would be to use a MemberImport instead, but this would create inconsistencies with other library imports, and we probably still need to indicate this is a library import (e.g., Python may want to process library imports differently at a later stage).

ncave:
@alfonsogarciacaro Ok, then what your last proposed version makes sense. We can just add some optional parameters to the `LibCall` helper for `isInstance` and `isModuleMember`. I don't think we need the `IsEntity` right now.

src/Fable.Transforms/Python/Replacements.fs
```diff
@@ -2231,7 +2231,7 @@ let sets (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: Exp
 
         let args = injectArg com ctx r "Set" mangledName i.GenericArgs args
 
-        Helper.LibCall(com, "set", mangledName, t, args, i.SignatureArgTypes, ?thisArg = thisArg, ?loc = r)
+        Helper.LibCall(com, "set", mangledName, t, args, i.SignatureArgTypes, isInstance=not isStatic, ?thisArg = thisArg, ?loc = r)
```
ncave, line 2234: @alfonsogarciacaro I stand corrected, perhaps one of optional `LibCall` arguments `isInstance` is not necessary, if we're also passing `?thisArg`, we can derive it in the helper body, no?

src/Fable.Transforms/Replacements.Util.fs
```diff
@@ -33,14 +33,17 @@ type Helper =
         makeImportLib com returnType coreMember coreModule
 
     static member LibCall(com, coreModule: string, coreMember: string, returnType: Type, args: Expr list,
-                           ?argTypes: Type list, ?genArgs, ?thisArg: Expr, ?hasSpread: bool, ?isConstructor: bool, ?loc: SourceLocation) =
+                           ?argTypes: Type list, ?genArgs, ?thisArg: Expr, ?hasSpread: bool,
+                           ?isInstance, ?isConstructor: bool, ?loc: SourceLocation) =
```
ncave, line 37: Perhaps we just need isModuleMember optional param.

alfonsogarciacaro:
Perfect, thanks @ncave! Please review latest changes. If this is OK with you we can merge and then I'll release Fable.AST v4 (no prerelease) so we can lock it and update the Feliz plugins 👍

src/Fable.Transforms/Python/Replacements.fs
alfonsogarciacaro: Ah, you're right 😅 Thanks!

alfonsogarciacaro (commented):
Thanks for reviewing @ncave! I will add your suggestions. Passing `isInstance` for maps/sets was just meant to be an example of how we could pass the info, but it's great you already spotted the changes for improvement ❤️

ncave:
@alfonsogarciacaro Can we also perhaps rename `LibraryImportInfo.IsInstance` to `IsInstanceMember` (cause that's what the property is called in `memberRef`, right?)

alfonsogarciacaro:
@ncave Actually in Fable AST we use `IsInstance` everywhere although in F# AST is `IsInstanceMember` 😅 I don't mind using one or the other here.

ncave:
@alfonsogarciacaro `IsInstanceMember` is much more clearer, imo.

alfonsogarciacaro:
Ok, let's change it 👍

BTW, @ncave, do you expect this information to be needed in user imports as well? Just in case we need to add parameters to the `Import` attribute (like we have `isStatement` for Emit) if Rust users need to import something as a module/static/instance member.

> Or maybe this information can be encoded in the selector; for example, for JS we allow accessing fields in the selector (`[<Import("foo.bar", from=...>]`) even if this is not standard syntax for JS imports.

ncave:
@alfonsogarciacaro I don't know, not at the moment. Anyway, it looks good now.

PR #3128, opened 2022-09-01
[Rust] Added StringBuilder and tests

ncave:


ncave:
@alexswan10k With this PR I've made some strides on support for BCL implementations in `fable-library-rust`.
It works, but we're hitting an issue that you cannot have the same module name in two different files.
It's not a problem if we put all of those implementations in one file and import from there (and they're not that many anyway), so that may work for now, just inconvenient. I wonder if there is a way that Rust can merge the contents of two modules with the same name, or is there a better way?

I initially went with a separate wrapping module, but that interferes with proper object names and overrides, so it's better to keep the BCL module structure if possible.

alexswan10k:
The module collision problem is indeed a tricky one. I briefly hit this myself too when trying to use common namespace patterns in a test project
```
open Core
open Core.Abc
open Core.Def.G
...etc
```

My first thoughts were barrel files too, but this creates problems?

Other than that, I wonder if we should consider treating namespaces and modules differently as Rust does not have the concept of a namespace.

What if they were merged, would that get you out of jail here?

```
module System {
    enum Enum {...}
}

module System_Text {
   ...
}

module System_Text_RegularExpressions {

}
```
etc

alexswan10k:
### Barrel files

Now that I have worked through an example, probably not.. but for reference I was thinking something like this
```
//system.rs
module System1 {
 pub something() {
}
}
//system.text.rs
module System2 {
  use System1::*;
  pub fn something() {

}
}

//system.io.fs
module System3 {
  use System1::*;
  use System2::*;
  pub fn somethingElseNew() {
  }
}
//barrel.rs
module System {
  pub use system1::*;
  pub use system2::*;
  pub use system3::*;
}
```

### Namespace merging

Interesting. Not entirely sure I follow your example, but the short of it is that you flatten out all of the namespaces and just keep the leaf modules.

You end up with stuff like this
```
/// system.io.text.rs
pub mod System_IO_Text {
  ... implementation
}

```

Ok I see the problem here, static methods can exist in namespaces. Damn. Let me go ponder on this :)

PR #3130, opened 2022-09-02
Rust - another problematic ref counting scenario + fix

alexswan10k:
@ncave  wouldn't mind your thoughts on this if you are happy to take a peek...

So I found a particularly gnarly bug when experimenting with v24, where counts were yet again not right - yes, a problem of my own creation, I know :)

I added in some extra hooks etc to track what is going on, but it seems that this might be due to the compiler generated ident ```matchValue```. Unfortunately my knowledge here is super limited of how this is generated, but it appears that some of the idents in the subtree are using the original name, and some are using the rewritten ```matchValue```, so it makes sense that they do not line up if you are just counting ```matchValue``` usages in isolation. Somehow these named idents are rewritten to also become matchValue calls in the output Rust AST, but I cannot work out where this is happening.

Perhaps the real problem here is actually that in F# it is not defined that two different idents are actually pointing to the same thing, and thus should perhaps be tracked together rather than as 2 independent idents?

Anyway, I went for a copout option of "we can't reason about matchValue, so 9999". It fixes the problematic scenario (which I have replicated in test in it's smallest form), but it kind of sweeps the real problem under the rug i think.

Any input is appreciated as always, although I do completely accept this is quite hard to follow, and the devil is in the detail.

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -80,100 +80,115 @@ module Helpers =
 module UsageTracking =
 
     type ConsumptionType =
-        | ConsByVal of string
-        | ConsByRef of string
+        {
+            Name: string
+            ByRef: bool
+            //Path: string list // for debugging purposes only
+        }
     let calcIdentConsumption (body: Fable.Expr) =
         // todo - handle shadowing of idents
-        let rec loop decTreeTargets consumingRef = function
+        let rec loop pathRev decTreeTargets consumingRef expr =
+            let mkUsage name= { Name = name
+                                ByRef = consumingRef
+                                //Path = pathRev |> List.rev //for debugging purposes only
+                                }
+            let loop pathComp =
+                loop []
+                //loop (pathComp::pathRev) // debugging purposes only
+            match expr with
             | Fable.IdentExpr id ->
-                if consumingRef then [ConsByRef id.Name] else [ConsByVal id.Name]
+                [mkUsage id.Name]
             | Fable.Sequential exprs ->
-                exprs |> List.collect (loop decTreeTargets consumingRef)
-            | Fable.Let(_, value, body) -> loop decTreeTargets false value @ loop decTreeTargets false body
+                exprs |> List.collect (loop "seq" decTreeTargets consumingRef)
+            | Fable.Let(_, value, body) -> loop "let" decTreeTargets false value @ loop "let" decTreeTargets false body
             | Fable.LetRec(bindings, body) ->
                 let bindingUsages =
                     bindings
                     |> List.map snd
-                    |> List.collect (loop decTreeTargets false)
-                bindingUsages @ loop decTreeTargets false body
+                    |> List.collect (loop "letrec" decTreeTargets false)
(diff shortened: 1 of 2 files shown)


ncave:
@alexswan10k Not sure, perhaps that's ok, we already make an exception for `matchValue` in the closure [cloning](https://github.com/fable-compiler/Fable/blob/main/src/Fable.Transforms/Rust/Fable2Rust.fs#L3030).

alexswan10k:
Agree. Happy to revisit when more issues become apparent. I am sure we will eventually get to a reasonable solution, it will likely just take some time. The basic ref counting approach seems to work well for simple cases, but there do still seem to be edge cases where the output transform is doing something complicated, and presumably it needs to be mirrored in the ref counting pass.

ncave:
@alexswan10k On a slightly different note, the whole reason why we do the `calcIdentUsage` is to avoid cloning in certain cases (if only seen once). Would it be so bad if we always clone? We're doing it anyway if the ident is shared. This way we wouldn't have to worry about corner edge cases no more. Then perhaps as a next step we can try to avoid some cloning using `alloc::borrow::Cow` as return type, if possible. Or not even that, it's just an optimization that may not be worth the added complexity, if our clones are supposedly cheap.

I know you've put a LOT of work in this, so we can keep it around. I've only commented it in #3131 as a proposal to see what you think, and we can bring it back if needed.

alfonsogarciacaro:
Sorry @alexswan10k, this may be my fault 😅 IIRC the F# generates a matchValue tuple when matching against multiple values. In JS this is usually unnecessary because the patterns just destructure the tuple again, so in bindingBetaReduction optimization I just try to access the underlying values directly.

I need to check the code, but it's possible the optimization sometimes only does half work and that's why you see sometimes values accessed directly and sometimes through matchValue. If you want, we can disable this optimization for Rust.

alexswan10k:
@alfonsogarciacaro this is an interesting find. It might be worth seeing what happens if we turn this off.

@ncave I appreciate the frustration on complexity, but I do feel this is necessary for longer-term acceptance. I would rather we had an escape hatch than throw out a solution that is mostly working.

Clones might be relatively cheap, but they are not free, and they can add up. It depends on what you are cloning - mostly it will be a smart pointer, but this could just as well be a struct of unknowably large size, or some other non-trivial deep clone indirection. Our rules apply to anything cloneable, not just smart pointers. To that point, there is a reason Rust distinguishes between a Copy and a Clone, otherwise we could achieve the same effect by just implementing Copy opaquely as Clone under the hood, and avoid all this fuss. This is generally frowned upon in the Rust community for a good reason as it is sub-optimal and can lead to potentially unpredictable outcomes. Better to be explicit.

With respect to function calls, understanding that a type is only used in a single ownership scenario, and thus transfers its ownership also opens up a potential Rust compiler optimization too, where sometimes moves or copy's can be entirely prevented across stack frames.

ncave:
@alexswan10k
> Our rules apply to anything cloneable, not just smart pointers. 

Yes, but all that is still in place, I only suggested removing the "used more than once" rule, which could be brittle with edge cases, that's all. Everything else is still in place.

Anyway, it was just a suggestion, to get your opinion. I'll switch the PR to a draft.

PR #3131, opened 2022-09-02
[Rust] Removed usage counting

ncave:
- Removed usage counting to avoid edge cases.

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -35,7 +35,7 @@ type ScopedVarAttrs = {
     IsRef: bool
     IsBox: bool
     // HasMultipleUses: bool
-    mutable UsageCount: int
+    // mutable UsageCount: int
 }
 
 type Context = {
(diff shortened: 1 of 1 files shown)


ncave:
@alexswan10k Just a proposal, let me know what you think.

ncave:
@alexswan10k Based on your feedback I will not be merging this PR, so I've converted it to a draft (to keep it around just in case we encounter an edge case that is hard to solve).

Just FYI, I've tested the performance of the additional cloning in hot loops and there is very little (if any) difference in performance, at least single-threaded, so that doesn't seem to be an issue.

alexswan10k:
Let's see if the edge cases overwhelm us. Genuinely happy to revisit and concede if so. Apologies for coming across as defensive about this, I just genuinely believe it will lead to a better solution, irrespective of my ego aha. 

Good news on the benchmarks, that is reassuring.

Not sure I understand how Cow will help beyond strings (which it is definitely a great fit for btw). Rc's are pretty good at not making unnecessary allocations, although real Gc's are uniquely efficient at batching up deallocations of course (we still might get to this mind you). Perhaps you have a specific example in mind?

ncave:
@alexswan10k 
> Perhaps you have a specific example in mind?

I don't, Cow's main purpose is to delay/avoid cloning, not allocations. And in our case clones are cheap anyway, cause every reference type is a smart pointer, so maybe there is no point (except for strings). Anyway, you're right about the GC, but it's encouraging that the raytracer demo manages to outperform managed .NET's GC despite running millions of allocations per second.

ncave:
@alexswan10k Turns out all those allocations were unnecessary (see #3133), so we've reclaimed some performance.
We're back to Rust native binary clearly outperforming managed .NET, and WebAssembly within a striking distance.
(at least for this demo which is numerically heavy, but still, I'd call that a progress).

alexswan10k:
Awesome progress. Another thing to beat in mind is that the biggest reason to not use managed .NET WASM is bloat, which rust has an edge if you turn on all the optimizations. Nice work!

ncave:
Just a note to myself that `cargo clippy --fix` can also eliminate some (most?) of unnecessary `.clone()`, so it's a good post-build step (along with `cargo fmt` for better formatting).

ncave:
Superseded by #3467

PR #3140, opened 2022-09-05
Rust - better datetime comparison + conversion. DateTimeOffset first …

alexswan10k:
More datetime progress.


alexswan10k:
Oh, @ncave, small discovery you might be interested in:

Turns out you can use the access modifier
```
pub(create) fn ... {}
```
which is basically the same as Internal. This was quite nice for DateTime as it allows us to pass around Chrono types under the hood between different internal modules, without exposing any Chrono api's to an end user. Chrono thus remains an implementation detail, allowing us to still change our minds on using it without breaking users.

PR #3142, opened 2022-09-06
Added `IsInternal` and `IsPrivate` properties in AST

ncave:
- Some fixes for constructors and inlined class members.
- Added `IsInternal` and `IsPrivate` properties in AST for entities and members.
- Vendored (copied locally) the `startup` crate as it was yanked from `crates.io`.

1) @alfonsogarciacaro
I know that the Fable AST v4 was already published, but I needed to add `IsInternal` and `IsPrivate` to handle visibility.
These are pure additions, so hopefully they wouldn't break any plugins, but let me know if you think there is an issue.
I would very much like to also rename `IsInstance` and `IsExtension` to `IsInstanceMember` and `IsExtensionMember`, so they would match the underlying property names (and the original names are much clearer), but decided to ask you first.

2) @alexswan10k
The `startup` crate was probably yanked because it makes it [way too easy](https://kerkour.com/rust-crate-backdoor#/run-code-before-main) to abuse execution before main (although yanking it can stop only the laziest of perps, so it's probably just about avoiding potential problems). In theory it's not any different than how `module do bindings` and `static initialization` work in .NET. It's all about how do you audit your dependency chain to make sure a dependency is not compromised.

    Still, we may have to replace it with a different way to support F# module [do bindings](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/functions/do-bindings), since Rust [frowns upon](https://doc.rust-lang.org/1.6.0/complement-design-faq.html#there-is-no-life-before-or-after-main-no-static-ctorsdtors) execution before main. Or perhaps not even support `module do bindings` and print a warning instead.


alfonsogarciacaro:
@ncave If the React tests pass it means the Feliz plugins are working so it should probably be fine :)

About renaming IsInstance and IsExtension, given that we've already published the "stable" AST, I would prefer to leave changes like this for a future release. Maybe we can keep an issue open to track potential improvements to the AST and implement them in bulk in a major (or minor if they're not breaking) release.

alexswan10k:
Makes sense. Maybe this ie another one for an opt in feature so it is off by default and leads to a build error like unsafe cells. Use at your own risk.

alfonsogarciacaro:
Just in case JS experience helps, when building JS apps support for "module do bindings" is also flaky, in the sense they generally work, but they can be lost if nothing in the module they're in is imported by another module in the bundle. Some users wanted to automatically import all files in the project to emulate static initialization in .NET, but I preferred not to do it to avoid problems with JS tree shaking tools.

Following Fable's philosophy of playing along the rules of the host ecosystem/runtime, it makes also sense here to raise a warning and discard execution before main (although this can be trickier for static constructors).

ncave:
Thank you @alfonsogarciacaro , added a warning in #3144.

PR #3152, opened 2022-09-09
[Rust] Updated module visibility

ncave:
- Updated module and member visibility.
- Enabled some operator tests.

For some reason F# member visibility properties are not independently set, they seem to be affected by their parent's visibility.

@alfonsogarciacaro I had to add another missing property to the AST to handle visibility, sorry about that.
But it's a new addition so it shouldn't affect existing plugins. The new property is `Entity.DeclaringEntity`.

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -127,6 +127,7 @@ type Entity =
     abstract DisplayName: string
     abstract CompiledName: string
     abstract FullName: string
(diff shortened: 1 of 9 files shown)


alfonsogarciacaro:
Yes, I think I reported this in the fsharp repo but I cannot find the issue now. For JS it doesn't matter because modules are flattened (so all members of a private module become private) but I guess it's a problem in Rust if you're maintaining nested modules.

PR #3156, opened 2022-09-09
Rust - some progress on timespans

alexswan10k:
* Timespan constructors
* Timespan static create
* DateTime add_ticks
* Refactor - move timespan + datetimeoffset to new .rs files
* Timespan does not depend on Chrono so I have removed it from the feature switch - It can always be available perhaps?

_Timespan.Zero is still rendering wrong even though I have redirected it to supposedly call TimeSpan.zero(). Not sure why this would be ?_


ncave:
> Timespan.Zero is still rendering wrong

I'll take a look.

ncave:
@alexswan10k Fixed in #3158.

PR #3169, opened 2022-09-13
Initial TS code cleanup

ncave:
- Initial TypeScript code cleanup to fix some issues: import extensions, printer, generic calls, etc.


ncave:
TypeScript support hasn't been touched in a while, so it's still in alpha and needs some love.
First we need to make `fable-library-ts` compile on TypeScript (i.e. `npm run build library-ts`).
Then we can continue with making tests compile on TypeScript, but that's a much bigger task.

PR #3173, opened 2022-09-15
[TypeScript] Fixed some array issues

ncave:
- Fixed some array annotation issues.
- Fixed inject replacements for some array funcs.
- Fixed `copyTo` to avoid exceptions (which are slow).


ncave:
@dbrattli I had to move the `Array` type alias from `types` to `util` to avoid circular dependencies, I hope that's ok.

PR #3174, opened 2022-09-15
Date fixes: prevent offset recalculation + wrong padding

stroborobo:
Hey Fable team,

I noticed some formatting issues with years before 1000 AD and changes in the time with years before 100 AD.

Turns out the set[UTC]FullYear() method, that's used for years 0-99, causes changes in historic time zones to be applied to the existing date. Javascript's handling of the year in the Date() constructor and Date.UTC() makes it fundamentally incompatible for these years to the API you'd expect from dotnet's date types. So since we're already using the iso string interface in some cases, I thought maybe it would be the most straight forward solution to just use it always, no mutation of the date object required.

Also I've added padding for custom date formatting according to the length of the specifier, which appears to be in line with the dotnet spec.

Please let me know what you think, I can change it of course if you don't agree.

Have a nice day :)

### DIFF
--- a/src/fable-library/Date.ts
+++ b/src/fable-library/Date.ts
@@ -263,13 +263,9 @@ export function create(
   year: number, month: number, day: number,
   h: number = 0, m: number = 0, s: number = 0,
   ms: number = 0, kind?: DateKind) {
-  const dateValue = kind === DateKind.UTC
-    ? Date.UTC(year, month - 1, day, h, m, s, ms)
-    : new Date(year, month - 1, day, h, m, s, ms).getTime();
-  if (isNaN(dateValue)) {
-    throw new Error("The parameters describe an unrepresentable Date.");
-  }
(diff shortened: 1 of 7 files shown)


stroborobo:
The more I think about it, the less sure I get. Maybe I should have stayed with the padding fixes. 😅

Starting to think the issue was just using setFullYear in some cases where it should have been setUTCFullYear. At least the test "DateTime UTC from Year 1 to 99 works" is fixed by considering that. I'll have look at it again later, maybe this fix could be much simpler.

stroborobo:
Reverted the iso string changes, using setUTCFullYear instead if it's UTC did help. Left the original commit for reference.

Please let me know if I should do it differently or squash.

alfonsogarciacaro:
Looks good @stroborobo, thanks a lot!

Actually I had this problem with years before 100AD in a project but used a patch and forgot to fix it in Fable 😅

PR #3178, opened 2022-09-17
[TypeScript] Added some interface annotations

ncave:
- Added some interface annotations.
- Added attached member return type.

Need to fix Python.


dbrattli:
Sorry, didn't know that "Fixes #" comments could close PRs as well 😬 Anyways, Python seems to be fixed now 😄

dbrattli:
@ncave Not sure it's relevant for TS or if you have already fixed it, but I had this issue in Python with interfaces where types of imported interfaces disappeared, since such interfaces are erased. Thus for code such as:

```fs
[<Import("Logger", "structlog")>]
type Logger =
    abstract msg: msg: string * [<ParamArray>] args: obj[] -> unit

type IExports =
    abstract getLogger : unit -> Logger

[<ImportAll("structlog")>]
let structlog : IExports = nativeOnly

let logger = structlog.getLogger()
``` 

Then the type `Logger` would end up being undefined (since it's erased). So I had to look it up again (in Fable2Python.fs) and import the imported type myself https://github.com/fable-compiler/Fable/pull/3175/files

ncave:
@dbrattli Thanks for fixing the build! I couldn't figure out what was the problem in `pyproject.toml`, it looked ok.

> Not sure it's relevant for TS or if you have already fixed it, but I had this issue in Python with interfaces where types of imported interfaces disappeared, since such interfaces are erased

Thanks, I'll keep it in mind.

PR #3185, opened 2022-09-20
Python Replacement Fixes

alfonsogarciacaro:
@dbrattli I need these fixes for the VS Code extension I'm working on to invoke Python interactive directly from F# code. I'm also removing some code that seems to be a remnant when porting the JS code. It should be fine but please have a look.

BTW, I need a couple of helpers to define cells and embed Python code. We can have them either in Fable.Core or in a separate library, where do you think it's best?

```fsharp
[<RequireQualifiedAccess>]
module Py

open Fable.Core.PyInterop

let inline python (template: string): unit = emitPyStatement () template
let inline NEW_CELL<'T> = emitPyStatement () "# %%"
```

The module needs to have `RequireQualifiedAccess` attribute (this is a requirement of the F# Template Highlighting extension), at the beginning I thought of `IPy` but `Py` is simpler, though overloaded, what do you think?

The code is to be used like this.

```fsharp
Py.NEW_CELL

printfn "Import matplotlib"

Py.python """
import matplotlib.pyplot as plt
plt.title("Line graph")
plt.xlabel("X axis")
plt.ylabel("Y axis")
plt.plot(x, y, color ="red")
plt.show()

"""

Py.NEW_CELL

printfn "Generate data in F#"

let x = Array.init 100 id 
let y1 = x |> Array.map (fun x -> x * x)

Py.NEW_CELL

printfn "Plotting..."

Py.python """
plt.title("Line graph")
plt.xlabel("X axis")
plt.ylabel("Y axis")
plt.plot(x, y, color ="red")
plt.show()
"""
```


dbrattli (approved):
Looks good. Thanks for cleaning up the code!

dbrattli:
Hi @alfonsogarciacaro. This is great work!! I was wondering about PyInterop vs Fable.Core.PY and how I should think about what should live where? There are a few things in Fable.Core.PY, mostly just the `DecoratorAttribute`, and some constructor stuff needed by Native.fs. But that file is listed before PyInterop.fs so it cannot depend on it. I would actually like to call this file `Fable.Core.Python` instead since we usually want spell out Python as with Dart. But anyways we cannot have both `PY` any `Py` modules in Fable.Core so not sure how we could solve this?

dbrattli:
Perhaps we could simply move the decorator into PyInterop instead since it's used for interop, and inline the array stuff in Native.fs (since we have Python specific version of this file). Then you are free to use Fable.Core.Py.

src/Fable.Core/Fable.Core.PY.fs
```diff
@@ -2,6 +2,7 @@ namespace Fable.Core
 
 open System
 
+[<RequireQualifiedAccess>]
 module PY =
```
dbrattli, line 6: Do you need to rename it to Py? or is PY fine. I actually would prefer `Py`. Then the file should also be renamed.

alfonsogarciacaro:
Thanks for checking @dbrattli! Actually, I forgot the Python bindings were in `Fable.Core.PY` module instead `Fable.Core.Python` 😅 In that case, although they're not exactly bindings, I think we can put the helpers in `Fable.Core.PY`. For that we need to add `RequireQualifiedAccess` to the module, but I think it's actually something good to do. This will prevent conflicts as we had we Fable.Core.JS, so we don't need the separate `Constructors` module. What do you think of the changes in the [last commit](f46b663a4cadff3df97ceeb3ad61f2448f14ba2e)?

src/Fable.Core/Fable.Core.PY.fs
alfonsogarciacaro: Either will work. If you prefer `Py` probably it's a good opportunity to rename the module 👍

dbrattli: Yes please, JS is JavaScript. Python should be Py

alfonsogarciacaro: Ok, [I've changed PY to Py](https://github.com/fable-compiler/Fable/pull/3185/commits/2971344f593df3fafde1a7ac32f7acffa081bbde). Doing a search & replace I also hit `Fable.PY` module within Fable code itself. Did you want to change that as well or only in Fable.Core?

dbrattli: Great. I can fix `Fable.PY` later since it's not exposed to the user. Feel free to merge 👍

PR #3190, opened 2022-09-21
[TypeScript] Fixed Set and Map annotations

ncave:
- Fixed Set and Map annotations
- Fixed constructor type params

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
 
 jobs:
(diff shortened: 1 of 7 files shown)


ncave:
With this PR the `fable-library-ts` finally compiles without errors with TypeScript.
More than a thousand type errors were squashed, but no type has been harmed.

The next step will be to gradually improve the TypeScript support by fixing tests, but that's a pretty big task, so that's probably going to take a while. Contributions are very welcome.

PR #3191, opened 2022-09-21
[Python] do not trim emitted statements

dbrattli:


alfonsogarciacaro:
@dbrattli IIRC the idea of the subsegments was trying to adjust the indentation so it fit better with the surrounding code, but if we're going to respect the indentation of the emitted code, we probably don't need to split the segment by line breaks at all. In fact, we probably shouldn't, as the printer may add extra indentation when adding a new line. So I think reducing the `printSegment` function to the following should work:

```fsharp
            let printSegment (printer: Printer) (value: string) segmentStart segmentEnd =
                let segmentLength = segmentEnd - segmentStart

                if segmentLength > 0 then
                    let segment = value.Substring(segmentStart, segmentLength)
                    printer.Print(segment)
```

PR #3197, opened 2022-09-22
Enable emitExpr/Statement with interpolation

alfonsogarciacaro:
This allows you to use interpolation when emitting native code. This PR enables the four for all four languages. Example in JS:

```fsharp
let doSomethingInJs (x: int) (y: float) (z: float): string =
    emitJsStatement () $"""
    let acc = 0;
    for (let i = 0; i < {min x 10}; i++) {{
        acc += {y ** 2} + {x + 3} * 2;
    }}
    return acc + {sprintf "%.2f" z};
    """
```

@dbrattli I've also renamed `emitPyExpr`/`emitPyStatement` to `emitExpr`/`emitStatement` to be in line with Dart and Rust (JS cannot change to avoid breaking changes, although maybe we can add the alias helpers). Please let me know if you prefer to undo the change.

### DIFF
--- a/src/Fable.Core/Fable.Core.Py.fs
+++ b/src/Fable.Core/Fable.Core.Py.fs
@@ -39,7 +39,8 @@ module Py =
     [<Emit("lambda *args: $0(args)")>]
     let argsFunc (fn: obj[] -> obj): Callable = nativeOnly
 
-    /// Embeds literal Python code into F#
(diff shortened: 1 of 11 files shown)


dbrattli:
This is great @alfonsogarciacaro. `emitExpr` / `emitStatement` are much better names. Thanks for fixing!!

Issue #2096, opened 2020-06-17
Improve TypeScript typing for unions


chrisvanderpennen:
### Description
The current generated TypeScript definitions for unions makes them difficult to correctly use or construct with TypeScript.

An example:
```fsharp
type Msg =
| ChangeValue of value: string
| Nothing
| Something of value: int * name: string
```
is compiled to:
```typescript
export class Msg extends Union {
  constructor(tag: int32, name: string, ...fields: Array<any>) {
    super(tag, name, ...fields);
  }
}
```
or a close equivalent with class types disabled.

I have been tinkering with this output and the fable-library types, and I've been able to combine a few additional definitions in fable-library with more explicit typing information on the generated  union types that enable TypeScript to safely use and construct Fable generated union types without changing the object structure.  It's not perfect, it creates an additional type declaration per union type in the compiled output and I'd like to find a way to avoid that, but I think it's good enough to share here for discussion.

The objectives I had were:
* TypeScript should be able to correctly represent the constructor signature of each union case
* TypeScript should be able to narrow a union instance to a specific case with switch/if statements
* TypeScript users should not need to know the mapping between tags and case labels
* The JavaScript objects needs to stay the same shape
* Pollute scopes as little as possible

Four additional declarations are added to fable-library Types.ts:
```typescript
// Provide an interface that supports declaring F# union cases using the form:
// {
//   0: ["label0", [types, of, fields]],
//   1: ["label1", [types, of, fields]],
// }
interface UnionCases {[tag: number]: [string, any[]]}

// Convert a union case declaration as above into the form:
// { tag: 0; label: "label0" } | { tag: 1; label: "label1" }
// This is then used in FSharpUnionType to type static members mapping case labels to their tag
type TagLabels<Cases extends UnionCases> = {[Tag in keyof Cases]: {tag: Tag; label: Cases[Tag & number][0]}}[keyof Cases]

// Given a shape type as above, produce Union<Cases, 0> | Union<Cases, 1>
export type FSharpUnionCase<Cases extends UnionCases> = {[Tag in keyof Cases]: Union<Cases, Tag & number>}[keyof Cases]

// Given a shape type as above, produce the following:
export type FSharpUnionType<Cases extends UnionCases> = {
  // A construct signature, generic on Tag, that narrows name to the string literal representing the case label and subsequent parameters to the case fields
  // It does _not_ return an instance narrowed to the specific case constructed
[...]

PR #3199, opened 2022-09-23
Add quicktest-ts

alfonsogarciacaro:
Hopefully this will allow us to quickly experiment with TS as we do with JS.


alfonsogarciacaro:
@ncave I'm getting errors because of the type annotations of local value declarations when interoping with native values. Sometimes our typings are less strict than Typescript (e.g. `string` vs `string | undefined`). Do you think we can skip the annotations for local declarations? I think I had a similar problem in Dart and removed them except in a few cases (e.g. when casting a value).

image

ncave:
@alfonsogarciacaro I fixed it in #3200, but IMO the real problem is the support for nullable reference types in F#.
I'm not sure we can do much about it until https://github.com/dotnet/fsharp/pull/6804 gets more traction.

> Do you think we can skip the annotations for local declarations?

No, that makes it worse, as some let bindings don't have enough type on the right (in the value) to properly resolve the type.
Also, it's not going to fix anything, as the (wrong) type will passed around next.

alfonsogarciacaro:
You're right @ncave. I tried to work around that for Dart by adding the `DartNullable` type which is the same as `Nullable` without the value type constraint. But in TypeScript the problem is compounded because there is both null and undefined and because we already have many existing bindings. I agree the best solution, at least for now, is what you propose and use the `??=` operator whenever TS compiler complaints. I've moved the operator to 

I still got some errors. Typescript was complaining that the type parameters in unions like Result and Choice were not actually used. Probably these could be skipped setting a flag in tsconfig.json but I decided to give it a go at [implementing this](https://github.com/fable-compiler/Fable/issues/2096#issuecomment-1251745366) and it seems to work. It complicates a bit dealing with unions in TS (particularly because the type declaration and the class/constructor are different now) but we can reach almost the same level of type-safety as F#. Please see below the code generated for `Result` (in JS it's basically the same as we have now). Though I broke a couple of tests in JS, I need to fix them 😅 

> I've tried to do some cleanup along the way, I hope it's ok.

```typescript
export const enum Result_Tag { Ok = 0, Error = 1 }
export type Result_Fields<T, TError> = [[T], [TError]];
export type Result<T, TError> = Result_Cons<T, TError, Result_Tag.Ok> | Result_Cons<T, TError, Result_Tag.Error>;

export class Result_Cons<T, TError, Tag extends keyof Result_Fields<T, TError>> extends Union {
    constructor(readonly tag: Tag, readonly fields: Result_Fields<T, TError>[Tag]) {
        super();
    }
    cases() {
        return ["Ok", "Error"];
    }
}

export function Result_Map<$a, $b, $c>(mapping: (arg0: $a) => $b, result: Result<$a, $c>): Result<$b, $c> {
    if (result.tag === Result_Tag.Ok) {
        return new Result_Cons<$b, $c, Result_Tag.Ok>(Result_Tag.Ok, [mapping(result.fields[0])]);
    }
    else {
        return new Result_Cons<$b, $c, Result_Tag.Error>(Result_Tag.Error, [result.fields[0]]);
    }
}
```

ncave:
@alfonsogarciacaro 
> Typescript was complaining that the type parameters in unions like Result and Choice were not actually used.

Yes, that's a problem for Rust as well, in theory the missing generic param type should needs to be replaced with `Unit`, but I still haven't figured out how to do that. Unfortunately looks like the F# compiler doesn't do that for us.

alfonsogarciacaro:
@ncave Sorry, actually I meant an issue that was general for all Unions. For example, before this PR, Choice was compiled as:

```typescript
export class FSharpResult$2<T, TError> extends Union {
    tag: int32;
    fields: any[];
    constructor(tag: int32, ...fields: any[]) {
        super();
        this.tag = (tag | 0);
        this.fields = fields;
    }
    cases() {
        return ["Ok", "Error"];
    }
}
```

Because `fields` is typed as `any[]` any generic parameter (`T` and `TError` here) remains always unused. Hopefully this is fixed by this PR, and we also get the proper field types in the `switch` branches when matching against a union tag.

PR #3202, opened 2022-09-26
Updated FCS-Fable to latest

ncave:
- Updated `FCS-Fable` to latest
- Updated CLI target framework to `net6.0`


ncave:
@alfonsogarciacaro I went ahead and updated the CLI target frameworks from `net5.0` to `net6.0`, since .NET 5.0 is no longer supported since 2022-05-10. I hope that's ok.

alfonsogarciacaro:
Yes, I was also wondering if we should do that. I think it makes sense for Fable 4. I will update the minimum requirements in the README :+1:

PR #3209, opened 2022-09-28
Use StringTemplate expr in Fable AST for Python

alfonsogarciacaro:
@dbrattli I just realized #3197 wasn't working with Python (I should have tested it 😅). After debugging I found out this was because `StringTemplate` was not being created for Python. So I have updated Python/Replacements to generate StringTemplate when possible as in JS (except for FormattableString) and in Fable2Python I'm just turning it into string concatenation.

Can you please check? Thanks!


dbrattli (approved):
This is great! Thanks for fixing!

PR #3230, opened 2022-10-12
[Python] Use eta-conversion to reduce arrow functions

dbrattli:
- This transform is made in the Babel printer, but we use it in the Fable2Python transformation step. Perhaps we will miss some cases?

### DIFF
--- a/src/Fable.Transforms/Python/Fable2Python.fs
+++ b/src/Fable.Transforms/Python/Fable2Python.fs
@@ -910,8 +910,9 @@ module Annotation =
             let resolved, stmts = resolveGenerics com ctx genArgs repeatedGenerics
             fableModuleAnnotation com ctx "mailbox_processor" "MailboxProcessor" resolved, stmts
         | "Fable.Core.Py.Callable", _ ->
-            let genArgs = [ Expression.ellipsis; Expression.any]
-            stdlibModuleAnnotation com ctx "typing" "Callable" genArgs, []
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Don't remember exactly why I added the optimization to the printer, but indeed it's probably because I couldn't detect all cases in Fable2Babel. Ideally, the printer shouldn't do optimizations so I think it's a good idea to do this in Fable2Python. If you detect some cases are missing you can revisit later if you can adapt Fable2Python or move the optimization.

dbrattli:
Thanks @alfonsogarciacaro. Discovered this when testing your simplified decorator since I got an extra arrow function. 

```py
def measure_performance(fn: Callable) -> Callable:
    return fn


def _arrow2(fn: Callable) -> Callable:
    return measure_performance(fn)


@_arrow2

def my_expensive_calculation(x: int, y: int) -> int:
    return x + y
``` 

This PR fixes that:


```py
def measure_performance(fn: Callable[..., Any]) -> Callable[..., Any]:
    return fn


@measure_performance

def my_expensive_calculation(x: int, y: int) -> int:
    return x + y
```

alfonsogarciacaro:
Yes, unfortunately the F# compiler does this when you use a method reference as a value, which makes sense in .NET but it's not necessary in languages like Python or JS. IIRC I did try once to remove this in FableTransforms but got issues with some edge cases and moved the optimization to the end of the pipeline.

Now that I think of it, it makes sense to do this in Fable2Python to avoid the need of declaring the extra arrow function. Glad that you managed to solve it!

PR #3233, opened 2022-10-12
[Python] Fix reference equal with literals

dbrattli:
@alfonsogarciacaro do you think it would it be possible to represent reference equals in the Fable AST? Current I need to handle it using Emit, and then do transformations on emit which is not very nice. The reason is that for Python a reference equals is written as `is` (except with literals that needs to use `==`) while equals is `==`. So for e.g `a == b` or `a is b` I have no way of knowing if it's one or the other when using `BinaryEqual`.

### DIFF
--- a/src/Fable.Transforms/Python/Fable2Python.fs
+++ b/src/Fable.Transforms/Python/Fable2Python.fs
@@ -1333,6 +1333,11 @@ module Util =
             match txt with
             | "$0.join('')" -> "''.join($0)"
             | "throw $0" -> "raise $0"
+            | "$0 is $1" ->
+                match args with
+                | [Constant(_); _]
+                | [_; Constant(_)] -> "$0 == $1"
+                | _ -> "$0 is $1"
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
@dbrattli In the AST we have BinaryEqual and BinaryEqualStrict, would that work for you?

dbrattli:
@alfonsogarciacaro That would work, but I cannot find it 😬 Has it been added?

alfonsogarciacaro:
@dbrattli Sorry, my mistake! I just remembered I removed BinaryEqualStrict when refactoring the AST 😅 It'd be nice to solve this with tags as they're meant to be used with language-specific features that don't need to be handled by other languages. But again, I forgot to add tags to the `Operation` expression 🤦 Would it work for you if the add the tags to `Operation`?

dbrattli:
Yes, @alfonsogarciacaro. I was initially looking for something like tags, so I could give a hint if it was strict or not.

PR #3234, opened 2022-10-13
Add tags to Operation expression

alfonsogarciacaro:
This way we can add specific info for each language. For example, if equality is strict in JS or by reference in Python.


dbrattli:
Very nice!

PR #3250, opened 2022-11-04
[Rust] Added Func type wrappers

ncave:
- Added Func type wrappers

### DIFF
--- a/src/Fable.Transforms/Rust/Fable2Rust.fs
+++ b/src/Fable.Transforms/Rust/Fable2Rust.fs
@@ -34,6 +34,7 @@ type ScopedVarAttrs = {
     IsArm: bool
     IsRef: bool
     IsBox: bool
+    IsFunc: bool
     // HasMultipleUses: bool
     mutable UsageCount: int
 }
@@ -49,7 +50,7 @@ type Context = {
     ScopedSymbols: FSharp.Collections.Map<string, ScopedVarAttrs>
     IsClassMember: bool
     IsInPluralizedExpr: bool //this could be a closure in a map, or or a for loop. The point is anything leaving the scope cannot be assumed to be the only reference
-    IsParamAnyType: bool
+    IsParameterType: bool
     IsParamByRefPreferred: bool
     RequiresSendSync: bool // a way to implicitly propagate Arc's down the hierarchy when it is not possible to explicitly tag
     ModuleDepth: int
@@ -68,8 +69,8 @@ type IRustCompiler =
 // TODO: Centralise and find a home for this
 module Helpers =
     module Map =
-        let except excludedKeys source =
-            source |> Map.filter (fun key _v -> not (excludedKeys |> Set.contains key))
+        let except excluded source =
+            source |> Map.filter (fun key _v -> not (excluded |> Map.containsKey key))
         let merge a b =
             (a, b) ||> Map.fold (fun acc key t -> acc |> Map.add key t)
         let mergeAndAggregate aggregateFn a b =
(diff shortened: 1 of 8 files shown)


ncave:
@alexswan10k On a side note, I have spent an exorbitant amount of time trying to ~hammer a square peg into a round hole~ make unboxed closures work to lower allocations, but I'm afraid it's not possible if we want the func types to be uniform to some extent. It doesn't look like we can define unboxed function return types (e.g. it looks like curried return types `a -> b  -> c -> d` can only be expressed as boxed closures in Rust).

I finally gave up and left them boxed (Lrc).

alexswan10k:
Yeah it was my understanding that any capturing closure has to be represented as an implicit struct somewhere, which means it needs a clear owner. To be able to pass them around (multiple owners) you basically are forced to box them in an Lrc. The exception is of course non-capturing closures, but I am not if we can abstract over both, but I guess impl on parameters helps a little here too (monomorphization).

Btw sorry I have not been much use recently, had quite a bit of other stuff going on. I will try and get back at some point though but am always around to bounce ideas off etc. Good to see things are still progressing though!

ncave:
@alexswan10k 
> I am always around to bounce ideas off

Thanks, I really appreciate that!

I take it that you don't have objections to the introducing of func wrapper types?

The only downside really is the doubling of function ptr size from 8 to 16 bytes because of the enum type taking another word, but the benefits are the same as with other wrapper types, we can implement traits for them etc. It also saves allocations on non-capturing lambdas.

We can technically have a feature to switch between enum and non-enum func wrapper type, but (unlike strings) function pointers are not usually stored around in large quantities, so the size only matters for passing into other functions.
What do you think?

alexswan10k:
Presumably it is 16 bytes due to alignment? 

Im honestly not sure really. What you have done makes sense, but I do agree that it is nice to have control to turn it on and off (although this adds complexity too, tradeoffs). Interestingly you could sort of apply this argument to any Rc derivatives, where a single owner use case could avoid the heap completely, although at the cost of extra memory and tag checks.

More than happy to go with this for now though.

ncave:
@alexswan10k Right, that's what `Cow` is supposed to do, keep it borrowed and postpone allocation until somebody needs to own it, but it does it with lifetimes, and I don't feel ready to introduce lifetimes in the generated code ;)

alexswan10k:
Yeah sort of. As you pointed out, Cow borrows. What I had in mind looks more like your function wrapper (always owned rather than borrowed)

I agree though that direct ownership is undoubtedly simpler to reason about, and I also think it is far closer to how .NET reference types can generally be thought about conceptually. From what I understand everything is passed by value (even if its reference by value) unless it is explicitly passed by reference (which we now correctly model too with inref and byref). I feel like this stuff is in a good place.

alexswan10k:
@ncave I imagine you saw it, but if not, [1.65 now supports GAT's](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html). This is kind of a big deal because it allows you to truly abstract over pointer types, among other things such as interfaces (iter used as an example).

[Here is a pointer example](https://www.reddit.com/r/rust/comments/ynvm8a/comment/ivayo42/?utm_source=share&utm_medium=web2x&context=3).

I am a little annoyed with myself I did not spot chat on this sooner (in unstable), as this could have simplified quite a few things, but I guess at least it is available to use now.

ncave:
@alexswan10k 
> I imagine you saw it, but if not, 1.65 now supports GATs.

I certainly did, it's a good tool to have in our disposal. I'm just not sure how much it can help for our current setup to switch between Rc/Arc with a feature. If you had something else in mind, please elaborate.

alexswan10k:
I agree the switch is probably still necessary from the point of view of object creation. I guess my thoughts were more that any of the built in .NET api's that accept a pointer could be rewritten to be entirely pointer agnostic. It may lead to generic param soup though, so it may be more hassle than it is worth. Just a vague idea right now, but I will let you know if I think of anything actionable!

PR #3252, opened 2022-11-05
Import dynamic call

alfonsogarciacaro:
This was supposed to be a small PR. The main purpose is to enable also calls with `Fable.Core.JsInterop.importValueDynamic`. So for example the following code:

```fsharp
open Util

importValueDynamic(doSomething(foo, bar))
```

becomes:

```js
import("./Util.js").then(m => m.doSomething(foo, bar))
```

But at the end I had to do some refactoring to enable this behavior and I also touched other languages. I took the chance to unify some of Fable.Core replacements and removed code that only worked for JS but was still present in the Python/Rust/Dart replacements. For Rust I had to leave `op_BangHat` to make the tests pass.

One important change of this PR is that functions like `importValueDynamic` can "capture" bindings from the scope. Many times these functions are not used directly but from another inline function in a library. Example:

```fsharp
type Sutil with
        static member inline lazyImport(call: SutilElement, ?onWaiting: SutilElement): SutilElement =
            Sutil.bindPromise(importValueDynamic call, id, ?onWaiting=onWaiting)
```

So far, I relied on beta reduction in FableTransforms to remove the binding after the capture, but this doesn't remove bindings when the value can have side effects. So the above example would be translated to something like the following in the call site:

```js
const call = doSomething(foo, bar)
return Sutil_bindPromise_60D7386A(import("./Component.js").then((m) => m.doSomething(foo, bar)), (x_2) => x_2, void 0);
```

This causes double evaluation and also defeats the purpose of the lazy import because `doSomething` is imported statically. This PR tries to [deal with these cases specifically](https://github.com/fable-compiler/Fable/blob/da82469ee3949cd0c8fb5615e98806cb2c997055/src/Fable.Transforms/FSharp2Fable.fs#L1846-L1854) when resolving inlined expressions.


ncave:
> This was supposed to be a small PR
🙂

PR #3263, opened 2022-11-09
Use .NET 7.0

ncave:
- Use .NET 7.0

@alfonsogarciacaro Now, I'm not saying we should actually merge this PR (although it works fine).

This is more of a test to see if that would resolve some .NET 7 issues that people are having.

Question: is there a way to version the Fable dotnet tool separately for different major .NET versions?


alfonsogarciacaro:
We can try, although I'd like to avoid forcing users to update the sdk just to update Fable. But it would be interesting to see if there are actual performance improvements (although we've been disappointed in the past). 

I am not aware of a way of using different sdk versions with dotnet tools. The only example I know of is Ionide and iirc they put different assemblies in the package and pick one based on the environment .

alfonsogarciacaro:
I think you can pass the dotnet exe location to Buildalyzer. So maybe we could pick dotnet 6 (if installed in the system) for project parsing .

alfonsogarciacaro:
> I was wondering if Fable dotnet tool built with .NET 7.0 will run ok if only .NET 6.0 is installed.

I would have to check but my understanding is it won't work, although you can build with dotnet sdk 7 and target net6.0, but not sure if this will help with the project parsing.

See #3265 for an alternative approach I'm trying.

PR #3268, opened 2022-11-13
[Python] Handle static getters

### LINKED ISSUE #3255: AttachMembers generates invalid Python for static fields
### Description

When adding a static field (`static member Foo = obj ()`) to a type marked with the `AttachMembers` attribute, compilation generates invalid python code: instead of emitting a `@property` it emits a simple field, which references the type before it exists.

### Repro code

repl: https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwChcDaAPAIIAuZcAxgBYCyYAtgmFgM4B8AurmQJ7hYwYFAAUASigBeXFDlRGTFligAPVACE4WCdKggsASwhkAZtABENYEjgBCC-nlQ2lMoaoKlrKAGVgijDCekHAEvihqP6BwpraEkA&html=Q&css=Q

For convenience, here is the code:

```fs
open Fable.Core

[<AttachMembers>]
type Foo () =
    member x.Bar() = printf "hello!"

    static member SomeFoo = Foo()

Foo.SomeFoo.Bar()
```

I am providing the full expected and actual results, and marking a comment on the problematic line, but please see "Explanation" below for a more readable explanation.

#### Expected result

```py
from typing import Literal
from fable_library.reflection import (TypeInfo, class_type)
from fable_library.string import (to_console, printf)

def _expr17() -> TypeInfo:
    return class_type("Test.Foo", None, Foo)


class Foo:
    def __init__(self, __unit: Literal[None]=None) -> None:
        pass

    def Bar(self, __unit: Literal[None]=None) -> None:
        to_console(printf("hola!"))

    # I would expect a property, so that the reference to Foo happens after the class is initialized
    @property
    def SomeFoo() -> Foo:
        return Foo()

Foo_reflection = _expr17

def Foo__ctor(__unit: Literal[None]=None) -> Foo:
    return Foo(__unit)


Foo.SomeFoo.Bar()

__all__ = ["Foo_reflection"]
```

#### Actual Result

```py
from typing import Literal
from fable_library.reflection import (TypeInfo, class_type)
from fable_library.string import (to_console, printf)

def _expr18() -> TypeInfo:
    return class_type("Test.Foo", None, Foo)


class Foo:
    def __init__(self, __unit: Literal[None]=None) -> None:
        pass

    def Bar(self, __unit: Literal[None]=None) -> None:
        to_console(printf("hola!"))

    # this causes an early reference to Foo (on the right-hand side of the assignment, the type is correctly handled thanks to the `from __future__ import annotations` 
    SomeFoo: Foo = Foo()

Foo_reflection = _expr18

def Foo__ctor(__unit: Literal[None]=None) -> Foo:
    return Foo(__unit)


Foo.SomeFoo.Bar()

__all__ = ["Foo_reflection"]
```

#### Explanation

Static member `SomeFoo` in Javascript is compiled (correctly) as follows:

```js
static get SomeFoo() {
    return new Foo();
}
```

In Python, instead, it compiles to:

```py
SomeFoo: Foo = Foo()
```

This causes an early invocation to `Foo`, which is not defined yet as we are inside its definition.

The correct code to be generated (also aligned to the behaviour in Javascript)  would be:

```py
@property
def SomeFoo() -> Foo:
    return Foo()
```

### Related information

Aside from the Fable tool version I don't think other versions matter at all in this specific case, but here is all I could find:

- Fable tool: 4.0.0-theta-015
- Fable.Core: 4.0.0-theta-006
- Python: 3.10.7
- dotnet: 6.0.402
- operating system: MacOS 12.6 (Monterey)

dbrattli:
Python do not support static getters (e.g for types with attached members) so rewrite to a static method instead. This fixes #3255

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -2046,7 +2046,10 @@ module Util =
                 $"Unexpected static interface/override call: %s{memb.FullName}"
                 |> attachRange r |> failwith
         let info = getAbstractMemberInfo com entity memb
-        if not info.isMangled && info.isGetter then
+
+        // Python do not support static getters, so we need to call the getter function
+        let pythonStaticGetter = com.Options.Language = Python && not memb.IsInstanceMember
+        if not info.isMangled && info.isGetter && not pythonStaticGetter then
             // Set the field as maybe calculated so it's not displaced by beta reduction
             let kind = Fable.FieldInfo.Create(
                 info.name,
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -2047,9 +2047,9 @@ module Util =
                 |> attachRange r |> failwith
         let info = getAbstractMemberInfo com entity memb
 
-        // Python do not support static getters, so we need to call the getter function
-        let pythonStaticGetter = com.Options.Language = Python && not memb.IsInstanceMember
-        if not info.isMangled && info.isGetter && not pythonStaticGetter then
+        // Python do not support static getters, so we need to call a getter function instead
+        let isPythonStaticMember = com.Options.Language = Python && not memb.IsInstanceMember
+        if not info.isMangled && info.isGetter && not isPythonStaticMember then
             // Set the field as maybe calculated so it's not displaced by beta reduction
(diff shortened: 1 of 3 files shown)


alfonsogarciacaro (approved):
@dbrattli Looks good to me! Should we merge?

alfonsogarciacaro:
Let' s merge and we can tweak it late 👍

PR #3271, opened 2022-11-14
Use target framework from .fsproj

alfonsogarciacaro:


ncave:
@alfonsogarciacaro I'm not sure why the switch to `netstandard2.1`, is that necessary? Even FCS is still on `netstandard2.0`.
IMO we shouldn't switch if there is no need for it.

alfonsogarciacaro:
@ncave Do you mean for libraries we are distributing through Nuget like Fable.AST? You're right, we should probably keep it targeting netstandard2.0 (Fable.Core still targets it).

The reason for the upgrade is that (I don't know why) Buildalyzer doesn't return the proper System references for netstandard2.0 when parsing .csproj. Given that netstandard is kind of deprecated I thought it wouldn't be a problem for users to upgrade their main projects if necessary (libraries and I think references too don't need to change). This is why I need to update Fable.Library projects, because they're self-compiled.

ncave:
@alfonsogarciacaro Since we're making a `.csproj` file copy, can we perhaps just replace the `netstandard2.0` with `netstandard2.1` in the `.csproj` file only?

ncave:
@alfonsogarciacaro 

> Note this only affects applications, not libraries

Then the error should only apply for applications, not libraries, no? 
```fs
        if result.OutputType = OutputType.Exe && result.TargetFramework = "netstandard2.0" then
            Fable.FableError "netstandard2.0 is not supported, please use netstandard2.1 or higher" |> raise
```

ncave:
@alfonsogarciacaro I guess what I'm trying to say is, it would be nice if we could avoid requiring that all libraries compiled with Fable have to use `netstandard2.1` (since `netstandard2.0` is [not deprecated](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-standard-not-deprecated) yet), but I understand we have limited options with the tools we using to instrument MSBuild. Technically the message should be "use `netstandard2.1` or `net6.0` or `net7.0`".

alfonsogarciacaro:
Sorry @ncave, my wording was not correct. It's actually about the main/entry project, not whether the output is an exe or a library. I identified non-main projects as libraries, because Fable packages are usually not distributed in precompiled form.

Fable.Library is the only example (I know of) of a Fable library that it is distributed in precompiled form. This is why we need to upgrade, because it becomes the main project when it is precompiled by Fable.

In any case, thanks a lot for checking! Ok, let's merge, release and see the user feedback.

PR #3277, opened 2022-11-19
Improve project parsing

alfonsogarciacaro:
Another attempt at improving project parsing by making sure we restore the "fake" .csproj


alfonsogarciacaro:
@ncave Funny enough this change seems to have fixed the issue with netstandard2.0 😆

PR #3279, opened 2022-11-20
Fix #3274 and other features

### LINKED ISSUE #3274: Wrong behavior if --outDir is a prefix of project dir
### Description

A simple project layout like so:

```
./fable_src
├── fable.fsproj
└── Program.fs
```

When compiling like this:

```bash
dotnet fable fable_src/fable.fsproj --verbose --outDir fable --extension .fs.js --noRestore --noCache
```

The transpiled file `Program.fs.js` appears in `fable_src` instead of `fable`. 

### Repro code

Please see this repo: https://github.com/njlr/fable-out-dir-repro

The GitHub Action demonstrates the issue.

I would expect the transpiled file to be in the `--outDir`, so `fable/Program.fs.js`

Instead it appears under the source directory at `fable_src/Program.fs.js`

The `_` in `fable_src` may be a significant factor.

### Related information

* Fable version: `3.7.20`
* Operating system: Ubuntu 22.04

alfonsogarciacaro:
1. Fix #3274: Output files are not put into outDir when this is a prefix of the project dir
2. Check `Condition` when manually parsing .fsproj in Fable packages. The condition is only evaluated if the property is one of the constants defined by Fable or user. Not sure if we will use this feature but I implemented it in case we decide to publish packages with [different files for each target language](https://github.com/fable-compiler/Fable/discussions/3219#discussion-4447898).
3. Support for `ValueOption` module

### DIFF
--- a/src/Fable.Cli/ProjectCracker.fs
+++ b/src/Fable.Cli/ProjectCracker.fs
@@ -4,6 +4,7 @@ module Fable.Cli.ProjectCracker
 
 open System
 open System.Xml.Linq
+open System.Text.RegularExpressions
 open System.Collections.Generic
 open FSharp.Compiler.CodeAnalysis
 open FSharp.Compiler.Text
@@ -248,11 +249,6 @@ let private getDllName (dllFullPath: string) =
     let i = dllFullPath.LastIndexOf('/')
     dllFullPath[(i + 1) .. (dllFullPath.Length - 5)] // -5 removes the .dll extension
 
-let (|Regex|_|) (pattern: string) (input: string) =
-    let m = Text.RegularExpressions.Regex.Match(input, pattern)
-    if m.Success then Some [for x in m.Groups -> x.Value]
-    else None
-
 let getBasicCompilerArgs () =
     [|
         // "--debug"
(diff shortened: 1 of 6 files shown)


alfonsogarciacaro:
@dbrattli If you want to identify `Fable.Core.U2` and friends, here's [how to do it](https://github.com/fable-compiler/Fable/blob/13e31b773ec57c0bfafd60751d639cba6e4c99be/src/Fable.Transforms/FSharp2Fable.Util.fs#L1289) (using a regex). I haven't added yet logic for identifying all types of erased union, which would be more complex, so for now is probably to still output `any` or its equivalent in those cases.

dbrattli:
@alfonsogarciacaro I just played with this example:

```fs
[<Erase>]
type ErasedUnion =
    | ErasedInt of int
    | ErasedString of string


let test () =
    let erased = ErasedInt 5
    ()
```

Which currently outputs:

```py
def test(__unit: None=None) -> None:
    erased: ErasedUnion = 5
```

So I need to figure out where to translate `ErasedUnion` into `int` since the type `ErasedUnion` cannot be used. So is that the reason for the types not being translated correctly?

alfonsogarciacaro:
Yes, but it's a bit complicated because erased unions can be compiled in different ways. I will update the documentation and add a helper tomorrow so it's clearer.

dbrattli:
PS: I btw managed to get `U2/U3/...` working with Python:

```fs
let test (arg: U2<string, int>) =
    ()

test (U2.Case2 10)
let a : U3<string, int, float> = U3.Case1 "hello"
```

```py
from typing import Union

def test(arg: Union[str, int]) -> None:
    pass


test(10)

a: Union[str, int, float] = "hello"
```

alfonsogarciacaro:
Because I realized I still need to fix reflection for erased records and unions and also Rust tests, maybe this cannot make it into Fable 4 release, so I've moved the erased record/union part to #3282

PR #3282, opened 2022-11-23
Enable erase unions with tags

### LINKED ISSUE #2914: Type of Erased types
Right now any type decorated with `Erase` becomes `Any` in Fable AST. This makes sense because originally `Erase` was thought for erased unions representing Typescript unions. However now there are cases where we want to erase the declaration but not necessarily the type. See this comment for example: https://github.com/fable-compiler/Fable/blob/8689186691ab4529c39b445ca85bc1b9fd20ddbb/tests/Main/ImportTests.fs#L113-L124

Another use case is when using a union just to wrap a primitive, then it should probably become the type of the primitive (although I'm not sure if Fable users frequently do this).

alfonsogarciacaro:
1. Allow building records with `Erase` attribute, compiling them as plain JS objects
2. Likewise, remove the limitation of `Erase` unions where multiple cases with multiple were not allowed. For this, I introduced the `tag` argument 
3. Fix #2914: Don't exclude erased types from Fable.AST.

Features 1 and 2 basically make possible what we [already tried in the past](https://github.com/fable-compiler/Fable/pull/2279) as an opt-in.

The last feature is the trickiest one because several patterns were relying on erased types being `Any` in Fable.AST. I managed to fix it for JS/TypeScript but for other languages I'm still [outputting `Any`](https://github.com/fable-compiler/Fable/blob/fa97e0822ba0330e449ed8db4e59485760675faf/src/Fable.Transforms/FSharp2Fable.Util.fs#L1160-L1163) until the tests are fixed. Another thing to note is I'm [wrapping new erased unions with a type cast](https://github.com/fable-compiler/Fable/blob/fa97e0822ba0330e449ed8db4e59485760675faf/src/Fable.Transforms/FSharp2Fable.fs#L62-L63) so the type info is not lost.

@dbrattli If you need the info from erased types, please [add Python here](https://github.com/fable-compiler/Fable/blob/fa97e0822ba0330e449ed8db4e59485760675faf/src/Fable.Transforms/FSharp2Fable.Util.fs#L1160-L1163) and check which tests fail. @ncave Unfortunately I broke the Rust tests again 😞 could you please have a look?

> NOTE: This has been split from #3279. Some previous discussion [here](https://github.com/fable-compiler/Fable/pull/3279#issuecomment-1321121386).

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -534,9 +534,8 @@ let testPython() =
         "--lang Python"
     ]
 
-    runInDir buildDir "poetry run pytest -x"
-    // Testing in Windows
-    // runInDir buildDir "python -m pytest -x"
+    if isWindows then runInDir buildDir "python3 -m pytest -x"
+    else runInDir buildDir "poetry run pytest -x"
 
 type RustTestMode =
     | SingleThreaded
@@ -764,7 +763,7 @@ match BUILD_ARGS_LOWER with
 | "test-integration"::_ -> testIntegration()
 | "test-repos"::_ -> testRepos()
 | ("test-ts"|"test-typescript")::_ -> testTypeScript()
-| "test-py"::_ -> testPython()
+| ("test-py"|"test-python")::_ -> testPython()
 | "test-rust"::_ -> testRust SingleThreaded
 | "test-rust-default"::_ -> testRust SingleThreaded
 | "test-rust-threaded"::_ -> testRust MultiThreaded
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -29,6 +29,7 @@ type AttachMembersAttribute() =
 type EraseAttribute() =
     inherit Attribute()
     new (caseRules: CaseRules) = EraseAttribute()
+    new (tag: bool) = EraseAttribute()
 
 /// Used for "tagged" union types, which is commonly used in TypeScript.
 type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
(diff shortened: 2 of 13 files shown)


alexswan10k:
Hi @alfonsogarciacaro , I had a quick look at this but not sure I have anything conclusive to add. A couple of observations:

Line 812 -``` hasAttribute Atts.erase```
should probably now be
```hasAttribute Atts.emit```

Previously the Erase;Emit type would not output the entity at all, but now it is creating a trait (it thinks it is now a real interface). I imagine this might be because the behaviour of Any has changed.

I wonder if we could go through instances of Any in the FableToRust transform, and where Any was previously used to handle the erased scenario, we replace them with some kind of active pattern to any entity that has an emit attribute? I imagine part of the problem is Any's are not considered to be wrapped in a RC but managed types are (by shouldBeRefCountWrapped). This might explain the as_ref call.

Sorry my time is a bit sparse at the moment, but I will try and dig in a bit further in the next few days.

ncave:
> Allow building records with Erase attribute, compiling them as plain JS objects

@alfonsogarciacaro Hmm, previously `Erase` meant erase, not `Plain`.
Is there an issue or thread I can read to understand what necessitates this change of attribute meaning?

shayanhabibi:
I initially was under the impression that Erasing a record type would default usage of it in JS/TS as POJOs. In my mind, it seems like it should be a natural conclusion. We can construct POJOs with Anon Records. If we erase a Record type from transpilation, then why would it not operate like a Anon Record would.

I'm probably missing some fundamental concept with erasure though :P

alfonsogarciacaro:
Because users expect declared records to be compatible with many of the functions available in F#/.NET like (explicit) interface implementation, type testing or reflection. And for this you need information about the type at runtime.

Issue #3298, opened 2022-12-14
Fable 3.7.10 incompatible with React.suspense


aaronmu:
We're using [Suspense](https://17.reactjs.org/docs/concurrent-mode-suspense.html) in combination with [Recoil](https://github.com/facebookexperimental/Recoil) in our project. After upgrading to Fable 3.7.10, every component that makes use of async recoil selectors (they support suspense) started crashing at runtime.

We've debugged the issue and have found that the introduction of `ensureErrorOrException` in `AsyncBuilder.js` is the source of these errors. https://github.com/fable-compiler/Fable/commit/1da89fe30f787129f5d3648b75152feb5c603771#diff-1220a10f73b73e8ff75e34b2fdeb2d8039767b0b4a92f7af993d483aab800b2c.

You see.. the thing about suspense is .. * drum roll * .. it throws promises 🤢. The `ensureErrorOrException` implementation will return this promise in an Error containing a string version of the promise. Which in turn breaks suspense.

Now. I don't know enough about `ensureErrorOrException` or the context in which this function is used but we played around with it and modifying it so that it checks for instance of promise seems to fix it for us.

alfonsogarciacaro:
Sorry for the late reply. Thank you @aaronmu for reporting and @jonasdw for the fix!

IIRC `ensureErrorOrException` is meant both to appease Typescript compiler and also to make sure the `try .. with ..` branches in async workflows always receive "proper" exceptions. It's unfortunate that React.suspense decides to throw promises but I assume we wouldn't change their minds so the "hack" on our side is the least evil.

I'm closing as I merged the fix and will release a new version. Please reopen if you still have issues. Thanks again!

PR #3300, opened 2022-12-15
potential fix for #3298

jonasdw:


alfonsogarciacaro:
Sorry, for the late reply. Thanks a lot for the fix!

PR #3302, opened 2022-12-16
Fixed #3301

### LINKED ISSUE #3301: Tail call optimization for partially applied functions and functions coming from a pattern match
### Description
Tail calls are not optimized in some cases where I hoped they would be, which can lead to stack overflows given large enough input.

### Repro code
Consider this simple function. It get's optimized to a while-loop:
```fs
let rec foo x a =
    if x <= 0 then a
    else
        foo (x-1) (a+1)
```
However, this does not, as the function is partially applied:
```fs
let rec foo x a =
    if x <= 0 then a
    else
        let bb = foo (x-1)
        bb (a+1)
```
And this also doesn't, because we had to pattern match to get the function:
```fs
let rec foo x a =
    if x <= 0 then a
    else
        let (bb, _) = (foo, 1)
        bb (x-1) (a+1)
```
In general, it seems any kind of delayed tail call isn't optimized. I assume this is just something nobody has implemented yet, but I am curious if it is tracked, and if there are any plans to do so? FWIW, Microsoft's main F# compiler currently optimizes all 3 of these.

All 3 functions are tail recursive, and should be optimized into while-loops to not overflow the stack. One might also consider using a trampoline in these cases, if this is deemed too difficult. That tends to be easier to implement, but slower.

ncave:
- Fixes #3301 issues with beta reduction.
- Beta reduction fix is causing some issues with `Dart/Python` tests, so for now is only enabled for `JS/TS/Rust`.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -594,7 +594,8 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) fsExpr =
         // we replace with immutable bindings instead which generates better code
         // and increases the chances of the tuple being removed in beta reduction
         | FSharpExprPatterns.NewTuple(tupleType, tupleValues) as tupleExpr
-            when var.IsCompilerGenerated && var.CompiledName = "matchValue" ->
+            when var.IsCompilerGenerated &&
+                (var.CompiledName = "matchValue" || var.CompiledName = "patternInput") ->
 
             let! tupleValues = transformExprList com ctx tupleValues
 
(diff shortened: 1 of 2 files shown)


ncave:
@dbrattli I didn't really want to bother you with this, so I disabled this fix for Python (and Dart) cause it was breaking some tests, but in theory it should work (at least for Python), we just have to figure out what's wrong.

ncave:
@alfonsogarciacaro Just to clarify, IMO this is ready for merging, Python/Dart support can be added later.

alfonsogarciacaro:
Agreed, thanks @ncave!

dbrattli:
@ncave Fixed for Python in #3310

PR #3303, opened 2022-12-17
[Dart] Moved tests output to build folder

ncave:
- Moved Dart tests output to build folder (for consistency).


alfonsogarciacaro:
Thanks! I think I did it like this at the beginning for simplicity. But it's indeed better to be consistent through languages

PR #3308, opened 2022-12-26
Fix list slice when indexes are out of bounds

### LINKED ISSUE #3295: slicing [1;2;3][..9] behaves differently in fable and dotnet F#
### Description
Slicing shorter lists throws an exception.
The following code works in dotnet F#, but fails with fable.

My actual use case is filtering a list by search string and truncating it to max 10 entries.
### Repro code

```f#
let a=[1;2;3]
printfn "A %A" a
let b=a[..1]
printfn "B %A" b
let c=a[..4]
printfn "C %A" c
```

#### expected (dotnet F#)
```
A [1; 2; 3]
B [1; 2]
C [1; 2; 3]
```
#### actual (fable https://fable.io/repl/)
```
A [1; 2; 3]
B [1; 2]
Uncaught Error: The index was outside the range of elements in the collection.\nParameter name: endIndex
    getSlice List.js:1234
```

### Related information

* current version of https://fable.io/repl/ 😉
* Fable version: `4.0.0-snake-island-alpha-026` on `ubuntu 22.10`

dbrattli:
Slicing should be tolerant and not throw exceptions. From: https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/slices. Fixes #3295

> C# developers may expect these to throw an exception rather than produce an empty slice. This is a design decision rooted in the fact that empty collections compose in F#. An empty F# list can be composed with another F# list, an empty string can be added to an existing string, and so on. It can be common to take slices based on values passed in as parameters, and being tolerant of out-of-bounds > by producing an empty collection fits with the compositional nature of F# code."

### DIFF
--- a/src/fable-library-dart/List.fs
+++ b/src/fable-library-dart/List.fs
@@ -620,11 +620,14 @@ let truncate count (xs: 'T list) =
 
 let getSlice (startIndex: int option) (endIndex: int option) (xs: 'T list) =
     let len = length xs
-    let startIndex = defaultArg startIndex 0
-    let endIndex = defaultArg endIndex (len - 1)
-    if startIndex < 0 then invalidArg "startIndex" SR.indexOutOfBounds
-    elif endIndex >= len then invalidArg "endIndex" SR.indexOutOfBounds
-    elif endIndex < startIndex then List.Empty
+    let startIndex =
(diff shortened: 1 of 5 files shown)


alfonsogarciacaro:
Thanks @dbrattli!

Issue #3311, opened 2022-12-27
JS tests fails in main


dbrattli:
### Description

JS tests fails consistently in `main` after https://github.com/fable-compiler/Fable/commit/76e538be77a81abef987a7b01d6281eb4f49e810. Not sure why we get `RangeError: Maximum call stack size exceeded`.  Does this mean that the code has not been tail call optimized in GH?

ncave:
@dbrattli Yes, looks like 2/3 of the new tests are not tail call optimized when compiled by Fable-JS.
I'll take a look.

ncave:
I temporarily disabled two of the tests for JS, so we can get back to a working build.
It could be that the F# AST is different between `Fable` and `Fable-FCS` somehow.

ncave:
Turns out F# AST was the same, it was just a matter of not defaulting to DEBUG mode for `fable-standalone`. See #3324.

PR #3312, opened 2022-12-27
More tests for #3302

ncave:


alfonsogarciacaro:
Thanks a lot @ncave! Sorry, I run the tests locally, but forgot to run the full build so I didn't check fcs-js. Given the tests only fail with fcs-js, maybe it makes sense to re-enable them again when compiling with dotnet fable so we are sure the tail call is working?

ncave:
@alfonsogarciacaro 
> maybe it makes sense to re-enable them again when compiling with dotnet fable

Perhaps, I just don't know which conditional symbol to use for that.
Also, once #3311 is resolved, we will re-enable them unconditionally.

PR #3321, opened 2023-01-02
Various improvements to PhP Output

entropitor:
1. Don't add unnecessary __ROOT__ call in php (because of path not matching `.fs.php` vs `.php`)
2. Copy fable-library-php over when building php target (instead of js output)
3. Add a very simple Random.php that supports some Random things from F#
4. Emit type information and add a Serialization.php to convert types from and to json compatible with [Fable.SimpleJson](https://github.com/Zaid-Ajaj/Fable.SimpleJson)

It makes sense to put Serialization.php in user land but it would still need the type information emitted by the fable compiler to make it work.


alfonsogarciacaro:
Really sorry for this, I had a busy start of the year and completely missed this PR. Currently PHP compilation is not maintained so I'm happy to accept any PRs that don't break any other language :) Keep them coming!

Issue #3328, opened 2023-01-09
Using a byref parameter with SRTP results in code which always fails


BillHally:
### Description

Using a `byref` parameter with SRTP results in code which always fails, because of 2 problems with the code generation.

### Repro code

```fsharp
type TypeWithByRefMember() =
  static member DoubleIntByRef (x: byref<int>) : unit = x <- x + x

let inline doubleIntByRef (x: ^a) (input: int) : int =
    let mutable value = input
    (^a: (static member DoubleIntByRef: byref<int> -> unit) &value)
    value

try
    let result = doubleIntByRef (TypeWithByRefMember()) 7
    printfn $"result: {result}"
with
| ex -> eprintfn $"ERROR: {ex}"
```

View in the [repl](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYFgBQXSSgMQEMAjAGzB3GiLLADoBhYAJws12pPMZfoCkEASQgpmoLFiRxwUACrSwAdQCWSABYAhOACUwAMwCyaYmGYAKAJRQAvFihRkhJMoDGUVMdNQAIsACutMJIWrp6UGYAHgBcUMRwrHoAPMoiAHxWMX4QqjZQEVCJALR5UADUeRKY5EhQKaQpYFAAJv6BIiH64dFQAHqEVmYpIH5IMSlIGbUiNnb2UNXuI1yNAG6EpH6N1lPDSLP2Zn0xZo7Obh6oJsw+reRBHXoxcQnJaVCFqVBZqlYAZGsbMAWfZQAGbSpIZhwEELVgIPykGrbFoBO7tHSdMzycAqdQPIyXUyWKwAdhBIGY4z00AAJAAiOEI0ZQADejMRAF86VgAO6qNRYAA+UDA+Q+IopVNpdIAotptAB5bQxFmirlYIA&html=Q&css=Q).

##### Javascript

Skipping the `TypeWithByRefMember` code, I think the following would be the expected code (at least, cutting and pasting it into the repl gives the expected output):

```javascript
try {
    let result;
    const x = TypeWithByRefMember_$ctor();
    let value = 7;
    const arg0_ = (new FSharpRef(() => value, (v) => {
        value = (v | 0);
    }));
    TypeWithByRefMember_DoubleIntByRef_Z3D2991E8(arg0_);
    result = value;
    toConsole(`result: ${result}`);
}
catch (ex) {
    toConsoleError(`ERROR: ${ex}`);
}
```

[...]

PR #3329, opened 2023-01-10
Address byref replacements and Babel optimization for const byrefs

weebs:
Issue for context: https://github.com/fable-compiler/Fable/issues/3328

These changes are a proposed fix for the byref placement behavior when byrefs are converted into their contents in instances the byref should remain unchanged. This behavior is most noticeable for inline functions that take a byref parameter, as the inlined expression will create a variable of a byref type, which gets de-referenced into its contents when it's used.

There is also a second change to address the behavior where a const byref variable will be optimized with `| 0` in the F# -> Babel replacements, casting the byref to 0. @BillHally has also included some tests to verify the behavior of byrefs in the scenarios that caused errors

I've used a string literal for the match comparison in the second commit, and the first commit includes an indexed access into the GenericArguments array after isByRefValue succeeds. Please let me know if there's any changes you'd like to see to the fix such as the FSharpRef match done differently or if the GenericArgs access should check for the length first as an added safety

### DIFF
--- a/src/Fable.Transforms/Replacements.Util.fs
+++ b/src/Fable.Transforms/Replacements.Util.fs
@@ -48,6 +48,10 @@ type Helper =
             | Some false | None -> None
         let info = CallInfo.Create(?thisArg=thisArg, args=args, ?sigArgTypes=argTypes, ?genArgs=genArgs, ?memberRef=memberRef, ?isCons=isConstructor)
         Call(callee, info, returnType, loc)
+        
+    static member FSharpRef(com, t, args, ?loc) =
+        let byrefType = DeclaredType({ FullName = Types.byref; Path = CoreAssemblyName "System" }, [ t ])
+        Helper.LibCall(com, "Types", "FSharpRef", byrefType, args, isConstructor=true, ?loc=loc)
 
     static member ImportedValue(com, coreModule: string, coreMember: string, returnType: Type) =
         makeImportUserGenerated None Any coreMember coreModule
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -949,7 +949,6 @@ module Util =
     /// Wrap int expressions with `| 0` to help optimization of JS VMs
     let wrapIntExpression typ (e: Expression) =
         match e, typ with
-        | NewExpression(Expression.Identifier (Identifier("FSharpRef", _, _, _, _)), _, _, _), _
         | Literal(NumericLiteral(_)), _ -> e
         // TODO: Unsigned ints seem to cause problems, should we check only Int32 here?
         | _, Fable.Number((Int8 | Int16 | Int32),_) ->
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -76,7 +76,7 @@ let setRefCell com r (expr: Expr) (value: Expr) =
     setExpr r expr (makeStrConst "contents") value
 
 let makeRefCell com r typ (value: Expr) =
-    Helper.LibCall(com, "Types", "FSharpRef", typ, [value], isConstructor=true, ?loc=r)
+    Helper.FSharpRef(com, typ, [value], ?loc=r)
 
 let makeRefFromMutableValue com ctx r t (value: Expr) =
     let getter =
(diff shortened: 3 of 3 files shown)


BillHally:
Ah - when testing on my machine I didn't notice that when run under CI the "test-js" target effectively runs the "test-standalone-fast" target as well.

If I explicitly run that locally:

```pwsh
dotnet fsi build.fsx test-standalone-fast
```

I also see the test failure reported by CI.

Specifically, it's the test `inline byref works (with separate binding for reference)` which uses a separate binding for the reference in order to ensure that the bitwise OR is no longer happening on the `FSharpRef` (because I'd only seen it happen when there was a separate binding).

Unfortunately, it appears that although it no longer happens with the normal build, with the "standalone" one it is still happening.

`build\tests\Js\Main\TypeTests.js`:

```fsharp
}), testCase("inline byref works (with separate binding for reference)", () => {
    let an_int_1 = 33;
    const intRef = new FSharpRef(() => an_int_1, (v_2) => {
        an_int_1 = (v_2 | 0);
    });
    intRef.contents;
    byrefFunc(intRef);
    equal(66, an_int_1);
})]);
```

`build\tests\Standalone\TypeTests.js`:

```fsharp
}), testCase("inline byref works (with separate binding for reference)", () => {
    let an_int_1 = 33;
    const intRef = (new FSharpRef_1(() => an_int_1, (v_2) => {
        an_int_1 = (v_2 | 0);
    })) | 0; // <--------------------------------
    intRef.contents;
    byrefFunc(intRef);
    equal(66, an_int_1);
})]);
```

As well as the re-appearance of the bitwise OR, there is another change - the typename is "FSharpRef_1" rather than "FSharpRef".

Given that the code fixing that particular issue is:

```fsharp
        | NewExpression(Expression.Identifier (Identifier("FSharpRef", _, _, _, _)), _, _, _), _
```

it looks as though the different name explains the bitwise OR reappearing, because the name is different to the one expected by that match case.

Therefore, it looks as though either:
1. the name "FSharpRef_1" is wrong, and needs to be fixed
2. the name "FSharpRef_1" is correct *and will always have that value*, in which case another match case can be added to fix the problem
3. the name "FSharpRef_1" is correct but may not always have that value, in which case the existing match case needs to be replaced with something a little more complex (depending on exactly how the name can vary)

I don't know which of these 3 applies though...

ncave:
@BillHally Technically, `FSharpRef_1` is correct, cause it's a generic type with one type parameter.

weebs:
I was having trouble finding enough type information at the Fable2Babel layer, but I was able to get the behavior working for the standalone tests by changing the return type of FSharpRef() constructor calls to a byref`1. Previously they were typed as the byref's parameter type. I'm not sure this is the right solution but it seems better than matching this info at the wrapIntExpression level. JS & Rust tests are passing for me locallly

src/Fable.Transforms/Replacements.Util.fs
```diff
@@ -48,6 +48,10 @@ type Helper =
             | Some false | None -> None
         let info = CallInfo.Create(?thisArg=thisArg, args=args, ?sigArgTypes=argTypes, ?genArgs=genArgs, ?memberRef=memberRef, ?isCons=isConstructor)
         Call(callee, info, returnType, loc)
+        
+    static member FSharpRef(com, t, args, ?loc) =
+        let byrefType = DeclaredType({ FullName = Types.byref; Path = CoreAssemblyName "System" }, [ t ])
```
ncave, line 53: @weebs Have you tried using `Types.refCell` intead of `Types.byref`?

ncave: That's fine, I can address it in another PR. I'll merge it as is and fix it later.

weebs: Thanks for doing the work to adjust the types @ncave , I tried using `Types.refCell` and ran into TypeScript creating FSharpRef<FSharpRef<>> types for ref constructors. I was able to the tests working by changing one of the makeRef calls but I realized it probably needed more attention and I'm not comfortable touching the various plugins :)

Your fixes helped me understand better how refs were being handled, thanks for the review and working out the remaining pieces!

ncave: @weebs You're very welcome, thanks for contributing!

PR #3336, opened 2023-01-31
Dart import issue

carne8:
When I used the ``print`` function, dart compiler said "The function 'print' isn't defined."
So I fixed this issue for the ``print`` function and also for the ``Future`` and ``Stream`` types.

### DIFF
--- a/src/Fable.Core/Fable.Core.Dart.fs
+++ b/src/Fable.Core/Fable.Core.Dart.fs
@@ -1,6 +1,7 @@
 module Fable.Core.Dart
 
 open System
+open Fable.Core
 
 type IsConstAttribute() =
     inherit Attribute()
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Sorry for the late reply, I was busy last month. Thanks a lot for this!

alfonsogarciacaro:
@carne8 What version of Dart/Flutter are you using? With this change compilation is failing for me.

image

carne8:
I had the same problem even without this change.
To solve this you just need to add `import "dart:core";` at the top of the Dart file.

carne8:
I'm new to Dart / Flutterm I don't know if there is a thing like this.
For my part, I wrote a little script that automatically add `import "dart:core"` to my generated files.
If you want to see, [here is my script](https://github.com/carne8/SoundBox/blob/main/PostFableBuild.fsx).

PR #3345, opened 2023-02-13
Golang (wip)

dbrattli:
Exploring the possibility of translating F# to Go. Currently forked out of Python branch, but not sure there will be much left of the Python legacy when (if) I get this finished one day. Added PR here for transparency. Will take many months before it would be possible to merge anything.

Discussion at #3346


dbrattli:
Abandoned in favor of using Rust instead

PR #3360, opened 2023-03-06
Moar TypeScript tests

alfonsogarciacaro:
I realized in the optimization phase we were moving lambdas with local generics without resolving them, which caused issues with the type system of some of the target languages. I've tried to solve it by [checking the arguments](https://github.com/fable-compiler/Fable/blob/f320c01a331b9253ce740fc0751d93ce93870065/src/Fable.Transforms/FSharp2Fable.fs#L698-L712) and [applied values](https://github.com/fable-compiler/Fable/blob/f320c01a331b9253ce740fc0751d93ce93870065/src/Fable.Transforms/FSharp2Fable.fs#L550) in the FSharp2Fable step. Then when inlining a value I make sure [its type matches the one of the replaced ident](https://github.com/fable-compiler/Fable/blob/f320c01a331b9253ce740fc0751d93ce93870065/src/Fable.Transforms/FableTransforms.fs#L71-L74).

This seems to work with TypeScript and Dart, but unfortunately I broke Rust again 😞 @ncave Could you please have a look if possible? It may happen that some corrections are not longer necessary. For example, in TypeScript we no longer need to [uncurry all lambda types just in case](https://github.com/fable-compiler/Fable/pull/3360/files#diff-5c1d432616b2869d7a3e9828fcf99f13d2e4eb3e88d6abbd135f5cced457a628R430).

> Example: In [this test](https://github.com/fable-compiler/Fable/blob/f320c01a331b9253ce740fc0751d93ce93870065/tests/Dart/src/OptionTests.fs#L267), the `x2` argument of the inner `test` function is generic, but when [it's applied](https://github.com/fable-compiler/Fable/blob/f320c01a331b9253ce740fc0751d93ce93870065/tests/Dart/src/OptionTests.fs#L304) it becomes `unit` so if inlining we need to have the proper type.

**Edit**: See discussion [below](https://github.com/fable-compiler/Fable/pull/3360#discussion_r1133028636).

### DIFF
--- a/src/Fable.Transforms/FableTransforms.fs
+++ b/src/Fable.Transforms/FableTransforms.fs
@@ -61,6 +61,20 @@ let replaceValues replacements expr =
             | None -> e
         | e -> e)
 
+let replaceValuesAndGenArgs (replacements: Map<string, Expr>) expr =
+    if Map.isEmpty replacements then expr
+    else
+        expr |> visitFromInsideOut (function
+            | IdentExpr id as e ->
+                match Map.tryFind id.Name replacements with
+                | Some e ->
+                    if typeEquals true e.Type id.Type then e
+                    else
+                        extractGenericArgs e id.Type
+                        |> replaceGenericArgs e
+                | None -> e
+            | e -> e)
+
 let replaceNames replacements expr =
     if Map.isEmpty replacements
     then expr
@@ -162,7 +176,7 @@ let noSideEffectBeforeIdent identName expr =
 
     findIdentOrSideEffect expr && not sideEffect
 
-let canInlineArg _com identName value body =
+let canInlineArg identName value body =
     (canHaveSideEffects value |> not && countReferences 1 identName body <= 1)
      || (noSideEffectBeforeIdent identName body
          && isIdentCaptured identName body |> not
@@ -202,11 +216,6 @@ let rec uncurryLambdaType (revArgTypes: Type list) (returnType: Type) =
     | t -> List.rev revArgTypes, t
 
 module private Transforms =
-    let (|LambdaOrDelegate|_|) = function
-        | Lambda(arg, body, name) -> Some([arg], body, name)
-        | Delegate(args, body, name, []) -> Some(args, body, name)
-        | _ -> None
-
     let rec (|ImmediatelyApplicable|_|) appliedArgsLen expr =
         if appliedArgsLen = 0 then None
         else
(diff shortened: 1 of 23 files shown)


src/Fable.Transforms/FableTransforms.fs
```diff
@@ -260,20 +260,21 @@ module private Transforms =
         match e with
         | Let(ident, value, letBody) when (not ident.IsMutable) && isErasingCandidate ident ->
             let canEraseBinding =
-                match value, com.Options.Language with
```
ncave, line 263: @alfonsogarciacaro You can leave the language check for Rust for now, just to make it pass, I'll fix it later.

alfonsogarciacaro: Thanks @ncave! At the end I've taken a different approach (see edited comment) bit unfortunately it's still breaking Rust :/

ncave: @alfonsogarciacaro I'll take a look, but can't we just keep the bindingBetaReduction as it was for Rust only? (just temporary)?

alfonsogarciacaro: You're right, that'd probably be the best solution. Unfortunately there have been several changes in the FSharp2Fable and FableTransforms steps. I've tried to undo some of them but Rust was still failing so I couldn't pinpoint the problem.

I've [disabled the 4 failing Rust tests](https://github.com/fable-compiler/Fable/pull/3360/commits/24df43cacb457d963e67998c8683e084dced097a). Do you think that'd be ok for now to merge this PR until we can investigate? It looks like an issue with the closures created by `sprintf` and friends. I had a similar problem in Typescript, but it was fixed by [adjusting the ident type after propagating uncurrying](https://github.com/fable-compiler/Fable/blob/24df43cacb457d963e67998c8683e084dced097a/src/Fable.Transforms/FableTransforms.fs#L423).

ncave: @alfonsogarciacaro Sure, not a problem, please merge and I'll take a look.

alfonsogarciacaro: Thanks for confirming @ncave! Given that it's only 4 related tests, hopefully it's easy to spot the problem. BTW, I just realized `kprintf` is failing sometimes for TypeScript too. I'll try to fix that, maybe it's related to the Rust regression.

There is still work to do but good news is we already have >1000 tests passing for TypeScript :tada:

ncave: @alfonsogarciacaro That sounds great!

The real challenge for all the typed languages (TS, Dart, Rust) are the Apllicative tests.
Hopefully your lambda changes will make things a bit easier there.

PR #3366, opened 2023-03-13
TypeScript applicative tests

alfonsogarciacaro:
With this, all applicative tests are passing now for TypeScript! :tada: Luckily it was not as hard as I expected. One important fix is that, although in my previous PR I resolved the uncurried types for expressions in the AST in the FableTransforms step, field types are still not uncurried, so I had to make a distinction when creating annotations for class/record/union fields: https://github.com/fable-compiler/Fable/blob/92358e7416013314b2e3ed698170b9eba7ef37fc/src/Fable.Transforms/Fable2Babel.fs#L459-L461

You may need to do something similar for Rust @ncave, if you haven't already, that is.


ncave:
Thanks @alfonsogarciacaro, we already uncurry the field types for Rust.

But I had to disable the field uncurry in anonymous records, as there is no way to do in Rust the parameter magic that `Util.checkArity` does.

PR #3367, opened 2023-03-13
[Rust] Added LrcPtr wrapper

ncave:
- [Rust] Added `LrcPtr` wrapper (behind feature flag).
- Avoid resolving `GenericParam` measures to `Fable.Any`.


alexswan10k:
Nice. I can see how this helps the standard library, but does this help you get around implementing a trait against LrcPtr<T> in a custom crate too?

ncave:
@alexswan10k It could potentially, I merged it behind a (disabled) feature flag, just to catch the 4.0.0 stable release train.

With this PR, now we're always implementing traits directly on the local struct, so the orphan rules don't apply. Technically we may not need the `LrcPtr` wrapper any more, but I spent so much time type-wrangling it that I decided to keep it around just in case.

The next steps would be to:
- properly support operators with different right-hand types
- add operators on non-value-types

I have some idea how to do that with some op macro reworking. We can also make non-value-type operators a bit more efficient by converting static operators where first arg is Self into instance calls.

alexswan10k:
Ah yes, that is a totally different way of approaching it. Fair play.

My concern was that, regrettably we might eventually need something like this to get around the lrc self problem:

```F#
interface SomeTraitThatNeedsSelfAsLrc with
    member x.Identity() = x

type SomeType = {
    x: int
}
with 
    interface SomeTraitThatNeedsSelfAsLrc with
        member x.Identity() = x
```
```Rust
pub struct SomeTypeRaw {
      pub x: i32,
  }

  #[derive(Clone)]
  pub struct SomeType(Lrc<SomeTypeRaw>);

  pub trait SomeTraitThatNeedsSelfAsLrc {
      fn identity(&self) -> Self;
  }

  impl SomeTraitThatNeedsSelfAsLrc for SomeType {
      fn identity(&self) -> Self {
          self.clone()
      }
  }
```

Hopefully this is avoidable though, although there is a good argument to be made that it might be a substantially better development experience from the Rust side. Anyway I am aware this is quite a fiddly massive sweeping change, so happy either way. I just wanted to log it before I forgot.

ncave:
@alexswan10k I know, you mentioned local wrappers for non-value types before.
It's similar to the `LrcPtr` wrapper, which is a bit more uniform, but local wrappers avoid orphan rules.

Anyway, we're handling fluent types by modifying the `&self` to `self: &Lrc<Self>` where needed for Fluent APIs, so that's already handled. And interface members are instance members, so non-fluent calls also work by auto deref-ing the &self. The only problem is the generic type constraints on traits (that need the trait to be implemented on the pointer, not the struct).

Technically from the point of view of the generated code, `LrcPtr` is external too, same as `Lrc`. The only point of using `LrcPtr` is cause we can implement some std lib traits on it.

Anyway, we can take it one step at a time and see how it goes. We do have options, just not very good ones. The orphan rules are a harsh mistress.

ncave:
@alexswan10k To clarify, in this PR the generic type constraints on traits are handled by [auto-implementing](https://github.com/fable-compiler/Fable/blob/main/src/Fable.Transforms/Rust/Fable2Rust.fs#L3693-L3706) each trait on the pointer as well, e.g.:

```rs
pub trait IHasAdd {
    fn Add(&self, x: i32, y: i32) -> i32;
}

impl<V: IHasAdd + Clone + 'static> IHasAdd for LrcPtr<V> {
    fn Add(&self, x: i32, y: i32) -> i32 {
        (**self).Add(x, y)
    }
}
```

alexswan10k:
That makes sense, thanks for explaining.

I guess the risk with fluent self-consumption is theoretically developers can do all sorts of weird things like this
```
let a = new FluentThing()
let b = a.With(config)
let c = b.With(otherConfig)
let b = b.With(differentThing) // will break because B is consumed 2 lines above
```
and not what you would expect, which looks more like
```
a.With(something).With(somethingElse).Build().Whatever()
```

It is kind of a terrible pattern anyway (just unfortunately it is everywhere in C# configuration land including aspnet core kestrel setup). Fluent api creators should learn to create new immutable instances rather than mutating self! Now that I say it, how can you know the difference between a side effect + return self, and a return new instance of TSelf with stuff changed? I guess you can work it out from the AST.

Ah well as long as the FP stuff works :)

Always exciting to see progress anyway. When I get a bit more time I would love to get back to the macroquad app I was writing.

PR #3372, opened 2023-03-17
Add ability to produce non-English identifiers

kant2002:
I would like to preserve non-English identifiers where possible in JS/TS/Python because these languages support Unicode identifiers. Dart does not support them and Rust seems to have only partial support. So this is partial work.

### DIFF
--- a/src/Fable.Transforms/Global/Naming.fs
+++ b/src/Fable.Transforms/Global/Naming.fs
@@ -46,6 +46,10 @@ module Naming =
         || (97 <= code && code <= 122)  // A-Z
         // Digits are not allowed in first position, see #1397
         || (index > 0 && 48 <= code && code <= 57) // 0-9
+        || match Compiler.Language with
+            | Dart -> false
+            | Rust -> false
+            | _ -> Char.IsLetter c
 
     let hasIdentForbiddenChars (ident: string) =
         let mutable found = false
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Ah, you're right. Actually the Naming module is one of the parts we haven't split per language (keywords are still the JS ones), but we store the current language in a static member of the Compiler type so maybe you can use it to decide if Unicode identifiers are accepted or not

kant2002:
Unfortunately Compiler included after Prelude 
https://github.com/fable-compiler/Fable/blob/0836ba3dda2f19f0d75153846bc3b8d99dc767b2/src/Fable.Transforms/Fable.Transforms.fsproj#L12-L13

I obviously can move `Compiler.Language` to some other location in prelude, but need your guidance or approval.

alfonsogarciacaro:
Hmm, there seems to be a dependency between Compiler and Prelude. I think it makes sense to move Naming to a separate file. Let me think about this.

alfonsogarciacaro:
Ok, I've move Naming to a separate file after Compiler :+1: #3376

kant2002:
Great, changes are now make sense. Please take a look

ncave:
@kant2002 IMO Rust works perfectly fine with Unicode identifiers, so you can enable it for Rust too.

kant2002:
Done. I try on some sandbox online and it give "scary" warnings. Because that's extent of my testing I decide to play safe.

PR #3378, opened 2023-03-20
Unify and improve (un)currying

alfonsogarciacaro:
**EDIT**: This PR doesn't try to inline (un)currying for all languages (only for Rust atm) because of the issue with side effects (see below) but it does try to make things more consistent and improve the mechanism.

-----------------------------------------

This is an attempt to resolve currying/uncurrying directly in Fable AST instead of using specific helpers for each language. This is basically what Rust already did only a bit improved (hopefully). The main advantage is we can share the complex mechanism and have proper typings. I'm almost there, it's working for the typed languages (interestingly something broke for the dynamic languages, I will check).

The bad news is I haven't been able so far to resolve the problem of side effects running multiple times when partial applying a function that has been previously uncurried. The way this is currently solved in currently solved for JS and Python is to add a property with a reference to the curried function so when partial applying we can [check if the curried version is available](https://github.com/fable-compiler/Fable/blob/746c9d98f14becc13b382aca315bddfa4a1fb616/src/fable-library/Util.ts#L741-L745).

However, this is not possible for the typed languages because we cannot add extra properties to the functions. Any ideas? @ncave @dbrattli @alexswan10k Maybe keep a global dictionary with the references to the curried functions instead?

### DIFF
--- a/src/Fable.Cli/Properties/launchSettings.json
+++ b/src/Fable.Cli/Properties/launchSettings.json
@@ -8,7 +8,11 @@
       // "commandLineArgs": "src/fable-library --outDir build/fable-library-ts --fableLib build/fable-library-ts --lang TypeScript --typedArrays false --exclude Fable.Core --define FX_NO_BIGINT --define FABLE_LIBRARY --noCache"
 
       // test-ts
-      "commandLineArgs": "tests/TypeScript --outDir build/tests/TypeScript --lang TypeScript --exclude Fable.Core --noCache"
+      //"commandLineArgs": "tests/TypeScript --outDir build/tests/TypeScript --lang TypeScript --exclude Fable.Core --noCache",
+
+      // quicktest-ts
+      "commandLineArgs": "src/quicktest --lang TypeScript --exclude Fable.Core --noCache"
+
     }
   }
 }
--- a/src/Fable.Cli/Properties/launchSettings.json
+++ b/src/Fable.Cli/Properties/launchSettings.json
@@ -2,8 +2,13 @@
   "profiles": {
     "Fable.Cli": {
       "commandName": "Project",
-      "commandLineArgs": "src/fable-library --outDir build/fable-library-ts --fableLib build/fable-library-ts --lang TypeScript --typedArrays false --exclude Fable.Core --define FX_NO_BIGINT --define FABLE_LIBRARY",
-      "workingDirectory": "../.."
+      "workingDirectory": "../..",
+
+      // fable-library-ts
+      // "commandLineArgs": "src/fable-library --outDir build/fable-library-ts --fableLib build/fable-library-ts --lang TypeScript --typedArrays false --exclude Fable.Core --define FX_NO_BIGINT --define FABLE_LIBRARY --noCache"
+
+      // test-ts
+      "commandLineArgs": "tests/TypeScript --outDir build/tests/TypeScript --lang TypeScript --exclude Fable.Core --noCache"
     }
   }
 }
--- a/src/Fable.Transforms/Dart/Fable2Dart.fs
+++ b/src/Fable.Transforms/Dart/Fable2Dart.fs
@@ -1479,7 +1479,7 @@ module Util =
         | Fable.Extended(kind, _range) ->
             match kind with
             | Fable.Curry(e, arity) ->
-                Dart.Replacements.curryExprAtRuntime com arity e
+                Replacements.Api.curryExprAtRuntime com arity e
                 |> transform com ctx returnStrategy
             | Fable.Throw(None, t) ->
                 [Expression.rethrowExpression(transformType com ctx t) |> Statement.ExpressionStatement], None
@@ -2283,6 +2283,7 @@ module Compiler =
             member _.OutputType = com.OutputType
             member _.ProjectFile = com.ProjectFile
             member _.SourceFiles = com.SourceFiles
+            member _.IncrementCounter() = com.IncrementCounter()
             member _.IsPrecompilingInlineFunction = com.IsPrecompilingInlineFunction
             member _.WillPrecompileInlineFunction(file) = com.WillPrecompileInlineFunction(file)
             member _.GetImplementationFile(fileName) = com.GetImplementationFile(fileName)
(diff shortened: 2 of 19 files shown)


alfonsogarciacaro:
I just found that Dart has a `Expando` type that is similar to `WeakMap` in JS. Does something similar exist in Rust?

```dart
final _curried = Expando();

TResult Function(T1, T2) uncurry2<T1, T2, TResult>(TResult Function(T2) Function(T1) f) {  
  f2(T1 a1, T2 a2) => f(a1)(a2);
  _curried[f2] = f;
  return f2;
}

TResult Function(T2) Function(T1) curry2<T1, T2, TResult>(TResult Function(T1, T2) f) {  
  var c = _curried[f];
  if (c == null) {
    return (T1 a1) => (T2 a2) => f(a1, a2);
  } else {
    return c as TResult Function(T2) Function(T1);
  }  
}

int main(List<String> args) {
  f1(int x) {
    print("Partial application: {x}");
    return (int y) => x + y;
  }
    
  var f2 = uncurry2(f1);
  var f3 = curry2(f2)(5);
  
  print(f3(5));
  print(f3(2));

  return 0;
}
```

ncave:
@alfonsogarciacaro 
> Does something similar exist in Rust?

I don't think so, perhaps we *could* add it to the function ptr wrapper, at the cost of making *each* func ptr fatter, or perhaps the global map can work (how global does it need to be)?

alfonsogarciacaro:
Uncurrying is happening when sending lambdas as arguments, so the problem is we don't know where partial application may happen. But if it's possible to make the dictionary private to the module where the currying helpers are, as in here, that should work: https://github.com/fable-compiler/Fable/blob/9c951782d49d009a636d7757614d265cf77bb7da/src/fable-library/Util.ts#L687-L693

alfonsogarciacaro:
@ncave These are the [two tests](https://github.com/fable-compiler/Fable/blob/d4e10c2fe21155d3a1e0ae8360900e46c0c02ff5/tests/Js/Main/ApplicativeTests.fs#L1250-L1290) to check if side effects are happening two many times when partial applying a function that has been previously uncurried.

@dbrattli Unfortunately I broke Python 😞 I'll try to check but if you have time could you please have a look? It seems to be an aggressive optimization that it's removing some expression but I don't know why it only affects Python.

BTW, if Python has a WeakMap-like structure (and thread-safe), you may want to use the [new pattern for currying/uncurrying](https://github.com/fable-compiler/Fable/blob/9c951782d49d009a636d7757614d265cf77bb7da/src/fable-library/Util.ts#L687-L697). It requires more helpers, but I assume it should be more performant than the fully dynamic curried function we're using now.

alfonsogarciacaro:
@ncave Actually you made me realize those tests are wrong because they are not checking the value of `counter` 😅 

If you do, they seem to fail for Rust (`counter` becomes `3`). I think it depends on whether you are compiling in Debug or Release mode. In Release mode, `adder` is inlined and it's probably not uncurried before the partial application. A way to be sure it's uncurried is to pass it to another function before partial application, like this:

```fsharp
let useAdder adder =
    let add = adder ()
    add 1 + add 2 + add 3

let ``Partially applied functions don't duplicate side effects locally`` () =
    let mutable counter = 0
    let next () =
        let result = counter
        counter <- counter + 1
        result
    let adder () =
        let add a b = a + b
        add (next())
    let result = useAdder adder
    equal 6 result
    equal 1 counter
```

BTW, in this case Rust fails because it applies the generated `arg0` in `unit` position. We should probably fix that.

```rust
            Main::useAdder(Func2::new({
                                          let adder = adder.clone();
                                          move |arg0: (), arg1: i32|
                                              adder(arg0)(arg1)
                                      }));
```

ncave:
@alfonsogarciacaro I updated the side effect tests to check the counter value.
But adding and using `useAdder`
```fs
let useAdder adder =
    let add = adder ()
    add 1 + add 2 + add 3
````
breaks the test for JS/TS too (`result = 9` instead of 6)

**Comment by alfonsogarciacaro** [comment]
I finally fixed the last issues so the PR should be ready to merge :tada:

@dbrattli The problem with Python was the same as the one for Rust mentioned above. Because now local generics are resolved, Python will complain if you try to apply an argument to a function expecting `unit`. I've solved it in Fable2Python the same way I did in Fable2Babel and seems to work. Please have a look if you've a moment.

@ncave This PR unifies currying/uncurrying (except temporarily Python) so in principle the methods `partialApplyAtRuntime` and `(un)curryExprAtRuntime` in Rust.Replacements are not needed anymore. But haven't deleted them so feel free to [bring them back](https://github.com/fable-compiler/Fable/blob/0b125d3c6e93c6ad3044fa71f16ad1948c30c1b4/src/Fable.Transforms/Replacements.Api.fs#L15) if needed.

I also found another issue with the beta reduction optimization and Rust. The test "Local_values_returning_lambdas_work" was compiling as seen below (in Release mode) and failing. My interpretation was that Fable2Rust needs to always clone mutable values captured by lambdas, but this doesn't happen if the lambda is immediately applied and inlined. So I added an extra check for Rust]

```rust
            #[test]
            pub fn Local_values_returning_lambdas_work() {
                let mutableValue_1: MutCell<i32> = MutCell::new(0i32);
                let localValueReturnsLambda = {
                    mutableValue_1.set(5i32);
                    Func0::new({
                        let mutableValue_1 = mutableValue_1.clone();
                        move || mutableValue_1.get() * 2i32
                    })
                };
                {
                    let actual: i32 = localValueReturnsLambda();
                    assert_eq!(10i32, actual,)
                }
                {
                    let actual_1: i32 = ({
                        mutableValue_1.set(7i32);
                        Func1::new({
                            let mutableValue_1 = mutableValue_1.clone();
                            move |j: i32| mutableValue_1.get() * j
                        })
                    })(9i32);
                    assert_eq!(63i32, actual_1,)
                }
                {
                    let actual_2: i32 = localValueReturnsLambda();
                    assert_eq!(14i32, actual_2,)
                }
            }
```

alfonsogarciacaro:
But we released a new REPL only a few days so the online version should already include your performance improvements 🤔

ncave:
@alfonsogarciacaro This is great then, unexpected performance gains are the best!

PR #3379, opened 2023-03-20
Add ability for compiler plugins to remove member declaration

### LINKED ISSUE #3377: Allow compiler plugins to tag a MemberDecl as a native import so that no JS is created
### Description

Right now with `[<ReactComponent(import="member", from="module")>]` syntax, the annotated function declaration still generates a dummy function in JS:

```fs
[<ReactComponent(import="Button", from="antd")>]
let AntdButton(children: string, onClick: unit -> unit) = React.imported()
```
generates:
```js
export function AntdButton(antdButtonInputProps) {
    return null;
}
```

Using `inline` just breaks the call site. Using `jsNative` or `nativeOnly` actually compile something:
```js
export function AntdButton(children, onClick) {
    throw new Error("A function supposed to be replaced by native code has been called, please check.");
}
```

It would be awesome if there is a way (maybe expect a special `Tag` in `MemberDecl` like `"nativeOnly"`) to make `[<ReactComponent>]` behave more like `[<Import>]` and not generate any _redundant_ JS code.`

### Related information

* Fable version: `dotnet fable --version` 4.0.1
* Operating system: macOS

Zaid-Ajaj:
fixes #3377

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -2497,7 +2497,8 @@ module Util =
                         |> Option.map (fun m -> hasAttribute Atts.exportDefault m.Attributes)
                         |> Option.defaultValue false)
 
-                if not isDefaultExport then decls
+                if List.contains "remove-declaration" decl.Tags then []
+                elif not isDefaultExport then decls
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks!

PR #3380, opened 2023-03-21
[Rust] Enabled some applicative tests

ncave:


alfonsogarciacaro:
Happy to see the latest changes are helping the Rust compilation!

ncave:
@alfonsogarciacaro Yes, well, this is only a very small subset that worked before too, and now I'm going to break two of them with your help ;) But hopefully all for the better.

PR #3382, opened 2023-03-22
Python currying

alfonsogarciacaro:
@dbrattli This is my attempt to unify Python uncurrying with JS/Dart after #3378. Please have a look if possible.

### DIFF
--- a/src/Fable.Transforms/Replacements.Api.fs
+++ b/src/Fable.Transforms/Replacements.Api.fs
@@ -13,21 +13,21 @@ type ICompiler = FSharp2Fable.IFableCompiler
 let curryExprAtRuntime (com: Compiler) arity (expr: Expr) =
     match com.Options.Language with
     // | Rust -> Rust.Replacements.curryExprAtRuntime com arity expr
-    | Python -> Helper.LibCall(com, "Util", "curry", expr.Type, [makeIntConst arity; expr])
+    // | Python -> Helper.LibCall(com, "Util", "curry", expr.Type, [makeIntConst arity; expr])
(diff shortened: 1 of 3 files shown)


ncave:
@alfonsogarciacaro I have a question about the curried weak map cache.
If I'm not mistaken, previously instead of a weak map the curried func version was embedded directly into the function object. Was this not working as expected, or was there another reason you changed it to a weak map (beyond potential allocation optimization from the weak ptrs)?

alfonsogarciacaro:
One reason is to try to make the solution consistent across languages including those which don't allow dynamic properties like Dart.

The other reason is the assumption, although we should benchmark it, that attaching a property to the function object may cause deoptimization in the JS frameworks engine. But maybe it's different in Python and still makes sense to attach the reference directly to the function object.

alfonsogarciacaro:
BTW, the only other place in the JS Fable library where we add an extra property to a native object is "kind" to Date. If WeakMap is indeed more optimized we could try it there too.

dbrattli (approved):
This is great work @alfonsogarciacaro. I would hire you as a Python programmer after reading this code 💯

alfonsogarciacaro:
Actually I just wrote (un)curry10 using your code as reference and let Copilot write the rest 😅 So do you think it's better for Python too to use the WeakDictionary instead of attaching the reference directly to the function?

dbrattli:
Yes, I'm a bit agnostic to that. But it feels cleaner not to patch the function object. WeakDictionary should do the trick. Copilot ftw!

alfonsogarciacaro:
Thanks @dbrattli, then let's merge this!

PR #3386, opened 2023-03-25
[Rust] Enable unused ident elimination

ncave:
This helps eliminate unused idents (e.g. computation expression builders, etc.) from appearing in the generated code:
```fs
module Identity =
    let inline run expr = expr
    type IdentityBuilder() =
        member inline __.Return(value) = value

let identity = Identity.IdentityBuilder()

let f() = identity { return 3 } |> Identity.run
```

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -377,6 +377,7 @@ type Ident =
     { Name: string
       Type: Type
       IsMutable: bool
+      IsModuleValue: bool
       IsThisArgument: bool
       IsCompilerGenerated: bool
       Range: SourceLocation option }
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -724,6 +724,11 @@ module Helpers =
         && isModuleValueForDeclarations memb
         && not(isModuleValueCompiledAsFunction com memb)
 
+    let isModuleValue com (memb: FSharpMemberOrFunctionOrValue) =
+        match memb.DeclaringEntity with
+        | Some ent -> isModuleValueForCalls com ent memb
+        | _ -> false
+
     let rec getAllInterfaceMembers (ent: FSharpEntity) =
         seq {
             yield! ent.MembersFunctionsAndValues
@@ -1659,6 +1664,7 @@ module Identifiers =
           Type = makeType ctx.GenericArgs fsRef.FullType
           IsThisArgument = fsRef.IsMemberThisValue
           IsCompilerGenerated = fsRef.IsCompilerGenerated
+          IsModuleValue = isModuleValue com fsRef
           IsMutable = isMutable
           Range = { makeRange fsRef.DeclarationLocation
                     with identifierName = Some fsRef.DisplayName } |> Some }
@@ -1984,7 +1990,10 @@ module Util =
 
         // If precompiling inline function always reference with Import and not as IdentExpr
         if not com.IsPrecompilingInlineFunction && file = com.CurrentFile then
-            { makeTypedIdent typ memberName with Range = r; IsMutable = memb.IsMutable }
+            { makeTypedIdent typ memberName with
+                Range = r
+                IsModuleValue = isModuleValue com memb
+                IsMutable = memb.IsMutable }
             |> Fable.IdentExpr
         else
             // If the overload suffix changes, we need to recompile the files that call this member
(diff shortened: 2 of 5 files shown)


ncave:
@alfonsogarciacaro I know this is a very late change to the AST, so I was wondering what options do I have at this point (besides just fixing the plugin). I want the reduction to eliminate unused idents, while at the same time convert module values to functions. Can this transformation of module values to functions be moved to after the reduction?

alfonsogarciacaro:
Sorry @ncave, there are several things here so I want to be sure I understand the goal.

- First there is the issue of adding extra information to the AST without breaking plugins, which I assume we will face multiple times.
    - We could use tags for that but my mistake is only added them to a few items, and unfortunately `Ident` is not one of them.
    - Another possibility is to do what we did for the uncurried functions and keep a dictionary on the compiler object to contain extra information about certain items.
    - And yet another alternative is what we do in other places: scan the file containing the ident to find the member the ident corresponds to.

- The second point is about removing unused idents. I assume you mean the `let identity = Identity.IdentityBuilder()` declaration that it's not necessary because the CE methods are inlined **and** the builder instance is not actually used (though I cannot see code in the PR that i's actually removing the declaration). This is part of a bigger topic we discussed in the past: dead code removal. So far we've been relying on JS tooling for that, although in this case it's probably not removed because the constructor call can potentially have side effects.

This particular optimization may indeed be difficult, the generated JS code for your example is:

```js

export class Identity_IdentityBuilder {
    "constructor"() {
    }
}

export function Identity_IdentityBuilder$reflection() {
    return class_type("Test.Identity.IdentityBuilder", void 0, Identity_IdentityBuilder);
}

export function Identity_IdentityBuilder_$ctor() {
    return new Identity_IdentityBuilder();
}

export const identity = Identity_IdentityBuilder_$ctor();

export function f() {
    return 3;
}
```

Here it looks obvious we can remove the `identity` declaration (and the class) but for more general cases, as mentioned above we need to be sure all CE members are inlined and the builder instance is not being used. In most CEs the builder is indeed not used but if all members are not inlined we need to go through all declarations and calls to remove the `this` arg, which may get complicated.

Is my understanding correct or did you have something else in mind?

ncave:
@alfonsogarciacaro 
No, sorry I was unclear, I didn't mean the extra declarations.
What I actually meant is that currently making module let bindings as function calls prevents beta reduction. You can see it if you make the builder mutable:

```fs
module Identity =
    let inline run expr = expr
    type IdentityBuilder() =
        member inline __.Return(value) = value

let mutable identity = Identity.IdentityBuilder() // <== making this mutable

let f() = identity { return 3 } |> Identity.run
```

Which results in the extra allocation (`builder$0040`) not removed:

```js
import { createAtom } from "fable-library/Util.js";

export class Identity_IdentityBuilder {
    "constructor"() {
    }
}

export function Identity_IdentityBuilder$reflection() {
    return class_type("Test.Identity.IdentityBuilder", void 0, Identity_IdentityBuilder);
}

export function Identity_IdentityBuilder_$ctor() {
    return new Identity_IdentityBuilder();
}

export let identity = createAtom(Identity_IdentityBuilder_$ctor());

export function f() {
    const builder$0040 = identity(); // <== unused, but not removed by beta reduction
    return 3;
}
```

alfonsogarciacaro:
Ah, ok. Actually this is "correct" behavior because binding to mutable idents are considered the same as side effects and they prevent beta reduction (because the referenced value may change right after the binding): https://github.com/fable-compiler/Fable/blob/552448ff909e47aab1170dbd8a7421c6d5827cab/src/Fable.Transforms/Transforms.Util.fs#L444

So in order to optimize this we would have to add extra logic to tell unused mutable idents apart from side effects, and then you can use a tag to identify the call as mutable module value. But I'm not sure if there will be many opportunities for this optimization.

BTW, the reason to transform mutable public idents into functions is because of a "limitation" of JS imports, which don't allow to modify imported values. But maybe this is not needed for Rust or other languages.

ncave:
@alfonsogarciacaro I somehow managed to communicate the wrong message again, sorry.

What I tried to show above was just an example to pinpoint where it's happening. The actual issue is about *non-mutable* idents which in Rust we also transform as function calls (cause you can't have global values except `const` and `static` which are very limited), and so it breaks beta reduction.

This PR goes around that by pushing the function call transformation to after the beta reduction. I was just wondering if there may be other ways of doing without pushing it so far down, perhaps changing the beta reduction to understand this transformation (only for non-mutable idents, point taken).

Thanks for sharing your thoughts, I'll think about it more.

alfonsogarciacaro:
Ok, I get it now 😅 

Then I'd suggest you [add a tag to those calls](https://github.com/fable-compiler/Fable/blob/552448ff909e47aab1170dbd8a7421c6d5827cab/src/Fable.Transforms/Replacements.Util.fs#L87-L90) (e.g. "no-side-effect") and then identify that tag in `canHaveSideEffects` to decide if `Call` expressions can have side effects, [as we do with FieldGet](https://github.com/fable-compiler/Fable/blob/552448ff909e47aab1170dbd8a7421c6d5827cab/src/Fable.Transforms/Transforms.Util.fs#L454-L456) (cannot have side effects if fiels is not mutable nor computed).

ncave:
Superseded by #3406.

PR #3395, opened 2023-03-31
Fix: Compiler Exception when `!!`, Anon Record, & aliased `Ux<...>`

Booksbaum:
```fsharp
type Value = U2<bool, string>
type I =
    abstract value: Value with get,set

let i: I = !!{|
    value = "foo"
|}
```
fails with compiler error:
> unreachable

[repl](https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwFgBQoksiKG2aAUgM4CSEALmFqAW-gwJ7hQBqcSAK5goAXigBVAEwAeBMGBIANFGoMsASwgBzAHwEuPWmIJQzURGqxwAxgygA3AcIBcfZyIDuGhgAso2mAMStRB7Cj2Gm7G4gCEsQDeAD6m5k5CIuIARABmClkESQC+QA&html=Q&css=Q)

Issue is: `value` is recognized as `U2`, but `GenericArguments` is empty because type is `Value` (no gen args) and not `U2<...>` ([location in code](https://github.com/fable-compiler/Fable/blob/a183c2995ed4966d6c58be9e811e531a99ec1208/src/Fable.Transforms/FSharp2Fable.Util.fs#L1361-L1404))

This PR fixes that by looking at `AbbreviatedType` (which is `U2<...>`) instead

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -1363,6 +1363,7 @@ module TypeHelpers =
                 // This is necessary because: `makeType` reduces Erased Unions (including Ux) to `Any` -> no type info any more
                 //
                 // Note: no handling of nested types: `U2<string, U<int, float>>` -> `int` & `float` don't get extract
+                let ty = ty.AbbreviatedType
                 match ty with
                 | UType tys ->
                     tys
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -1380,8 +1380,8 @@ module TypeHelpers =
                 match ty with
                 | TypeDefinition tdef ->
                     match FsEnt.FullName tdef with
-                    | Types.valueOption -> Some(ty.GenericArguments[0], true)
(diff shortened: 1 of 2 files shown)


src/Fable.Transforms/FSharp2Fable.Util.fs
```diff
@@ -1363,6 +1363,7 @@ module TypeHelpers =
                 // This is necessary because: `makeType` reduces Erased Unions (including Ux) to `Any` -> no type info any more
                 //
                 // Note: no handling of nested types: `U2<string, U<int, float>>` -> `int` & `float` don't get extract
+                let ty = ty.AbbreviatedType
```
alfonsogarciacaro, line 1366: There is a `nonAbbreviatedType` helper in FSharp2Fable.Util that deals with edge cases (e.g. alias of an alias) that could be used here. Latest FCS also has a `StripAbbreviations` method, not sure if it can be used.

alfonsogarciacaro (approved):

alfonsogarciacaro:
Thanks @Booksbaum!

Interesting timing. Actually I just sent a PR that affects this code because it doesn't erase erased types (pun unintended) from Fable AST: https://github.com/fable-compiler/Fable/pull/3396

Issue #3394, opened 2023-03-31
Fable issues with reflection in Fable-compatible library


Freymaurer:
### Description

Hey everyone, i am working on a library ([FsSpreadsheet](https://github.com/Freymaurer/FsSpreadsheet/tree/fable) and want to make one project fable compatible. I followed the official [guide lines](https://fable.io/docs/your-fable-project/author-a-fable-library.html) and started testing around with local nuget packages (`./build.cmd pack` to create nuget packages). I then created a fresh SAFE stack to test the whole thing ([here](https://github.com/Freymaurer/Fable-reflection-library-issue)). I used a `nuget.config` to add a local feed for my nuget packages and tested around. Now to my issue. [This](https://github.com/Freymaurer/FsSpreadsheet/blob/635f3d707ab7df959669264a28123a3fd9b3c4e4/src/FsSpreadsheet/Cells/FsCell.fs#L231) member function errors in fable with:

![image](https://user-images.githubusercontent.com/39732517/229076231-c8b09b7b-5657-4557-a9a1-6158ec2b69dd.png)

I then removed this function for fable using the `#if FABLE_COMPILER` logic you can see in the current version of the repo.
I added the package again and it works with the function removed. I then added the code to the test SAFE stack directly [here](https://github.com/Freymaurer/Fable-reflection-library-issue/blob/e2be2ba60b700aa687843d10d39b5e33d4e6851c/src/Client/Test.fs#L9) and it works. 

So to summarize, the same function returns an error in a Fable library, but runs if written directly into the Fable compiled project.

### Repro code

**Fable library**: https://github.com/Freymaurer/FsSpreadsheet/tree/fable
**Test SAFE stack**: https://github.com/Freymaurer/Fable-reflection-library-issue

The test SAFE stack contains to local nuget packages in `lib`. 
Version 1.1.9 has the relevant code in the library removed (therefore it works in the SAFE stack),
version 1.2.0 contains the relevant code (therefore it does not work in the SAFE stack).
Can switch between these versions [here](https://github.com/Freymaurer/Fable-reflection-library-issue/blob/e2be2ba60b700aa687843d10d39b5e33d4e6851c/src/Client/Client.fsproj#L15).

### Expected and actual results

It should work in both versions or not work in both versions.

### Related information

* Fable version: `dotnet fable --version`:  **3.7.5**
* Operating system: Windows **10**

Please let me know if you need further information 👍

MangelMaxime:
[...]

PR #3396, opened 2023-03-31
Emit Fable erased unions as TS union types

alfonsogarciacaro:
This PR stops compiling erased types as `Any` so they can be dealt with in the Fable2XXX step. @ncave @dbrattli Not sure how it affects Rust and Python, although tests are passing. For JS/TS I'm [skipping the declaration](https://github.com/fable-compiler/Fable/blob/6987bc34aefdb0a38ca2e68f89c32fc2276a0309/src/Fable.Transforms/Fable2Babel.fs#L2647-L2652) and for TS I'm transforming the annotations to TS unions/tuples when possible: https://github.com/fable-compiler/Fable/blob/6987bc34aefdb0a38ca2e68f89c32fc2276a0309/src/Fable.Transforms/Fable2Babel.fs#L617-L638

@Booksbaum An unfortunate consequence is this breaks some of the [integration tests for anonymous records casted to interfaces](https://github.com/fable-compiler/Fable/pull/3396/commits/63268122957e3a6a3c2d8d1d40f4b579a940a652#diff-fd3722e75ea03ecc628b598d94926d181420b87393e44a0bfc26e8b10945370b) because U2, U3... are not compiled as `Any` anymore. For now I just commented out the test because I think it's not a common pattern to have an anonymous record with a field like U2<string, int> (you either have string or int) but we can try to bring them back if needed. I also moved the code to deal with this from FSharp2Fable.Util to Replacements.Util, as it's only use from JS and Python replacements.

> Also, at the end I added a [commit to optimize compile-time equality](https://github.com/fable-compiler/Fable/pull/3396/commits/6987bc34aefdb0a38ca2e68f89c32fc2276a0309) to reduce generated code for patterns [like this](https://github.com/Freymaurer/FsSpreadsheet/blob/635f3d707ab7df959669264a28123a3fd9b3c4e4/src/FsSpreadsheet/Cells/FsCell.fs#L231) (see #3394).


src/Fable.Transforms/Replacements.fs
```diff
@@ -469,6 +469,7 @@ let rec equals (com: ICompiler) ctx r equal (left: Expr) (right: Expr) =
         Helper.LibCall(com, modName, "equals", Boolean, [left; right], ?loc=r) |> is equal
     | Builtin (BclGuid|BclTimeSpan|BclTimeOnly)
     | Boolean | Char | String | Number _ | MetaType ->
+        // Helper.LibCall(com, "Util", "physicalEquality", Boolean, [left; right], ?loc=r) |> is equal
```
ncave, line 472: @alfonsogarciacaro I thought the `physicalEquality` wrapper was needed, to allow equality tests like `111 === 222`, which the TS compiler frowns upon, if not wrapped in a call. Perhaps we should leave it.

alfonsogarciacaro: It's tricky, I went back to the equality operator to enable the optimization included in this PR to reduce equality checks at compile time. It seems this optimization is what prevents Typescript from complaining, but we can also "fix" the test by making the variables mutable and adding a potential side effect between the declaration and the assertion to prevent Fable from inlining them... The optimizer has become so good that now we need to fight to have "proper" tests 😂

ncave: @alfonsogarciacaro We can fix the tests to avoid it, but we can't fix user code.
So if the user does `11 = 22`, it's not going to compile for TypeScript.

alfonsogarciacaro: Sorry, by "fixing" I meant if we want to test the actual runtime equality. After this PR, if we end up with 11 = 22, either by user definition or after inlining, this will be reduced to "false" (and the if/else branch potentially discarded, see operationReduction transform in this PR) so Typescript won't get to see the equality and it should be fine.

ncave: I see, that's ok then, thanks.

PR #3400, opened 2023-04-01
[JS/TS] Use native bigint type

ncave:
- Use native JS `bigint` type in `fable-library`.
- Enabled ConvertTests, UtilTests for TS.
- Enabled ComparisonTests for TS.

2190 tests passing for TS.

Note: `bigint` needs `es2020`, but perhaps that's ok, it's 2023 already.

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -317,7 +317,7 @@ module Reflection =
         | Fable.Boolean -> jsTypeof "boolean" expr
         | Fable.Char | Fable.String _ -> jsTypeof "string" expr
         | Fable.Number(Decimal,_) -> jsInstanceof (libValue com ctx "Decimal" "default") expr
-        | Fable.Number(BigInt,_) -> jsInstanceof (libValue com ctx "BigInt/z" "BigInteger") expr
+        | Fable.Number(BigInt,_) -> jsTypeof "bigint" expr // jsInstanceof (libValue com ctx "BigInt/z" "BigInteger") expr
         | Fable.Number((Int64|UInt64),_) -> jsInstanceof (libValue com ctx "Long" "default") expr
         | Fable.Number _ -> jsTypeof "number" expr
         | Fable.Regex -> jsInstanceof (Expression.identifier("RegExp")) expr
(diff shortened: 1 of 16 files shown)


alfonsogarciacaro:
This is fantastic @ncave! I've been thinking of using native bigint for a while. It'd be great if we could use also bigint to represent int64 and decimals, so we can remove the external dependencies. For int64 we would need to [incorporate bound checks to operations and conversions](https://stackoverflow.com/a/20130355) and for decimal a [light wrapper to keep track of the decimal places](https://stackoverflow.com/a/54409977). Do you think that would be possible?

ncave:
@alfonsogarciacaro We can, but I'm not sure that we need to, it's not like int64 and decimals need much maintenance. If there are any performance reasons, sure, but I haven't seen any perf tests.

MangelMaxime:
Great work and always happy to see Fable use native solutions 😍

A wild guess from me about another benefit of using native BigInt and removing external dependencies is that the impact of using these types on the bundle size would be reduce.

I know that @alfonsogarciacaro advice to disable by default some types in Thoth.Json to reduce the bundle size impact:

CleanShot 2023-04-02 at 16 56 58@2x

alfonsogarciacaro (approved):

alfonsogarciacaro:
Makes sense @ncave. Thanks again!

ncave:
@alfonsogarciacaro To clarify, although using `bigint` for `int64`/`uint64` (instead of `long.js`) seems logical, when you add the bounds checking to each operation it becomes *2x slower* than `long.js`.

The best I could come up with was to let the runtime do the bounds checking using typed array, but that only improves it to about *1.5x slower* than `long.js`.

So if *1.5x slower* for arithmetic operations is acceptable price to get rid of that `long.js` dependency, we can use these simple wrappers for bounds checking on each `int64`/`uint64` op:

```ts
const int64buf = new BigInt64Array(1);
const uint64buf = new BigUint64Array(1);

function toInt64(x: bigint) {
    int64buf[0] = x;
    return int64buf[0];
}

function toUInt64(x: bigint) {
    uint64buf[0] = x;
    return uint64buf[0];
}
```

ncave:
@alfonsogarciacaro Actually, I stand corrected, looks like the static `bigint` wrappers `BigInt.asIntN` / `BigInt.asUintN` are performant enough (faster than `long.js`), so let me try it in a separate PR.

alfonsogarciacaro:
That'd be awesome @ncave! Thanks a lot for taking the time to measure the performance of the different options! ❤️

PR #3401, opened 2023-04-04
[JS/TS] int64/uint64 as native bigint

ncave:
- Added `int64`/`uint64` as native `bigint`.


ncave:
@alfonsogarciacaro Ready for review/merge of you're ok with it.

alfonsogarciacaro (approved):
This is great @ncave, thanks a lot!

alfonsogarciacaro:
BTW, not sure if it makes sense to use BigInt for decimals too. I found [this](https://github.com/srknzl/bigdecimal.js). It claims to be faster than big.js, but it's quite a bit of code. I was hoping for such a library to be only a thin wrapper (a bigint plus the number of decimal digits).

kerams:
@MangelMaxime, does this mean Thoth could stop requiring extra encoders for `int64` when compiling with Fable 4.0.5+?

MangelMaxime:
@kerams I guess so.

The following types are the one that need the extra coders addition right now:

- decimal
- int64
- uint64
- bigint

I created an issue on Thoth.Json repo https://github.com/thoth-org/Thoth.Json/issues/173

PR #3403, opened 2023-04-05
Optional fields and arguments in TypeScript

alfonsogarciacaro:
This PR fixes several issues with type annotation in TypeScript, but the biggest change is my attempt to make optional arguments and fields more "idiomatic" in TS. The issue with the optional arguments is they were still typed as `Option`, this didn't cause problems with compiled F# code (it could be problematic if sending a wrapped `Some(x)` option to native code) but it was not nice when exposing an API. I'm now removing the `Option` when possible, but this means sometimes we need to "unwrap" values passed as arguments. Example:

> Here we understand "unwrap" as the following [TS type conversion](https://github.com/fable-compiler/Fable/blob/6f6aba3e309cbeef54d5c072f5cc790a9910f798/src/fable-library/Option.ts#L16-L18): `Some<T> | T | undefined => T | undefined`

```fsharp
type StaticClass =
    static member Add(?x: int, ?y: int) = ...

let v: int option = ...
StaticClass.Add(5, ?y=v)
```

Becomes:

```ts
// Before it was (x?: Option<int32>, y?: Option<int32>)
function Static_Add(x?: int32, y?: int32) { ... }

StaticAdd(5, unwrap(v))
```

> **Note** This does not affect options that are always "wrapped", like nested options or `unit option`.

Likewise Interface or Anonymous Record properties with Option type are compiled as TypeScript optional fields:

```fsharp
let f (p: {| foo: int option; bar: string option |}) = ...
f {| foo = Some 5; bar = None |}
```

Becomes:

```ts
function f(p: { foo?: int32, bar?: string }) { ... }
f({ foo: 5 })
```

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -1438,14 +1438,13 @@ let private addUsedRootName (com: Compiler) name (usedRootNames: Set<string>) =
     Set.add name usedRootNames
 
 // Entities that are not output to other languages
-// Interfaces should be part of the AST, see #2673
 let private isIgnoredLeafEntity (ent: FSharpEntity) =
-    // ent.IsInterface
     ent.IsEnum
     || ent.IsMeasure
-    || ent.IsFSharpAbbreviation //&& com.Options.Language <> Rust
+    || ent.IsFSharpAbbreviation
     || ent.IsDelegate
     || ent.IsNamespace // Ignore empty namespaces
+    || ent.Attributes |> hasAttribute "Microsoft.FSharp.Core.MeasureAnnotatedAbbreviationAttribute"
 
 // In case this is a recursive module, do a first pass to get all entity and member names
 let rec private getUsedRootNames (com: Compiler) (usedNames: Set<string>) decls =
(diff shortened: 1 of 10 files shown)


src/Fable.Transforms/FSharp2Fable.fs
```diff
@@ -1438,14 +1438,13 @@ let private addUsedRootName (com: Compiler) name (usedRootNames: Set<string>) =
     Set.add name usedRootNames
 
 // Entities that are not output to other languages
-// Interfaces should be part of the AST, see #2673
 let private isIgnoredLeafEntity (ent: FSharpEntity) =
-    // ent.IsInterface
     ent.IsEnum
     || ent.IsMeasure
-    || ent.IsFSharpAbbreviation //&& com.Options.Language <> Rust
```
ncave, line 1446: @alfonsogarciacaro It's fine, but we may bring back abbreviations for Rust in the future, as they're supported by the language.

alfonsogarciacaro: @ncave Ah, sorry for that. I was just trying to clean up. But please feel free to bring it back. Actually TypeScript could also benefit from abbreviations because some types can get too verbose in APIs. The tricky part is all types are "de-abbreviated" when entering the Fable AST. We can undo that but we will have to fix all transformations that assume types are never aliases.

PR #3414, opened 2023-04-10
[TS] Sanitize class fields

ncave:
- [TS] Sanitize member names == `constructor`.
- [TS] Fix generic parameters starting with `?`.
- [TS] Get generic types of generated members.
- [TS] Fixed arrow function type annotation.


alfonsogarciacaro:
Thanks!

Issue #3411, opened 2023-04-09
[TS] Slow compilation of some union cases


ncave:
@alfonsogarciacaro I tried to make a relatively small repro, only happens for TS, not JS.

The attached file compiles 10x slower for TypeScript (than for JavaScript).
Can't really figure out why, seems to be just deeply nested ifs. There must be some additional work that's called for TypeScript quite often that slows it down.

One thing I noticed is if we remove (comment) all 3 occurrences of the case guard `when relaxWhitespace2`,
it makes it speeds up for TS as well (somehow the resulting decision tree gets compiled faster?). Not sure that helps though.

[LexFilter.fs.txt](https://github.com/fable-compiler/Fable/files/11184740/LexFilter.fs.txt)

alfonsogarciacaro:
Wow, it's slow indeed. I need to investigate but my guess is this operation takes long time and repeats a lot with many union tests: https://github.com/fable-compiler/Fable/blob/31a92cd90196dbe726845d14452b953028a6ea83/src/Fable.Transforms/Fable2Babel.fs#L1759-L1774

Probably there's an opportunity to cache the reference to the `tagIdent` as it will be retrieved many times when compiling big decision trees to TypeScript.

ncave:
@alfonsogarciacaro That was my first guess too, but unfortunately no, even if you comment the typescript portion of `transformUnionCaseTag` it's still slow. It helps a little bit, but that's not the main reason.

ncave:
@alfonsogarciacaro I tried profiling it, it's completely different profile between compiling to TS and JS.
For TS it spends a *lot* of time in `transformIfStatement` and appending arrays, and in JS, it doesn't.
I know, not very helpful. It's probably something very obvious, I just don't see it.

alfonsogarciacaro:
Thanks a lot for this! Actually I just realized the generated code is 1630 lines of code for JS... and 813,431 for TypeScript! 😮 

I've tried going back as far as 4.0.0-theta-009, which is the first tag that accepts the `quicktest-ts` target, and it also generates too much code. So there seems to be an issue remaining from the early stages of TypeScript compilation that causes code duplication in certain situations.

ncave:
> generated code is 1630 lines of code for JS... and 813,431 for TypeScript! 😮

Amazing, must be a new record :)

alfonsogarciacaro:
[...]

PR #3420, opened 2023-04-12
Use library-ts also for JS compilation

alfonsogarciacaro:
This unifies fable-library(-js) and fable-library-ts. The idea is to publish fable-library as JS but with .d.ts declarations to make sure F# code compiled both as JS and TS is compatible.

I ama having some trouble with the const enums for the union tags, so we may need to drop them and use plain ints again.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -57,6 +57,9 @@ module Util =
     let runTypeScript projectDir =
         run ("npm run tsc -- --project " + projectDir)
 
+    let runTypeScriptWithArgs projectDir args =
+        run ("npm run tsc -- --project " + projectDir + " " + String.concat " " args)
+
     let runFableWithArgs projectDir args =
         run ("dotnet run -c Release --project src/Fable.Cli -- " + projectDir + " " + String.concat " " args)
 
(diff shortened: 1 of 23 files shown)


ncave:
@alfonsogarciacaro Yes, it will be great to have .d.ts in `fable-library`, but we still want to keep the `fable-library-ts` for lang=TypeScript, right?

alfonsogarciacaro:
Yes, when embedding the fable-library-ts files we can keep doing it the same way. The use case I've in mind is publishing fable-library, then you write a library in F# with the fable-library dependency and compile it to Typescript, then to .js + .d.ts for publishing. And finally you can consume it either from a JS or TS project.

alfonsogarciacaro:
As commented above, finally this PR is giving up using `const enum` to represent union tags. This is because it caused problems with [isolated modules](https://www.typescriptlang.org/tsconfig#isolatedModules) which apparently is a requirement of many alternate TS compilers used in different toolchains. On the plus side, now TS consumers can also safely pattern match against the union name.

I have added casts in several places to make things work. @ncave I have also included your code from #3411 in the tests to make sure it compiles with TypeScript, but it'd be great if you could add [some actual assertions](https://github.com/fable-compiler/Fable/blob/603f2c63e1dbdf2cee03f3a81d9607430d4c85e6/tests/Js/Main/NestedAndRecursivePatternTests.fs#L131-L135).

ncave:
@alfonsogarciacaro It's not my code actually, it's a heavily reduced repro sourced from FCS.
I'm not sure including it will be a good test if we don't also test for file size or compile time.

alfonsogarciacaro:
It's not only the decision tree explosion, the code also gave me a chance to fix other typing issues (like [casting non-ident expressions when accessing fields](https://github.com/fable-compiler/Fable/blob/744b52c60e99350294d296b793bb48035e09b8c2/src/Fable.Transforms/Fable2Babel.fs#L1752-L1762)). But it's true it should be part of the integration tests. Although mocha has a timeout when running tests, maybe it's hit when the file takes too long to load. For now I've rebased the commit to add a simple assertion to make sure the code works.

PR #3421, opened 2023-04-13
[Python] Implemented many missing bigint functions

johannesmols:
First of all thank you for this amazing package.

I'm porting my F# code into TS and Python, and am using bigint's in a few places. I saw that much better support was recently added in #3400. For the Python version though, many basic functions are still missing, like multiplication or comparing.

I added most functions from the TS implementation, but didn't bother with the conversions. Getting those basic functions likely cover most use cases of bigint.

I also didn't spend time writing or running tests, so be warned that it may contain errors. I followed the Python naming pattern, and could see from my own project that it generates the references to e.g. `op_multiply`, but the implementation is missing.

Hope this helps kickstart better bigint support for Python! :)

### DIFF
--- a/src/fable-library-py/fable_library/big_int.py
+++ b/src/fable-library-py/fable_library/big_int.py
@@ -288,6 +288,7 @@ __all__ = [
     "op_left_shift",
     "op_bitwise_and",
     "op_bitwise_or",
+    "op_exclusive_or",
     "op_less_than",
     "op_less_than_or_equal",
     "op_greater_than",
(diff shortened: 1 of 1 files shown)


ncave:
@johannesmols This is a great first contribution, thanks!
I think you missed exporting `op_exclusive_or`, but otherwise looks good.

johannesmols:
Hi @ncave, thanks! I added the missing method now.
Is that all that is required to make this work? I'm not familiar with the codebase, just found the Python file and thought that looked like the right place :)

PR #3434, opened 2023-04-25
Improve switch optimization, TS string enums and JsDocs

alfonsogarciacaro:
### Enable annotations for `StringEnum` and `TypeScriptTaggedUnion` unions

~~For now I'm just inlining the annotation everywhere and not emitting the declaration, although this may make the code quite verbose sometimes.~~

### Transform XmlDocs as JsDocs

This is activated also for JS (better only for TS?). For now it's the **summary** of classes, dettached members and interface members. I am not adding the docs to attached members because that requires telling interface implementations apart to prevent comment duplication.

### Improve nested ifs to switch optimization

Cleanup of the code for decision trees and unification of switch optimization with general IfThenElse compilation (sometimes there are also ifs nested in a decision tree that were missed before)

### DIFF
--- a/src/Fable.AST/Common.fs
+++ b/src/Fable.AST/Common.fs
@@ -8,7 +8,8 @@ type Position =
 type SourceLocation =
     { start: Position
       ``end``: Position
-      /// We added the display name here because it seemed to be used by Babel source map generation
+      /// Used to pass the display name of the identifier, but also the name of the file
+      /// as a hack in order not to break the API
       identifierName: string option }
     static member (+)(r1, r2) =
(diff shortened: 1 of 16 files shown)


ncave:
@alfonsogarciacaro It's probably hard to completely reverse the [lossy transformation](https://github.com/fable-compiler/Fable/blob/main/src/fcs-fable/src/Compiler/Symbols/Exprs.fs#L1271) that FCS does (as [discussed before](https://github.com/fable-compiler/Fable/pull/2859#issuecomment-1105090552)), but I hope you're improveming on the status quo.

PR #3462, opened 2023-05-21
Some small php fixes

entropitor:
This are some more fixes for php (similar to https://github.com/fable-compiler/Fable/pull/3321). I think I'm now able to fully transpile a small fsharp code base into working php code


alfonsogarciacaro:
Awesome, thank you!

PR #3467, opened 2023-05-27
[Rust] Updated usage counting

ncave:
- [Rust] Updated ident usage counting, to lower the number of unnecessary clones.
- [Rust] Enabled some applicative tests.


ncave:
@alexswan10k Just FYI, the updated usage counting in this PR adds a small number of net new clones in certain cases, but removes a much larger number of clones in other cases, so overall is IMO a win.

I kept the usage client mutation to remove the last ident clone, but I still think it's brittle, as it may very easily break if somehow the removed ident clone is not the last.

PR #3469, opened 2023-06-06
[Rust] Added const, unsafe, extern func attributes

### LINKED ISSUE #3468: proposal for rust: make it easy to define `#[repr(C)]` structs and `#[no_mangle] extern "C" fn` constructs from F# code
### Description

I am playing with the rust target and modified this repository to enable:

```fsharp
[<ReprCAttr;Struct>]
type ExternStruct =
    {
        a: int
        b: int
    }
[<NoMangleExternCAttr>]
let extern_c_function data =
    printfn $"a: {data.a} b: {data.b}"
```

to be supported by rust target, turning it into functions that can bind to C:

```rust
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd, Hash, Eq,
         Ord,)]
pub struct ExternStruct {
    pub a: i32,
    pub b: i32,
}
#[no_mangle]
pub extern "C" fn extern_c_function(data: Main::ExternStruct) {
    println!("a: {} b: {}", &data.a, &data.b,);
}
```

Currently, with current compiler, you can only add `#[repr(C)]` and `#[no_mangle]` but can't produce `extern "C"`.

Wondering if there is a way around that I didn't find.

@ncave if this makes sense, I can make a PR for you to review and help me adjust the changes if wanted.

ncave:
- [Rust] Added const, unsafe, extern func attributes.
- Fixes #3468.


smoothdeveloper:
Looking good @ncave, thanks!

PR #3499, opened 2023-07-30
add test case for issue with bigint to decimal conversion

ThisFunctionalTom:
Converting `Decimal.MinValue` to `bigint` and back to decimal returns -1 instead of `Decimal.MinValue`


ncave:
Superseded by #3846

Issue #3496, opened 2023-07-20
python target is confused with class taking a `length: int` argument


smoothdeveloper:
```fsharp
type Class(length: int) =
    member x.Length = length

let c = Class(10)
printfn $"{c.Length}"
```

AFAIU, it confuses access to `length` value and turns it into `len(c)`.

```python
def _expr126() -> TypeInfo:
    return class_type("Program.Class", None, Class)

class Class:
    def __init__(self, length: int) -> None:
        self.length: int = length or 0

Class_reflection = _expr126

def Class__ctor_Z524259A4(length: int) -> Class:
    return Class(length)

def Class__get_Length(x: Class) -> int:
    return len(x) # this does not look right

to_console(("" + str(Class__get_Length(Class__ctor_Z524259A4(10)))) + "")
```

workaround is to bind `length`:

```fsharp
type Class(length: int) =
    let length = length
    member x.Length = length
```

related: https://github.com/matthewcrews/SliceMap/pull/4

cc: @dbrattli

dbrattli:
Bug identified and working on fix.

dbrattli:
Fix is merged and will be available in the next version of Fable.

PR #3505, opened 2023-08-13
WIP fix for #3496

smoothdeveloper:
Not to be merged as it fails ``test Can type test interfaces decorated with Global`` in TestPyInterop.

I'm not sure how to deal with this, it would be needed to resolve the `[<Global("Array")>]` attribute adorned to `PyArray` or some other approach.

cc: @dbrattli


dbrattli:
I'm currently trying to refactor things so `x,Length` is not translated to `len(x)` since length is not an interface member and should not be translated to `__len__` protocol in Python. Perhaps restrict to `ICollection.Count` to be automatically translated to `__len__` and `x.Count()` to become `len(x)`.

dbrattli:
I've merged the test from this PR into #3540 so closing this PR in favour of #3540

Issue #3528, opened 2023-09-17
Function hidden by a signature file should not be exported 


nojaf:
### Description

When using signature files to hide implementation details, I would expect every function that is not in the signature file, not to do be exported in JavaScript

### Repro code

Lib.fsi:

```fsharp
module Lib

val f: int -> int
```

Lib.fs:

```fsharp
module Lib

let g a  = a + 1
let f a = g a
```

### Expected and actual results

Expected:
```js
function g(a) {
    return a + 1;
}

export function f(a) {
    return g(a);
}
```

Actual:
```js
export function g(a) {
    return a + 1;
}

export function f(a) {
    return g(a);
}
```

The workaround is to mark `g` as private in the implementation file, but given the signature that should not be necessary.

### Related information

* Fable version: `4.1.4`
* Operating system: Ubuntu (WSL2)

I would be interested in contributing a fix for this if I get a pointer where to start.

MangelMaxime:
Hello @nojaf,

The way I would go to fix this is issue is to is to make the `src/quicktest/QuickTest.fs` contains your minimal reproduction.

And then, I would either use the debugger or `printfn` to track down which portion of the printer is responsible for generating the code.

My guess is that this is this portion:

https://github.com/fable-compiler/Fable/blob/dd4d6c986c1083495044ed29e9f9d11abe4dfd66/src/Fable.Transforms/BabelPrinter.fs#L602-L605

If that's the case, I would then look which portion of the code is responsible for generating `ExportNamedDeclaration` which seems to be:

https://github.com/fable-compiler/Fable/blob/dd4d6c986c1083495044ed29e9f9d11abe4dfd66/src/Fable.Transforms/Fable2Babel.fs#L2455-L2457

And then, I would check if the `isPublic` is determine correctly at each invocation places.

Sorry, I don't have better pointer to gives you. If you need more help or information, don't hesitate to ping me here or on Slack.

PR #3530, opened 2023-09-18
Consider functions hidden by a signature file as private.

nojaf:
Hi @MangelMaxime, thanks for the pointer in https://github.com/fable-compiler/Fable/issues/3528.

I was able to resolve the problem, but I'm unsure about how I can add a proper test for this.
Any thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -213,3 +213,6 @@ project.lock.json
 __pycache__/
 .python-version
 .hypothesis
+
+# Compilation tests
+tests/**/*.actual
--- a/build.fsx
+++ b/build.fsx
@@ -500,9 +500,8 @@ let testProjectConfigs() =
     )
 
 let testIntegration() =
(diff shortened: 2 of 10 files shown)


MangelMaxime:
Hum, it is true that we don't have syntax tests, you can perhaps create one for this one. Like you store in a file somewhere the expected generated file and compare it to the actual generated file.

The other solution feels a lot more hacky to me.

1. Have a signature file + function definition
2. Use dynamic import to import the non exported definition
3. Catch the exception and say that all good this is expected to not be exposed.

nojaf:
@MangelMaxime, thanks I made an attempt at having something that asserts expected output.
I'm not that familiar with Expecto, so feel free to point out where things can improve.

MangelMaxime:
Thank you @nojaf,

I will have a look at it later this week.

PR #3543, opened 2023-10-14
Move build/Build project to src/Fable.Build

ncave:
- Move `build/Build` project to `src/Fable.Build`

@MangelMaxime Feel free to rollback if you disagree, but please consider these:
- `build` folder has been used before for build outputs (and still probably is for some builds).
- it's probably more intuitive for all sources (except tests) to be in the `src` folder.


MangelMaxime:
In theory, I changed all the reference to the `build` output to use `temp`.

If this is not the case, we should update those reference to use `temp` so everything is consistent. Regarding the location of the `build` project, in general people use `build` folder from what I saw in other projects.

I also had the idea of add sub folder to the `src` folder because I find that right now, the flat arborescence makes it a bit difficult to navigate. I wanted to propose to add sub folders like `quicktest` and `fable-library` to try to organise the projects a bit more. What do you think ?

ncave:
@MangelMaxime
> in general people use build folder

Perhaps, but not always. IMO the majority of projects on GitHub don't have a build folder, or if they do, it's used for [build artifacts](https://github.com/kriasoft/Folder-Structure-Conventions) (or, more rarely, build scripts).

> add sub folder to the src

Personally I like the flat project structure under `/src`, as it shows you everything at a glance, and you don't have to dig further down to find things. But again, it's a personal preference, so if you feel strongly about it, feel free to make changes. I really don't think it's the folder structure that prevents people from contributing more.

And I don't know the right answer, perhaps a better contributor documentation of what each project does can help increase participation, i.e. what are the moving parts, how would someone start when trying to solve an issue, etc.

Or maybe that's not the reason either, given that Fable (and F# itself) are somewhat niche, so maybe trying to increase exposure in any way possible is a better way to increase participation.

My only motivation was to try to avoid drastic changes that might perhaps inhibit other existing contributor's understanding of the project, that's all, but again, it's subjective.

TL;DR: IMO if you feel strongly about some change, by all means feel free to change things around.

PS: Thanks again for the build infrastructure rework, looks good.
And thanks for picking up the torch and keeping the releases going!

MangelMaxime:
Thank you for the kind words ❤️

> Perhaps, but not always. IMO the majority of projects on GitHub don't have a build folder, or if they do, it's used for [build artifacts](https://github.com/kriasoft/Folder-Structure-Conventions) (or, more rarely, build scripts).

True indeed

---

I don't believe neither that reworking the folder structure is a magical solution to get more contributors. I would be really surprised if this was the case. 

My idea was to enforce a bit more of consistency, for example you have some folders who use abbreviation of the languages, others the full name, and some nothing at all.

- `quicktest`
- `quicktest-rust`
- `quicktest-py`
- `quicktest-dart`

Same goes for `fable-library`.

And I was thinking that having the rule `quicktest/<full_lang>` could help enforce consistency.

I also had the idea of moving NPM related packages under a `npm` folder, like that if in the future we have Python (poetry), Rust (cargo), etc. packages they can go under their respective folders.

I don't have a strong feeling about it right now. I think it would be clearer but there are others things that we can do who will have more impact. Like finish completing the new build system by fixing fable-standalone/compiler-js compilation and working on the Fable.Core API rework to better isolate languages.

PR #3545, opened 2023-10-15
Fixes #3541

### LINKED ISSUE #3541: Interface imported in generated javascript when DU has an interface type constraint
### Description

When an interface is defined in one file and the interface is used as a generic interface type constraint for a discriminated union in another file, the generated JavaScript file containing the DU attempts to import the interface even though no code is generated for the interface.

This bug is not present in 4.1.2; it was introduced in 4.1.3 and it is still present in 4.2.2.

### Repro code

https://github.com/sasmithjr/fable-interface-union for a repo that demonstrates the error.

The interface is defined here https://github.com/sasmithjr/fable-interface-union/blob/7eedf8c90b4d7c45319f32a6e3bfc0c2d35d2e38/Source.fs

The DU with the interface type constraint is defined here https://github.com/sasmithjr/fable-interface-union/blob/7eedf8c90b4d7c45319f32a6e3bfc0c2d35d2e38/Program.fs#L16C1-L17C22

The generated `Program.fs.js` will contain `import { IErrorInterface } from "./Source.fs.js";` even though no `Source.fs.js` is created, and if a `Source.fs.js` was created (e.g. the file contains functions), there wouldn't be an `IErrorInterface` in it to import anyways.

In `Program.fs`, you can uncomment the types defined above `ErrorUnion` and then comment out `ErrorUnion`, and you'll be able see that the other types do not generate code which attempts to import the interface.

The generated JavaScript does not import the interface.

### Related information

* Operation system: macOS 14.0
* .NET SDK: 7.0.401 and 8.0.100-rc.2.23502.2
* Fable compiler: 4.1.3 and 4.2.2

ncave:
- Fixes #3541

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -83,7 +83,10 @@ module Lib =
     let tryJsConstructorFor purpose (com: IBabelCompiler) ctx (ent: Fable.Entity) =
         let isErased =
             match purpose with
-            | Annotation -> ent.IsMeasure || (FSharp2Fable.Util.isErasedOrStringEnumEntity ent && not ent.IsFSharpUnion)
+            | Annotation ->
+                ent.IsMeasure
+                || (ent.IsInterface && not com.IsTypeScript)
+                || (FSharp2Fable.Util.isErasedOrStringEnumEntity ent && not ent.IsFSharpUnion)
             // Historically we have used interfaces to represent JS classes in bindings,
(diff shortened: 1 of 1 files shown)


MangelMaxime:
Thank you for the fix @ncave .

The new build system supports Unreleased section in the CHANGELOG.md files (previously RELEASES_NOTES.md). For future PRs can you please update the changelog files in the PR too?

This is to make the release process easier and avoid forgetting about an item where preparing the release. Nothing is needed on your side for this PR, I updated the CHANGELOG file in the main branch for this PR

Issue #3514, opened 2023-08-26
curriing limit to 10 is low


slyshykO:
### Description

I've tried to create a form with 11 fields with Fable.Form .
When compiling to JavaScript Fable generates the wrong function call & export

```js

import { createObj, curry11 } from "./fable_modules/fable-library.4.1.4/Util.js";
...
return append(extra0Field, append(extra0Field, append(extra0Field, append(extra0Field, append(extra0Field, append(extra0Field, append(extra1Field, append(extra0Field, append(rememberMe, append(passwordField, append(emailField, succeed(curry11(onSubmit)))))))))))));
...
```

Fable 4.0.1 generate the working code.

### Repro code

https://github.com/slyshykO/big-form

### Expected and actual results

Fable compile working code.

### Related information

* Fable version: 4.1.4
* Operating system: win11 x64

MangelMaxime:
When running fable in non watch mode (`dotnet fable src`) the generated is correct and support infinite curry as requested in https://github.com/fable-compiler/Fable/issues/2525

However, if running it in watch mode (`dotnet fable watch src`) then it tries to use `curry11` which is invalid.

Fable `4.0.2` is the version which introduced this regression.

If you revert to Fable `4.0.1` then your application should works, you will have a few bug fixes missing but most of the changes are related to the others Fable target like TypeScript, Rust, Python, etc. So I hope for now it can be a workaround for your project.

MangelMaxime:
@ncave I am looking into fixing this issue because having infinite currying/uncurrying is really important for libraries like Thoth.Json and Fable.Form.

I found the previous implementation of the `curry` function:

**fable-library/Util.ts**

```ts
const CURRIED = Symbol("curried");

function _curry(args: any[],  arity: number, f: Function): (x: any) => any {
  return (arg: any) => arity === 1
    ? f(...args.concat([arg]))
    // Note it's important to generate a new args array every time
    // because a partially applied function can be run multiple times
    : _curry(args.concat([arg]), arity - 1, f);
}

export function curry(arity: number, f: Function): Function | undefined {
  if (f == null || f.length === 1) {
    return f;
  }
  else if (CURRIED in f) {
    return (f as any)[CURRIED];
  } else {
    return _curry([], arity, f);
  }
}
```

[...]

PR #3555, opened 2023-10-19
Support (un)curry up to 20 arguments

MangelMaxime:
#3514

@dbrattli Can you please have a look at the `fable-library-py/Util.py`.

I don't understand why Pylance generates errors when the function as more that 10 arguments.

CleanShot 2023-10-19 at 22 15 42@2x

I tried to check for syntax errors, but could not find any.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -8,9 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+#### JavaScript
+
+* Support (un)curry up to 20 arguments (by @MangelMaxime)
+
 #### Python
 
 * Remove support for Python 3.9. Add GH testing for Python 3.12 (by @dbrattli)
+* Support (un)curry up to 20 arguments (by @MangelMaxime)
+
+#### Dart
+
+* Support (un)curry up to 20 arguments (by @MangelMaxime)
 
 ## 4.3.0 - 2023-10-14
 
--- a/src/fable-library-dart/Util.dart
+++ b/src/fable-library-dart/Util.dart
@@ -235,6 +235,692 @@ TResult Function(T10) Function(T9) Function(T8) Function(T7) Function(T6)
   }
 }
 
+TResult Function(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11)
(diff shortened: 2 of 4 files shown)


dbrattli:
@MangelMaxime the thing here is that Python is a bit different with typing than e.g. F#. A generic type argument in Python must be used more than one time in a function signature to make sense (which makes sense for F# as well when you think about it). So, if a function has e.g a type `_T11` that is used only once, you can just as well replace it with `Any`. 

So, for the outer function signature you can see that the type arguments make sense because it locks the type of the input arguments to the output arguments. But for the inner functions e.g:

```py
def f2(a1: _T1, a2: _T2, a3: _T3, a4: _T4, a5: _T5) -> _TResult:
        return f(a1)(a2)(a3)(a4)(a5)
```

You can see that the type arguments are all different. Thus, they do not provide any additional value. No input is locked to any output. They do not inherit the outer type arguments, so they are not locked to the outer function declaration either. E.g a function like:

```py
def id(value: _T1) -> _T2:
   return value
```

is no different than
```py
def id(value: Any) -> Any:
   return value
```

Thus, the inner functions should be written like this instead to avoid the type errors you see:

```py
def f2(a1: Any, a2: Any, a3: Any, a4: Any, a5: Any) -> Any:
        return f(a1)(a2)(a3)(a4)(a5)
```

MangelMaxime:
Thanks for the explanation but why does it work for `uncurry2` to `uncurry10`?

And doesn't anymore starting from `uncurry11`?

If I understand correctly the function below should have the errors too on `f2` definition but that's not the case.

```py 
def uncurry2(
    f: Callable[[_T1], Callable[[_T2], _TResult]]
) -> Callable[[_T1, _T2], _TResult]:
    def f2(a1: _T1, a2: _T2) -> _TResult:
        return f(a1)(a2)

    _curried[f2] = f
    return f2
```

Another interrogation is why would the error only be on `_T11`, `_T12`, `_T13`, etc. and not all the types starting from `_T1` ?

https://github.com/fable-compiler/Fable/assets/4760796/9913dc22-5d7d-4927-8a42-343e5c06a142

Replacing `_TX` with `Any` for `uncurry11` and plus, makes the error goes away. If you are fine with that I can make the change.

dbrattli:
Yes, it's fine. We have no clue about what`TX` is and it doesn't really matter. It's just important that the signature of the outer function locks input types to output types.

Issue #3552, opened 2023-10-18
Fable as a library


nojaf:
### Description

It is related to https://github.com/fable-compiler/Fable/issues/3549.

Could we compile F# code using Fable in a library scenario?
Similar to how FCS works today. The things I would be most interested in are getting Fable AST and compiling JS programmatically.

I envision this would be a carve-out of what `Fable.Cli` does today.
Things like project cracking, the file watcher and writing files to disk should remain in `Fable.Cli`.

MangelMaxime:
I have a difficult time seeing what would be the separations in term of features/API between the packages.

I suppose for `Fable.Cli`, the pipeline would be something like:

1. You provide an fsproj to it
2. It handles project cracking
3. Call `Fable.Compiler.transformFile`
4. Call `Fable.Compiler.Transforms.transformFile`
5. Call `Fable.Compiler.JavaScript.transformFile`
6. Write output to the disk

Still need to find where the `FableCompiler` instances needs to live in.

> Things like project cracking, the file watcher and writing files to disk should remain in `Fable.Cli`.

I don't get that part, if you keep the project cracking inside of `Fable.Cli` and not in the `Fable.Compiler` library. How will you restore the dependencies inside of your Vite plugin ?

Things would be different when thinking of the MSBuild integration because I believe inside of MSBuild, we can have access to the restored dependencies etc.

nojaf:
Project cracking would still need to be solved for the Vite plugin. However, there are situations where you want to go with a hardcoded or custom project. I'm thinking about some serverless function that would convert F# into Fable AST or JS. The compilation should not need to care where the input information came from.

There are multiple ways of solving the project cracking. The MSBuild route would probably get all the information first-hand, so you don't want to take any dependency on what `Fable.Cli` uses.

As for the Vite Plugin, I was actually thinking about trying the new MSBuild capabilities announced in dotnet 8 rc 2. It could be convenient to run a design-time build from the dotnet cli and capture all the arguments from there. 

Similar to how FCS works, you also need to provide the `FSharpProject` and they also don't provide a good way to construct those.

ncave:
@nojaf Just FYI, the JS-only `fable-compiler-js` uses a simple [ProjectCracker](https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/src/ProjectParser.fs) using Regex instead of MSBuild, in order to be usable on systems that don't even have .NET installed.

[...]

Issue #3549, opened 2023-10-17
MSBuild/NuGet integration: custom TargetFramework


jwosty:
### Description

There should be a Fable MSBuild SDK (read: https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2022) which provides tasks and targets to compile Fable code. If this were done, setting up a Fable project could be as simple as referencing the SDK, setting `<TargetFramework>fable</TargetFramework>`, then running `dotnet build`.

A proof-of-concept of exactly this grew out of a discussion we had in the FSSF Slack channel: https://github.com/jwosty/Fable.Sdk

This POC mainly answers the question: "Can we integrate Fable into the dotnet build process, with its own TargetFramework, and make it work with built-in MSBuild mechanisms (like package restoring)"? I think this shows the answer is: **Yes**. The main question left is whether or not NuGet package support is possible without NuGet having to know about a Fable TFM (see: section below)

# Technical details

Some better integrations with the Fable compilation process would be awesome, and would allow for a much better implementation of this SDK. Currently, the POC accomplishes compilation through [a `FableCompile` task](https://github.com/jwosty/Fable.Sdk/blob/7cb537fbd8092705274d6e887b23c7778e5d989f/src/Fable.Sdk/Build/Core.targets#L70) which just [runs the Fable CLI](https://github.com/jwosty/Fable.Sdk/blob/7cb537fbd8092705274d6e887b23c7778e5d989f/src/Fable.Sdk/FableCompile.fs#L156). For example, if Fable either exposed some of its internal functions which perform compilation publicly, or if the CLI allowed you to manually provide all the inputs (F# source files, package references, etc, a la `fsc`), then the SDK could completely bypass having to deal with an extra fsproj (no need for Buildalyzer when we have the real thing!).

# Open questions

Below are some unanswered questions, as well as some design decisions that could go one of multiple ways:

1. Does NuGet.org allow custom target frameworks in packages? Or are we going to have to work with Microsoft to get it added to a hardcoded list for restoring to work?
2. Should the TargetFramework be completely standalone (i.e. `fable`), or should it extend a .NET TFM (i.e. `net8.0-fable`, like `net8.0-mac`? There are pros and cons to each.
3. Should the SDK concern itself with JS package managers (i.e. yarn, npm)? If so, how? Should it have special knowledge?
4. Should the SDK concern itself with JS bundlers (i.e. webpack, vite)? If so, how? Should it have special knowledge?
[...]

PR #3561, opened 2023-10-21
Compiling as library

nojaf:
This is a very rough proof of concept for https://github.com/fable-compiler/Fable/issues/3552.
It worked on my machine, I think it is kinda cool.
I duplicated a lot of code from `Fable.Cli`, you should imagine `Fable.Cli` takes a dependency on `Fable.Compiler.Service` and the right shared core components live there.

This is just an attempt to illustrate what I think could be valuable. 
Checkout `tests/FCSTest/Program.fs` to see what a third-party tool could do with this.
    
image

I believe the MSBuild experiment in https://github.com/fable-compiler/Fable/issues/3549 could benefit from this as well.


ncave:
@nojaf I agree with the sentiment expressed in other threads that it makes more sense to rename `Fable.Compiler.Service` to just `Fable.Compiler`, otherwise it's visually very confusing.

nojaf:
Hi @ncave , I would like to polish this PR and wrap it up so we can ship the first alpha version of `Fable.Compiler`.

Two open questions:
- Could you sync your F# fork one more time to include https://github.com/ncave/fsharp/pull/2/commits/3a553a78d5b5853bcfc23ced2dee5087a0e814e7. The PR was merged but is not synced with the Fable repo yet.
- How do we wish to package the dependencies of `Fable.Compiler`? Do we just want to include the `FSharp.Core` and `FCS` lib as part of the same NuGet package? This is maybe a little less clean but would be more convenient I think.

ncave:
@nojaf 
> Could you sync your F# fork one more time?

Sure, if you don't have any more changes for it.

> Do we just want to include the `FSharp.Core` and FCS lib as part of the same NuGet package?

Probably yes, same dependencies as the `Fable.Cli` tool.

> Fable.Standalone.fsproj appear to be using fable-fcs.fsproj instead of the (FCS) binary.

Yes, `Fable.Standalone` is using a [different branch](https://github.com/ncave/fsharp/pull/1) of the FCS fork, which is more heavily modified so it can be compiled with Fable itself, mostly for the `REPL` and pure JS version of `fable-compiler-js` running on Node.js.

jwosty:
I'm the case of FSharp.Compiler, since it's a library package, shouldn't FSharp.Core be a proper dependency? I think bundling it only makes sense for the tool. FCS of course is an exception since it's a custom fork.

ncave:
@nojaf Rebased FCS in #3646, let e know if there is anything else you need.

nojaf:
Thanks @ncave! No need to do an extra sync for this, but could you include the `.pdb` files in the future?

nojaf:
@MangelMaxime even though this still has some todos, would it be ok to have this PR merged?
I'd like to follow up with future PRs instead of keeping this PR in sync any longer.

nojaf:
> The PDBs are already embedded in the assemblies, no?

Ah, yes indeed, I believe you are right. Didn't think of that.

src/Fable.Compiler/Fable.Compiler.fsproj
```diff
@@ -0,0 +1,56 @@
...
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>
+        <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+        <RootNamespace>Fable.Compiler</RootNamespace>
+        <Version>0.1.0</Version>
+    </PropertyGroup>
+
+    <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+        <!-- Generate PDB files in release mode -->
+        <DebugType>portable</DebugType>
```
ncave, line 14: @nojaf Perhaps we can keep the PDBs embedded here too (seems like that's the default, at least on .NET 8.0).

ncave:
@nojaf
> even though this still has some todos ...

I assume this will eventually support the full set of languages that Fable supports.

MangelMaxime:
> @MangelMaxime even though this still has some todos, would it be ok to have this PR merged? I'd like to follow up with future PRs instead of keeping this PR in sync any longer.

Looking at the discussion history it seems like this PR is not yet ready for a release. 

You said above:

> Needs a changelog file. Would like to introduce [KAC](https://github.com/ionide/keepachangelog) first.

We can have the KAC integration done later, current system is working fine so Fable.Compiler could just use the same publishing system as the others.

>  Needs some more though on creating the nupkg.

What more though are needed?

These questions needs to be answered because we need to be able release Fable.CLI, otherwise we will not be able to releases fixes for the different bug or improvements made to Fable.

ncave:
@nojaf A good question for the `Fable.Compiler` as a library is what is the best way to drive some of the different Fable behaviors/options that are currently passed in as command line arguments.

src/Fable.Compiler/Fable.Compiler.fsproj
nojaf: Where are you getting that? I believe the default is `portable`.
I'm fine with embedded, all I really want to do is be able to step through the `Fable.Compiler` and `FCS` code when debugging.

ncave: @nojaf I think you're right, the default seems to be `portable`.
Not sure how FCS sets it to `embedded`, but that's not important.

I guess we don't need to set it to `portable` here then, if `portable` is the default.

nojaf:
Hi everyone,

I'm keen on getting this PR merged to avoid further rebasing, which is becoming quite tedious. Just to be clear, I'm not suggesting an immediate release on NuGet. Using it locally works for me. What I'm really aiming for is to contribute more regularly but in smaller chunks.

> I assume one of the todos is to eventually support the full set of languages that Fable supports.

Eventually yes, but right now, there is no use case for anything different than JS.

> What more though are needed?

I want to run this by Chet in terms of the NuGet creation. Whether I have all the right props in place and I am adhering to the best practices. And also hook this up in the `Fable.Build` project to publish the NuGet. All things that can be done later I think.

MangelMaxime:
> Just to be clear, I'm not suggesting an immediate release on NuGet.

I understand that you are not suggesting an immediate release. 

Any fixes we want to do for Fable, require a new release of Fable.CLI. Which with this PR requires Fable.Compiler to be released meaning that it needs to have a Changelog and be hooked to the actual build process.

That's why I suggested to not use KAC for now and just use the way things are for now. KAC integration can be done later if/when needed.

Personally, I am ok with releasing a version of Fable.CLI that use the current Fable.Compiler. Even if Fable.Compiler, is not stable yet. As long as the surfacing Fable.CLI args and requirements in term of SDKs/Tools installed on the user machine doesn't change compared to what we have now, I am fine with it.

nojaf:
> Personally, I am ok with releasing a version of Fable.CLI that use the current Fable.Compiler. Even if Fable.Compiler, is not stable yet. As long as the surfacing Fable.CLI args and requirements in term of SDKs/Tools installed on the user machine doesn't change compared to what we have now, I am fine with it.

Yup, that would be the short-term plan for sure. If anything were to change somewhere in `Fable.Compiler` that affects `Fable.Cli`, it should be listed in the `Fable.Cli` changelog.
For example, a fix happens in `Glob.fs`.

src/Fable.Transforms/Fable.Transforms.fsproj
```diff
@@ -49,7 +49,8 @@
     <ProjectReference Include="Rust/AST/Rust.AST.fsproj" />
   </ItemGroup>
   <ItemGroup>
-    <Reference Include="../../lib/fcs/FSharp.Core.dll" />
-    <Reference Include="../../lib/fcs/FSharp.Compiler.Service.dll" />
+      <Reference Include="../../lib/fcs/FSharp.Compiler.Service.dll" />
+      <Reference Include="../../lib/fcs/FSharp.DependencyManager.Nuget.dll" />
```
MangelMaxime, line 53: Is it normal that we have a new "dependency" here?

I asking in case this is something that has been forgotten.

nojaf: I won't be used in Transforms, so I will remove it.
If we were using vanilla FCS that ref will be there:
image

In this case, we are not using it, so I will remove it.

MangelMaxime:
@nojaf I had to revert the merge of this PR because I could not pack `Fable.Compiler`.

```
Unhandled exception. System.AggregateException: One or more errors occurred. (The command exited with code 1.

Standard output (stdout):

MSBuild version 17.8.3+195e7f5a3 for .NET
  Determining projects to restore...
  Restored /workspaces/Fable/src/Fable.Compiler/Fable.Compiler.fsproj (in 213 ms).
  3 of 4 projects are up-to-date for restore.
  Rust.AST -> /workspaces/Fable/src/Fable.Transforms/Rust/AST/bin/Release/netstandard2.0/Rust.AST.dll
  Fable.AST -> /workspaces/Fable/src/Fable.AST/bin/Release/netstandard2.0/Fable.AST.dll
  Fable.Transforms -> /workspaces/Fable/src/Fable.Transforms/bin/Release/netstandard2.0/Fable.Transforms.dll
  Fable.Compiler -> /workspaces/Fable/src/Fable.Compiler/bin/Release/net6.0/Fable.Compiler.dll
/home/vscode/.dotnet/sdk/8.0.100/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(221,5): error NU5019: File not found: '/workspaces/Fable/src/Fable.Compiler/bin/Release/net6.0/Fable.Transforms.pdb'. [/workspaces/Fable/src/Fable.Compiler/Fable.Compiler.fsproj]


Standard error (stderr):
```

PR #3563, opened 2023-10-22
[Python] Test printing unicode characters

dbrattli:
- Print unicode character to make sure printing does not blow up
- Code cleanup and formatting

### DIFF
--- a/src/fable-library-py/fable_library/time_span.py
+++ b/src/fable-library-py/fable_library/time_span.py
@@ -146,6 +146,7 @@ def to_string(ts: timedelta, format: str = "c", _provider: Any | None = None) ->
 __all__ = [
     "create",
     "to_milliseconds",
+    "to_string",
     "from_ticks",
     "from_milliseconds",
     "from_hours",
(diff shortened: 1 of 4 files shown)


TheAngryByrd:
Well one thing I just noticed is there doesn't seem to be any python tests actually running on the windows runs

image

MangelMaxime:
If they don't run there are no errors, so it means that it works 😏

More seriously, I have PR testing a fix for the Python tests on windows here #3569 

Waiting for it to run and will merge it if all is fine.

TheAngryByrd:
While it doesn't [show the character](https://github.com/fable-compiler/Fable/actions/runs/6644757140/job/18054573658#step:10:33) it doesn't seem to blow up. Puzzling.

MangelMaxime:
@dbrattli Which formatter are you using for Python code?

This is so I can install it on my machine too when contributing to the python code.

dbrattli:
@MangelMaxime I'm using `ruff format`, https://github.com/astral-sh/ruff#usage

MangelMaxime:
Ok thank you @dbrattli.

I think one of the next thing I want to add to the repo is auto format of the different source code when possible, just so we all write the same things. And it will also clean up some of the files who use odd formatting rules

dbrattli:
@MangelMaxime I'll open an issue on this topic so we can discuss more

PR #3572, opened 2023-10-27
Add applicative test for accessor function shorthand

MangelMaxime:
@dbrattli @ncave 

Is it the correct way and place to add these tests? 

I was not able to find the RFC, so I don't know if there are others cases to cover.

Edit: The CI is broken because I don't know how to setup .NET 8 on it. Do you know if this is possible to install .NET 8 even if it is still in preview?

When running the tests on my machine, the tests are passing.


ncave:
@MangelMaxime 
> Do you know if this is possible to install .NET 8 even if it is still in preview?

Yes, it's [possible](https://learn.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual#example), basically download the [correct binary](https://dotnet.microsoft.com/en-us/download/dotnet/8.0), unzip in a folder and set env variable to point to that folder. Something like this:
```bash
wget https://download.visualstudio.microsoft.com/download/pr/9144f37e-b370-41ee-a86f-2d2a69251652/bc1d544112ec134184a5aec7f7a1eaf9/dotnet-sdk-8.0.100-rc.2.23502.2-linux-x64.tar.gz

DOTNET_FILE=dotnet-sdk-8.0.100-rc.2.23502.2-linux-x64.tar.gz
export DOTNET_ROOT=$(pwd)/.dotnet

mkdir -p "$DOTNET_ROOT" && tar zxf "$DOTNET_FILE" -C "$DOTNET_ROOT"

export PATH=$PATH:$DOTNET_ROOT
```
But perhaps we can wait two more weeks for the official .NET 8.0 release and then add those tests, it will be easier to do the CI matrix.

MangelMaxime:
Let's way for the officiel release of .NET 8.0.

I don't think this is worth the effort to add custom CI rules just to support .NET 8 for 2 weeks.

MangelMaxime:
Not all F# 8 features has been added yet, but this is a good start I believe.

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.100",
+    "version": "8.0.0",
```
ncave, line 3: @MangelMaxime Should be 8.0.100, no?

MangelMaxime: You are right indeed, I suppose because we have `"rollForward": "latestMinor"` it didn't break on me.

But better to use a real version

PR #3578, opened 2023-10-31
[Python] Format generated files using ruff

dbrattli:
Use [Ruff formatter](https://astral.sh/blog/the-ruff-formatter) on the generated Python output files. This makes the generated files easier to read and also makes it easier to spot any issues with the generated code. Ruff format is Black compatible but super fast so it doesn't really add much to the total compile time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,10 +109,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install dependencies
+      - name: Install poetry
         run: |
           pip install poetry
-        # poetry install
 
       - name: Fable Library - Python (linux)
         if: matrix.platform == 'ubuntu-latest'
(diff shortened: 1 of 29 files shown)


dbrattli:
@MangelMaxime Any ideas about this? I'm not familiar with the new build system, so perhaps there's something obvious I'm doing wrong? Works locally btw on my mac. https://github.com/fable-compiler/Fable/actions/runs/6703457876/job/18214036253?pr=3578#step:6:52

MangelMaxime:
@dbrattli It seems like the CI is fixed now.

For me the problem seems to be that you were trying to run `poetry` before it was installed on the CI.

And you also tried to invoke it from inside the `temp/fable-libary-py/library_py` folder when the tool is installed at the root of the repository.

One question that I have is do we still need to have `fable-library` reference in the top level `pyproject`?

https://github.com/fable-compiler/Fable/blob/5d309a0adff213c22de2fed9cf08db3d8fbe478a/pyproject.toml#L15-L16

Because of that line, I had to not restore the `poetry` dependencies directly in the CI commands but only after `fable-library-py` has been built otherwise the path is missing.

MangelMaxime:
No problem, happy to help on what I can. :)

PR #3600, opened 2023-11-19
Improvements to the Datetime support on Python

MangelMaxime:
The module still don't support all the formatting of the date, but the change made are enough for `Thoth.Json` to encode the `datetime` correctly.


dbrattli:
@MangelMaxime Looks great, but I took the liberty of rewriting the if-else logic into a match statement that we can now use from Python 3.10+. Not as powerful as F# match expressions, but quite nice nonetheless

dbrattli (approved):

MangelMaxime:
Oh great thank you @dbrattli

Issue #3596, opened 2023-11-17
Setup auto formatting of source code


MangelMaxime:
It would be nice to have auto formatting of the code so we can unify the styling in the project.

For F# we can use Fantomas

Here is a setting configuration that I often use

```yml
[*.{fs,fsx}]
max_line_length = 80
fsharp_alternative_long_member_definitions = true
fsharp_multi_line_lambda_closing_newline = true
fsharp_bar_before_discriminated_union_declaration = true
fsharp_multiline_bracket_style = aligned
fsharp_keep_max_number_of_blank_lines = 2
fsharp_record_multiline_formatter = number_of_items
fsharp_array_or_list_multiline_formatter = number_of_items
fsharp_align_function_signature_to_indentation = true
fsharp_multi_line_lambda_closing_newline = true
fsharp_max_infix_operator_expression = 30
fsharp_max_if_then_else_short_width = 0
```

For Python, we can use `ruff`

ncave:
@MangelMaxime Please exclude the `src/fcs-fable` folder, as it needs to retain the original formatting from upstream.

MangelMaxime:
@ncave I will make sure to ignore that folder 👍

MangelMaxime:
@nojaf It looks like the old setting is a subset of the one I proposed.

So for me, I am in favour of replacing them we what I proposed but obviously I am biased ^^.

To explains the motivations of my setting is that I try to have a code formatting which is consistent independently from the numbers of elements. If the array/record/member has 1, 2 or 10 arguments I prefer to have it formatted in the same way.

---

The implementation I had in mind is to have the CI check if the code is formatted correctly. I don't know if all the tools allows that but I know that fantomas does.

And also, try to leverage git hooks to auto format the code on commits. Like that this is something we can "forget" about as it is done automatically.

nojaf:
> @nojaf It looks like the old setting is a subset of the one I proposed.

Well, it has `fsharp_multiline_bracket_style = stroustrup` and you are proposing `aligned`.
Which one do you want to go with?

MangelMaxime:
Oh my bad, I missed that difference.

Now I questioning my whole life as I always though I was using `stroustrup`, let's go with `aglined` as you did in the PR

cr3wdayt5p:
@MangelMaxime Seems like this can be closed :) Fantomas has been enabled.

PR #3609, opened 2023-11-23
Format fsharp code

nojaf:
See https://github.com/fable-compiler/Fable/issues/3596

When it comes to git hooks, you could add [this one](https://fsprojects.github.io/fantomas/docs/end-users/GitHooks.html). However, a git hook is typically something personal which you don't put in source control (as it lives in the `.git` folder).

I've added the [commands](https://github.com/dotnet/fsharp/blob/main/.github/workflows/commands.yml) the F# team is using to format their code by commenting `/run fantomas` on the PR.


MangelMaxime:
For the git hooks, we could use [Husky.Net](https://alirezanet.github.io/Husky.Net/).

This is a tool which allows you to commit the hooks information and then you can use `dotnet husky attach` to automate the installation.

We can run `dotnet husky attach` at the top of the build process:

```fs
[<EntryPoint>]
let main argv =
    let argv = argv |> Array.map (fun x -> x.ToLower()) |> Array.toList

	// Run dotnet husky attach here

    match argv with
    | "fable-library" :: args ->
```

What do you think?

nojaf:
Interesting idea, I never heard of Husky.
Would be nice to play around with.
If the experience is positive, we might even recommend it in the Fantomas docs.

MangelMaxime:
Thank you @nojaf

PR #3618, opened 2023-11-28
Fix comparaison between optional list when one is None

### LINKED ISSUE #3617: [Python] Comparison with None fails
### Description

Transpiled comparison of `Some List` and `None` fails, as the equality tries to compare fields which don't exist on None.

### Repro code

```fsharp
Some [1;2;3] = None
```

Expected result: should return true

Actual result: `AttributeError: 'NoneType' object has no attribute 'tail'`

### Related information

* `dotnet fable --version`: 4.5.0
* `dotnet tool list/update/install`: 4.6.0
* Windows11

@freymaurer

MangelMaxime:
Fix #3617

The error actually comes from:

CleanShot 2023-11-28 at 11 34 08@2x

Because the code expect `xs_1` and `ys_1` to be list. But because the `def equal`  in `util.py` doesn't capture the `Some .. vs None` difference early, then we try to access `.tail` on a `None` value.

I believe JavaScript implementation does something similar to capture these case early (line 474 - 476)

https://github.com/fable-compiler/Fable/blob/8ee247a9f9aa3c25155e5390bf0a369a8a03f4da/src/fable-library/Util.ts#L471-L489

I am not sure what `a is b` does in Python, so I let it there and disabled the `None, None` check as it was already handled by `a is b`.

Also, I can't "fix" `List.equal` definition because this is coming from `List.fs` so I expect that it's implementation is correct and behaving the same as in FSharp.Core

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+#### Python
+
+* Fix #3617: Fix comparaison between list option when one is None
+
 ## 4.6.0 - 2023-11-27
 
 ### Changed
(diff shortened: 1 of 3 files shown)


dbrattli (approved):
Very nice. I took the liberty to move all the test logic inside the match statement

MangelMaxime:
> I took the liberty to move all the test logic inside the match statement. 

You did good.

Thanks for explanation for `a i b`

PR #3620, opened 2023-11-28
minor cleanup: Removed double-dictionary lookups

Thorium:
Tiny code cleanup: Removed double-dictionary lookups and other structures a bit better


MangelMaxime:
Thank you, looks good to me

PR #3643, opened 2023-12-05
[Python] Add literal kind field to constant

dbrattli:
This should hopefully avoid type checking to decide between integers and floats which fails in the REPL (js)

Screenshot 2023-12-02 at 09 44 20

### DIFF
--- a/src/Fable.Transforms/Python/Python.fs
+++ b/src/Fable.Transforms/Python/Python.fs
@@ -27,7 +27,7 @@ type Expression =
     /// A constant value. The value attribute of the Constant literal contains the Python object it represents. The
     /// values represented can be simple types such as a number, string or None, but also immutable container types
     /// (tuples and frozensets) if all of their elements are constant.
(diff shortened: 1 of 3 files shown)


src/Fable.Transforms/Python/Python.fs
```diff
@@ -43,6 +43,16 @@ type Expression =
         upper: Expression option *
         step: Expression option
 
+type Literal =
+    | FloatLiteral of float
+    | IntLiteral of obj
```
MangelMaxime, line 48: @dbrattli 

Should it not be of type `int`?

dbrattli: @MangelMaxime No it was intentionally since we need to store every kind of int from (u)int8 to (u)int64 and even BigInteger. Int will not do since we will get both sign and overflow problems

MangelMaxime: Oh I see

PR #3645, opened 2023-12-07
Add `TimeSpan.Parse` and `TimeSpan.TryParse` support to Python

MangelMaxime:
The tests have been ported from the JavaScript implementation, there are some still marked as todo like in the JavaScript implementation didn't look at why there is a todo yet.

The only test that is different is one that is marked as not passing on JavaScript, than pass with our Python implementation. I believe this is related to the fact that we have a better precision in our Python implementation.


src/fable-library-py/fable_library/time_span.py
```diff
@@ -178,6 +182,74 @@ def to_string(ts: TimeSpan, format: str = "c", _provider: Any | None = None) ->
...
+                        ms = int(g_8) / 1000
+                    case 7:
+                        ms = int(g_8) / 10000
+                    case _:
+                        raise Exception("String '%s' was not recognized as a valid TimeSpan." % string)
+            return multiply(create(d, h, m, s, ms), sign)
+    raise Exception("String '%s' was not recognized as a valid TimeSpan." % string)
+
+
+def try_parse(string: str, def_value_or_format_provider: FSharpRef[TimeSpan] | Any, def_value : FSharpRef[TimeSpan] | None = None) -> bool:
```
dbrattli, line 238: Missing format (line too long)

src/fable-library-py/fable_library/time_span.py
```diff
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
-from math import ceil, floor, fmod
+import re
+from math import ceil, floor, fmod, isnan
```
dbrattli, line 4:
```suggestion
from math import ceil, floor, fmod
```

src/fable-library-py/fable_library/time_span.py
```diff
@@ -178,6 +182,74 @@ def to_string(ts: TimeSpan, format: str = "c", _provider: Any | None = None) ->
...
+
+    # If we have 3 arguments, it means that the second argument is the format provider
+    if def_value is not None:
+        out_value = def_value
+
+    try:
+        out_value.contents = parse(string)
+        return True
+    except Exception:
+        return False
```
dbrattli, line 250: Not important, but I would move the return out of the try block
```suggestion
    except Exception:
        return False
    
    return True
```

MangelMaxime:
@dbrattli Thanks you for the review.

I added ruff format to the git hooks, so now we should have the python code auto formatted before commit to avoid dealing with it manually.

PR #3651, opened 2023-12-11
[Python] List sort with 

dbrattli:
Make own copies of `List.fs` and `Array.fs` so we can fix the current we have problems with sorting for Python. Not sure if the intention was to have a single implementation for all languages, but see that Rust already have separate implementations.

Also imports `Callable` from `collections.abc` which is the preferred way instead of `typing`.

Fixes #3649
Fixes #3638

PR #3652, opened 2023-12-12
Add markdownlint check to enforce consistency in Changelog.md files

MangelMaxime:
@dbrattli This PR is the continuation of the discussion started in #3651 

The setup in the PR, will check that the CHANGELOG.md are compliant with the rules we have and prevent commit if an error is detected.

According to `markdownlint` [documentation](https://github.com/igorshubovych/markdownlint-cli#fixing-errors) it is possible to use `--fix` to format/fix simples errors but it also mention that it doesn't work for all the errors. I think we can give it a try without `--fix` for now and if it is too annoying we can give a try to give `--fix` a try later.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,7 @@
 indent_style = space
 trim_trailing_whitespace = true
 insert_final_newline = true
+indent_size = 4
 
 # TS files
 [*.ts]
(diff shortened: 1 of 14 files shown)


dbrattli (approved):
Thanks, this PR fixes all the squirrely lines I got with Markdownlint in vscode 😄

nojaf (approved):
Works for me!

PR #3659, opened 2023-12-13
[Python] Fix file output naming

dbrattli:
Generate the right Python output file name earlier in the pipeline so that commands such as `runScript` also gets the correct filename.

Fixes #3655

PR #3662, opened 2023-12-14
Change outFile extension later otherwise the modules is incorrectly computed because of an offset of 1 char length

MangelMaxime:
@dbrattli 

#3659 introduce a bug because there is an different lenght of `1` between `fsx` and `py`.

This changes fix that, I think it should be possible to change the extension of the file sooner in this function but I am not sure if this could have others side effects.


dbrattli (approved):
Good catch!!

Issue #3654, opened 2023-12-12
Rework `DateTime.Parse/TryParse` and `DateTimeOffset.Parse/TryParse`


MangelMaxime:
### Description

Hello @dbrattli,

The last failing test I have in Thoth.Json currently are related to DateTime/DateTimeOffset and I work on a fix for that. While working on it, I discovered that the parse function is using `dateutil` which is not a Python standard API and so that function doesn't work.

https://github.com/fable-compiler/Fable/blob/927652842e3e609f2db582fb5f5d4872bf9e0fbf/src/fable-library-py/fable_library/date.py#L226-L229

Currently, I am doing something really naive to use the native API and support at least a subset of the .NET formatting date:

```py
def parse(string: str, detectUTC: bool = False) -> datetime:
    # from dateutil import parser

    try:
        return datetime.fromisoformat(string)
    except ValueError:
        try:
            return datetime.strptime(string, "%H:%M:%S")
        except ValueError:
            try:
                return datetime.strptime(string, "%I:%M:%S %p")
            except ValueError:
                # IMPORTANT: This format needs to be tested after %I:%M:%S %p
                # Otherwise, it happens that %H:%M:%S %p incorrectly parses some
                # of the string that can overlap with %I:%M:%S %p
                # Example: 1:5:34 PM
                try:
                    return datetime.strptime(string, "%H:%M:%S %p")
                except ValueError:
                    raise ValueError("Unsupported format by Fable")
```

Really ugly, I know ^^. This is just a temporary solution instead of something that crash rightnow.

If we want to use `dateutil`, we need to publish fable-library to PyPi (or whatever is the package repository in Python) and check if this doesn't create others issues like we discussed on Slack regarding the Rust distribution of the library.

Or embed the `dateutil` source code inside of Fable, like it has been done in the past for `BigInt` in JavaScript.

How do you envision the DateTime/DateTimeOffset parsing functions to be implemented?

dbrattli:
@MangelMaxime I would be happy if we could get rid of the dateutil dependency. Here's a copilot refactor of your code:

```py
def parse(string: str, detectUTC: bool = False) -> datetime:
    formats = {
        # Matches a date-time string in the format "YYYY-MM-DDTHH:MM:SS"
        r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}": "%Y-%m-%dT%H:%M:%S",

        # Matches a time string in 24-hour format "HH:MM:SS"
[...]

PR #3663, opened 2023-12-14
Work on DateTime for Python

MangelMaxime:
Related to #3654


src/fable-library-py/fable_library/date.py
```diff
@@ -224,9 +260,26 @@ def op_addition(x: datetime, y: timedelta) -> datetime:
 
 
 def parse(string: str, detectUTC: bool = False) -> datetime:
-    from dateutil import parser
```
dbrattli, line 227: Also remove `python-dateutil` from `pyproject.toml` and do a `poetry update` to update `poetry.lock` remove the dependency

MangelMaxime:
This is strange the CI is failing on one of the test while I can run the tests locally. I tested with the same version of Python 3.10 on the dev container which is also on ubuntu.

dbrattli:
CPython is the same as Python. Almost all Pythons are CPython, except PyPy, Cython, Jython, ... Usually (every) time you talk about Python, you talk about CPython.

But can there be any locale format differences between your local computer and CI?

MangelMaxime:
Thanks for the explanations.

About the locales, I am not sure. Because, in theory `2014-09-10T13:50:34.0000000` should be handled by `datetime.fromisoformat(string).astimezone()` as this is an ISO format date.

If it was related to locales, I would expect the error to happens later like when we try to compare the date value or something like that.

dbrattli:
It is related to Python 3.10. It parses for more recent versions of Python

dbrattli (approved):
Looks good. Should we merge?

MangelMaxime:
I wanted to have a look at some others improvements to the date module but we can do it incrementally indeed.

I will revisit this PR at the beginning of next week, I need to update the changelog.

MangelMaxime:
@dbrattli I didn't forget about this PR.

While checking it I discovered that their was some errors in our implementation.

For example, time-only date in F# are using the now as the reference while in Python with the current implementation it is using 1900-01-01.

I have a fix for that and trying to fix a few others tests that are failing.

src/fable-library-py/fable_library/date.py
```diff
@@ -197,6 +238,10 @@ def utc_now() -> datetime:
     return datetime.utcnow()
 
 
+def today() -> datetime:
+    return datetime.replace(now(), hour=0, minute=0, second=0, microsecond=0)
+
+
 def to_local_time(date: datetime) -> datetime:
     return date.astimezone()
```
dbrattli, line 246: I'm worried about this one compared to #3691 since I think you will get problems when localtime is the same timezone as utc (+00). This is why I also removed the tzinfo in my PR

src/fable-library-py/fable_library/date.py
```diff
@@ -231,36 +276,224 @@ def add(x: datetime, y: TimeSpan) -> datetime:
...
 def add_milliseconds(d: datetime, v: int) -> datetime:
     return d + timedelta(milliseconds=v)
 
 
+def add_microseconds(d: datetime, v: int) -> datetime:
+    return d + timedelta(microseconds=v)
+
+
+def kind(d: datetime) -> DateKind:
+    if d.tzinfo == timezone.utc:
```
dbrattli, line 405: I don't think this will be able to see the difference between localtime and utc if localtime = utc

MangelMaxime: You are right

src/fable-library-py/fable_library/date.py
MangelMaxime: I believe the range of the comment is off ?

But yes, there is something to think about here

MangelMaxime:
@dbrattli I believe you are correct when pointing out that my proposition for supporting `DateTimeKind` has some limitations.

The difficulty is that in .NET we want to represents 3 states:

- `DateTimeKind.Local`
- `DateTimeKind.Utc`
- `DateTimeKind.Unspecified`

If I try to summarise our current 2 propositions: 

|   | Your implementation |  My implementation |
|---|---|---|
| `DateTimeKind.Local`  |  ✅ `date.tzinfo is None` |  ⚠️ `date.tzinfo is not None and date.tzinfo != timezone.utc` |
| `DateTimeKind.Utc`  | ✅  `date.tzinfo == timezone.utc` | ⚠️  `date.tzinfo == timezone.utc`  |
| `DateTimeKind.Unspecified`  |  ❌  |  ✅ `date.tzinfo is None`  |

In my case, I added ⚠️ because like you mentioned if your local timezone is `UTC` then we are in grey area because I don't know if we should return `DateTimeKind.Local` or `DateTimeKind.UTC`.

Because of that, I think if we want to stick with using native `datetime` to represents `DateTime` we will need to take a decision on which limitations we want.

1. We say that for Python dates are always treated as local unless `DateTimeKind.Utc` is passed at its creation (or when 
setting it). ==> Leads to your implementation

	- Should we allow creating date without specifying `DateTimeKind` in Python?
	- Should we generate errors if the user tries to use `DateTimeKind.Unspecified`?
	
2. We wants to stick "closer" (sorry for the lack of another word...) to .NET behaviour and decide that for all timezones we have the same behaviour but if the user is located in UTC timezone then we always report it as `DateTimeKind.UTC`.

Do you think my analysis is correct ? Which solutions do you prefer?

The others solutions I see are:

1. Find a way to attach `kind` property to the `datetime` type but I don't think Python allows that
2. Wrap `datetime` inside of a custom class so we can have our `kind` property

I will try to give a try to implements `DateTimeOffset` because I believe some of the question we currently have will also occurs for it. Hopefully, it will be less difficult because `DateTimeOffset` always needs the `offset` except for **1 constructor**

CleanShot 2024-01-07 at 11 14 50@2x

ncave:
@MangelMaxime 
> The others solutions I see are:
    1) Find a way to attach kind property to the datetime type but I don't think Python allows that
    2) Wrap datetime inside of a custom class so we can have our kind property


Both are viable options, JS is using `#1` and Rust is using `#2` (custom class with `kind` property).

MangelMaxime:
Indeed, I am personally in favour of 

> Wrap datetime inside of a custom class so we can have our kind property

If Python, doesn't allows us to add an additional property to `datetime`. This lose the transparence between `DateTime` (from Fable/.NET) and the native type from Python. But it garante that the same F# code behave the same way everywhere. 

If people, need a native Python `datetime` then you create it via binding API and if needed convert from to another thanks to an helpers. This something, I did in JavaScript from time to time when working with JavaScript Date libraries.

dbrattli:
What if we have native datetime as unspecified and then have a custom timezone for local time e.g:

```py
import time as _time

class LocalTimezone(tzinfo):
    def utcoffset(self, dt: datetime | None = None):
        # Return offset of local timezone from UTC
        return timedelta(seconds=-_time.timezone)

    def dst(self, dt: datetime | None = None):
        # Return the daylight saving time (DST) adjustment
        if dt is not None and self._isdst(dt):
            return timedelta(seconds=-_time.altzone)
        else:
            return timedelta(seconds=-_time.timezone)

    def tzname(self, dt: datetime | None = None) -> str:
        # Return the name of the time zone
        return "localtime"

    def _isdst(self, dt: datetime):
        # Determine whether or not DST is in effect
        tt = (dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second, dt.weekday(), 0, 0)
        stamp = _time.mktime(tt)
        tt = _time.localtime(stamp)
        return tt.tm_isdst > 0

# Example usage (perhaps make a singleton object to reuse)
dt = datetime.now(LocalTimezone())

# We can now detect it's local timezone 
assert dt.tzinfo.tzname() == "localtime"
```

Then I think we can handle unspecified (tzinfo=None), localtime (tzinfo=localtime) and utc (tzinfo=utc)

MangelMaxime:
@dbrattli I will give this approach a try 👍

MangelMaxime:
@dbrattli Using your approach I was able to implement `DateTimeKind` supports.

I have all the tests green locally but 4 are failing on the CI related due the CultureInfo being different. I am checking which Culture is used on the CI.

src/fable-library-py/fable_library/util.py
```diff
@@ -716,7 +716,7 @@ def f2(a1: _T1, a2: _T2, a3: _T3, a4: _T4, a5: _T5) -> _TResult:
 
 
 def curry5(
-    f: Callable[[_T1, _T2, _T3, _T4, _T5], _TResult]
+    f: Callable[[_T1, _T2, _T3, _T4, _T5], _TResult],
```
dbrattli, line 719: I think you need to update ruff to the latest version? 😉

src/fable-library-py/fable_library/singleton_local_time_zone.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def _isdst(self, dt: datetime):
+        # Determine whether or not DST is in effect
+        tt = (dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second, dt.weekday(), 0, 0)
+        stamp = _time.mktime(tt)
+        tt = _time.localtime(stamp)
+        return tt.tm_isdst > 0
+
+
+local_time_zone = LocalTimezone()
```
dbrattli, line 34:
```suggestion
local_time_zone = LocalTimezone()

__all__ = ["local_time_zone"]
```

MangelMaxime:
@dbrattli This is finally ready for review 🎉, I think I covered most of DateTime features and if need I believe this is a good foundation to add more API over time if needed. 

# Compatibility

*To be added to Fable website for reference*

Because Fable doesn't support Globalisation,`DateTime.ToString` behave has if `CultureInfo.InvariantCulture` is passed.

Python `datetime` is precise up to 6 digit microseconds meaning that we don't support `fffffff` or `FFFFFFF` custom format.

Similarly when parsing an ISO date format, it is possible that there is a lost of precision for the microseconds value. Especially for Python 3.10, the date will be adapted to use exactly 6 digits for the microseconds value. Python 3.10 doesn't support ISO date format out of the box.

- `2009-06-15T13:45:30.6175425` is transformed to `2009-06-15T13:45:30.617542`
- `2009-06-15T13:45:30.05` is transformed to `2009-06-15T13:45:30.050000`
-`2009-06-15T13:45:30.` is transformed to `2009-06-15T13:45:30.000000`

Because of number precision it is possible for `Ticks` to differs a little from .NET value.

`DateTime(1, 1, 1, 0, 0, 0, 0, 99, DateTimeKind.Utc).Ticks` should be `990` but Python returns `1040`
because `d.timestamp()` returns `-62135596799.9999` instead of `-62135596800000`.

To fix that we would need to compute our own `timestamp` value but I am not sure if we could do better. If someone wants to give it a try later they can contribute via a PR.

src/fable-library-py/fable_library/date.py
```diff
@@ -60,98 +96,321 @@ def year(d: datetime) -> int:
...
+        if current_char == quote_char:
+            found_quote = True
+            break
+        elif current_char == "\\":
+            if pos < format_length:
+                pos += 1
+                result += format[pos]
+            else:
+                # This means that '\'is the last character in the format string.
+                raise Exception("Invalid string format")
```
dbrattli, line 196:
```suggestion
                raise ValueError("Invalid string format")
```
Ref: https://docs.python.org/3/library/exceptions.html#ValueError

src/fable-library-py/fable_library/date.py
```diff
@@ -60,98 +96,321 @@ def year(d: datetime) -> int:
...
+        else:
+            result += current_char
+
+    if not found_quote:
+        # We couldn't find the matching quote
+        raise Exception(f"Invalid string format could not find matching quote for {quote_char}")
```
dbrattli, line 202:
```suggestion
        raise ValueError(f"Invalid string format could not find matching quote for {quote_char}")
```

src/fable-library-py/fable_library/date.py
```diff
@@ -231,43 +498,273 @@ def add(x: datetime, y: TimeSpan) -> datetime:
...
+        header_text = m.group("header")
+
+        if len(microseconds) == 7:
+            microseconds = microseconds[:-1]
+
+        return header_text + microseconds.ljust(6, "0")
+
+    if re.fullmatch(iso_format_regex, string):
+        adapted_string = re.sub(iso_format_regex, adapt_microseconds, string)
+        print(f"{string} is transformed to {adapted_string}")
```
dbrattli, line 562: Print statement should be removed

src/fable-library-py/fable_library/date.py
```diff
@@ -231,43 +498,273 @@ def add(x: datetime, y: TimeSpan) -> datetime:
...
+        return datetime.strptime(adapted_string, "%Y-%m-%dT%H:%M:%S.%f")
+
+    raise ValueError("Unsupported format by Fable: %s" % (string))
+
+
+def try_parse(string: str, defValue: FSharpRef[datetime]) -> bool:
```
dbrattli, line 568: Use snake casing i.e `def_value`

dbrattli:
Looks great. This is really fantastic work @MangelMaxime!! Just a few comments and suggestions. PS: don't forget to change the title (remove WIP) before merging.

dbrattli (approved):
Great work!!

MangelMaxime:
Thank you and thanks for reviewing the PR over time.

PR #3668, opened 2023-12-17
Normalize fable-library argument

### LINKED ISSUE #3653: Enforce `/` separator in the JavaScript import
### Description

@nojaf Reported issues on Windows when mixing `\` and `/` in the import of JavaScript.

It was discovered because the build system use `Path.Combime` to generate paths and on windows it use `\`.

Full command:

```
dotnet run -c Release --project C:\Users\nojaf\Projects\Fable\src\Fable.Build\..\Fable.Cli -- src\fable-library --outDir temp\fable-library-ts --fableLib .\temp\fable-library-ts --lang typescript --exclude Fable.Core --define FABLE_LIBRARY --noCache --typedArrays false --define FX_NO_BIGINT
```

We should normalize the import path to using `/` as it works across all the platforms and avoid custom code for a specific OS.

* Fable version: 4.7.0
* Operating system: Windows

nojaf:
Fixes #3653

The problematic code is, I believe, that https://github.com/fable-compiler/Fable/blob/38c8e10293af02adb25a9c82336ff4b940a81fad/src/Fable.Transforms/Global/Prelude.fs#L394-L427

expects the paths already to be in `/` notation. 
Converting this earlier on will ensure the relative paths in the import statements are fixed.

Also, trying out Windows CI again for JS. It worked on my machine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,10 @@ jobs:
 
   # Separate build job for JavaScript
   build-javascript:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.platform }}
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


MangelMaxime:
Thank you

Issue #2727, opened 2022-01-13
Fable CLI 3.7 watch stales


AngelMunoz:
### Description
Hey! sorry to bother again
When invoking the fable cli `3.7.0-beta-*` in watch mode through a library like `CliWrap` it never goes into watch mode this behavior is not observed in `v3.6.3` using the same conditions

```
Fable: F# to JS compiler 3.7.0-beta-009
Thanks to the contributor! @fbehrens

Parsing src\fable37-cli.fsproj...
src> cmd /C dotnet restore fable37-cli.fsproj
  Determining projects to restore...
  All projects are up-to-date for restore.
Project and references (1 source files) parsed in 2125ms

Started Fable compilation...
```

### Repro code
Here's a reproduction repository: https://github.com/AngelMunoz/fable37-cli

### Expected and actual results
I expect to observe the same behavior as Fable `3.6.3`, it should compile and start watching for file changes

### Related information

* Fable version: `dotnet fable --version`: `3.7.0-beta-009`
* Operating system: `Windows 11`

AngelMunoz:
**Note**: invoking the CLI manually `dotnet fable watch ./src/project.fsproj` doesn't fail... so I'm not sure if it is related to something with a stdout redirect

AngelMunoz:
I think they are generated, I need to check further, but CliWrap allows me to set env vars for the process perhaps I can add the CI env var, I'll test in a couple of hours and let you know

AngelMunoz:
Well it weren't a few hours I could find a spot

> I see. Does watch mode doesn't work at all (no JS files are generated)

I just checked and no, no files are generated at all

Setting the CI flag (`.WithEnvironmentVariables(fun env -> env.Set("CI", "1") |> ignore)`) does indeed make it work! it isn't as pretty but works nicely
![image](https://user-images.githubusercontent.com/8684875/149438494-d3bef91a-60ce-4278-9692-2b411fc6dd89.png)

alfonsogarciacaro:
Hmm,  interesting. So I guess that trying to move the cursor when being called from CliWrap is breaking/freezing the process somehow. Maybe opening an issue about that in CliWrap? Or you may need to spawn the process with its own shell.

AngelMunoz:
Had a talk with CliWrap's author on discord he was kind enough to give this issue a glance

[...]

Issue #3631, opened 2023-11-30
`Fable` freezing with `--run vite`


lukaszkrzywizna:
### Description

Hi,

After upgrading from `Webpack` to `Vite.js`, I've noted that `Fable` freezes during `HMR/Fast-Refresh`. The problem is **critical** because it happens quite often for big projects/files (like my production application). 

I've been able to reproduce the error for a small project:

### Repro code

1. Create an application using a template from [SAFEr.Template](https://github.com/Dzoukr/SAFEr.Template)
2. Quickly add code and save the file alternately (for my production project I don't have to do this, updating code randomly causes freeze) 

The video shows the problem:

https://github.com/fable-compiler/Fable/assets/12534253/2e90a2e5-424b-41f8-8992-0a3780497e41

You can see that at some point `Fable` stops responding with the last message of `Started fable compilation...`, however, Vite and the application still work.

### Related information

* Fable version: `4.5.0`
* Operating system: `Mac OS 14 Sonoma`

MangelMaxime:
Hum, this kind of issue remind me of the system being out of inotify watcher but I think it was happening to me on Linux.

Can you reproduce without running vite?

Can you also try to use set Vite to use polling?

https://vitejs.dev/config/server-options.html#server-watch

Could you please try to use set `export DOTNET_USE_POLLING_FILE_WATCHER=1`?

lukaszkrzywizna:
After a few minutes of testing, I can confirm that:
- `export DOTNET_USE_POLLING_FILE_WATCHER=1` **doesn't** work
- use polling **doesn't** work
- without `vite` **works**
- running `fable` and `vite` as two separated processes **works**

Based on that, I suspect the problem is with running `vite` using the `fable's` `--run` parameter.
I will continue testing my production app but so far, the last option gives me no freezes.

lukaszkrzywizna:
I confirm - running `fable` and `vite` separately resolve the issue.

MangelMaxime:
Thank you for sharing the work around.

I am not sure why there is a problem when running it via Fable. In theory Fable, just spawn the process and forget about it especially on Unix system where we don't need to kill the child process ourself compared to windows.

MangelMaxime:
I am experiencing the same problem but I have also also when running the processes separately...

laurentpayot:
[...]

PR #3671, opened 2023-12-18
Revist logging

nojaf:
Currently, a lot of logging happens directly to the standard output. 
This is unwanted in the `Fable.Compiler`, and I wish to introduce a logging abstraction.
Inside my use case, I would like to log in to memory or disk.

The only thing I'm not quite sure about is what to do with https://github.com/fable-compiler/Fable/issues/2727. It seems like some code was written to avoid this problem and I'm not sure it will still be problematic after this change.
@AngelMunoz, is this still an active use-case of yours?


MangelMaxime:
Reading the linked issue in your command, ringed a bell to me.

I believe we still have some issues when running Fable in watch mode.

I don't know if this is related to the logger or not but, when running a sub process via Fable sometimes the processes are freezing.

Example of command: `dotnet fable --run npx vite`

https://github.com/fable-compiler/Fable/issues/3631

I observed a similar behaviour when working on fantomas-tool or even worth sometimes for that particular project where `vite` process would start and just stop completely at some point.

AngelMunoz:
@nojaf The main problem there was that the logs were being re-written in the same line via cursor manipulation, this could cause hang ups if the output was redirected (e.g. calling Fable via Process APIs or in my case CLIWrap).

From a quick glance on this PR, looks like the main change is that logging now goes through an `ILogger` right?

I'm guessing that leaves using the cursor to write in the same line/position out of the window, if that is the case, I believe it should not break anyone calling fable from another process via stdio.

nojaf:
Ok, thanks for confirming. For now, I think I'd like to remove that cursor manipulation thing, as it is probably tricky to achieve using the ILogger.

MangelMaxime:
The cursor manipulation allows for a much smaller / easier to read output in the log when Fable does all the recompilation of the files. So it would be nice to keep it.

There is currently the `--verbose` flag which I believe allow people to disable it if it needed.

nojaf:
Ok, I restored that cursor update functionality.
Ready for review.

MangelMaxime:
I don't have much time this week, review will most likely happen next weeks.

ncave:
@nojaf Minor CI error to fix:
```
Error: /home/runner/work/Fable/Fable/tests/Integration/Compiler/Util/Compiler.fs(36,9): error FS0764: No assignment given for field 'Verbosity' of type 'Fable.Compiler.Util.CliArgs'
```

nojaf:
> Minor CI error to fix:

Thanks, is this project not part of the `Fable.sln`?

ncave:
@nojaf 
> is this project not part of the Fable.sln?

I think [it is](https://github.com/fable-compiler/Fable/blob/main/Fable.sln#L47).

nojaf:
Alright, another crucial part of `Fable.Compiler` is that **no code writes to stdout**! 
This messed up the `JSON RPC` communication I'm using for the Vite plugin.

To guarantee that nobody ever sins against this, I've created an analyzer that detects calls to `prinfn` and `Console.Write`. The results are logged [here](https://github.com/fable-compiler/Fable/security/code-scanning?query=pr%3A3671+is%3Aopen+severity%3Aerror). And it is possible to configure a PR build to fail if errors are present. (Needs to be configured in GitHub).

`build / analyzers` is green because the process of running the analyzers succeeded. This won't fail if the analyzers detect something bad.

`Code scanning results / Ionide.Analyzers.Cli` is green because no **new** code violated the rules. In future PRs, we want this to be red when someone breaks the rules.

I still need to address the actual `Console.Write` usage in `Fable.Transform`, so please don't merge this PR yet. No need to convert to draft, I should be able to address this in the next days.

src/Fable.Analyzers/StdOutAnalyzer.fs
```diff
@@ -0,0 +1,63 @@
...
+open FSharp.Analyzers.SDK
+
+let namesToAvoid =
+    set
+        [
+            "System.Console.Write"
+            "System.Console.WriteLine"
+            "System.Console.WriteAsync"
+            "System.Console.WriteLineAsync"
+            "Microsoft.FSharp.Core.Operators.print"
```
ncave, line 16: @nojaf Shouldn't the last 3 be these instead?
```fs
    "Microsoft.FSharp.Core.Printf.PrintFormat"
    "Microsoft.FSharp.Core.Printf.PrintFormatToError"
    "Microsoft.FSharp.Core.Printf.PrintFormatLine"
    "Microsoft.FSharp.Core.Printf.PrintFormatLineToError"
```

nojaf: Thanks, those names are wrong indeed: https://github.com/dotnet/fsharp/blob/main/src/FSharp.Core/fslib-extra-pervasives.fs#L267-267

I'll check them.

ncave:
@nojaf
> I still need to address the actual Console.Write usage in Fable.Transform

I moved the `Console.WriteLine` usage out of `Fable.Transforms`, please rebase.
You can now move the `Log` module back to `Fable.Compiler`.
This should fix your `Fable.Transforms` build issue.

nojaf:
@ncave thanks! I've rebased and all is green now.

@MangelMaxime with the `getPlugin` changes in here I'd like to have this PR merged.
I was able to use my local Vite plugin for a small Fable project I have.
It would be great to have another `Fable.Compiler` release with this.

MangelMaxime:
> @MangelMaxime with the `getPlugin` changes in here I'd like to have this PR merged.

Noted, I will work on the next release of Fable.

> It would be great to have another `Fable.Compiler` release with this.

To trigger the release for `Fable.Compiler` can you please update its Changelog file? You can update the `Unreleased` section for now, I will bump the version when doing the release process.

If you could update the `Unreleased` section of the different Changelog files impacted by this PR it would be great  🙏

nojaf:
Done!

src/Fable.Transforms/Fable.Transforms.fsproj
```diff
@@ -45,6 +45,9 @@
     <Compile Include="Rust/Fable2Rust.fs" />
     <Compile Include="State.fs" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
```
ncave, line 49: @nojaf Perhaps this can be removed.

tests/Js/Adaptive/Fable.Tests.Adaptive.fsproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
ncave, line 3: @nojaf Please change to
```
    <OutputType>Exe</OutputType>
```
for consistency with the other test projects.

src/Fable.Analyzers/Fable.Analyzers.fsproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
```
ncave, line 4: @nojaf Can this target `net8.0` if it only run in CI?

MangelMaxime:
Thank you for the contribution and to the reviewers

PR #3677, opened 2023-12-20
Update CHANGELOG.md about breaking `static member`.

smoothdeveloper:


smoothdeveloper:
Not sure if it is easy to update the nuget packages description (assuming it is where most people would read this content).

MangelMaxime:
You can't update a nuget package after it has been published.

If we merge this PR, the only place that will be updated is the CHANGELOG itself and I can also fix the Release on Github.

PR #3685, opened 2023-12-29
Python: added 'Double.IsPositiveInfinity'

PierreYvesR:
`System.Double.IsPositivieInfinity` was only missing from Python.


dbrattli (approved):
Thanks a lot for fixing!!

PR #3688, opened 2024-01-04
Queue collection bug fix

PierreYvesR:
In certain case, `tail` would point past the end of inner storage of the queue, leading to incorrect behaviour.


MangelMaxime:
Thank you

PR #3691, opened 2024-01-05
[Python] Add kind property for datetime

### LINKED ISSUE #3690: [Python] How to works with `Kind` property on the `DateTime` type?
### Description

Currently, to represent the `DateTime` type we use `datetime.datetime` in Python.

This works fine until, we need to work with the `kind` property. Is there a way in Python to add a "property" to an existing class?

I tried to look only for how to dynamically add attributes to a class and I found  `setattr(date, "kind", kind or DateKind.Unspecified)` but it doesn't works:

```
AttributeError: 'datetime.datetime' object has no attribute 'kind'
```

I believe this is because `datetime.datetime` doesn't implements `__dict__` or `__slot__` or something like that. 

Does python support implicit conversion? For example, if we do something like `class MyDate(datetime):` does it means that `MyDate` can be use anywhere `datetime` can?

> Note, I was not able to implements `class MyDate(datetime):` I having trouble trying to call the parent constructor.

@dbrattli

dbrattli:
Fixes #3690

### DIFF
--- a/src/Fable.Transforms/Python/Replacements.fs
+++ b/src/Fable.Transforms/Python/Replacements.fs
@@ -5053,7 +5053,9 @@ let dates
             ?loc = r
         )
         |> Some
-    | "get_Kind"
+    | "get_Kind" ->
+        Helper.LibCall(com, "date", "get_kind", t, [ thisArg.Value ], ?loc = r)
+        |> Some
     | "get_Offset" ->
         Naming.removeGetSetPrefix i.CompiledName
         |> Naming.lowerFirst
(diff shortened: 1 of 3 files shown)


MangelMaxime:
Thank you I am closing.

Based on your suggestion, I was able to make an implementation of `kind` which supports `DateTimeKind.Unspecifed` in #3663

dbrattli:
I would like to see the two tests from this PR in the new. There are details in this PR that is lost. I think we should have merged this and then you should rebase yours.

PR #3715, opened 2024-01-24
[JS] Add support for System.Array.Resize

chkn:
This adds support for [System.Array.Resize](https://learn.microsoft.com/en-us/dotnet/api/system.array.resize?view=net-8.0) to Fable for JS/TS.

I had to add 2 injected parameters to the library function: one for the array constructor, which seems standard, and another for the "zero" value for the array's type, in case the length of the array is being extended. If there is a nicer way to handle this, let me know.


ncave (approved):
LGTM

PR #3716, opened 2024-01-26
[JS] System.Array.Resize: also handle the case where the array is null

chkn:
In my [initial PR](https://github.com/fable-compiler/Fable/pull/3715), I missed that [the docs](https://learn.microsoft.com/en-us/dotnet/api/system.array.resize?view=net-8.0) say the array can be `null`, in which case a new array is allocated.

### DIFF
--- a/src/fable-library/Array.fs
+++ b/src/fable-library/Array.fs
@@ -1418,6 +1418,9 @@ let resize
 
     let len =
         if isNull xs then
+            if newSize = 0 then
+                xs <- allocateArrayFromCons cons 0
+
             0
         else
             xs.Length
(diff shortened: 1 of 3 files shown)


src/fable-library/Array.fs
```diff
@@ -1416,14 +1416,23 @@ let resize
...
 
-    let len = xs.Length
+    let len =
+        if isNull xs then
+            if newSize = 0 then
+                xs <- allocateArrayFromCons cons 0
+
+            0
+        else
+            xs.Length
```
ncave, line 1427: @chkn I know it already works, but perhaps simplify the logic a bit like this (untested):

instead of:
 ```fs
    let len =
        if isNull xs then
            if newSize = 0 then
                xs <- allocateArrayFromCons cons 0
            0
        else
            xs.Length
    ....
```

perhaps this:
```fs
if isNull xs then
    xs <- allocateArrayFromCons cons newSize
else
    let len = xs.Length
....
```

ncave (approved):

PR #3720, opened 2024-01-28
Extract Builalyzer

nojaf:
I've been vocal about this in several places, I'm not a fan of Buildalyzer and the state of project cracking. The whole temporary `*.csproj` thing is really a turn off.

That is why I would like to be able to swap this out in `Fable.Compiler`.
I have a custom implementation of `ProjectCrackerResolver` in my [plugin](https://github.com/nojaf/vite-plugin-fable/blob/main/Fable.Daemon/CoolCatCracking.fs).

Ready for review.


MangelMaxime:
> I've been vocal about this in several places, I'm not a fan of Buildalyzer and the state of project cracking. The whole temporary `*.csproj` thing is really a turn off.

I believe most of the maintainers agree on that and we want to remove that hack.

I am going to review the PR today as I need to make a changes in the watch mode support too of Fable.

PR #3734, opened 2024-02-05
[Python] Fixes custom equality on records

### LINKED ISSUE #3717: [Python] Nested type with Custom Equality gives false negative equality
### Description

If a record type with custom equality and hashcode is included in another type, calling the equality on the outer type returns false, possibly because custom equality of inner type is not used.

Possibly related to #3674, which works like a charm now, thank you!

#### test.fsx
```fsharp
[<CustomEquality; NoComparison>]
type X = 
    {
        ID : int 
        Name : string 
    }

    override x.GetHashCode() = x.ID

    override this.Equals(other : obj) = 
        match other with
        | :? X as other -> this.ID = other.ID
        | _ -> false

// Record type here (unlike issue #3674)
type Y = 
    {X : X}

// Should be equal according to custom equality
let x1 = {ID = 1; Name = "a"}
let x2 = {ID = 1; Name = "b"}

// Should all be equal according to custom equality of inner type
let y1 = {X = x1}
let y2 = {X = x2}
let y3 = {X = x1}



printfn $"x1=x2:{x1 = x2}" // true

printfn $"y1=y2:{y1 = y2}" // incorrectly false
printfn $"y1=y3:{y1 = y3}" // correctly true, possibly because equal by non-custom equality 
```
#### test.cmd
```shell
dotnet fable . --lang python
python test.py
```

Expected: Everything should be true. 

Actual: Two records containing a only a record each that should be equal by custom equality still returns false

### Related information

* `dotnet fable --version`: 4.10.0
* `dotnet tool list/update/install`: 4.10.0
* Windows11

@freymaurer

dbrattli:
- Nested type with Custom Equality gives false negative equality, fixes #3717 
- Generate assert statements for `assert` expressions in debug mode

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -37,6 +37,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Remove `Choice.d.ts` from source code of `fable-library` (by @MangelMaxime)
 
+### Fixed
+
+#### Python
+
+* [GH-3717](https://github.com/fable-compiler/Fable/issues/3717) Nested type with Custom Equality gives false negative equality (by @dbrattli)
+* Generate assert statements for `assert` expressions in debug mode (by @dbrattli)
+
 ## 4.11.0 - 2024-01-30
 
 ### Changed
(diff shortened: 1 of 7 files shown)


MangelMaxime:
@dbrattli I rebased the PR on main because we made some changes to the formatting rules.

PR #3736, opened 2024-02-05
Set max_line_length to 120

### LINKED ISSUE #3735: Change Fantomas max_line_length and reformat
IMO the [current Fantomas setting](https://github.com/fable-compiler/Fable/blob/main/.editorconfig#L19) of `max_line_length = 80` is perhaps too restrictive and leads to a lot of vertical code separation that makes the code a bit hard to read.

Perhaps changing it to the [default Fantomas setting](https://fsprojects.github.io/fantomas/docs/end-users/Configuration.html#max_line_length) of `max_line_length = 120` can help make the code more readable everywhere.

I understand this is one of those personal preferences so feel free to disagree.

ncave:
- Changed Fantomas `max_line_length` to the default setting (120).
- Fixed some analyzer warnings.
- Fixes #3735.


MangelMaxime:
As always with formatting, for some stuff I prefer the one element on a line approach (forced by a smaller width) and for other having it on a single line makes more sense.

Let's give it a try with a width of `120` as it is closer to how the code was shaped before I took the lead on Fable project.

Thank you

Issue #3743, opened 2024-02-07
New `logger` doesn't display the same information as before when using `--verbose` 


MangelMaxime:
For example, running `dotnet fable --verbose` before logged messages like:

- `Type checked: src/tk.Client/App.fs`
- `Type check finished`

The new `verbose` pollute less the output but it would be nice to still have a way to see debug information.

@nojaf Any idea?

### Related information

* Fable version: 4.11.0
* Operating system: OSX

PR #3744, opened 2024-02-08
Set verbosity in Fable.Cli

nojaf:
This is the easy fix for https://github.com/fable-compiler/Fable/issues/3743.
A better fix would probably be to revisit `let mutable private verbosity = Fable.Verbosity.Normal`.


nojaf:
Hi @MangelMaxime, did you have a chance to look at this PR?

nojaf:
> mimic the behaviour from before.

Yup, it does, I would merge this one in just to get the original behaviour back.
We can revisit it later. Your suggestion makes sense.

MangelMaxime:
Thanks for your answer, I will include it in the next release.

Issue #3700, opened 2024-01-19
Support ListCollector


nojaf:
### Description

It would be great if Fable could support [ListCollector<'T>](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-compilerservices-listcollector-1.html)

### Repro code

```fsharp
    let cutOffLast list =
        let mutable headList = ListCollector<'a>()

        let rec visit list =
            match list with
            | []
            | [ _ ] -> ()
            | head :: tail ->
                headList.Add(head)
                visit tail

        visit list
        headList.Close()
```

### Expected and actual results

Expected:

```js
// Something along the lines of
// let mutable headList = ListCollector<'a>()
const headList = [];

// headList.Add(head)
headList.push(head);

// headList.Close()
import { ofArray } from "fable-library/List.js";
ofArray(headList);
```

Actual: `error FABLE: Microsoft.FSharp.Core.CompilerServices.ListCollector`1.Close is not supported by Fable`.

### Related information

* Fable version: `dotnet fable --version`: 4.9.0
* Operating system: Windows

I might wanna dig into this if approved.

MangelMaxime:
It seems to be a small enough type to write [Source](https://github.com/dotnet/fsharp/blob/712016d1754fbabc8c102777fb3a9e346351fc8f/src/FSharp.Core/seqcore.fs#L534-L593) so fo me it is ok to add it.

Because this is an F# type, what we do in general is we take ~heavy~ inspiration from the FSharp.Core implementation and include it inside of the fable-library folder. For example, this is what we did for [StringBuilder](https://github.com/fable-compiler/Fable/blob/1b40a3fd9509812582e1986cf82c54ecb49d1b26/src/fable-library/System.Text.fs).

The only things I have question about is that ListCollector is using `RuntimeHelpers.FreshConsNoTail` which is defined as:

```fs
let inline FreshConsNoTail head = head :: (# "ldnull" : 'T list #)
```

so it is emitting IL code and we will not be able to do that same goes for `SetFreshConsTail`. But I don't think this is blocking.

ncave:
Closing as implemented by #3745, feel free to reopen if needed.

PR #3745, opened 2024-02-08
ListCollector

nojaf:
An initial naive implementation for https://github.com/fable-compiler/Fable/issues/3700.

I'm a little confused that `FSharp.Core.CompilerServices.js` looks fine:
```js
import { addRangeInPlace } from "./Array.js";
import { toList } from "./Seq.js";
import { class_type } from "./Reflection.js";
export class ListCollector$1 {
    constructor() {
        this.collector = [];
    }
    Add(value) {
        const this$ = this;
        void (this$.collector.push(value));
    }
    AddMany(values) {
        const this$ = this;
        addRangeInPlace(values, this$.collector);
    }
    AddManyAndClose(values) {
        const this$ = this;
        addRangeInPlace(values, this$.collector);
        return toList(this$.collector);
    }
    Close() {
        const this$ = this;
        return toList(this$.collector);
    }
}
export function ListCollector$1_$reflection(gen0) {
    return class_type("Microsoft.FSharp.Core.CompilerServices.ListCollector`1", [gen0], ListCollector$1);
}
export function ListCollector$1_$ctor() {
    return new ListCollector$1();
}

```

but the usage in `QuickTest.js` does not:

```js
import { some } from "./fable_modules/fable-library.4.11.0/Option.js";
import { defaultOf, equals } from "./fable_modules/fable-library.4.11.0/Util.js";
import { startsWith, isNullOrEmpty, toFail, printf, toConsole } from "./fable_modules/fable-library.4.11.0/String.js";
import { startImmediate } from "./fable_modules/fable-library.4.11.0/Async.js";
import { singleton } from "./fable_modules/fable-library.4.11.0/AsyncBuilder.js";
import { ListCollector$1__Close, ListCollector$1__Add_2B595 } from "./fable_modules/fable-library.4.11.0/FSharp.Core.CompilerServices.js";

// ...

testCase("meh2", () => {
    let copyOfStruct_1;
    const c = defaultOf();
    let copyOfStruct = c;
    ListCollector$1__Add_2B595(copyOfStruct, 1);
    (copyOfStruct_1 = c, ListCollector$1__Close(copyOfStruct_1, void 0));
});

```

Am I missing something obvious @MangelMaxime @ncave?

### DIFF
--- a/tests/Js/Main/ListCollectorTests.fs
+++ b/tests/Js/Main/ListCollectorTests.fs
@@ -1,7 +1,6 @@
 module Fable.Tests.ListCollector
 
 open Util.Testing
-open Fable.Tests.Util
 open Microsoft.FSharp.Core.CompilerServices
 
 let cutOffLast list =
@@ -23,4 +22,32 @@ let tests =
         testCase "ListCollector.Add and .Close" <| fun () ->
             let result = cutOffLast [ 1; 2; 3 ]
             result |> equal [ 1; 2 ]
+
+        testCase "ListCollector.Close works for empty list" <| fun () ->
+            let mutable l = ListCollector<_>()
+            let result = l.Close()
+
+            result |> equal []
+
+        testCase "ListCollector.AddMany works" <| fun () ->
+            let mutable l = ListCollector<_>()
+            l.AddMany([ 1; 2; 3 ])
+
+            let result = l.Close()
+
+            result |> equal [ 1; 2; 3 ]
+
+        testCase "ListCollector.AddMany works for empty list" <| fun () ->
+            let mutable l = ListCollector<_>()
+            l.AddMany([])
+
+            let result = l.Close()
+
+            result |> equal []
+
+        testCase "ListCollector.AddManyAndClose works" <| fun () ->
+            let mutable l = ListCollector<_>()
+            let result = l.AddManyAndClose([ 1; 2; 3 ])
+
+            result |> equal [ 1; 2; 3 ]
     ]
--- a/tests/Js/Main/ListCollectorTests.fs
+++ b/tests/Js/Main/ListCollectorTests.fs
@@ -19,8 +19,8 @@ let cutOffLast list =
     headList.Close()
 
 let tests =
-  testList "ListCollector" [
-    testCase "ListCollector.Add and .Close" <| fun () ->
-        let result = cutOffLast [ 1; 2; 3 ]
-        result |> equal [ 1; 2 ]
-  ]
+    testList "ListCollector" [
+        testCase "ListCollector.Add and .Close" <| fun () ->
+            let result = cutOffLast [ 1; 2; 3 ]
+            result |> equal [ 1; 2 ]
+    ]
--- a/tests/Js/Main/ListCollectorTests.fs
+++ b/tests/Js/Main/ListCollectorTests.fs
@@ -1,0 +1,26 @@
+module Fable.Tests.ListCollector
+
+open Util.Testing
+open Fable.Tests.Util
+open Microsoft.FSharp.Core.CompilerServices
+
+let cutOffLast list =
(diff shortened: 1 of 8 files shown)


MangelMaxime:
This is probably because you used `[<AttachMembers>]` on the `ListCollector` type definition.

This tells Fable attach members to the JavaScript class instead of generated a function outside of the type like `ListCollector$1__Add_2B595`.

**With `[<AttachMembers>]`**

```js

export class ListCollector$1 {
    constructor() {
        this.collector = [];
    }
    Add(value) {
        const this$ = this;
        void (this$.collector.push(value));
    }
}

export function ListCollector$1_$reflection(gen0) {
    return class_type("Test.ListCollector`1", [gen0], ListCollector$1);
}

export function ListCollector$1_$ctor() {
    return new ListCollector$1();
}
```

**Without `[<AttachMembers>]`**

```js

export class ListCollector$1 {
    constructor() {
        this.collector = [];
    }
}

export function ListCollector$1_$reflection(gen0) {
    return class_type("Test.ListCollector`1", [gen0], ListCollector$1);
}

export function ListCollector$1_$ctor() {
    return new ListCollector$1();
}

export function ListCollector$1__Add_2B595(this$, value) {
    void (this$.collector.push(value));
}
```

If you remove `[<AttachMembers>]` it will probably works.

In general `[<AttachMembers>]` is only use for interop with JavaScript so I don't think we are using it in `fable-library`. 

The alternative would be to specifically rewrite how Fable generates the code for this type/members.

For info using `[<AttachMembers>]` comes with limitations, as for example it does not support method overload because JavaScript method doesn't support it.

ncave:
@nojaf See [PR#1](https://github.com/nojaf/Fable/pull/1)

tests/Js/Main/ListCollectorTests.fs
```diff
@@ -0,0 +1,26 @@
...
+
+    visit list
+    headList.Close()
+
+let tests =
+  testList "ListCollector" [
+    testCase "ListCollector.Add and .Close" <| fun () ->
+        let result = cutOffLast [ 1; 2; 3 ]
+        result |> equal [ 1; 2 ]
+  ]
```
MangelMaxime, line 26: Can we add tests for each `method` available and also edges cases like calling `.close()` on an empty `ListCollector`?

tests/Js/Main/ListCollectorTests.fs
```diff
@@ -0,0 +1,26 @@
+module Fable.Tests.ListCollector
+
+open Util.Testing
+open Fable.Tests.Util
+open Microsoft.FSharp.Core.CompilerServices
+
+let cutOffLast list =
+    let mutable headList = ListCollector<'a>()
```
MangelMaxime, line 8: Looking at the code it seems like we can remove the `mutable` here

MangelMaxime (changes requested):

tests/Js/Main/ListCollectorTests.fs
nojaf: That would not reflect the actual usage of shared code. In regular F# the collector needs to be mutable.

MangelMaxime: Didn't know about that.

tests/Js/Main/ListCollectorTests.fs
MangelMaxime: I don't agree yes for this module it is probably enough, but it happens several times that doing so allowed us to catch behaviour we were not aware of.

There are still some APIs in fable-library that don't behave correctly because they didn't have a test associated for every usage. Having tests also allows us to catch on regressions.

I added the tests to cover the most common cases.

MangelMaxime:
Thanks for the implementation, I will include it in the next release which should come out this weekend.

nojaf:
Thanks for going the extra mile here!

PR #3746, opened 2024-02-10
Extract type-checking from compilation

nojaf:
It is slightly more convenient to have a separate type-check function and be able to compile a specific set of files.


nojaf:
I'm going to merge this one in as it really only affects `Fable.Compiler`.

PR #3748, opened 2024-02-10
Accessing an array out of bounds, should generates an exception like on .NET

### LINKED ISSUE #3729: IndexOutOfRangeException for bad index?
I am expecting an `IndexOutOfRangeException` here, but it just returns `undefined`
```fsharp
let xs: string[] = [||]

printfn $"{xs[0]}"
```
See [REPL](https://fable.io/repl/#?code=DYUwLgBAHgzgXBGYBOBLAdgcwNoF0IC8E2APibgFAUAOa6YAZuhACQBEA3rNgAy4C+bIA&html=DwCwLgtgNgfAsAKAAQqaApgQwCb2ag4CdMTJcMABwFp0BHAVwEsA3AXgCIBhAewDsw6AdQAqAT0roOSAMb9BAzoIAeYAPThoAbhkhMAJwDOJNgzAAzagA4OeQhqy5EhAEY9sYu6mBq3HvD6asEA&css=Q)
I know this is the default JS behavior. Should Fable stick to .NET semantics? 
Or could it offer a debug mode where indices are checked?

MangelMaxime:
Fix #3729

@ncave Could you please have a look at what I did? 

By some trial and error, I found that these places are where we handle array index access + `Array.item` which have the same behaviour.

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -1948,8 +1948,34 @@ let arrayModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Ex
         Helper.LibCall(com, "List", "ofArray", t, args, i.SignatureArgTypes, genArgs = i.GenericArgs, ?loc = r)
         |> Some
     | ("Length" | "Count"), [ arg ] -> getFieldWith r t arg "length" |> Some
-    | "Item", [ idx; ar ] -> getExpr r t ar idx |> Some
-    | "Get", [ ar; idx ] -> getExpr r t ar idx |> Some
+    | "Item", [ idx; ar ] ->
+        Helper.LibCall(
+            com,
+            "Array",
+            "item",
+            t,
+            [
+                idx
+                ar
+            ],
+            i.SignatureArgTypes,
+            ?loc = r
+        )
+        |> Some
+    | "Get", [ ar; idx ] ->
+        Helper.LibCall(
+            com,
+            "Array",
+            "item",
+            t,
+            [
+                idx
+                ar
+            ],
+            i.SignatureArgTypes |> List.rev,
+            ?loc = r
+        )
+        |> Some
     | "Set", [ ar; idx; value ] -> setExpr r ar idx value |> Some
     | "ZeroCreate", [ count ] -> createArray count None |> Some
     | "Create", [ count; value ] -> createArray count (Some value) |> Some
@@ -2500,8 +2526,21 @@ let intrinsicFunctions (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisAr
         |> withTag "downcast"
         |> Some
     | "MakeDecimal", _, _ -> decimals com ctx r t i thisArg args
-    | "GetString", _, [ ar; idx ]
-    | "GetArray", _, [ ar; idx ] -> getExpr r t ar idx |> Some
+    | "GetString", _, [ ar; idx ] -> getExpr r t ar idx |> Some
+    | "GetArray", _, [ ar; idx ] ->
+        Helper.LibCall(
+            com,
+            "Array",
+            "item",
+            t,
+            [
+                idx
+                ar
+            ],
+            i.SignatureArgTypes |> List.rev,
+            ?loc = r
+        )
+        |> Some
     | "SetArray", _, [ ar; idx; value ] -> setExpr r ar idx value |> Some
     | ("GetArraySlice" | "GetStringSlice"), None, [ ar; lower; upper ] ->
         let upper =
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -1922,6 +1922,10 @@ let arrays (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (thisArg: E
         |> Some
     | _ -> None
 
+let arrayGetItem (com: ICompiler) r (t: Type) (args: Expr list) (signatureArgTypes: Type list) =
+    Helper.LibCall(com, "Array", "item", t, args, signatureArgTypes, ?loc = r)
+    |> Some
+
 let arrayModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Expr option) (args: Expr list) =
     let newArrayAlloc size t =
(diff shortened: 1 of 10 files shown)


MangelMaxime:
Oups I didn't test TypeScript locally, I will have a look at it.

ncave (changes requested):

src/fable-library/Array.fs
```diff
@@ -877,13 +877,23 @@ let tail (array: 'T[]) =
...
 
-let item index (array: _[]) = array.[index]
+let item (index: int) (array: 'T[]) : 'T =
+    if index < 0 || index >= array.Length then
+        invalidArg "index" "Index was outside the bounds of the array."
+    else
+        // We need to emit JavaScript directly here
+        // otherwise we will call `item` function recursively
+        // and it will cause a stack overflow
+        emitJsExpr (index, array) "$1[$0]"
```
ncave, line 887: @MangelMaxime Perhaps move this to `Native.fs`.

ncave (changes requested):

src/fable-library/Array.fs
MangelMaxime: I didn't know about this file.

I will look to add an helper in that module

src/Fable.Transforms/Replacements.fs
```diff
@@ -2582,8 +2582,34 @@ let arrayModule (com: ICompiler) (ctx: Context) r (t: Type) (i: CallInfo) (_: Ex
         Helper.LibCall(com, "List", "ofArray", t, args, i.SignatureArgTypes, genArgs = i.GenericArgs, ?loc = r)
         |> Some
     | ("Length" | "Count"), [ arg ] -> getFieldWith r t arg "length" |> Some
-    | "Item", [ idx; ar ] -> getExpr r t ar idx |> Some
-    | "Get", [ ar; idx ] -> getExpr r t ar idx |> Some
+    | "Item", [ idx; ar ] ->
+        Helper.LibCall(
+            com,
+            "Array",
+            "item",
```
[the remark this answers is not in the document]
MangelMaxime: I thought about it but didn't pursue in that direction for some reason.

I will make the changes

MangelMaxime:
Failing test is this one:

https://github.com/fable-compiler/Fable/blob/53238bc4278d33de8b2548af747aa18fef234474/tests/Js/Main/MiscTests.fs#L491-L499

Failing on this line `equal<float64>(80, item(1, ((parsed["widget"])["text"])["size"])["height"]);` because TypeScript infers the type of `item<'T>` to `item<unknown>`.

Should we provide the generic type to `item` ? Also looking at the compilers directives for that test they feels strange. Is it expected that this test is only for TypeScript or should it run for JavaScript too ?

@ncave

ncave:
@MangelMaxime Also, on an unrelated note, recently we tend to merge a lot of commits with every PR.
Perhaps we should always "squash and merge" the PRs instead, otherwise the commit history becomes somewhat unusable to figure out what changed and when.

MangelMaxime:
> Can you paste the TS code that was generated for the test?

Generated code looks like this:

```ts
testCase("Fable.JsonProvider works", (): void => {
    const parsed: any = JSON.parse(ANOTHER_JSON);
    equal<boolean>(false, (parsed["widget"])["debug"]);
    equal<string>("lots of", ((parsed["widget"])["text"])["data"]);
    equal<float64>(80, item(1, ((parsed["widget"])["text"])["size"])["height"]);
    equal<float64>(500, ((parsed["widget"])["text"])["vOffset"]);
})
```

> @MangelMaxime Also, on an unrelated note, recently we tend to merge a lot of commits with every PR.
> It may be a good idea to always "squash and merge" the PRs instead, otherwise the commit history becomes somewhat unusable to figure out what changed and when.

Noted, I agree and it will avoid commits like `Update changelog`... I will give it a try and we will see how it looks like.

MangelMaxime:
I didn't think about it.

I will also have a look at the setter code.

MangelMaxime:
I manage to add the fix for `Array.set` and `ar[idx] <- value` syntax.

For the second one, I add to unwrap `fableExpr` in `transformSet` quite a bit but I didn't see any other places to capture the `ExprSet`. And it seems like some other module do make some call to `libCall` even if this is not the norm in that file.

ncave (approved):

tests/Js/Main/MiscTests.fs
```diff
@@ -489,7 +489,12 @@ let tests =
         to_str () |> equal $"{()}"
 
 #if FABLE_COMPILER
-#if !FABLE_COMPILER_JAVASCRIPT
+#if !FABLE_COMPILER_TYPESCRIPT
+    // The tests seems to cause problems in TypeScript because of the typing
+    // system. We suppose this is probably due to the TypeProvider generation
+    // so for now we just ignore that situation hoping this is not
+    // indicative of other potential issues like this
+    // See: https://github.com/fable-compiler/Fable/pull/3748
```
ncave, line 497: @MangelMaxime Looks like that test is failing for JS too

MangelMaxime: This seems to be because I forgot to update the compiler directive for the `open` statement.

I have it working locally now, let's hope this is the final CI run and we can merge this PR.

MangelMaxime: Looks like not, the test was disable for JavaScript because when testing using `fable-standalone` we don't support TypeProviders. 

But ok, that's not related to my PR changes will fix it with an additional compiler directive.

tests/Js/Main/MiscTests.fs
```diff
@@ -468,7 +468,7 @@ type Order =
         quantity : int<kg>
     }
 
-#if !FABLE_COMPILER_TYPESCRIPT
+#if FABLE_COMPILER && !FABLE_COMPILER_TYPESCRIPT && !NPM_PACKAGE_FABLE_COMPILER_JAVASCRIPT
```
ncave, line 471: @MangelMaxime I wonder if we can simplify this to be just:
```fs
#if !FABLE_COMPILER
```
instead of introducing yet another define:
```fs
#if FABLE_COMPILER && !FABLE_COMPILER_TYPESCRIPT && !NPM_PACKAGE_FABLE_COMPILER_JAVASCRIPT
```

PR #3749, opened 2024-02-10
Prepare for release of NPM packages as part of @fable-org

MangelMaxime:


MangelMaxime:
Ok this is almost ready, now I would just like to adapt `fable-standalone` internal path to use `fable-library-js` instead of `fable-library` for consistency.

MangelMaxime:
@ncave I invited you to be a member of the `@fable-org` NPM organisation.

I don't expect you to make release I am happy to continue to handle that process but I am trying to give access to this organisation to Fable maintainers, with the goal to avoid the situation where we can't release a package.

PR #3757, opened 2024-02-14
Updated fable-library TFM

ncave:
- Updated `fable-library` projects TFM to `net8.0`.


MangelMaxime:
I think there is no harm in upgrading the TFM for these as we don't publish them.

For projects that we publish on NuGet, we need to be more conservative because the TFM is going to impact the dependency resolution + SDKs required on user machine.

PR #3758, opened 2024-02-17
[Fable Compiler] AST endpoint

nojaf:
I would like to programmatically get access to `Fable.AST` from a project.

Moving `MSBuildCrackerResolver` to Fable.Compiler.
Adding a new API `compileFileToFableAST`.

I want this because I'm playing around with building a simple AST viewer: 
Sample

//cc @damianreeves


src/Fable.Compiler/MSBuildCrackerResolver.fs
```diff
@@ -0,0 +1,179 @@
...
+                    "/p:RestoreUseStaticGraphEvaluation=False"
+                    // Avoid restoring with an existing lock file
+                    "/p:RestoreLockedMode=false"
+                    "/p:RestorePackagesWithLockFile=false"
+                    // We trick NuGet into believing there is no lock file create, if it does exist it will try and create it.
+                    " /p:NuGetLockFilePath=Fable.lock"
+                    // Avoid skipping the CoreCompile target via this property.
+                    $"/p:NonExistentFile=\"%s{nonExistentFile}\""
+                ]
+                |> List.filter (String.IsNullOrWhiteSpace >> not)
```
MangelMaxime, line 118: To me it looks like this filter always return the full array as we don't have any empty string in it.

We the change of scope of `configuration` we will probably need to keep this line.

src/Fable.Compiler/MSBuildCrackerResolver.fs
```diff
@@ -0,0 +1,179 @@
...
+        dotnet_msbuild_with_defines fsproj args List.empty
+
+    let mkOptionsFromDesignTimeBuildAux (fsproj: FileInfo) (options: CrackerOptions) : Async<ProjectOptionsResponse> =
+        async {
+            let! targetFrameworkJson =
+                let configuration =
+                    if String.IsNullOrWhiteSpace options.Configuration then
+                        ""
+                    else
+                        $"/p:Configuration=%s{options.Configuration} "
```
MangelMaxime, line 74: We can move this variable in the scope above as we have a similar logic on line 102-103.

MangelMaxime:
@nojaf I left 2 comments but other than that it looks good to me

DamianReeves:
Thanks for this (was on vacation last week so I didn't get to play with this, but going to dive into it now)

PR #3760, opened 2024-02-18
WIP: Add support for `StringBuilder.Chars` for Rust

MangelMaxime:
Hello @ncave ,

While trying to add support for `StringBuilder.Charts` to Rust I run into an issue where `fable-library-rust` cannot be compiled.

```
Checking fable_library_rust v0.1.0 (/Users/mmangel/Workspaces/Github/fable-compiler/Fable/temp/fable-library-rust)
error[E0599]: no method named `Clear` found for reference `&module_5bc0559d::System::Text::StringBuilder` in the current scope
   --> src/./System.Text.rs:210:37
    |
210 | ...                   (_self_.Clear()).Append_Z372E4D23(bufferText_1);
    |                               ^^^^^ method not found in `&StringBuilder`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
note: `ICollection_1` defines an item `Clear`, perhaps you need to implement it
   --> src/./Interfaces.rs:245:17
    |
245 |                 pub trait ICollection_1<T: Clone + 'static> {
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0599`.
error: could not compile `fable_library_rust` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `fable_library_rust` (lib test) due to previous error
```

While looking at the generated source code I discovered that `member` have `self: &Lrc<Self>` generated but `getter/setter` have `&self`.

Manually changing

```rs
pub fn set_Chars_413E0D0A(self, index: i32, value: char) {
```

to

```rs
pub fn set_Chars_413E0D0A(self: &Lrc<Self>, index: i32, value: char) {
```

seems to make the error go away at least in the IDE.

I found that the line responsible for generating `&Lrc<Self>` is 

https://github.com/fable-compiler/Fable/blob/f96d458f044e98cf158dad8567691a7fab127b7e/src/Fable.Transforms/Rust/Fable2Rust.fs#L3352-L3374

But it doesn't work in the current situation because the `ident.Type` and `returnType` are different. I am not sure what is the path of action to take from here.

### DIFF
--- a/src/fable-library-rust/src/System.Text.fs
+++ b/src/fable-library-rust/src/System.Text.fs
@@ -44,6 +44,24 @@ type StringBuilder(value: string, capacity: int) =
         buf.Clear()
         x
 
+    member x.Chars
+        with get (index: int) =
+            let bufferText = x.ToString()
+
+            if index < 0 || index >= bufferText.Length then
+                failwith "Index was outside the bounds of the array"
+            else
+                bufferText.[index]
+        and set (index: int) (value: char) =
+            let bufferText = x.ToString()
+
+            if index < 0 || index >= bufferText.Length then
+                failwith "Index was outside the bounds of the array"
+            else
+                let bufferText = bufferText.ToCharArray()
+                bufferText.[index] <- value
+                x.Clear().Append(bufferText) |> ignore
+
     member x.Length = buf |> asArray |> Array.sumBy (fun s -> s.Length)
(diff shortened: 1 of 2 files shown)


ncave:
@MangelMaxime Thanks, I'll take a look.

ncave:
Superseded by #3761.

PR #3769, opened 2024-02-25
Dependent signature files

nojaf:
I noticed some gaps when working with signature files in my project:

- A type was not hidden by the signature file (similar to https://github.com/fable-compiler/Fable/pull/3530)
- The dependent files of the current file should be detected for the signature file if there is one present. (This is due to the nature of graph-based checking)

### DIFF
--- a/src/Fable.Compiler/Library.fs
+++ b/src/Fable.Compiler/Library.fs
@@ -263,6 +263,9 @@ module CodeServices =
             let! dependentFiles =
                 checker.GetDependentFiles(currentFile, crackerResponse.ProjectOptions.SourceFiles, sourceReader)
 
+            let combinedDependentFiles = String.concat "\n" dependentFiles
+            Log.info $"Dependent files for %s{currentFile} are:\n%s{combinedDependentFiles}"
+
             let lastFile =
                 if Array.isEmpty dependentFiles then
                     currentFile
--- a/src/Fable.Compiler/Library.fs
+++ b/src/Fable.Compiler/Library.fs
@@ -260,11 +260,23 @@ module CodeServices =
         : Async<CompileResult>
         =
         async {
-            let! dependentFiles =
-                checker.GetDependentFiles(currentFile, crackerResponse.ProjectOptions.SourceFiles, sourceReader)
+            let signatureFile =
+                crackerResponse.ProjectOptions.SourceFiles
+                |> Array.tryFind (fun f -> f = String.Concat(currentFile, "i"))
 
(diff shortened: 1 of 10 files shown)


nojaf:
Ready for review

src/Fable.Compiler/Library.fs
```diff
@@ -260,8 +260,23 @@ module CodeServices =
...
+                    checker.GetDependentFiles(currentFile, crackerResponse.ProjectOptions.SourceFiles, sourceReader)
+                | Some signatureFile ->
+                    checker.GetDependentFiles(signatureFile, crackerResponse.ProjectOptions.SourceFiles, sourceReader)
+
+            Log.info (
+                sprintf
+                    "Dependent files for %s are:\n%s"
+                    (Option.defaultValue currentFile signatureFile)
+                    (String.concat "\n" dependentFiles)
+            )
```
MangelMaxime, line 279: I think we should remove this log as we don't have yet a debug mode for the log and it will "pollute" the already really verbose `-- verbose` mode.

And I think this log is mostly for debug purpose.

What do you think?

src/Fable.Compiler/Util.fs
```diff
@@ -460,7 +460,9 @@ module Process =
 
     let runSyncWithOutput workingDir exePath args =
         let p = startProcess true [] workingDir exePath args
-        p.WaitForExit()
+        // Don't wait indefinitely to run process
+        // This call is used to build local plugins, if the binary is used by another process this process will never end.
+        p.WaitForExit 7000 |> ignore
```
MangelMaxime, line 465: Does this means Fable will crash or will continue without the new DLLs being built?

MangelMaxime (changes requested):

src/Fable.Compiler/Library.fs
nojaf: I added this log for a reason, it is incredibly useful and this is not called by `Fable.Cli`.
It happens exclusively in `Fable.Compiler` API endpoint. Users will never see this and I want to have this log.

src/Fable.Compiler/Util.fs
nojaf: It would continue without the new DLL being built.
This can btw only happen when a user is referencing a local project that contains a Plugin.

src/Fable.Compiler/Library.fs
MangelMaxime: Ah ok no problem, I thought it was going to go to the end user too.

PR #3782, opened 2024-03-12
[JS] Add ConditionalWeakTable

chkn:
Adds an implementation of [ConditionalWeakTable](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2?view=net-7.0) for JS/TS using [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap).

Missing/different behavior to .NET:

- The [GetOrCreateValue](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2.getorcreatevalue?view=net-7.0) method is not implemented. It might be possible to invoke the default constructor of the value type if the type info for the generic argument is saved when the `ConditionalWeakTable` is constructed, but I did not implement that.
- It is not iterable. Though `WeakMap` is not enumerable, it may be possible to support this by keeping an array of `WeakRef`s in addition to the `WeakMap`. But this seems like a lot of overhead, especially since the .NET version only implements `IEnumerable` as an explicit implementation.
- You cannot use JS primitive types as keys, including strings. The workaround is to manually box them using, e.g. `new String(str)`

With this PR, any `WeakMap` obtained from JS should be usable in F# as a `ConditionalWeakTable`, but the `Clear` method will only work on `ConditionalWeakTable`s constructed from F#.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* [JS/TS] Add `ConditionalWeakTable` (by @chkn)
+
 ## 4.15.0 - 2024-03-18
 
 ### Fixed
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -2636,6 +2636,60 @@ let dictionaries (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Exp
       Some c -> Helper.InstanceCall(c, methName, t, args, i.SignatureArgTypes, ?loc = r) |> Some
     | _ -> None
 
+let conditionalWeakTable (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr option) (args: Expr list) =
+    match i.CompiledName, thisArg with
+    | ".ctor", _ ->
+        match i.GenericArgs with
+        | [ keyType; _ ] ->
+            match keyType with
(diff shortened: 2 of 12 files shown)


MangelMaxime:
Hello thanks for the PR, 

I am not familiar with ConditionalWeakTable so I not use what they are used for.  Do you need `ConditionalWeakTable` for a specific usage?

It seems like the user needs to be really careful when using it the implementation from this PR, and have a high change of shooting himself in the foot for this reason if we are to accept this PR, it would be nice to also have a new paragraph on [Fable documentation here](https://fable.io/docs/javascript/compatibility.html).

Could you please write a documentation for it, with examples close to a real usage? So people can refers to it if needed?

chkn:
> I am not familiar with ConditionalWeakTable so I not use what they are used for. Do you need `ConditionalWeakTable` for a specific usage?

It is analogous to `WeakMap` in JS. Basically, a map data structure where keys are always compared by reference, and being present in the map doesn't keep key objects alive. If the key is garbage collected, the key-value pair is removed from the map. It's handy for adding "expando" properties to objects without leaking them.

> It seems like the user needs to be really careful when using it the implementation from this PR, and have a high change of shooting himself in the foot for this reason if we are to accept this PR, it would be nice to also have a new paragraph on [Fable documentation here](https://fable.io/docs/javascript/compatibility.html).
> 
> Could you please write a documentation for it, with examples close to a real usage? So people can refers to it if needed?

Sure, how about this? https://github.com/fable-compiler/fable-compiler.github.io/pull/182

MangelMaxime:
Thank for the explanation and PR.

It has been released as part of Fable 4.16.0

Issue #3794, opened 2024-03-26
Better support for fable-compiler-js


ken-okabe:
### Description

Fable and [fable-compiler-js](https://github.com/fable-compiler/Fable/tree/main/src/fable-compiler-js) are astonishing projects/products.

According to the [Changelog of fable-compiler-js](https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/CHANGELOG.md), recently, the project status has achieved to stable.

 >1.0.0 - 2024-02-12
>Release stable version

I'm happy to see the stable version release and really appreciate your continued development.

I’m writing articles on F# and planning to introduce Fable. Especially, I observe that fable-compiler-js is promising as an AltJS solution with an easy setup using just Node.js, without the need for a .NET installation.

**Here, I would like to request better support for fable-compiler-js.**

[README.md](https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/README.md)

The information:

>This version is somewhat experimental

looks outdated, and

```
npm install fable-compiler-js
npx fable <PROJECT_PATH> <OUT_DIR> [--options]
```

although the installation succeeded, the `npx fable` command fails: 
[fable-compiler-js package.json dependencies are broken #3793](https://github.com/fable-compiler/Fable/issues/3793)

After manual modification by myself, `npx fable` seems still unusable.

`$ npx fable --help`

```
file:///home/ken/Documents/p/c/fable/counter/node_modules/@fable-org/fable-compiler-js/dist/app.min.js:1
import{getAssembliesDir as Js}from"@fable-org/fable-metadata";import*as te from"fs";import*as he from"path";import Zs from"child_process";import{createRequire as Us}from"node:module";import{init as
...
...
SyntaxError: The requested module '@fable-org/fable-standalone' does not provide an export named 'init'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:134:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:217:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:120:12)

Node.js v21.7.1
```

### Repro code

https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/README.md
https://www.npmjs.com/package/@fable-org/fable-compiler-js

```
npm install fable-compiler-js
npx fable --help
```

### Expected and actual results

[...]

Issue #3793, opened 2024-03-26
fable-compiler-js `package.json` dependencies are broken


ken-okabe:
### Description

https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/package.json

```
  "dependencies": {
    "fable-metadata": "^2.0.0",
    "fable-standalone": "^3.0.0"
  },
```

is outdated, and  since

https://github.com/fable-compiler/Fable/blob/main/src/fable-compiler-js/src/app.fs

tries to call new npm package: `@fable-org/fable-metadata` etc., `npx fable` fails to launch.

```
node:internal/modules/esm/resolve:845
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@fable-org/fable-metadata' imported from /home/ken/Documents/p/c/fable/counter/node_modules/@fable-org/fable-compiler-js/dist/app.min.js
    at packageResolve (node:internal/modules/esm/resolve:845:9
```

---

Also, 

`dirname-filename-esm` dependency is missing.

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'dirname-filename-esm' imported from /home/ken/Documents/p/c/fable/counter/node_modules/@fable-org/fable-compiler-js/node_modules/@fable-org/fable-metadata/index.js
```

### Repro code

```
npm install fable-compiler-js
npx fable <PROJECT_PATH> <OUT_DIR> [--options]
```

https://www.npmjs.com/package/@fable-org/fable-compiler-js

### Expected and actual results

`npx fable` command should work after the `npm` installation, but actually fails because the `package.json` dependencies are broken 

### Related information

@fable-org/fable-compiler-js
1.1.0

PR #3795, opened 2024-03-27
WIP: Fixing fable-compiler-js when run from NPM

MangelMaxime:
#3794 #3793 

How to run:

1. `./build.sh compiler-js`
2. `cd src/fable-compiler-js`
3. `npm link @fable-org/fable-metadata ../fable-metadata`
4. `node index.js`

`node index.js --version` works but `node index.js test.fsx` fails ATM

To avoid the problem with having it working locally but not when packaged, we should probably remove the compiler directives and make use of `npm link` so it behave the same locally as if published.


ncave:
Thanks @MangelMaxime!

PR #3804, opened 2024-04-10
[Python] Use Rust (pyo3) for core part of Fable Library

### LINKED ISSUE #3766: [Python] Boxing bytes looses type information
### Description

When boxing a `byte` value, it looses its underlying type information.

#### test.fsx
```fsharp
let intValue = box 1
let byteValue = box 2uy

let matchType (v : obj) =
    match v with
    | :?int as i -> printfn "int: %d" i
    | :?byte as u -> printfn "byte: %u" u
    | :?uint8 as u -> printfn "uint8: %u" u // equals byte, just for testing against Fable
    | _ -> printfn "other"

matchType intValue // int: 1
matchType byteValue // other

printfn $"{2uy.GetType().Name}" // Byte
printfn $"{byteValue.GetType().Name}" // Object
```
#### test.cmd
```shell
dotnet fable . --lang python
python test.py
```

Expected result: Boxed byte value should match the reflective match case, and calling `GetType().Name` should return "Byte"

Actual result: Match case is not hit and "Object" is returned

### Related information

* `dotnet fable --version`: 4.13.0
* `dotnet tool list/update/install`: 4.13.0
* Windows11

@freymaurer

### LINKED ISSUE #3832: [Python] `--fableLib fable-library` crashing dependencies.
I am not sure if this is intended or not, but using `--fableLib fable-library` will crash the dependencies of the published files:

This is the command i use for transpilation 

`dotnet fable <path> -o <out_path> --noCache --lang py --fableLib fable-library --noReflection`

This results in an empty `fable_modules` folder and the following error message:

```
Traceback (most recent call last):
  File "c:\Users\Kevin\Desktop\test\py.py", line 1, in <module>
    from siren_dsl import siren, git
  File "C:\Users\Kevin\Desktop\test\.venv\Lib\site-packages\siren_dsl\__init__.py", line 1, in <module>
    from .siren_types import SirenElement
  File "C:\Users\Kevin\Desktop\test\.venv\Lib\site-packages\siren_dsl\siren_types.py", line 5
    from .fable-library.list import (FSharpList, of_array, singleton as singleton_1, empty as empty_1, fold, map)
               ^
SyntaxError: invalid syntax
```

This is what the python file looks like

```python
from collections.abc import Callable
from dataclasses import dataclass
from typing import (Any, TypeVar)
from .fable-library.list import (FSharpList, of_array, singleton as singleton_1, empty as empty_1, fold, map)
from .fable-library.option import (value as value_2, map as map_1, default_arg)
from .fable-library.reflection import (TypeInfo, union_type, string_type, list_type, option_type, record_type, bool_type, class_type)
from .fable-library.seq import (to_list, delay, append, singleton, collect, empty)
from .fable-library.string_ import (join, to_text, printf)
from .fable-library.types import (Array, Union, Record, to_string, seq_to_string)
from .fable-library.util import IEnumerable_1
from .yaml import (Yaml_line, Yaml_level, Yaml_AST, Yaml_root, Yaml_write)
```

This is what the dependencies look like:

image

dbrattli:
## Why

There are some edge cases where it's really hard to mimic .NET e.g., for numeric types (`uin16, int8`, ...) and arrays. It will also enable Fable Python to have much better performance i.e run at native speed when running Fable library code. In future we could consider sharing code with Fable Rust, but this is not a short-term goal since PyO3 is quite different from normal Rust code (e.g. PyO3 cannot use generic classes, macros etc., since types are not known until runtime)

For Python library designers it would also be great to be able to install Fable Library from PyPI. This has been requested several times before (#3832), and is now a must requirement when going for library backed by Rust since we need to pre-compile it for all architectures.

## How

Rewriting of fable-library for Python in Rust using [PyO3](https://pyo3.rs/) and [Maturin](https://www.maturin.rs/).  Note that this will be a gradual migration, not a big-bang release. 

Also start auto-publish Fable Library (python) to PyPI. We must do this since Rust extensions needs to be built for all architectures, not possible to do on local machine. The idea is to do this in a way where we have [no breaking changes](https://github.com/fable-compiler/Fable/blob/c9152364dc8615c20df2ab57fe79954e0d448172/src/fable-library-py/README.md), i.e any version of Fable can use the latest fable-library on PyPI.

Allows use of fable-library from PyPI when compiling with `--fableLib pypi`. This will look for fable-library in the default Python search path i.e. use fable-library installed into `site-packages`.

Fixes #3766
Fixes #3832

### Task list

- [ ] Fable Library Core
  - [x] Use sized and signed integers from Rust instead of Python int
  - [x] Float32 and Float64 from Rust
  - [x] Option module written in Rust
  - [ ] Use Rust array instead of Python list/array/bytearray
- [x] Enable Fable Python to be used with bundled fable-library (default) or installed from PyPI
  - [ ] But `--fableLib` is currently a hidden argument. How should we document this?
- [x] CI/CD now handles and publishes Python wheels for all architectures.
- [ ] Add PYPI token as repository secret. Add @MangelMaxime as additional maintainer of [fable-library](https://pypi.org/project/fable-library/) on PyPI.


ncave:
@dbrattli You can convert this to draft PR so it won't be merged until ready.

ncave:
@dbrattli Just to clarify, while this is a great idea, do you have any concerns about losing some of the (potential) benefits of easy interoperability with other Python code by not using "native" Python types (or at least `numpy` types), or is that not a big deal if it can be provided through some interop layer?

dbrattli:
@ncave It was so hard to find the convert to draft link 🙈 but fixed now 😊  The custom Rust sized and signed integers will be implemented so they appear as regular ints to other Python code. Python will also coerce them to normal Python int when needed, e.g a function that expects a regular Python int, or if you try to do any arithmetic using them.  E.g:

```py
>>> from array import array # This is a Python array that expects Python ints
>>> from core import uint16 # This is a `pub struct uint16(u16);` from Rust.
>>> x = uint16(42)
>>> array("H", [x]) # works where Python expects a Python int
array('H', [42])
>>> x+5
47
>>> 5+x
47
```

Most of Python (and NumPy) is already implemented the same way using extension modules. NumPy is implemented in C++ and Python (cpython) is implemented in C. So interop will not be a problem since other Python code will not know the difference.  Everything will appear as regular Python modules and can be used by any Python code. A lot of new Python libraries are in fact implemented in Rust these days, e.g Pydantic v2. We just need to provide .pyi files so that static type checkers knows what is going on.

src/Fable.Transforms/Python/Fable2Python.fs
```diff
@@ -438,14 +438,14 @@ module Reflection =
...
+            | _, Fable.Type.Number(Int8, _) -> pyTypeof "<class 'fableInt8'>" expr
+            | _, Fable.Type.Number(Int16, _) -> pyTypeof "<class 'fable.Int16'>" expr
+            | _, Fable.Type.Number(UInt16, _) -> pyTypeof "<class 'fable.UInt16'>" expr
             | _, Fable.Type.Number(Int32, _) -> pyTypeof "<class 'int'>" expr
-            | _, Fable.Type.Number(UInt32, _) -> pyTypeof "<class 'fable_modules.fable_library.types.uint32>" expr
-            | _, Fable.Type.Number(Int64, _) -> pyTypeof "<class 'fable_modules.fable_library.types.int64'>" expr
-            | _, Fable.Type.Number(UInt64, _) -> pyTypeof "<class 'fable_modules.fable_library.types.uint32'>" expr
+            | _, Fable.Type.Number(UInt32, _) -> pyTypeof "<class 'fable.UInt32>" expr
+            | _, Fable.Type.Number(Int64, _) -> pyTypeof "<class 'fable.Int64'>" expr
+            | _, Fable.Type.Number(UInt64, _) -> pyTypeof "<class 'fable.UInt36'>" expr
```
ncave, line 448: @dbrattli typo

dbrattli: Thanks! I'll fix 🙈

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,280 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheels-sdist
+          path: ./temp/fable-library-py/dist
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'release' || github.event_name == 'workflow_dispatch' }}
+    needs: [linux, musllinux, windows, macos, sdist]
```
MangelMaxime, line 258: If I understand correctly, we use Github CI to build the binaries and then publish?

How do we trigger the release workflow? 

How is the version number determined? Is it based on the Github release version?

.github/workflows/build.yml
```diff
@@ -145,17 +145,15 @@ jobs:
...
-
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
 
       - name: Install poetry
         run: |
-          pip install poetry
+          pipx install poetry
```
MangelMaxime, line 155: Is `pipx` available out of the box or is it something we need to install?

If this is the latest, we need to update [Dockerfile](https://github.com/fable-compiler/Fable/blob/5a59e3a06e26902dbb6b6f14ebca83ed3fe54ca8/.devcontainer/Dockerfile)

.github/workflows/publish-pypi.yml
dbrattli: - Yes, we use Github Ci to build the binaries for every platform and architecture
- The idea is that the release is triggered by a new Fable release e.g [5.0.0-alpha.12](https://github.com/fable-compiler/Fable/releases/tag/5.0.0-alpha.12). So we build a new fable-library for every Fable release. 
- The version is generated from the tag of the release so 5.0.0-alpha-12 becomes fable-library-5.0.0a12 (because of [PEP-440](https://peps.python.org/pep-0440/))

MangelMaxime:
@dbrattli I read through the code, I think I understand most of it but not in details.

What is the current status/next steps of this PR?

Is it something we can merge and make changes incrementally or do we need to merge it only once all the APIs have been ported to the new Rust optimisation?

I am asking because, some people are using Fable 4/5 with the Python target and we shouldn't make release of Fable where Python stop working.

dbrattli:
@MangelMaxime The status of this PR is green, so we can make changes incrementally. I'm currently working on migrating Array.fs/py to Rust. Things should be good, but GitHub actions are always a bit complicated so it might take a few iterations to get everything right.  So, if we break things it will only be for a few days until we fine tune the GH PyPI publish action.

Using Fable locally will be the same as before. The breaking change is that library makers will have to use fable-library from PyPI, since building using Fable locally only generates for your own local platform/architecture e.g windows-x86. The PyPI version is built for all platforms/architectures.

Next steps are to complete the array migration, then take on other parts of fable-library, e.g list, string, .. which should give considerable performance improvements.

PS: I need to add a PyPI token as a repo secret to the Fable repository. How should we do this? Since I'm not an admin, you will have to do this, so either I give you the PYPI_TOKEN or I give you admin access to PyPI for fable-library so you can create it yourself.

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,280 @@
...
+        run: |
+          pipx install poetry
+          pipx install maturin
+      - name: Transpile F# to Python
+        run: ./build.sh fable-library --python
+      - name: Install Dunamai
+        run: pipx install dunamai
+      - name: Set version
+        run: |
+          VERSION=$(dunamai from git --format "{base}{stage}{revision}" --pattern "^(?P<base>\d+\.\d+\.\d+)((-(?P<stage>alpha|beta|theta))\.(?P<revision>\d+))?$" --latest-tag)
```
dbrattli, line 34: @MangelMaxime Here is the version generated from the latest tag

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,280 @@
...
+          # Convert stage names to PEP440 equivalents
+          VERSION=${VERSION/alpha/a}
+          VERSION=${VERSION/beta/b}
+          echo "Setting version to $VERSION"
+          cd temp/fable-library-py
+          poetry version $VERSION
+          mkdir -p fable_library
+          echo "__version__ = \"$VERSION\"" > fable_library/_version.py
```
dbrattli, line 42: Versioni is written back into the source code and package config

ncave:
@dbrattli I wonder if it is an option to compile the fable-library-py on the fly the first time it compiles a project, to match the target architecture.

dbrattli:
@ncave Yes, that is the idea. If you use Fable locally, you will build and use it as before. Only requirement is that for Python you also need to have `cargo` installed, so same dependencies as for Rust. Nothing new there, ref dependencies in README.

Now, the "challenge" is when you are the author of a Python library built using Fable that you want to distribute on PyPI. Then the fable-library for that library cannot be built locally, and you need to depend on the PyPI version of fable-library. You can build it locally using `--fableLib pypi` which will produce code that do not bundle fable-library, but expects it to be installed in PYTHON_PATH, (usually site-packages) from PyPI using uv, pip or poetry etc. That library then needs to have `fable-library` as a PyPI dependency. Does this answer your question?

.github/workflows/publish-pypi.yml
MangelMaxime: Thanks

MangelMaxime:
> The status of this PR is green, so we can make changes incrementally. I'm currently working on migrating Array.fs/py to Rust. Things should be good, but GitHub actions are always a bit complicated so it might take a few iterations to get everything right.

Yes, this is fine if Github actions takes a few iterations we all know the pain.

> PS: I need to add a PyPI token as a repo secret to the Fable repository. How should we do this? Since I'm not an admin, you will have to do this, so either I give you the PYPI_TOKEN or I give you admin access to PyPI for fable-library so you can create it yourself.

I can make you an admin, but I would also like to be part of the fable-library on PyPi this is to have redundancies in the access to avoid the situation we had to deal with for NPM when Alfonso was not available.

Do you know if PyPi as an org feature similar to NPM or NuGet. This is handy to share access or to also identify which packages we (as Fable maintainer) official maintains.


> @ncave Yes, that is the idea. If you use Fable locally, you will build on-the-fly and use it as before. Only requirement is that for Python you also need to have `cargo` installed, so same dependencies as for Rust. Nothing new there, ref dependencies in README.
> 
> Now, the "challenge" is when you are the author of a Python library (or app) built using Fable that you want to distribute on PyPI (i.e for non-Fable users). Then the fable-library for that Python library cannot be built locally, and you need to depend on the PyPI version of fable-library. You can now (this PR) build it locally using `--fableLib pypi` which will produce code that do not bundle fable-library, but expects it to be installed normally in PYTHON_PATH (usually site-packages) from PyPI using uv, pip or poetry etc. That library then needs to have `fable-library` as a PyPI dependency. Does this answer your question?

This is actually similar (not identical) to what happens for libraries build using with Fable and published to NPM.

Fable libraries published to NPM included their own version of `fable-library-js` but we can also make then depends on `@fable-org/fable-library-js` using `--fableLib`.

@dbrattli Regarding the merge of this PRI will wait for it to be green again. And if you want we could have a call on Discord or other, while we try to make the full release.

dbrattli:
@MangelMaxime I'll try to fix the PR. I just added a few files I should have waited with 🙈 I'll try to revert them. I almost have Array working as well, but we can add that in a separate PR. If you sign up for PyPI, and give me your username then I will add you as maintainer to the package. Would be nice with a call on discord. I'll add you so we can find a time that would work for both.

PS: What is the `--fableLib <??>` command you give to configure that you want to use fable-library from NPM? Then I can try to keep the syntax similar if possible. Is it actually the dot i.e `.`?

MangelMaxime:
> PS: What is the `--fableLib <??>` command you give to configure that you want to use fable-library from NPM? Then I can try to keep the syntax similar if possible. Is it actually the dot i.e `.`?

I believe for JavaScript at least (perhaps other targets too?) you can provide `--fableLIb XXX` and Fable will use `XXX`  instead of the default `./fable_modules/fable-library-js.4.14.0/` 

So instead of generating:

```js
import { toString, Union, Record } from "./fable_modules/fable-library-js.4.14.0/Types.js";
```

it will generate 

```js
import { toString, Union, Record } from "XXX/Types.js";
```

If possible, I think it makes sense to keep the same logic for Python as it seems to be the same usage. By default, you want `"./fable_modules/fable-library-py.4.14.0/` (pseudo code) unless the user provide `--fableLib` to consume the package from PyPi.

dbrattli:
Ok, I have changed it so if you give `--fableLib fable-library` then you will consume it from PyPI. Not really true, but you will consume it from PYTHON_PATH which in most cases is site-packages inside your Python installation / environment that is where packages from PyPI will be installed. I think that should align better.

Issue #3480, opened 2023-06-27
Function decorated with `[<NamedParams>]` fails at runtime if all the arguments are optional and none are provided


Freymaurer:
### Description

Hey i am working on a library we want to offer for js as well as f#. So to improve JS syntax we wanted to use `[<NamedParams>]` on members.
I got a very strange error of:
![image](https://github.com/fable-compiler/Fable/assets/39732517/976e4088-1b73-4783-87c8-a76aca69f0be)

.. when running my Fable.Mocha tests. To add to this:

- If i use `[<NamedParams(fromIndex=0)>]` or `[<NamedParams>]` it runs for Example 1.
- Example 2 raises a similiar error no matter if i use `(fromIndex)` or not. So i would suspect the issue being `?param` f# syntax
- Fable Repl replicating Example 2 runs.

### Repro code

**Fable Repl**

[...]

Issue #2614, opened 2021-11-15
Char addition is implemented as string concatenation


jwosty:
### Description

Adding two chars together is supposed to give one char as a result whose value is the addition of the numeric values of the two inputs.

### Repro code

Take the following code:
```fsharp
for i in 0..25 do
    printfn "%c" i ('A' + char i)
```

### Expected and actual results

CLR F# will the characters A-Z.

Fable mostly prints garbage:

```
A
A
A
A
A
A
A
A
A
A	
A
A
A
A
A
A
A
A
A
A
A
A
A
A
A
A
```

A workaround is to explicitly do the math as ints, then convert to char afterward:

```fsharp
for i in 0..25 do
    printfn "%i -> %c" i (char (int 'A' + i))
```

The compiled JS appears to be performing string concatenation instead of char addition:

```javascript
import { printf, toConsole } from "fable-library/String.js";

for (let i = 0; i <= 25; i++) {
    const arg10 = "A" + String.fromCharCode(i);
    toConsole(printf("%c"))(arg10);
}
```

### Related information

* Fable version: 3.4.7
* Operating system: Windows 10

alfonsogarciacaro:
I've added automatic conversions so char addition works as expected in .NET. However, please note the operation won't be very performance won't be very good as it's necessary to convert back and forth between number and string because Fable represents chars as strings of length 1 in runtime.

jwosty:
Interesting, and thanks for the fix.

Any reason that chars are represented as JS strings instead of numbers? Mostly for displaying them to "just work" most of the time, I assume (i.e. printing to console, etc)?

alfonsogarciacaro:
If I remember correctly, originally it was because enumerating a string in JS gives you 1-length strings, and it was difficult to make the conversion to uint16 in generic code dealing with enumerables. Although now we're indeed transforming the string when iterating to solve some other issues, I believe.

PR #3817, opened 2024-05-18
Fix broken --precompiledLib switch

DunetsNM:
This fixes two problems that broke `--precompiledLib` switch after upgrading from Fable 3 to Fable 4, both caused by change in data format that affects deserialization of precompiled inline_exprs_0 json files. 

* `MemberRefInfo.Attributes` is a sequence of abstract Attribute type, which can't be deserialized even when collection is empty.  Build crashes on attempt to link precompiledLib  with:

>  error EXCEPTION: Deserialization of interface types is not supported. Type 'Fable.AST.Fable.Attribute'. Path: $[0] | LineNumber: 0 | BytePositionInLine: 2.
>    at System.Text.Json.ThrowHelper.ThrowNotSupportedException(ReadStack& state, Utf8JsonReader& reader, NotSupportedException ex)

* `ValueKind.NumberConstant` stores the value as `obj` which deserializes from precompiledLib as ... `JsonElement`, compilation proceeds but at the end fails complaining that `JsonElement` is not a good type for a numeric value (rightfully so).

This pull request fixes both issues (at least now project builds with `--precompiledLib` switch)

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -482,6 +482,29 @@ type ArrayKind =
     | MutableArray
     | ImmutableArray
 
+[<RequireQualifiedAccess>]
+type NumberValue =
+    | Int8 of sbyte
+    | UInt8 of byte
+    | Int16 of int16
+    | UInt16 of System.UInt16
+    | Int32 of System.Int32
+    | UInt32 of System.UInt32
+    | Int64 of System.Int64
+    | UInt64 of System.UInt64
+    | Int128 of System.Int64 // System.Int128
+    | UInt128 of System.UInt64 // System.UInt128
+    | BigInt of bigint
+    | NativeInt of nativeint
+    | UNativeInt of unativeint
+    | Float16 of System.Single // System.Half
+    | Float32 of System.Single
+    | Float64 of System.Double
+    | Decimal of System.Decimal
+    // this will be coerced to type later / depends on lang
+    // TODO: review KindOfChar, it reeks of poor modelling
+    | KindOfChar of NumberKind * System.Char
+
 type ValueKind =
     // The AST from F# compiler is a bit inconsistent with ThisValue and BaseValue.
     // ThisValue only appears in constructors and not in instance members (where `this` is passed as first argument)
@@ -497,7 +520,7 @@ type ValueKind =
     /// String interpolation with support for JS tagged templates
     /// String parts length should always be values.Length + 1
     | StringTemplate of tag: Expr option * parts: string list * values: Expr list
-    | NumberConstant of value: obj * kind: NumberKind * info: NumberInfo
+    | NumberConstant of NumberValue * info: NumberInfo
     | RegexConstant of source: string * flags: RegexFlag list
     | NewOption of value: Expr option * typ: Type * isStruct: bool
     | NewArray of newKind: NewArrayKind * typ: Type * kind: ArrayKind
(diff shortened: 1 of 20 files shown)


DunetsNM:
while I'm here, maybe it makes sense to separate part of Fable.AST module that meant to be json-serializable from everything else? How to?

src/Fable.AST/Fable.fs
```diff
@@ -482,6 +482,29 @@ type ArrayKind =
...
+    | BigInt of bigint
+    | NativeInt of nativeint
+    | UNativeInt of unativeint
+    | Float16 of System.Single // System.Half
+    | Float32 of System.Single
+    | Float64 of System.Double
+    | Decimal of System.Decimal
+    // this will be coerced to type later / depends on lang
+    // TODO: review KindOfChar, it reeks of poor modelling
+    | KindOfChar of NumberKind * System.Char
```
ncave, line 506: I agree that doesn't make sense, when there is `CharConstant`.
Conversions from number to char can be expressed as `TypeCast`.

ncave (commented):
@DunetsNM Please keep in mind that any Fable AST change is a breaking change for plugins so it requires a major version update.

DunetsNM:
> @DunetsNM Please keep in mind that any Fable AST change is a breaking change for plugins so it requires a major version update.

@ncave  Should I change the version in Fable.AST.fsproj right in this PR ? Judging by history version increases committed separately / along with releases

If it has to be in this PR ,  I suppose we keep major version at 4 (as it seems to be anchored to Fable 4 ?) i.e. 4.4.0 => 4.5.0 ? Again, here I assume that based on previous history (quick search reveals examples of technically breaking changes  that were not versioned by a strict SemVer e.g. see  diff between "Release 3.7.1" and "Release 3.7.2" commits , it has breaking changes in diff)

DunetsNM:
@ncave I restored the char/number workaround in its original, tests should pass now.  I tried to fix it via TypeCast which broke the build and found myself in a rabbit hole trying to understand how it works. So sticking with good old "if it works don't touch it"

ncave:
@DunetsNM See https://github.com/DunetsNM/Fable/pull/1

ncave:
LGTM, the only outstanding issue is the (potential) breaking change for Fable plugins, as mentioned above.
Not sure what semver change that requires.

++ @MangelMaxime, if you know of any list of Fable plugins (looking up in GitHub only brings very old ones that probably have to be updated anyway to work with Fable 4).

ncave (approved):

DunetsNM:
These plugins were updated to Fable 4 (I use a fork of those) however they shouldn't break as they don't inspect `Attributes` or `NumberConstant` case directly

https://github.com/Zaid-Ajaj/Feliz/tree/master/Feliz.CompilerPlugins

MangelMaxime:
The only Fable plugin I am aware of is Feliz.CompilerPlugins.

If the tests are passing then I think it means that this plugin does not need an update because we have it tested as part of [tests/React/Fable.Tests.React.fsproj](https://github.com/fable-compiler/Fable/blob/856f9bad2fd705cf08dc95a1f0a32b6847696cc6/tests/React/Fable.Tests.React.fsproj)

For the version in the past, I upgraded only the major version so in this case it would become 4.5.0.

DunetsNM:
@MangelMaxime @ncave gentlemen when can I expect a new Fable nuget to be published (with this fix)?  Don't want to be pushy, just wondering if I have time to wait for it or rather should make a private nuget from my fork in the meantime

MangelMaxime:
I was looking into the PR and I have questions.

`MemberRefInfo.Attributes` is not the only place which currently have the signature `abstract Attributes: Attribute seq`.

For this reason, I suspect there are still situation where the `--precompiledLib` could still fails, no ?

Reading the exception error it seems like the problem is that the problem is because `Attribute` is an interface?

```fs
type Attribute =
    abstract Entity: EntityRef
    abstract ConstructorArgs: obj list
```

By looking into the Fable.AST this is not the only interface we have does that mean we don't have the problem with the others interfaces because the generated JSON doesn't include all the AST but only a subset of the AST which consist only of records, DUs and primitives types ?

Should we try to use a record instead of an interface for `type Attribute`? This way we would have the same signature for all `Attributes: ...` members/properties ?

[Looking online](https://khalidabuhakmeh.com/serialize-interface-instances-system-text-json), it also seems possible to create a custom converter to teach `System.Text.Json` how to deal with an interface. However, I am not sure how well it would played with the `abstract ConstructorArgs: obj list` property.

I am asking all this questions, in case we have a way to fix the issues in a consistent way and also without loosing information in the AST. I unsure if this is important or not, if you should it does not harm or it is too complex to do please feel free to tell me.

DunetsNM:
> because the generated JSON doesn't include all the AST but only a subset of the AST which consist only of records, DUs and primitives types ?

I believe this is the case.   Serialized json data has type of `(string * Fable.InlineExpr)[]`  which consists only of primitive and F# algebraic types. I did only informal analysis, perhaps we can write a property-based test that roundtrips random data?

Addition of `Attributes`  is traced back to #3480 fix and introduction of `obj` in `NumberConstant` is the fix of #2614

DunetsNM:
Speaking about losing the information in AST - my impression is that full information about the `Attributes` is not required in AST. I don't know if there's already 3rd party code that relies on the full Attributes info (probably not) but it was definitely an overkill for its original purpose: they were bolted on just to detect presence of `Fable.Core.ParamObjectAttribute`. 
Arguably it could be even less coupling than that e.g. `HasParamObjectAttribute: bool`, I'd prefer to expose bare minimum in published types and enrich it only when there's a use case for it. 

In similar manner `NumberConstant` was relaxed to a have `obj` field only to hack in `char` value into it (which is replaced with a cleaner way here thanks to @ncave )

MangelMaxime:
> Addition of `Attributes` is traced back to #3480 fix and introduction of `obj` in `NumberConstant` is the fix of #2614

I now remember indeed working on this issue (I mean git history doesn't lie 🤣).

And indeed, I used the existing `Attribute` interface because it was available and never though it would break something as the CI passed etc.

I agree that I don't think people rely on the `ConstructorArgs` information. Fable plugin is not something really common. Because, this PR don't break anything at least according to the CI I think we can merge it @ncave was already ok with it yesterday.

I will probably just add a comment on the `AttributeFullNames` property to link back to this PR.

Thanks a lot for the time you spend on this issue.

MangelMaxime:
@DunetsNM This is now available in Fable 4.18.0

PR #3819, opened 2024-05-19
[Python] Fix quotation of union string cases

dbrattli:
Fixes https://github.com/fable-compiler/Fable.Python/issues/46

PS: @MangelMaxime Looks like JS have the same issue


MangelMaxime:
I confirm after testing in the REPL this test is failling #3820

PR #3823, opened 2024-05-23
Fixed #3566

### LINKED ISSUE #3566: F# 8 and static member in interface
Hi, @MangelMaxime I just wanted to know if there are any plans to support F# 8. Especially I'm interested in [static member in interface](https://github.com/fsharp/fslang-suggestions/issues/1191). This would be I huge thing for the whole `Fable/Feliz` community since we could build strong typing for base `HTMLElement/React` + any other custom component with `list of props` as a param. Look at this example (@Zaid-Ajaj I think this could interesting for you):

```fsharp
open Browser.Types
open Fable.Core
open Feliz
open Fable.Core.JsInterop

// we want to distinguish between props of different components and all mkAttr() will return an extra generic arg 
// in the alternate scenario we can omit this and reuse everywhere IReactProperty to have more flexibility
[<Erase>]
type IProp<'t> = interface end

module Interop =
  let mkAttr<'comp> (name: string) (value: obj) = unbox<IProp<'comp>> (name, value)
  let createElement (name: string) (props: IProp<'comp> list) = Interop.createElement name !!props

// MOCK: Typescsript for React has generic FocusEvent<T> type, but Browser.Types don't support it
type FocusEvent<'t> = {| target: 't|}

// base props interface
[<Interface>]
type HtmlAttributes<'comp, 'element> =
  static member inline className(x: string) = Interop.mkAttr<'comp> "className" x
  static member inline onFocus(x: FocusEvent<'element> -> unit) = Interop.mkAttr<'comp> "onFocus" x 

// input props interface = Input + htmlAttributes
[<Interface>]
type input =
  inherit HtmlAttributes<input, HTMLInputElement>
  static member inline isChecked(x: bool) = Interop.mkAttr<input> "checked" x

[<Erase>]
type HTML =
  // notice: IProp<input> list - strong type check!
  static member inline input (props: IProp<input> list) = Interop.createElement "input" props
  
module Test =
  let test =
    HTML.input [
      // both HtmlAttributes and input props are available
      input.className "test"
      input.isChecked true
      // here we have a target of HTMLInputElement. no need to cast!
      input.onFocus(fun x -> printf $"{x.target.``checked``}")
    ]
```

Currently, we can achieve a similar effect by using `type-extensions`, however, this is not ideal - the user could be forced to import an extra thing to have all functions available (if there are in separated `assemblies/files/namespaces`).

ncave:
- [JS/TS/Python] Fixed interface static members (#3566)

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [JS/TS] Fixed TimeSpan.FromMilliseconds (#3815) (by @ncave)
 * [Python] Fixed quotation for union string cases (by @dbrattli)
 * [Python] Fixed casing issues with identifiers and reflection info (#3811) (by @dbrattli)
+* [JS/TS/Python] Fixed interface static members (#3566) (by @ncave)
 
 ## 4.17.0 - 2024-04-23
 
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -2681,7 +2681,7 @@ module Util =
 
         // Check if this is an interface or abstract/overriden method
         | _, Some entity when
-            entity.IsInterface
+            entity.IsInterface && memb.IsInstanceMember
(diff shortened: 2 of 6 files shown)


MangelMaxime:
Thanks a lot 🙏

PR #3825, opened 2024-05-23
Fixed `fable-compiler-js` includes

ncave:
- Fixed `fable-compiler-js` includes.

@MangelMaxime This fixes the incorrect `fable-compiler-js` includes.
Can you please push a new `fable-compiler-js` version with this fix, if you agree with it? Thanks!

### DIFF
--- a/src/fable-compiler-js/src/util.js
+++ b/src/fable-compiler-js/src/util.js
@@ -11,7 +11,7 @@ export function getVersion() {
 
 export function getFableLibDir() {
   const require = createRequire(import.meta.url);
-  return Path.join(Path.dirname(require.resolve("@fable-org/fable-standalone")), "fable-library-ts");
+  return Path.join(Path.dirname(require.resolve("@fable-org/fable-compiler-js")), "fable-library-js");
 }
 
 export function getDirFiles(dir) {


MangelMaxime:
Sure I am with the changes, I think the `@fable-org/fable-standalone` is relic of the past because `fable-library` was perhaps not distributed in the compiler package.

However, I think there is still an issue with the path of the `fableLib` because when looking into the package structure on NPM we have

CleanShot 2024-05-24 at 15 04 17

Note that the `fable-library-js` is under `/@fable-org/fable-compiler-js/dist/`.

So I suppose we should use `return Path.join(Path.dirname(require.resolve("@fable-org/fable-compiler-js")), "dist", "fable-library-js");` no ?

ncave:
@MangelMaxime
> Note that the fable-library-js is under /@fable-org/fable-compiler-js/dist/. 

Yes, but I think `require.resolve("@fable-org/fable-compiler-js")` does that for you, i.e. it resolved to `node_modules/@fable-org/fable-compiler-js/dist`. So adding another dist would break it, no?

Anyway, that's just a quick fix to make the already released compiler work at least for JS.
You are correct that we should be bundling other libraries too, thanks for opening an issue for it.

MangelMaxime:
> Yes, but I think `require.resolve("@fable-org/fable-compiler-js")` does that for you, i.e. it resolved to `node_modules/@fable-org/fable-compiler-js/dist`. So adding another `/dist/` would break it, no?

I didn't know about that. I will make a release and if this is not enough we can investigate more.

MangelMaxime:
I think `require.resolve("@fable-org/fable-compiler-js")` resolve to the package folder and not the dist folder.

```fs
import { printf, toConsole } from "/home/vscode/.npm/_npx/bfa75a35966fbb71/node_modules/@fable-org/fable-compiler-js/fable-library-js/String.js";

toConsole(printf("Hello wolrd!"));
```

When I try to run this code, it fails to import the module.

If I manually add `dist` to the path it is all good.

Reproduction:

1. Create a `test.fsx` file with `printfn "Hello"` in it
2. `npx @fable-org/fable-compiler-js test.fsx`
3. `node test.fs.js`

ncave:
@MangelMaxime You are right, sorry about that. 
I got fooled by the previous resolving to `fable-standalone`.

MangelMaxime:
@ncave No problem, I will update it and make a new release.

ncave:
@MangelMaxime Now I'm not sure how it works either.
Using `fable-compiler-js` 1.2.0 results in `/node_modules/@fable-org/fable-standalone/dist/fable-library-ts/String.js`, so `resolve` must be doing something different for `@fable-org/fable-standalone`.

MangelMaxime:
> Using `fable-compiler-js` 1.2.0 results in `/node_modules/@fable-org/fable-standalone/dist/fable-library-ts/String.js`, so `resolve` must be doing something different for `@fable-org/fable-standalone`.

I suspect this is because of where the entry point of the package is located:

- `@fable-org/fable-standalone` -> `"main": "./dist/bundle.min.js",`
- `@fable-org/fable-compiler-js` -> `"main": "./index.js",`

I made a new release for `@fable-org/fable-compiler-js` and according to my test above it is working.

Issue #3833, opened 2024-06-07
Question about compile time file ordering of dependent fable nuget packages


ThisFunctionalTom:
### Description

I have a fable compatible nuget packages `Constants` and `Library` where `Library` depends on `Constants`. 
When I add package reference to Library in our SPA app and compile it with fable I get F# compiler error while compiling sources of `Library` that the namespace of `Constants` is not known.

I executed fable compiler with --verbose flag and have seen that the source files of `Library` come before the source files of `Constants`. Both of the package sources can be found in fable_modules so I guess the dependencies resolved correctly.

Can I somehow influence the compile time ordering of files from fable_modules? Am I doing something wrong? How does the Fable order the files from referenced fable nuget packages? 

### Expected and actual results

I would expect the fable compiler to respect the nuget package dependencies order and sort the files when compiling accordingly. In my case first all the files from `Constants` should be compiled before the files of `Library` and at the and the files from the SPA app.

### Related information

* Fable version: 4.18.0 
* Operating system: windows 11

ThisFunctionalTom:
I just found that there is a project_cracked.json file. In this file the files are in the correct order but in the --verbose log the files are in different order.

MangelMaxime:
Hello @ThisFunctionalTom,

Are you trying to consume the packages via NuGet or using `ProjectReference`.

If this is the former, you need to make sure to release a Fable compatible package. The main things being you need to include the F# source files in a `fable` folder.

In order to do so, you can use the all new shiny [Fable.Package.SDK](https://github.com/fable-compiler/Fable.Package.SDK) which automate a lot of things for you and guide you.

Or you can do it the manual way by using [Fable documentation](https://fable.io/docs/your-fable-project/author-a-fable-library.html) (in the future it will be updated to use Fable.Package.SDK as it more friendly to use).

ThisFunctionalTom:
Hi Maxime,

Yes both nuget packages `Constants` and `Library` are fable compatible with sources in fable folder. I also find this sources in fable_modules of the SPA app. As I said, I guess it is the order of the files on the command line of F# compiler that is wrong. Here the files of the `Constants` come after the files of the `Library` and F# compiler fails because in `Library` we use `Constants`.

What's strange is that in the project_cracked.json file the order of the files is correct.

Maybe if you could navigate me where the F# compiler is called in Fable I could maybe find the error myself or maybe find what am I doing wrong.

[...]

PR #3835, opened 2024-06-09
use case insensitive comparison when searching for dependencies between fable packages

ThisFunctionalTom:
This should fix the [Issue 3833](https://github.com/fable-compiler/Fable/issues/3833)

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * [TS] Fixed interface indexers (#3830) (by @ncave)
+* [GH-3835](https://github.com/fable-compiler/Fable/pull/3835) [All] Use case insensitive comparison when searching for dependencies between fable packages (by @ThisFunctionalTom)
 
 ## 4.18.0 - 2024-05-23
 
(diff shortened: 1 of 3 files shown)


MangelMaxime:
@ThisFunctionalTom Where you able to test it using a locally build version of Fable (via `./build.sh package`)?

Running `./build.sh package` will give you instruction on how to use that local version of Fable. I am asking, because this fix is not something we can catch easily using the CI I think.

ThisFunctionalTom:
> @ThisFunctionalTom Where you able to test it using a locally build version of Fable (via `./build.sh package`)?
> 
> Running `./build.sh package` will give you instruction on how to use that local version of Fable. I am asking, because this fix is not something we can catch easily using the CI I think.

I debuged fable.cli on my project with a breakpoint and looked at the file . I thought about adding tests for `sortFablePackages` which shouldn't be that hard because it is a pure function but I didn't find the right place where to put the tests.

I will try running it with `build.sh` later but I would probably need to use `build.cmd` because I am on windows.

MangelMaxime:
> I will try running it with `build.sh` later but I would probably need to use `build.cmd` because I am on windows.

Yes should be the same.

> Do you have an idea where should I put the tests?

Perhaps the `tests/Integration/Integration/Fable.Tests.Integration.fsproj` project could be the place for it? It seems like we are testing some internal APIs from Fable in it.

ThisFunctionalTom:
Oh, I cannot test it because the function `sortFablePackages` and type `FablePackage` is not public (they are not in the fsi file).

We used to make it internal and add `[assembly:InternalsVisibleTo("MyTestAssembly")]` to expose internal functions and types to test projects.

Should I do this? Is this compatible with fable repository coding style? Is there another way to test private functions?

ncave:
I think it's fine as proposed, the [spec](https://learn.microsoft.com/en-us/nuget/reference/nuspec#id) says that package ids are case-insensitive.

MangelMaxime:
Let's keep it simple and merge the PR as is.

MangelMaxime:
This has been released

PR #3844, opened 2024-06-13
Fixes #3837

### LINKED ISSUE #3837: Missing discriminator field when using TypeScriptTaggedUnion
### Description

When creating an enum value from a TypeScriptTaggedUnion, the type field doesn't get added if a case has a single argument.

### Repro code

```fsharp 
open Fable.Core

[<TypeScriptTaggedUnion("type")>]
type AuthState =
| Zero
| One of string
| Two of string * string

let zero = AuthState.Zero
let one = AuthState.One "a"
let two = AuthState.Two ("a", "b")

let check = function 
  | Zero -> "zero"
  | One _ -> "one"
  | Two _ -> "two"
```

https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwFgBQBA2gDwAqAnuAMoDGWAliAC5lwDm7YAJgKoQNgEABQAiZlTCiAlAD4AugQngoAQQCuzABbVmcZmCgBeAgB8oALTBZgZqAHkIh4ADMoAZ2aMI7O2QDuwFCuHl4MPlAAVKHevoT4KMxQAF7WQUZqmjp6BqhWNgSJwU7Gmdq6+miOhqJwooVgScyBpRrlOWgBQWJ1ADRQoggyBA1JtFpgtADWpS7qELTMgtAEUFDm+UEAtLIDqTb1+Gvm1VAA+lA7A0JSq+tQXeeXu+KBokA&html=Q&css=Q

The `One` case gets compiled to the raw value, without the discriminator field "type", while the pattern match is checking for a field that will never be present:

```js
export const one = "a";
....
export function check(_arg) {
    switch (_arg.type) {
        case "one":
....
```

I would expect it to look similar to `Two`:

```js
export const one = {
    type: "one",
    Item1: "a"
};
```

### Related information

* Fable version: 4.18.0

ncave:
- [JS/TS] Fixed missing field using TypeScriptTaggedUnion (#3837)

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [All] Fixed Fable compiler hanging on some errors (#3842) (by @ncave)
 * [JS/TS] Fixed DateTime.MinValue, DateTime.MaxValue (#3836) (by @ncave)
+* [JS/TS] Fixed missing field using TypeScriptTaggedUnion (#3837) (by @ncave)
 
 ## 4.19.0 - 2024-06-10
 
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -134,8 +134,9 @@ type FsUnionCase(uci: FSharpUnionCase) =
             | _ -> None
         )
 
-    static member HasNamedFields(uci: FSharpUnionCase) =
-        not (uci.Fields.Count = 1 && uci.Fields[0].Name = "Item")
+    // static member HasNamedFields(uci: FSharpUnionCase) =
+    //     not (uci.Fields.Count = 1 && uci.Fields[0].Name = "Item")
+    //     true
 
     interface Fable.UnionCase with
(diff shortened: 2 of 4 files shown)


ncave:
@MangelMaxime For some reason that I don't quite understand, there seems to be a special case for single unnamed union case fields (when used with `TypeScriptTaggedUnion`).

Disabling it fixes the linked issue, tests are still passing, but please chime in if you know more about it.

MangelMaxime:
I don't know much about `TypeScriptTaggedUnion`.

I never really used the TypeScript target unless for writing the documentation about it. Which is kind of ironic 🤣

I suppose you left the previous code commented on purpose?

ncave:
@MangelMaxime 
> I suppose you left the previous code commented on purpose?

Yes, as I don't understand what was it supposed to do (some sort of optimization perhaps for certain cases of unions with only one unnamed field?).
And it's not just for TypeScript. Using `TypeScriptTaggedUnion` attribute will add a tag property for JS too.

MangelMaxime:
Perhaps, it was trying to do something similar to `StringEnum` which allows to map `DUs` case to a string.

But if someone, wants to do that they can use `StringEnum` and now `TypeScriptTaggedUnion` is consistent for all type of cases. We will see, if someone report a problem with it that we are not aware of.

PR #3872, opened 2024-07-23
add missing unicode categories in python library

joprice:
This adds the missing unicode categories to fix the error `ValueError: Fable error, unknown Unicode category: Ps` when calling for example, `Char.IsLetterOrDigit` with left paren '('. 

I used the values defined in the referenced doc https://docs.microsoft.com/en-us/dotnet/api/system.globalization.unicodecategory?view=net-6.0, and also found that `No` was assigned to `UnicodeCategory.OtherLetter` instead of `UnicodeCategory.OtherNumber`.  

I'm not sure how to test the surrogate category, as I hit an error I left a note about that I didn't get a chance to look into.

The test data was created by running the following script

```python
import sys
import unicodedata
from collections import defaultdict

unicode_category = defaultdict(list)
for c in map(chr, range(sys.maxunicode + 1)):
    unicode_category[unicodedata.category(c)].append(c)

for value in unicode_category.values():
    c = value[0]
    e = c.encode("unicode_escape")
    print(repr(e), unicodedata.category(c))
```

which groups chars by category and then prints each category with a sample value.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [All] Updated Fable-FCS to latest F# 9.0 (by @ncave)
 * [All] Updated metadata to latest .NET 9.0 (by @ncave)
 
+### Fixed
+
+* [Py] Add missing unicode categories in python library (by @joprice)
+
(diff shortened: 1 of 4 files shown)


ncave:
@MangelMaxime LGTM

MangelMaxime:
@joprice Sorry for the long wait, seems like I missed this PR 😅

PR #3890, opened 2024-09-18
Handle escaping of { and } in FormattableString

### LINKED ISSUE #3889: System.FormattableString results in doubled '{' in JavaScript output
### Description

When writing bindings referring to `System.FormattableString` the resulting JavaScript string includes `{{` and `}}` where only single characters are expected. This results in broken styles in e.g. Fable.Lit.

### Repro code

[REPL](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYFgBQpJQGICGARgDZgB0AwsAE5hZYDaAPAKI0EJgB8AulkjjgoAIQCWEACYSA5gAoAlFAC8WKOqgtWqMUjkAiAMYIEcgCQAGADRQzARgX6FfNRuQEkYw1FRoiYGigJEgkwKBAaMQA3DzDjBABJCAgAuWQaBAAuKHTZRl4bFgAFAg5UAEEaDjg+KFKZLKhgIgArfKVs5paVKBaEADkPaPpMV3V3T29fVH9A4NCoeLTsvFpUDyRiMhgkSIgZJVVMDRPRCWl98niklMC08gBxMCQdvYbFGwRH58qZAFdfBAkKYFAoGJhUMBJH8yFAABo9VxkJBQMAADzsUEUKjGUGRi2AJFoPX0pD+YH0uPEUlkVxMUFxJzM+hZmQAFsBkFAAN7cxmndSGQm0bLcoVEmgAXwA3Pz1JLJSzKaNjnjnqi0QAmLGHKnnWnxBmqgVmZmsjlcvnGgUacUiqBksCy60nRUs8HqfHogDMOpx1upFxkdIQRpttn07M5KN5ctOdpo2UdzvDCuVJyAA&html=DwCwLgtgNgfAUHUBTAhgE3gAm54ElgqbhgAOAtEgI4CuAlgG4C8A5AMID2AdmEj+QBUAnqSQtMAY268erXgA8wAenDQA3JJAoATgGcCTGmABm5ABwssOYLona6pMJl3aJTAEQB9TwAkA8gDKAt5KUHQARrpK2qgSYAB0pNocaDRxdNzxEHRc8QBWuu4wwEq29o5W2DZ2Dk4ubl6+gcGeoRFRMShx5GgcEInJqemZ2bkFRSVltZWYiFOOzq4eJKS6AFxKSjRcpADWAObxUhBKAAIA7kjhx6TcfGBRl9d9t1z3ugWnAEzxAAzxAGYlBIaLowH1KFAkPgeLpKLoAKzkdAoRxIbT5QrFUo1CpzXF1JbuFbrTbbPaHY5nJ43O6w4Gg8EQSHQ96nACM8R+vwZYIhSChMIeWRymImOPKYHg1UliwaJI2Wx2ByOfWpV1pb3pui0aCEvQgHPinKBOvQ+r6IrGWMmBPgCBKIFQGAd4RSQkkUBQul0HlRFCkPBQOXRRTgOFwECE5HCRnBXEwXvCAo8HA44qjMbj3GlSjdevtjsgsDgQA&css=BYFwtgNgNAUARgewCYE8AEBvGadoGYIB2IAtAM4CWAXgKYBcaAjAGwAOAHgNwwC+MMAOjABDCoRIBjIiFGEaAJ0zZcSCmVYRhKBngg0uynAHcKSEMAaMADFYCk3XGmA0KAc1CWb9w2gBWAVzIQCjwUSWkaYgYJSJAFB1xhCDdxCjiwMmjY+J9dfRJVeRoJYKJohAh-MEJuPkExVn8QJUdWYSRVQlcGAQAmAFYisAScAmJyanomNgNHEzMLNAGrDhG0EXlXMRJEEBAEMEsh2pggA)

```fs
open System
open Fable.Core

[<Erase>]
type Binding() =
    [<Emit("css($0, $1)")>]
    static member inline private cssInner(strs: string[], [<ParamArray>] args: obj[]) : obj = jsNative

    static member inline css(s: FormattableString) =
        Binding.cssInner (s.GetStrings(), s.GetArguments())

module X = 
  let ex1 () =
    let color = "blue"
    Binding.css 
        $""":host {{
            color: {color};
        }}"""

  let ex2 () =
    Binding.css 
        $$""":host {
            color: blue;
        }"""

  let ex3 () =
    Binding.css 
        $":host {{
            color: blue;
        }}"
```

#### Actual

```js
export function X_ex1() {
    const color = "blue";
    const s = fmt`:host {{
            color: ${color};
        }}`;
    return css(s.strs, ...s.args);
}
```

#### Expected

```js
export function X_ex1() {
    const color = "blue";
    const s = fmt`:host {
            color: ${color};
        }`;
    return css(s.strs, ...s.args);
}
```

**or**

```js
export function X_ex1() {
    const color = "blue";
    return css(`:host {
            color: ${color};
        }`);
}
```

### Related information

* Fable version: `dotnet fable --version 4.20.0`

roboz0r:
Fixes #3889 by replacing `{{` and `}}` with `{` and `}`. Adds tests to confirm correct unescaping.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [Python] Remove `$` sign when reporting an error from `assert_equal` and `assert_not_equal` (#3878) (by @joprice)
 
+### Fixed
+
+* [JS/TS] Fix escaping of `{` and `}` in FormattableString (#3890) (by @roboz0r)
+
 ## 4.20.0 - 2024-09-04
 
 ### Added
--- a/src/Fable.Transforms/BabelPrinter.fs
+++ b/src/Fable.Transforms/BabelPrinter.fs
@@ -573,7 +573,12 @@ module PrinterExtensions =
             | Literal.DirectiveLiteral(literal) -> printer.Print(literal)
             | StringTemplate(tag, parts, values, loc) ->
                 let escape str =
-                    Regex.Replace(str, @"(?<!\\)\\", @"\\").Replace("`", @"\`")
+                    Regex
+                        .Replace(str, @"(?<!\\)\\", @"\\")
+                        .Replace("`", @"\`")
+                        // FormattableString replaces { with {{
+                        .Replace("{{", "{")
+                        .Replace("}}", "}")
 
                 printer.AddLocation(loc)
                 printer.PrintOptional(tag)
--- a/src/fable-library-ts/CHANGELOG.md
+++ b/src/fable-library-ts/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* [JS/TS] Fix escaping of `{` and `}` in FormattableString (#3890) (by @roboz0r)
+
 ## 1.4.3 - 2024-09-04
 
 * [JS/TS] Fixed Decimal comparisons (#3884) (by @ncave)
--- a/src/fable-library-ts/String.ts
+++ b/src/fable-library-ts/String.ts
@@ -87,7 +87,7 @@ export function indexOfAny(str: string, anyOf: string[], ...args: number[]) {
   }
   const endIndex = startIndex + length
   const anyOfAsStr = "".concat.apply("", anyOf);
-  for (let i=startIndex; i<endIndex; i++) {
+  for (let i = startIndex; i < endIndex; i++) {
     if (anyOfAsStr.indexOf(str[i]) > -1) {
       return i;
     }
@@ -588,6 +588,8 @@ export function fmtWith(fmts: string[]) {
 
 export function getFormat(s: FormattableString) {
   return s.fmts
-    ? s.strs.reduce((acc, newPart, index) => acc + `{${String(index - 1) + s.fmts![index - 1]}}` + newPart)
-    : s.strs.reduce((acc, newPart, index) => acc + `{${index - 1}}` + newPart);
+    ? s.strs.map((s, _) => s.replace('{', '{{').replace('}', '}}'))
+      .reduce((acc, newPart, index) => acc + `{${String(index - 1) + s.fmts![index - 1]}}` + newPart)
+    : s.strs.map((s, _) => s.replace('{', '{{').replace('}', '}}'))
+      .reduce((acc, newPart, index) => acc + `{${index - 1}}` + newPart);
 }
--- a/src/fable-library-ts/String.ts
+++ b/src/fable-library-ts/String.ts
@@ -587,9 +587,11 @@ export function fmtWith(fmts: string[]) {
 }
 
 export function getFormat(s: FormattableString) {
+  const strs = s.strs.map((value) => value.replace(/{/g, '{{').replace(/}/g, '}}'));
+
   return s.fmts
-    ? s.strs.map((s, _) => s.replace('{', '{{').replace('}', '}}'))
+    ? strs
       .reduce((acc, newPart, index) => acc + `{${String(index - 1) + s.fmts![index - 1]}}` + newPart)
-    : s.strs.map((s, _) => s.replace('{', '{{').replace('}', '}}'))
+    : strs
       .reduce((acc, newPart, index) => acc + `{${index - 1}}` + newPart);
 }
--- a/tests/Js/Main/StringTests.fs
+++ b/tests/Js/Main/StringTests.fs
@@ -1183,6 +1183,12 @@ let tests = testList "Strings" [
         s3.GetStrings() |> toArray |> equal [|"I have no holes"|]
 
     testCase "FormattableString fragments handle { and }" <| fun () ->
+// TypeScript will complain because TemplateStringsArray is not equivalent to string[]
+#if FABLE_COMPILER_TYPESCRIPT
+        let toArray = Seq.toArray
+#else
+        let toArray = id
+#endif
         let classAttr = "item-panel"
         let cssNew :FormattableString = $$""".{{classAttr}}:hover {background-color: #eee;}"""
         let strs = cssNew.GetStrings() |> toArray
--- a/tests/Js/Main/StringTests.fs
+++ b/tests/Js/Main/StringTests.fs
@@ -1179,5 +1179,37 @@ let tests = testList "Strings" [
         s2.GetStrings() |> toArray |> equal [|""; "This is \""; "\" awesome!"|]
         let s3: FormattableString = $"""I have no holes"""
         s3.GetStrings() |> toArray |> equal [|"I have no holes"|]
+
+    testCase "FormattableString fragments handle { and }" <| fun () ->
+        let classAttr = "item-panel"
+        let cssNew :FormattableString = $$""".{{classAttr}}:hover {background-color: #eee;}"""
+        let strs = cssNew.GetStrings()
+        strs |> equal [|"."; ":hover {background-color: #eee;}"|]
+        let args = cssNew.GetArguments()
+        args |> equal [|classAttr|]
+
+        let cssNew :FormattableString = $""".{classAttr}:hover {{background-color: #eee;}}"""
+        let strs = cssNew.GetStrings()
+        strs |> equal [|"."; ":hover {background-color: #eee;}"|]
+        let args = cssNew.GetArguments()
+        args |> equal [|classAttr|]
+
+        let cssNew :FormattableString = $".{classAttr}:hover {{background-color: #eee;}}"
+        let strs = cssNew.GetStrings()
+        strs |> equal [|"."; ":hover {background-color: #eee;}"|]
+        let args = cssNew.GetArguments()
+        args |> equal [|classAttr|]
+
+        let another :FormattableString = $$"""{ { } {{classAttr}} } } }"""
+        let strs = another.GetStrings()
+        strs |> equal [|"{ { } "; " } } }"|]
+        let args = another.GetArguments()
+        args |> equal [|classAttr|]
+
+        let another :FormattableString = $"""{{ {{{{ }}}}}} {classAttr} }}}} }} }}}}"""
+        let strs = another.GetStrings()
+        strs |> equal [|"{ {{ }}} "; " }} } }}"|]
+        let args = another.GetArguments()
+        args |> equal [|classAttr|]
 #endif
 ]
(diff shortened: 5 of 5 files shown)


roboz0r:
Not sure how I missed this last night but this change achieves the expected output for JS template literals but it breaks an existing test:

```fs
    let s5: FormattableString = $"I have {{escaped braces}} and %%percentage%%"
    s5.Format |> equal "I have {{escaped braces}} and %percentage%"
```

There is a difference between `_.Format` and `_.ToString()` 

```fs
open System
let s:FormattableString = $"""I have {{escaped braces}} and %%percentage%%"""
s.Format
s.ToString()

--- FSI Output ---

- s.Format;;
val s: System.FormattableString = I have {escaped braces} and %percentage%
val it: string = "I have {{escaped braces}} and %percentage%"

> s.ToString();;
val it: string = "I have {escaped braces} and %percentage%"
```

I am unsure whether it is simply I need to change the implementation of `_.Format` or if I have misunderstood the problem more fundamentally?

MangelMaxime:
I am not sure what direction to take either. I never played deeply with string interpolation escaping or the custom `GetStrings` from Fable.

While looking for a fix my approach was different.

Instead of having a direct access to the underlying `strs`. What I did, is create a `getStrings` function in JavaScript which handles the call to `cssNew.GetStrings()`.

My reasoning is it looks like the escaping problem only happens when using `GetStrings` for fixing it only for this call make sense to me.

And it also, fix it at a single place where your solution needs to modify the string/parts at 2 different places.

CleanShot 2024-09-18 at 18 21 18

What do you think of this approach ?

roboz0r:
How does that approach affect the generated template literals? Specifically, the template literal shouldn't include `{{` 

I think that StringTest.js should look like:

```js
equal("I have {{escaped braces}} and %percentage%", getFormat(fmt_2`I have {escaped braces} and %percentage%`));
```

MangelMaxime:
> How does that approach affect the generated template literals? Specifically, the template literal shouldn't include `{{`

It doesn't affect it and the tests are all greens.

CleanShot 2024-09-18 at 18 34 49

I am not sure what we want as an output actually 😅

I would have liked if it was as simple as we want the user enter is what we want in the output. But it seems like it will be more difficult than that

```fs
let formatString0 : FormattableString = $"""{{}}"""
let formatString1 : FormattableString = $"""{{}}"""
let formatString2 : FormattableString = $$"""{}"""
let formatString3 : FormattableString = $$$"""{}"""
let formatString4 : FormattableString = $$$"""{{}}"""
```

**Current implementation (compatible with my proposition)**

```js
export const formatString0 = fmt`{{}}`;
export const formatString1 = fmt`{{}}`;
export const formatString2 = fmt`{{}}`;
export const formatString3 = fmt`{{}}`;
export const formatString4 = fmt`{{{{}}}}`;
```

**Your implementation**

```js
export const formatString0 = fmt`{}`;
export const formatString1 = fmt`{}`;
export const formatString2 = fmt`{}`;
export const formatString3 = fmt`{}`;
export const formatString4 = fmt`{{}}`;
```

MangelMaxime:
Looking at Fable code, contrary to my primary fear it seems like `String.get_Format` is only used when working with`FormattableString`.

This is the only place I see it being used:

https://github.com/fable-compiler/Fable/blob/9cf4849fa3da4e7f53138aa468622f1b6987a2ca/src/Fable.Transforms/Replacements.fs#L1658-L1705

Something to not too, is that how we decide to represents output in this PR will probably not only have impact on JavaScript but also the others targets. Because it seems like they are using the same output format.

@roboz0r Is there a case where not using the current output format cause issue?

@ncave @dbrattli Any opinion on how we should handle `FormattableString` representation ?

roboz0r:
I think that my implementation is the right way to go then. I don't think template literals require escaping of `{` so having 

```js
export const formatString0 = fmt`{{}}`;
```

Doesn't give the right semantics as

```js
// JS
`{{}}` === "{{}}" // true
`{{}}` === "{}" // false
```

```fs
//F#
$"""{{}}""" = "{}" // true
```

This would be possibly breaking existing code but if existing template literals were always wrapped in `fmt()` then it wouldn't affect anything. 

If we make the change then `fmt` wouldn't be required for js functions accepting template literals directly.

ncave:
@MangelMaxime IMO we should try to keep string interpolation as similar to F# on .NET as possible:
https://github.com/dotnet/docs/blob/main/docs/fsharp/language-reference/interpolated-strings.md#extended-syntax-for-string-interpolation

roboz0r:
One of the things I don't like about doing unescaping in `getStrings` is it's causing runtime string allocations each time it is called. Ideally, we should just be passing through the runtime optimized `TemplateStringsArray` unchanged when passing a format string to a JS API. 

Trying to stick closer to the .NET API makes me think we should reexamine the `FormattableString` type definition in String.ts

https://github.com/fable-compiler/Fable/blob/9cf4849fa3da4e7f53138aa468622f1b6987a2ca/src/fable-library-ts/String.ts#L575-L579

I sketched up a type that more closely matches the .NET type definition of [`FormattableString`](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/FormattableString.cs). I'm really not a JS dev so please change any non-idiomatic usage.

```cs
public abstract class FormattableString : IFormattable
{
    [StringSyntax(StringSyntaxAttribute.CompositeFormat)]
    public abstract string Format { get; }
    public abstract object?[] GetArguments();
    public abstract int ArgumentCount { get; }
    public abstract object? GetArgument(int index);
    public abstract string ToString(IFormatProvider? formatProvider);

    string IFormattable.ToString(string? ignored, IFormatProvider? formatProvider) =>
        ToString(formatProvider);

    public static string Invariant(FormattableString formattable)
    {
        ArgumentNullException.ThrowIfNull(formattable);
        return formattable.ToString(Globalization.CultureInfo.InvariantCulture);
    }

    public static string CurrentCulture(FormattableString formattable)
    {
        ArgumentNullException.ThrowIfNull(formattable);
        return formattable.ToString(Globalization.CultureInfo.CurrentCulture);
    }

    public override string ToString() =>
        ToString(Globalization.CultureInfo.CurrentCulture);
}
```

**Concept**

```ts
abstract class FormattableString {
    // Should this just be an interface?
    // Unsure about default naming conventions for JS/TS.
    abstract get Format(): string;
    abstract get Arguments(): any[] | undefined;
    abstract get ArgumentCount(): number;
    abstract getArgument(index: number): any;
    // toString() should return the formatted string. Currently in fable it returns [object Object]
    abstract toString(): string; // TODO: How to add IFormatProvider?
}

// Long Name. Consider "TemplatedString", "FableString", etc.
class TemplatedFormattableString extends FormattableString {
    readonly strs: TemplateStringsArray
    readonly fmts?: ReadonlyArray<string> // The constant portion of the string is truly immutable.
    readonly args?: any[] // The arguments are mutable, to match the .NET API.

    constructor(strs: TemplateStringsArray, fmts: string[], args: any[]) {
        super();
        this.strs = strs;
        this.fmts = fmts;
        this.args = args;
    }

    static createWithFmts(fmts: string[]) {
        return (strs: TemplateStringsArray, ...args: any[]) => {
            if (args.length !== strs.length - 1 || fmts.length !== args.length) {
                throw new Error("The number of format strings must match the number of expressions.");
            }
            let x = new TemplatedFormattableString(strs, fmts, args);
        }
    }

    static create(strs: TemplateStringsArray, ...args: any[]) {
        if (args.length !== strs.length - 1) {
            throw new Error("The number of format strings must match the number of expressions.");
        }
        return new TemplatedFormattableString(strs, [], args);
    }

    get Format(): string {
        // Inefficient, but it works to return the .NET style format string.
        // Consider caching the result in a private field.
        return this.fmts
            ? this.strs.map((s, _) => s.replace('{', '{{').replace('}', '}}'))
                .reduce((acc, newPart, index) => acc + `{${String(index - 1) + this.fmts![index - 1]}}` + newPart)
            : this.strs.map((s, _) => s.replace('{', '{{').replace('}', '}}'))
                .reduce((acc, newPart, index) => acc + `{${index - 1}}` + newPart);
    }

    get Arguments(): any[] | undefined {
        return this.args;
    }

    get ArgumentCount(): number {
        return this.args?.length ?? 0;
    }

    getArgument(index: number): any {
        return this.args![index];
    }

    private static applyFormat(format: string, arg: any): string {
        // TODO: Use the format string to format the argument.
        return arg.toString();
    }

    toString(): string {
        if (this.args === undefined) {
            return this.strs[0];
        } else {
            return this.strs.reduce((acc, newPart, i) => acc + (TemplatedFormattableString.applyFormat(this.fmts![i - 1], this.args![i - 1])) + newPart);
        }
    }
}

function fmt(strs: TemplateStringsArray, ...args: any[]) { return TemplatedFormattableString.create(strs, args); }
function fmtWith(fmts: string[]) { return (strs: TemplateStringsArray, ...args: any[]) => TemplatedFormattableString.createWithFmts(fmts)(strs, args); }
```

**Usage**

```fs
let speedOfLight = 299792.458  
let frmtStr = $"The speed of light is {speedOfLight:N3} km/s." : FormattableString
```

Becomes

```ts
const speedOfLight = 299792.458;
const frmtStr = fmtWith([":N3"])`The speed of light is ${X_speedOfLight} km/s.`;
```

---

```fs
let color = "blue"
Lit.css 
        $""":host {{
            color: {color};
        }}"""
```

Becomes

```ts
function css(strs: TemplateStringsArray, ...args: any[]) {
    // As imported by e.g. lit
}

const color = "blue";
const s = fmt`:host {
            color: ${color};
        }`;
// Default path for template literals has no extra string allocations provided we unescape "{{" to "{" at compile time.
return css(s.strs, s.args);
```

MangelMaxime:
We avoid creating classes in `fable-library` so bundler have a chance to remove dead code. I don't know if now days they can do it for classes or not.

> Trying to stick closer to the .NET API makes me think we should reexamine the FormattableString type definition in String.ts

What we mean by that in general, is that the behaviour at runtime should be same. Not necessary that we use a class if they use a class, etc.

MangelMaxime:
Note, I just said that we avoid using class in fable-library but working on another fix I just found that we do use class sometimes 😅.

So let's say it depends and can be used.

@roboz0r I merged the PR as is because it seemed to fix your original issue and from what I understood using a class would just re-structured but with the same behaviour.

If I was wrong feel free to tell me and we can merge another PR.

roboz0r:
@MangelMaxime  Thanks for your help getting this one completed. The class version didn't change any behaviour but was just trying to better capture the semantics of the .NET API. I'll trust your judgement whether to go with a class def or leave the interface as is. 

While I was working on it, I did identify that `FormattableString.ToString()` needs an implementation to produce the formatted string instead of returning `[object Object]`. That can be a separate issue/task.

MangelMaxime:
Yes, there a few places where if you use `printfn "%A" myValue` then you get `[object Object]`.

But if you use `JS.console.log(myValue)` you often get something more useable. I never really dig this issue and always assume it was a limitation to how some types are represented or how Fable handle printing values. But if we can improve it that's good :) 

Thanks for the work done in this issue, I want to fix a few other ones before making a release but should not take too long.

PR #3892, opened 2024-09-18
[JS/TS] OrdinalIgnoreCase overload for .EndsWith

goswinr:
Since[ `string.StartsWith`  supports a `StringComparison.OrdinalIgnoreCase` overload](https://github.com/fable-compiler/Fable/blob/9cf4849fa3da4e7f53138aa468622f1b6987a2ca/src/fable-library-ts/String.ts#L66), I think `.EndsWith` should too

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,5 +2,3 @@
 
 dotnet tool restore
 dotnet run --project src/Fable.Build/Fable.Build.fsproj -- $@
-# dotnet run --project src/Fable.Build/Fable.Build.fsproj -- test javascript
-# dotnet run --project src/Fable.Build/Fable.Build.fsproj -- test typescript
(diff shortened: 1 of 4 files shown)


goswinr:
I am not sure where to fix these failing tests, in the .fs file or in .ts ? 
image

ncave:
@goswinr Perhaps make the third argument optional, with a sensible default.

goswinr:
@ncave Do you know why the Rust build fails? I tried to do it properly. So went for the transformers.

MangelMaxime:
Seems like .NET 8.0.401 introduced a bug in NuGet package resolution: https://github.com/NuGet/Home/issues/13729

It seems like downgrading to 8.0.304 will fix this issue, so perhaps we are going to consider doing it at the CI level.

Issue #3864, opened 2024-07-09
[TypeScript] Async type signature always returns any


leononame:
### Description

Async does not generate correct type signature.

### Repro code

[REPL](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYDoAqALATmAQwBMBLCAcywIQGsEAoegGzCSiOAEEE4IBjKAF4o9KGKjVeAgN6jx8-EgCuuaACJsJNXLEBfIA&html=DwQgJg9gxgLgngBwKYAIAWMC2AbAfAKGAxwMLSQEMwCUVgYBLGbJXAUQA8olsUBmAGwBWFAwB2KChLQQArgCdgAekbNW+WsExIYFdDBgIAtEgCOshgDcAvAHIAwhDEwkzowBVESWyihOXznYuHDBKxNgA3L5oFPIAzjrWsjAAZkYAHLY0dNq6KGIU2tYARJYMSADuCBDyMMW+-q4wJRUMYDBo1mBIZdxGre1oADSiYkwMFNhGcVCTSNYAjMXZwDPyDAgwKHHyUCUYhnEAXEpKUGBiAHQAVnHd2Fbyl2I6SmIImEpltbKTRpCYAACACZLgswUowAw4qFvjBflMATc4vUoPIIHE4jUGABzcS4ZRrDYwFYwuAsbK0ABGEDAcBQAG8NLQWQgqFCxDijighAAGBAcCLMlkoFL+IwpQoMbBwblxKRxaZIdYpIUigC++GFKF0VJYjO11Jq3Xk0zZUHEXJQ-MFhpQNPkJqMNIMEEw3IWAu2EAeYBQAGIkLyg0G1SL7cblUZ1jiMB6vVjfQGQ8HeWHaJrtTAwEMs2gDeGUAMOty+QL0yyHU6WCkYPGON6k4HU6G7VWozAIAh6422smW2m7WywByrUJy3aIJZlSlsBAKty0G1umIK7RgjAjFOZ3OFygeA8EHFoRXMyLs5cEixYEg-UzC8PR9ybaetee-eIEMkC+Hi2hubM2BQAAFAsvK8gApCgRg8gKACUa7oEguJxigwLAhOhaYLEeJiM+mEspqmhKGSFKEGElDUFqwA0nSKxQpYohgCU2HiMsygMQQyi0XApDKOEBBAA&css=Q)

### Expected and actual results

[...]

PR #3903, opened 2024-09-27
[TS] Make `AsyncBuilder` returns a real type instead of `Any`

MangelMaxime:
I was looking at fixing #3864, and it seems like my fix works for the code provided in the issue:

```fs
let doAsync =
    async {
        return "hi"
    }
```

However, it fails on more complex code and I don't understand why.

```text
Argument of type '(ctx: IAsyncContext<number | undefined>) => void' is not assignable to parameter of type 'IAsync<number>'.
  Types of parameters 'ctx' and 'x' are incompatible.
    Type 'IAsyncContext<number>' is not assignable to type 'IAsyncContext<number | undefined>'.
      Type 'number | undefined' is not assignable to type 'number'.
        Type 'undefined' is not assignable to type 'number'.ts(2345)
```

I don't understand where TypeScript infer the type to `XX | undefined` we never seems to use it inside of `AsyncBuilder`.

@ncave Do you have an idea?

### DIFF
--- a/src/fable-library-ts/Async.ts
+++ b/src/fable-library-ts/Async.ts
@@ -9,9 +9,6 @@ import { protectedReturn } from "./AsyncBuilder.js";
 import { FSharpChoice$2_$union, Choice_makeChoice1Of2, Choice_makeChoice2Of2 } from "./Choice.js";
 import { TimeoutException } from "./SystemException.js";
 
-// Implemented just for type references
-export class Async<_T> { }
-
 function emptyContinuation<T>(_x: T) {
   // NOP
 }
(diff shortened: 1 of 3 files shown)


src/fable-library-ts/Async.ts
```diff
@@ -9,9 +9,6 @@ import { protectedReturn } from "./AsyncBuilder.js";
 import { FSharpChoice$2_$union, Choice_makeChoice1Of2, Choice_makeChoice2Of2 } from "./Choice.js";
 import { TimeoutException } from "./SystemException.js";
 
-// Implemented just for type references
-export class Async<_T> { }
```
MangelMaxime, line 13: Remove because we never use that type, we seems to always refers to `IAsync`

ncave:
Superceded by #3904

MangelMaxime:
Oh no problem, the solution was so simple... 

I was so focused on looking for `| undefined` that I didn't see the optional argument in the method 🤦‍♂️

And that's probably what caused me to try to restrict where we typed Async in this PR too.

Thank you for taking a look.

PR #3946, opened 2024-10-27
[JS] Add `ReadonlyArray` to `Fable.Core`

MangelMaxime:
This is not a real JS type, but TypeScript makes extensive use of this type.

In the future, Glutinum will probably take over the responsibility of exposing such types but I am not sure yet how Glutinum packages should be shaped.

For now, I think exposing it via Fable.Core is fine. Looking at the git history it was originally part of Fable.Core but removed at some point not sure why.

Later on, we will be able to either remove the type again or keep it as no one will really use it outside of bindings.

@ncave @dbrattli 

Do you think that's ok?

### DIFF
--- a/src/Fable.Core/CHANGELOG.md
+++ b/src/Fable.Core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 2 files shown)


ncave:
@MangelMaxime I don't see a problem, perhaps add a note that it's only for TypeScript target.

MangelMaxime:
After a night of thinking, I think I will start a Glutinum.Types NuGet package because I am not sure how much more custom type I will need in the future.

And doing so, gives me the ability to experiments with module structure more freely by make breaking changes.

PR #3963, opened 2024-11-25
Allow Fable 5 to be used with Fable 4 plugins

### LINKED ISSUE #3962: Fable 5, not compatible with Fable 4 plugins
### Description

When releasing Fable 5.0.0-alpha.1, the tests didn't capture the fact that upgrading Fable version would make Fable 4 plugins not supported.

This happened because when running the tests, we don't upgrade Fable internal version so it was still checking against version 4.24.0

### Solutions

1. We ask Fable plugins to upgrade their `FableMinimumVersion` to `5.0`.

	**Pro** we don't need to do anything on Fable side.

	**Cons** this will create a new shift in Fable ecosystem, because then Feliz will need to create a new major version. Which could cause a cascade of upgrade needed for others packages that depends on it.

	The problem is that package authors are not as active as before, so this cascade change can take some time.

2. We adapt Fable logic to allows plugins requesting `FableMinimumVersion = "4.0"` to works with Fable 5.
	
	**Pro** Fable compiler plugins and others NuGet packages depending on them will need to do nothing. This seems also more aligned with the name of the property `FableMinimumVersion`. It says minimum version and not major version.

	**Cons** we will start to make custom logic allowing to support different span of Fable compiler plugins. This would allows us to say as long as Fable.AST stay stable we can support the same plugin for Fable 4, 5, 6, etc.
	
	This question also if we should not have used the version of Fable.AST assembly (I think .NET as version info in the assembly) instead of an arbitrary `FableMinimumVersion` property.

What do you think @ncave @dbrattli ?

ncave:
- Fixes #3962

### DIFF
--- a/src/Fable.Transforms/Global/Compiler.fs
+++ b/src/Fable.Transforms/Global/Compiler.fs
@@ -99,7 +99,7 @@ type InlineExprLazy(f: Compiler -> InlineExpr) =
 
 [<AutoOpen>]
 module CompilerExt =
-    let private expectedVersionMatchesActual (expected: string) (actual: string) =
+    let expectedVersionMatchesActual (expected: string) (actual: string) =
         try
             let r = System.Text.RegularExpressions.Regex(@"^(\d+)\.(\d+)(?:\.(\d+))?")
 
--- a/src/Fable.Transforms/Global/Compiler.fs
+++ b/src/Fable.Transforms/Global/Compiler.fs
@@ -119,7 +119,7 @@ module CompilerExt =
             actualMajor > expectedMajor
             || (actualMajor = expectedMajor
                 && (actualMinor > expectedMinor
-                    || actualMinor = expectedMinor && actualPatch >= expectedPatch))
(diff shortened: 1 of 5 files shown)


MangelMaxime:
Thank you @ncave.

I added some simple tests to covert our needs.

I think in the future, we will need a more specific logic to manual control the range of which version of Fable support which version of Fable.AST but for our current needs it should work fine.

Waiting for the CI to confirm that that all is still good before making a new release

MangelMaxime:
@ncave I added what I think was a missing parenthesis but even with that a test is still failing on the CI.

It works on my machine... do you have an idea?

ncave:
@MangelMaxime 
> I added what I think was a missing parenthesis

No, that's not going to change anything, the operator precedence is the same.
It must be something else.

MangelMaxime:
I think I was just tired, I was not running the correct tests locally.

And it appears that in my tests, I had the argument order inverted.

Waiting for the CI to confirm that all is good now 🤞

Issue #3961, opened 2024-11-24
Remove and deprecate `Fable.PublishUtils`


MangelMaxime:
### Description

Fable.PublishUtils consist on list of utility tools which are used to release some of Fable packages.

It was created a long time ago because FAKE and other solution were considered too complex to put in place.

Today, with projects like [Fun.Build](https://github.com/slaveOftime/Fun.Build), [EasyBuild](https://github.com/easybuild-org) I think Fable.PublishUtils don't make much sense.

Also no one is really maintaining it, for example myself I invested my time in EasyBuild ecosystem.

For this reason, I think we should just delete the folder and mark the project as deprecated on both NuGet and NPM.

PS: I believe Fable.PublishUtils is mostly used by project under Fable org, so if previous version breaks or need to be update to new tool chain I will make the upgrade.

MangelMaxime:
@ncave `Fable.PublishUtils` is not used by Fable release pipeline.

It is used by other repository inside of Fable org like [fable-browser](https://github.com/fable-compiler/fable-browser) or [Fable.Node](https://github.com/fable-compiler/fable-node). And I am planning to move these project to using EasyBuild because EasyBuild allows to generate changelog based on the git history and it make my life as a maintainer easier. I don't need to manually patch Changelog file etc.

So removing the folder would cause no harm. I wanted to remove the folder, to have 1 less folder to care about especially, if we don't plan to make a release for it anymore (it is saved in the `fable4` branch if we need it for some reason).

ncave:
@MangelMaxime 
> Fable.PublishUtils is not used by Fable release pipeline.

I know, that's why keeping it would cause no harm either. But if you feel really strongly about it, sure.
My personal preference is to keep things minimal and not introduce too many external dependencies.

In that line of thought, the current build system may be a bit too complicated to navigate and contribute to.
I'm just saying that in hope to attract future contributors we may want to keep things simple.

To clarify, I really appreciate your efforts and taking on the burden of publishing and I want to help as much as I can. So anything that can simplify that task, I'm all for it, hopefully without introducing too many dependencies we need to keep up with.

MangelMaxime:
> To clarify, I really appreciate your efforts and taking on the burden of publishing, and I want to help as much as I can. So anything that can simplify that task, I'm all for it, hopefully without introducing too many dependencies that we'll have to keep up with.

Thank you :) I think we both found our balance on working on Fable and I also appreciate your help and pointer regarding Fable internals.

[...]

PR #3964, opened 2024-11-27
chore: simplify code of Publish target

MangelMaxime:
Hello @ncave,

This my proposition to simplify the code of the `Publish` target that I mentioned in #3961.

Benefits are:

- Simplified logic overall (a little), thanks to automatic version setting in the `fsproj` (no more manual patching)
- Remove 5 files from the build process with one being especially complex (ChangelogParser)


ncave (approved):
Looks good, I was just wondering if this <PackageReference Include="EasyBuild.PackageReleaseNotes.Tasks" Version="2.0.0"> can be run as a tool during the build, instead of adding it as a package dependency for all versioned binaries. Just wondering, if not possible it's fine.

MangelMaxime:
> Looks good, I was just wondering if this can be run as a tool during the build, instead of adding it as a package dependency for all versioned binaries. Just wondering, if not possible it's fine.

Unfortunately, this is not possible.

I tried it in the past, and the issue when running as a tool, is that we need to use CLI arguments to provide MSBuild properties like `Version`, `PackageReleaseNotes`, etc.

When doing that, if one of your project use a `ProjectReference` then the CLI arguments will impact both the direct `fsproj` and the ones referenced via `ProjectReference` which can mess up the dependency version if the libraries don't have their version in sync.

PR #3982, opened 2024-12-18
[TS] Make discriminated union `.Is*` properties works

MangelMaxime:
I am not sure why but casting the `tag` to `number` make some others scenarios fails:

```fs
type Test<'T> =
    | A
    | Value of 'T

let testResult = Value "test"

printfn "%A" testResult.IsA
printfn "%A" testResult.IsValue

let apply (f: 'A -> 'B) (value : Test<'A>) : Test<'B> =
    match value with
    | A -> A
    | Value v -> Value (f v)
```

```ts
(function () {
    const arg: boolean = testResult.tag === (/* A */ 0 as int32);
    toConsole(printf("%A"))(arg);
})();

(function () {
    const arg: boolean = testResult.tag === (/* Value */ 1 as int32);
    toConsole(printf("%A"))(arg);
})();

export function apply<A, B>(f: ((arg0: A) => B), value: Test$1_$union<A>): Test$1_$union<B> {
    if (value.tag === (/* Value */ 1 as int32)) {
        const v: A = value.fields[0]; // <===================== Error here
        return Test$1_Value<B>(f(v));
    }
    else {
        return Test$1_A<B>();
    }
}
```

```
Type 'A | undefined' is not assignable to type 'A'.
```

But without the cast then in some scenario `Is*` don't works with TypeScript is able to compare the literal value directly.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [Python] Fix type testing against `uint8`, `uint32`, `uint64`, `decimal` (@MangelMaxime)
 * [JS/TS] Workaround source map generation bug (deteriorate them a little) (@MangelMaxime)
+* [TS] Make discriminated union `.Is*` properties works (@MangelMaxime)
 
 ## 5.0.0-alpha.2 - 2024-11-25
 
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -2438,7 +2438,17 @@ module Util =
                 expr
         | Fable.UnionCaseTest tag ->
             let expected = transformUnionCaseTag com range expr.Type tag
-            let actual = getUnionExprTag com ctx None expr
+
+            let unionExprTag = getUnionExprTag com ctx None expr
+
+            let actual =
+                if com.IsTypeScript then
+                    // In TypeScript, we need to cast the union case tag to a number
+                    // otherwise TypeScript compiler is too smart and test against the literal value (0, 1, 2, etc.)
+                    let number = Fable.Number(Int32, Fable.NumberInfo.Empty)
+                    (AsExpression(unionExprTag, makeTypeAnnotation com ctx number))
+                else
+                    unionExprTag
 
             Expression.binaryExpression (BinaryEqual, actual, expected, ?loc = range)
 


MangelMaxime:
By using `as any` I am able to make the compiler happy (no I am not hiding stuff under the rug).

Looking at the generated code it seems like even if we use `as any` it happens in places that are internal to Fable code and not expose to the outside world.

```fs
type Test<'T> =
    | A
    | Value of 'T

let apply (value : Test<'A>)  =
    match value with
    | A -> failwith "Cannot apply function to A"
    | Value v -> v
```

generates

```ts
export function apply<A>(value: Test$1_$union<A>): A {
    if ((value.tag as int32) === /* Value */ 1) {
        const v = value.fields[0] as any;
        return v;
    }
    else {
        throw new Error("Cannot apply function to A");
    }
}
```

You can see that the `v` end up typed in via the output of the function.

I tried to cast the value to the concrete type of the `fields[x]` but it caused issues at other places mostly because we are generating curried function instead of standard function.

joprice:
Does the `as int32` widen the discriminator type and lose the more specific type information that would allow the flow inference to figure out what type the field should be?

MangelMaxime:
> Does the `as int32` widen the discriminator type and lose the more specific type information that would allow the flow inference to figure out what type the field should be?

`int32` is an alias to `number` and it does indeed widen the type.

This is need because TypeScript is "too smart" and when we do `value.tag === (/* Value */ 1` it does not consider `1` as a number but a literal type which has only `1` as the possible value.

This is problematic because if you do something like:

```fs
type Test =
    | A
    | Value of string

let testA = A

testA.IsA
testA.IsValue
```

then we generate

```ts
(testA.tag) === /* A */ 0;

(testA.tag) === /* Value */ 1;
```

resulting in this error

CleanShot 2024-12-19 at 08 47 12@2x.

So my idea was to say to TS that `tag` is actually a `number` to allow `IsValue` to work in this situation. But changing that caused issue to the `apply` generation (see original comment).

MangelMaxime:
Strangely if the union has a generic the problem does not occur because TS infer `tag` to be `0 | 1` and not just `0` or `1` depending on the situation.

The code generation is little different with the generic:

```fs
type Test =
    | A
    | Value of string

type TestWithGeneric<'T> =
    | A
    | Value of 'T

testGenericA.IsValue
testA.IsValue
```

```ts
(testGenericA<any>().tag as int32) === /* Value */ 1;
// Hover on tag gives:
// (property) TestWithGeneric$1<T, Tag extends keyof TestWithGeneric$1_$cases<T>>.tag: 0 | 1

(testA.tag as int32) === /* Value */ 1;
// Hover on tag gives:
// (property) Test<0>.tag: 0
```

ncave (approved):
LGTM

Issue #3985, opened 2024-12-21
Fable library: TimeSpan.FromSeconds and .NET9


zelenij:
### Description

.NET9 added a new overload for the `TimeSpan.FromSeconds` which expects `int64`.  This overload is missing from the library implementation. However, fable client code compilation doesn't fail. However, during runtime this message is generated by JS:

```
TypeError: can't convert BigInt to number
```

### Repro code

```
TimeSpan.FromSeconds 1L
```

### Expected and actual results

Expected: execute correctly

Actual: a JS exception is thrown

### Related information

* Fable version: 4.24.0
* Operating system: MacOs, Linux

MangelMaxime:
After lookin into this issue, it seems like they added a bunch of new overload, working on a fix for them.

PR #3960, opened 2024-11-22
Updated Fable-FCS to F# 9

ncave:
- Updated Fable-FCS to F# 9
- Updated metadata to .NET 9.0

PR #3994, opened 2025-01-01
Add new `TimSpan` overload support coming from .NET 9

MangelMaxime:
#3985

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -24,7 +24,7 @@ RUN apt-get update && apt-get install -y xz-utils
 
 # Options
 # Let it here to avoid conflicts with Microsoft options !!!
-ARG DOTNET_VERSION=8.0.404
+ARG DOTNET_VERSION=9.0.101
 ARG NODE_VERSION=18.8.0
 ARG PYTHON_VERSION=3.10.0
 ARG DART_VERSION=3.1.2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: '8.0.x'
+        dotnet-version: '9.0.x'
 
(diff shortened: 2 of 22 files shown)


MangelMaxime:
@ncave I could look into adding support for the new overload to Rust too but I am not sure how you would translate this kind of code in Rust:

```ts
export function fromMinutes(m: number | bigint, s: bigint = 0n, ms: bigint = 0n) {
  return create(0, 0, Number(m), Number(s), Number(ms));
}
```

I think Rust don't have function overload or union type. Would you create a function per overload?

MangelMaxime:
The remaining failing CI probably needs #3960, we should be able to work around this limitation by using compiler directives.

MangelMaxime:
I found out that the Rust target create one function per overload, this can be seen here for example:

https://github.com/fable-compiler/Fable/blob/a61b29d4c4a98a95668ea060141a6bc383b4a03e/src/fable-library-rust/src/TimeSpan.rs#L38-L61

I will do something similar for supporting the TimeSpan overload.

MangelMaxime:
@ncave Is `./build.sh quicktest rust` working for you ?

When trying to run it I have this error:

```text
error[E0599]: no function or associated item named `try_seconds` found for struct `chrono::Duration` in the current scope
   --> /workspaces/Fable/temp/fable-library-rust/src/./DateTime.rs:70:39
    |
70  |         let subsecond = d - Duration::try_seconds(seconds).unwrap();
    |                                       ^^^^^^^^^^^ function or associated item not found in `Duration`
    |
```

Even if the IDE in VSCode is able to find the `try_seconds` function.

ncave:
@MangelMaxime 
> Is ./build.sh quicktest rust working for you ?

Fixed in #4001

MangelMaxime:
> Yes, using FABLE_COMPILER_5 should work. You can merge it as is, I can add the Rust part in a separate PR if you prefer.

Thank you, I had a look at adding Rust support but this is not add straightforward as it seemed to me. I will look at your PR to learn from it.

PR #4007, opened 2025-01-08
[TS] Emit `interface` declaration when encountering a `ParamObject` pattern

### LINKED ISSUE #4003: [TypeScript] POJO with `[<ParamObject>]` generates missing type hint.
### Description

I am working on Fable transpiled React components for usage from typescript. I wanted to create the best possible native feeling for typescript users. My component uses a type which i had defined as record type before:

```fsharp
type Term = {
    Name: string option
    Id: string option
    Description: string option
    Href: string option
    IsObsolete: bool option
    Source: string option
    Data: obj option
} with
    static member init(?name, ?id, ?description, ?source: string, ?href, ?obsolete: bool, ?data) = {
        Name = name
        Id = id
        Source = source
        IsObsolete = obsolete
        Href = href
        Description = description
        Data = data
    }
```

To improve usability from ts I changed it according to the docs to a js pojo using `[<ParamObject>]` [here](https://fable.io/docs/javascript/features.html#paramobject).

```fsharp
[<AllowNullLiteral>]
[<Global>]
type Term
    [<ParamObjectAttribute; Emit("$0")>]
    (?name, ?id, ?description, ?source: string, ?href, ?isObsolete: bool, ?data) =
    member val name: string option = jsNative with get, set
    member val id: string option = jsNative with get, set
    member val description: string option = jsNative with get, set
    member val source: string option = jsNative with get, set
    member val href: string option = jsNative with get, set
    member val isObsolete: bool option = jsNative with get, set
    member val data: obj option = jsNative with get, set
```

Which works as intented but produces a type hint without reference:

image

### Repro code

Here is the code with an example: 

[repl](https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwFgBQoksiKG2aAUgM4CSEALmFqAQQNoA8AgkksAHcAcgFc+AGQCWTLHCQA+ALodOAcX4I5SggwCe4KABVmAWwJQLULgAU4skwHkEAKzABjBtwYMskhCKYAbigAURNpAAoAIgASAAYogEptfEsoCIB+CDgTMAAaKAzJABMCjOKwajdfEAZJYAgy6mARLDcwAC4oah9JCABzMoALHAAzMslqJ2aUJi6EYGAkMuK4BjhEqABec0tckwRmKAA3OShs3K6e3wGoUDqG7ahnaiE1yWOwKAFpIah+sAMArUQG7Cz7Q5YE5nEpXXq3e71aBbZ6vd6fb6-f6A4Gg1J7MAHI6nJBQCpVGoPCBwm79O61JFPF5vOoYn4MP4AoHdPFpCHEs7NVrtGl9OmIx4o5nor7szk4nkMMFQflQklQEZgUaihEMyWolkfWVYrm4pX48GEyHQ0mTaZLQGdKALJb0qlMtGs40c7HckHmvlWgWk1brLrAFxuxlSz1GzE+02Ktj4WZQGQmJ7GLAmaIAWV0RlMUDeuSiBTtCBmjq2PhEYESyYAQiwBCCsKgACLAEyoNwNKuofh09NAA&html=Q&css=Q)

I would expect fable to do one of the following transpilations with decreasing favorability:

1. Transpile type to typescript `interface` (best solution)
2. Transpile with pojo type hint (as such: `{ data?: Option<any>, description?: string, href?: string, id?: string, isObsolete?: boolean, name?: string, source?: string }`)
3. transpile using `any`

Instead it produces a type hint towards the non-generated class, which can lead to huge issues, as having types with similiar names can lead to shadowing at worst.

### Related information

* Fable version: `4.19.0` and repl version `4.23.0`
* Operating system: Windows 11

MangelMaxime:
Fix #4003 

```fs
[<AllowNullLiteral>]
[<Global>]
type User
    [<ParamObjectAttribute; Emit("$0")>]
    (id : int, ?name : string, ?age: int) =
    member val name: string option = jsNative with get, set

let term =
    User(1, "test")
````

transpile to

```ts
export interface User {
    id: int32,
    name?: string,
    age?: int32
}

export const term: User = {
    id: 1,
    name: "test",
};
```

Still need to add integration tests,  supports generics, inheritance.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -817,6 +817,35 @@ module Helpers =
             | None -> false
         )
 
+    // When compiling to TypeScript, we want to captuer classes that use the
+    // [<Global>] attribute on the type and[<ParamObject>] on the constructors
+    // so we can transform it into an interface
+
+    /// <summary>
+    /// Check if the entity is decorated with the <code>Global</code> attribute
+    /// and all its constructors are decorated with <code>ParamObject</code> attribute.
+    ///
+    /// This is used to identify classes that should be transformed into interfaces.
+    /// </summary>
+    /// <param name="entity"></param>
+    /// <returns>
+    /// <code>true</code> if the entity is a global type with all constructors as param objects,
+    /// <code>false</code> otherwise.
+    /// </returns>
+    let isParamObjectClassPattern (entity: Fable.Entity) =
+        let isGlobalType =
+            entity.Attributes |> Seq.exists (fun att -> att.Entity.FullName = Atts.global_)
+
+        let areAllConstructorsParamObject =
+            entity.MembersFunctionsAndValues
+            |> Seq.filter _.IsConstructor
+            |> Seq.forall (fun memb ->
+                memb.Attributes
+                |> Seq.exists (fun att -> att.Entity.FullName = Atts.paramObject)
+            )
+
+        isGlobalType && areAllConstructorsParamObject
+
     let tryPickAttrib attFullNames (attributes: FSharpAttribute seq) =
         let attFullNames = Map attFullNames
 
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -840,8 +840,23 @@ module Helpers =
             entity.MembersFunctionsAndValues
             |> Seq.filter _.IsConstructor
             |> Seq.forall (fun memb ->
-                memb.Attributes
-                |> Seq.exists (fun att -> att.Entity.FullName = Atts.paramObject)
+                // Empty constructors are considered valid as it allows to simplify unwraping
+                // complex Union types
+                //
+                // [<AllowNullLiteral>]
+                // [<Global>]
+                // type ClassWithUnion private () =
+                //     [<ParamObjectAttribute; Emit("$0")>]
+                //     new (stringOrNumber : string) = ClassWithUnion()
+                //     [<ParamObjectAttribute; Emit("$0")>]
+                //     new (stringOrNumber : int) = ClassWithUnion()
+                //
+                // Without this trick when we have a lot of U2, U3, etc. to map it is really difficult
+                // or verbose to craft the correct F# class. By using, an empty constructor we can
+                // "bypass" the F# type system.
+                memb.CurriedParameterGroups |> List.concat |> List.isEmpty
+                || memb.Attributes
+                   |> Seq.exists (fun att -> att.Entity.FullName = Atts.paramObject)
             )
 
         isGlobalType && areAllConstructorsParamObject
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -2031,10 +2031,25 @@ module Util =
 
     let tryGlobalOrImportedAttributes (com: Compiler) (entRef: Fable.EntityRef) (attributes: Fable.Attribute seq) =
         let globalRef customName =
-            defaultArg customName entRef.DisplayName
-            |> makeTypedIdent Fable.Any
-            |> Fable.IdentExpr
-            |> Some
+            let name =
+                // Custom name has precedence
+                match customName with
+                | Some name -> name
+                | None ->
+                    let entity = com.GetEntity(entRef)
+
+                    // If we are generating TypeScript, and the entity is an object class pattern
+                    // we need to use the compiled name, replacing '`' with '$' to mimic
+                    // how Fable generates the compiled name for generic types
+                    // I was not able to find where this is done in Fable, so I am doing it manually here
+                    if com.Options.Language = TypeScript && isParamObjectClassPattern entity then
+                        entity.CompiledName.Replace("`", "$")
+                    // Otherwise, we use the display name as `Global` is often used to describe external API
+                    // and we want to keep the original name
+                    else
+                        entRef.DisplayName
+
+            name |> makeTypedIdent Fable.Any |> Fable.IdentExpr |> Some
 
         match attributes with
         | _ when entRef.FullName.StartsWith("Fable.Core.JS.", StringComparison.Ordinal) -> globalRef None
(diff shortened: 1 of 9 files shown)


MangelMaxime:
@ncave 

When generating type for 

```fs
[<AllowNullLiteral>]
[<Global>]
type TestWithGenerics<'T, 'A>
    [<ParamObjectAttribute; Emit("$0")>]
    ( _name : 'T, ?age: int) =
    member val name: string option = jsNative with get, set

let test2 : TestWithGenerics<string, obj> =
    TestWithGenerics("", 0)
```

I get this output:

```ts
export interface TestWithGenerics$2<T, A> {
    _name: T,
    age?: int32
}

export const test2: TestWithGenerics<string, any> = {
    _name: "",
    age: 0,
};
```

Do you have an idea on why we don't get the same name for `TestWithGenerics` class ?

ncave:
@MangelMaxime Possibly a difference between using a type's `CompiledName` vs `DisplayName`

MangelMaxime:
I will try to investigate more.

@Freymaurer @joprice 

Basic cases seems to be easy enough to support, but there is more complexes cases caused by the fact that `ParamObject` use the name of the argument for generating the POJO property but nothing in F# force the user to use the same name between inheritance or multiple constructors.

## Inheritance

```fs
[<AllowNullLiteral>]
[<Global>]
type BaseClass
    [<ParamObjectAttribute; Emit("$0")>]
    ( name : string) =
    member val name: string option = jsNative with get, set

[<AllowNullLiteral>]
[<Global>]
type DerivedClass
    [<ParamObjectAttribute; Emit("$0")>]
    ( name : string, age : int) =
    inherit BaseClass(name)
    member val name: string option = jsNative with get, set
```

Which should generate something like:

```ts
export interface BaseClass {
    name: string
}

export interface DerivedClass extends BaseClass {
    age: int32
}

export const derived: DerivedClass = {
    name: "name",
    age: 42,
};
```

However, this F# code is also valid:

```fs
[<AllowNullLiteral>]
[<Global>]
type BaseClass
    [<ParamObjectAttribute; Emit("$0")>]
    ( name : string) =
    member val name: string option = jsNative with get, set

[<AllowNullLiteral>]
[<Global>]
type DerivedClass
    [<ParamObjectAttribute; Emit("$0")>]
    ( thisIsAnotherName : string, age : int) =
    inherit BaseClass(thisIsAnotherName)
    member val name: string option = jsNative with get, set
```

and now the `ParamObject` with generate a property of name `thisIsAnotherName` instead of `name`. I guess, we could add some Fable validation to force the user to use the same argument name? 

Depending on how difficult adding Fable validation is, we can also mark it to do later.

## Multiple constructors

Another situation is how to support multiple constructors:

```fs
[<AllowNullLiteral>]
[<Global>]
type Multiple
    [<ParamObjectAttribute; Emit("$0")>]
    private ( numberOrString : U2<string, int>) =

    [<ParamObjectAttribute; Emit("$0")>]
    new (numberOrString : string, arg: int) =
        Multiple(U2.Case1 numberOrString)

    [<ParamObjectAttribute; Emit("$0")>]
    new (numberOrString : int) =
        Multiple(U2.Case2 numberOrString)

let multiple1 =
    Multiple("test", 1)

let multiple2 =
    Multiple(1)
```

We could generate something like that:

```ts
export type Multiple = {
  numberOrString: string | int32
} | {
  numberOrString: string,
  arg: int32,
} | {
  numberOrString: int32
}

export const multiple1: Multiple = {
  numberOrString: "test",
  arg: 1,
};

export const multiple2: Multiple = {
  numberOrString: 1,
};
```

Freymaurer:
This looks already very promising!

> argument for generating the POJO property but nothing in F# force the user to use the same name between inheritance or multiple constructors.

Not sure how important this is, but if this is the case then we would require more fable magic, as the current docs case with

```fsharp
[<AllowNullLiteral>]
[<Global>]
type User
    [<ParamObjectAttribute; Emit("$0")>]
    (id : int, ?name : string, ?age: int) =
    member val name: string option = jsNative with get, set
```

already assumes that `name` will be called the same for parameter and member.

> ```fsharp
> [<AllowNullLiteral>]
> [<Global>]
> type Multiple
>     [<ParamObjectAttribute; Emit("$0")>]
>     private ( numberOrString : U2<string, int>) =
> 
>     [<ParamObjectAttribute; Emit("$0")>]
>     new (numberOrString : string, arg: int) =
>         Multiple(U2.Case1 numberOrString)
> 
>     [<ParamObjectAttribute; Emit("$0")>]
>     new (numberOrString : int) =
>         Multiple(U2.Case2 numberOrString)
> ```

This is difficult for me to evaluate. Not sure if we must display all alternative constructors. I would have expected only this:

```ts
export type Multiple = {
  numberOrString: string | int32
}
```

As the constructors are only for the f# side and the pojo is only based on the fields? But this might be my misunderstanding

Freymaurer:
@MangelMaxime Do you think it would be possible to keep comments on the interface members? For the output to be like this?

```ts
export interface ButtonProps {
  /** Is this the principal call to action on the page? */
  primary?: boolean;
  /** What background color to use */
  backgroundColor?: string;
  /** How large should the button be? */
  size?: 'small' | 'medium' | 'large';
  /** Button contents */
  label: string;
  /** Optional click handler */
  onClick?: () => void;
}
```

src/Fable.Transforms/FSharp2Fable.Util.fs
```diff
@@ -2002,10 +2031,25 @@ module Util =
...
+                    // how Fable generates the compiled name for generic types
+                    // I was not able to find where this is done in Fable, so I am doing it manually here
+                    if com.Options.Language = TypeScript && isParamObjectClassPattern entity then
+                        entity.CompiledName.Replace("`", "$")
+                    // Otherwise, we use the display name as `Global` is often used to describe external API
+                    // and we want to keep the original name
+                    else
+                        entRef.DisplayName
+
+            name |> makeTypedIdent Fable.Any |> Fable.IdentExpr |> Some
```
MangelMaxime, line 2052: > @MangelMaxime Possibly a difference between using a type's CompiledName vs DisplayName

@ncave You were right this was because for `Global` we always used the `DisplayName` as in general this is used to map to native API.

MangelMaxime:
Ok after spending way too much time trying to make "strict" inheritance works via `extends` keyword, I remembered that in TypeScript you can also inherit via "shape".

```ts
export interface BaseClass {
    name: string
}

export interface DerivedClass {
    name: string
	age: int32
}

export const derived: DerivedClass = {
    name: "name",
    age: 42,
};

export const baseClass : BaseClass = derived;
```

I believe in the scenario of this PR this more than sufficient and will help keep the implementation simple enough compared. The implementation for `extends` started to touch too much different places in Fable in my opinion and at least like that it allows to deliver the feature by covering all cases.

Issue #2636, opened 2021-12-04
StringBuilder missing some Append overloads


jwosty:
### Description

```StringBuilder.Append``` is missing replacements for the float and decimal overloads. There may be more; those two are all I checked.

```fsharp
stringBuilder.Append (string someFloatValue)
```

### Related information

* Fable version: 3.1.4
* Operating system: Windows 10.0.19043

alfonsogarciacaro:
Yes, atm not all overloads of `Append` are supported. If you want to add more, it should be possible to just add the overload to the replacement here: https://github.com/fable-compiler/Fable/blob/1dfdd94c54375fb6c82bfff5ed93b0a031bfc0cc/src/fable-library/System.Text.fs#L3-L16

And tests here: https://github.com/fable-compiler/Fable/blob/1dfdd94c54375fb6c82bfff5ed93b0a031bfc0cc/tests/Main/StringTests.fs#L80-L88

jwosty:
Yep, seems easy enough. Expect a PR soon

PR #4009, opened 2025-01-11
[All] Add StringBuilder.Append char * int overload

roboz0r:
- https://github.com/fable-compiler/Fable/issues/2636
- https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.append?view=net-8.0#system-text-stringbuilder-append(system-char-system-int32) 

Perhaps a lesser used overload but I was using it in some .NET code I'm working on making Fable compatible


MangelMaxime:
Thank you

Issue #3887, opened 2024-09-13
Rolling out nullness annotations


kerams:
.NET 9 is around the corner, and with it support for consuming and emitting nullness metadata in F# projects. JS APIs will often return `null`/`undefined`, and while Fable bindings often capture this information by using options, [sometimes we give no indication that `null`s are a possibility](https://github.com/fable-compiler/fable-browser/issues/99). I think F# 9 will be really useful here, so I'd like to start a discussion about the adoption plan for these annotations.

From the technical point of view, binding projects need to be compiled using .NET 9 SDK, `<LangVersion>preview</LangVersion>` (for now) and with the `--checknulls+` flag (also for now - I think this will be aligned with C#'s `<Nullable>enable</Nullable>`). TFMs don't matter, so `netstandard2.0` assemblies will contain nullness metadata too.

Finally, every API will need to be manually reviewed and annotated as required. In the case of existing bindings that use options with reference types, there's the question of whether we want to preserve options or replace them with `null` annotations, just so that a single, unified approach is used.

MangelMaxime:
I decided to move this issue into Fable repository because there is more to decide that the aesthetic of the API.

@ncave As you probably know more about FCS + Fable internals, what is your vision on how nullness could be added to Fable?

MangelMaxime:
> but then we'll also need to add `net9.0` TFMs so that nullness warnings show up in the IDE, and users also need to retarget their Fable projects to `net9.0` so that nullness is checked during compilation (once the Fable compiler is updated).

> I wonder why for this new feature people need to use target `net9.0`?
> 
> ~I mean in previous addition to F# people where able to keep using `netstandard2.1` for example without issues.~
> 
> After thinking, I think I understand why you want to do that. This is to have support for compiler directives to branch out the API with or without nullness.

To avoid the need for targeting `net9.0`, we could also decide to add a custom compiler directive in Fable like `FABLE_NULLNESS`. And we could make it active by default or not starting a certain version of Fable.

kerams:
> If I remember how TFM works, in order to use net9.0 in your project you need to use .NET 9.0 SDK

You need .NET 9 to use nullness in the first place, so I don't see why requiring `net9.0` to make use of nullness-aware Fable bindings is a problem.

MangelMaxime:
> > Duplicating the API surface means depending on your configuration you have a different API
> 
[...]

PR #4018, opened 2025-01-22
WIP: Nullable/Nullness support

MangelMaxime:
I am starting this PR early to avoid duplicate work regarding nullable/nullness support #3887


ncave:
@MangelMaxime I would suggest doing separate PR for supporting nullable value type (i.e. `System.Nullable<T>`) and a different PR for nullable ref types (`T | null`), and keeping them as small as possible to be able to iterate on them.

Perhaps this PR can be renamed to `Added support for FSharp.Linq.NullableOperators`.

MangelMaxime:
> I would suggest doing separate PRs,

I agree this PR will need to be split because locally I have even more different fixes for others part of Fable because I am discovering some unsupported APIs.

I will make specialized PRs for the different changes.

PR #4020, opened 2025-01-22
Keep comments for ts interface generation :sparkles:

Freymaurer:
I updated ts interface generation to keep xmlDoc comments from f# code. I tried to uphold existing patterns, but please check if everything is fine with you:

The changes transpile the following f# code:

```fsharp
/// This is a user, a very strange individium which will find each and every edge case existing.
[<AllowNullLiteral>]
[<Global>]
type User [<ParamObjectAttribute; Emit("$0")>] (id: int, ?name: string, ?age: int) =
    /// Id used for id stuff
    member val id = id with get, set
    /// name used for naming conventions
    member val name: string option = jsNative with get, set
    /// age because you never tell age
    member val age: int option = jsNative with get, set
```

to:

```ts
/**
 * This is a user, a very strange individium which will find each and every edge case existing.
 */
export interface User {
    /**
     * Id used for id stuff
     */
    id: int32,
    /**
     * name used for naming conventions
     */
    name?: string,
    /**
     * age because you never tell age
     */
    age?: int32
}
```

I kept the example in `src/quicktest/QuickTest.fs`.

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -3788,9 +3788,6 @@ module Util =
         (classDecl: Fable.ClassDecl)
         (ent: Fable.Entity)
         =
-        printfn "[classDecl] %A" classDecl
-        printfn "[ent] %A" ent
-        printfn "[ctx] %A" ctx
 
         let members =
             ent.MembersFunctionsAndValues
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -3788,6 +3788,10 @@ module Util =
         (classDecl: Fable.ClassDecl)
         (ent: Fable.Entity)
         =
+        printfn "[classDecl] %A" classDecl
+        printfn "[ent] %A" ent
+        printfn "[ctx] %A" ctx
+
         let members =
             ent.MembersFunctionsAndValues
             |> Seq.filter _.IsConstructor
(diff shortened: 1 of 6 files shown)


src/quicktest/QuickTest.fs
```diff
@@ -94,7 +94,23 @@ let measureTime (f: unit -> unit) : unit =
 
 printfn "Running quick tests..."
 
+
+/// This is a user, a very strange individium which will find each and every edge case existing.
```
ncave, line 98: *individuum

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -3788,6 +3788,10 @@ module Util =
         (classDecl: Fable.ClassDecl)
         (ent: Fable.Entity)
         =
+        printfn "[classDecl] %A" classDecl
```
ncave, line 3791: I assume these were debug statements, can they be removed?

Freymaurer:
Is there some contribution guide i can follow or just try to follow the existing patterns?

MangelMaxime:
> Is there some contribution guide i can follow or just try to follow the existing patterns?

Just try to follow the existing convention for this PR it would be something like:

```md
* [TS] Include XML Doc comment on interface properties (by @Freymaurer)
```

src/Fable.Transforms/Fable2Babel.fs
Freymaurer: Yes sorry!

Freymaurer:
@ncave done!

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -3797,6 +3798,12 @@ module Util =
                 |> List.mapi (fun index arg ->
                     let name = defaultArg arg.Name $"arg{index}"
 
+                    /// Try to find getter/setter in f# syntax for POJOs. If found propagate its xml doc to interface.
+                    let tryXmlDoc =
+                        ent.MembersFunctionsAndValues
```
ncave, line 3803: This lookup may be a bit brittle when there are overloads. I know we don't normally allow overloads on interfaces, unless the mangle attribute is added, so it should be mostly fine, at least as a first pass.

ncave (approved):
LGTM

MangelMaxime:
Thank you

Issue #4014, opened 2025-01-17
Github job fails with Fable for OSX runners


halcwb:
Build fails using the new Fable compiler but only when running as a github job, with the following stack trace: 

> client: Unhandled exception. System.Text.Json.JsonReaderException: 'W' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
> client:    at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
> client:    at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
> client:    at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
> client:    at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
> client:    at System.Text.Json.Utf8JsonReader.Read()
> client:    at System.Text.Json.JsonDocument.Parse(ReadOnlySpan`1 utf8JsonSpan, JsonReaderOptions readerOptions, MetadataDb& database, StackRowStack& stack)
> client:    at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 utf8Json, JsonReaderOptions readerOptions, Byte[] extraRentedArrayPoolBytes, PooledByteBufferWriter extraPooledByteBufferWriter)
> client:    at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 json, JsonDocumentOptions options)
> client:    at Fable.Compiler.MSBuildCrackerResolverModule.mkOptionsFromDesignTimeBuildAux@80-1.Invoke(String targetFrameworkJson) in /workspaces/Fable/src/Fable.Compiler/MSBuildCrackerResolver.fs:line 82
> client:    at Microsoft.FSharp.Control.AsyncPrimitives.CallThenInvokeNoHijackCheck[a,b](AsyncActivation`1 ctxt, b result1, FSharpFunc`2 userCode) in /home/dev/Projects/fsharp/src/FSharp.Core/async.fs:line 528
> client:    at Microsoft.FSharp.Control.Trampoline.Execute(FSharpFunc`2 firstAction) in /home/dev/Projects/fsharp/src/FSharp.Core/async.fs:line 112
> client: --- End of stack trace from previous location ---
[...]

PR #4021, opened 2025-01-22
[All] Suppress `Detected package downgrade` warning and log json output if we fails to parse MSBuild result

MangelMaxime:
#4014 

[All] Log JSON output if we fail to parse MSBuild result


MangelMaxime:
> Nitpick, but did this suppress `Detected package downgrade`?

No because this was not the issue we faced with the CI for Fable.

If you are encountering an issue with `Detected package downgrade`, we can add the rule to suppress that message.

PR #4027, opened 2025-01-28
[Py] - Print root module and module function comments

alfonsogarciacaro:
Hey there! Cool project ;)

Just a small PR to print root module and module function comments in Python so the linter stops complaining.

Example:

```fsharp
/// A gratuitous example of a module that depends
/// on another module
module Module2

/// Adds two numbers
/// and returns the result
let add x y = 
    Module1.add x y + 4
```

becomes:

```python
"""A gratuitous example of a module that depends on
another module
"""
from Module1.module1 import add as add_1

def add(x: int, y: int) -> int:
    """Adds two numbers
    and returns the result
    """
    return add_1(x, y) + 4
```

I had to modify the signature of `FSharp2Fable.fsi` and the `Compiler` interface, I hope that's ok, but let me know if another way is preferred.

Cheers!

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [TS] Include XML Doc comment on interface properties (by @Freymaurer)
 * [TS] Generate `interface` type when using the "ParamObject" class pattern (by @MangelMaxime)
 * [Rust] - Initial support for null values (by @ncave)
+* [Py] - Print root module and module function comments (by @alfonsogarciacaro)
 
 ## 5.0.0-alpha.7 - 2025-01-23
 
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* [Python] - Print root module and module function comments (by @alfonsogarciacaro)
+
 ## 5.0.0-alpha.9 - 2025-01-28
 
 ### Fixed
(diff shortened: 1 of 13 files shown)


pkese:
Omg @alfonsogarciacaro 

It's soooooo nice to see you here again.

Warmest regards. I hope you're doing well wherever you are.

ncave:
I'm very glad to hear from you again, @alfonsogarciacaro, I hope all is well with you and best wishes!

This looks good, just needs the linting (`dotnet fantomas .`)

MangelMaxime:
Hello @alfonsogarciacaro, glad to hear from you.

Tips if you run `./build.sh` at least once, then the formatting will happen automatically on authoring commit.

TheAngryByrd:
👋 Glad good to see you again! @alfonsogarciacaro

ncave:
Perhaps we could retain the comments for every module, instead of just the root module, with a slight update to the Fable AST:

```fs
type ModuleDecl =
    {
        Name: string
        Entity: EntityRef
        Members: Declaration list
        XmlDoc: string option   // <--- new optional property
    }
```

It is an optional property so it shouldn't be technically a breaking change.
I know it's not that useful for target languages that flatten their modules like JS/TS,
but it can still be helpful to output the module comments.

alfonsogarciacaro:
Thank you @pkese, @ncave, @MangelMaxime, @TheAngryByrd for the warm welcome ❤️ ...and for accepting the PR.
The youngest is finally about to start primary school and work is a bit less crazy now, so stress is slowly coming back to manageable levels. Not sure if I can dedicate much time to open source again, but playing with Fable 5 is a lot of fun!

@MangelMaxime 
Wow, there is a lot of useful stuff in the repo now! Actually, I did run build.sh once but Husky gave me an error, so I just commented out the line that activates it for now. I need to look into that.

@ncave 
That makes a lot of sense, the only reason I didn't add the comment to the Module declaration is Fable AST erases it, so I had to find another way to pass the info. But adding the XmlDoc to the ModuleDecl as well sounds like the right thing to do.

Issue #4025, opened 2025-01-27
[TS] Generate imports rule when using a "ParamObject" from another file


MangelMaxime:
@Freymaurer reported to me that we don't have `import` the type when consuming a "ParamObject" defined in another file which results in a TS compilation error.

Related to the work done in #4007

Freymaurer:
@MangelMaxime  Can you point me to some examples on the import rules? Then i can have a look at it.

alfonsogarciacaro:
A tricky thing here is the semantics of the attributes is being challenged. `Global/Import` were intended to import types from the outside to F#, and now we want to use it export code from F# to the outside.

At this point, if #4007 has only been released in alpha the `Global/ParamObject/Emit` combination could be reduced to a single attribute, say `PojoDefinedByConsArgs` or similar. Then we should get exports automatically and we can also establish constraints (no secondary constructors, no inheritance, no interface implementation).

@MangelMaxime If you agree with that I can try to adapt your work in #4007 in this sense.

MangelMaxime:
@Freymaurer Should be fixed with 5.0.0-alpha.10

Freymaurer:
Awesome will test it tomorrow! 🚀

Freymaurer:
Seems like fable expects to find a type reflection information for my `[<Pojo>]`:

```fsharp
[<Pojo; AllowNullLiteral>]
type Term (?name: string, ?id: string, ?description: string, ?source: string, ?href: string, ?isObsolete: bool, ?data: obj) =
    member val name: string option = name with get, set
    member val id: string option = id with get, set
    member val description: string option = description with get, set
    member val source: string option = source with get, set
    member val href: string option = href with get, set
    member val isObsolete: bool option = isObsolete with get, set
    member val data: obj option = data with get, set

```

![Image](https://github.com/user-attachments/assets/f2549193-1c08-464b-a36a-003270e850e1)

This was triggered by using the pojo as field in a record type:

```fsharp
type private TermSearchResult = {
    Term: Term
    IsDirectedSearchResult: bool
}
```

producing the following reflection for the record type (note `Term_$reflection()`):

```js
function TermSearchResult_$reflection(): TypeInfo {
    return record_type("Swate.Components.TermSearchResult", [], TermSearchResult, () => [["Term", Term_$reflection()], ["IsDirectedSearchResult", bool_type]]);
}
```

Issue #3531, opened 2023-09-18
`sync-fcs-repo` and `copy-fcs-repo` broken


MangelMaxime:
### Description

Hello @ncave, `sync-fcs-repo` target seems to be broken to me.

Our `build.fsx` si trying to update a `build.fsx` file from the FCS repo which doesn't seems to exist. And even, when trying to run commands manually, it seems like I am missing steps or not running from the correct directory if I follow what our `build.fsx` tell me.

What are the steps to build the service_slim and the Fable metadata packages?

What is the usage of service_slim?

ncave:
@MangelMaxime I don't think they're broken, but they may be expecting you to have this `dotnet/fsharp` forked [repo](https://github.com/ncave/fsharp) locally, in a specific folder. There are quite a few branches in there, but most of them are only kept for historic reasons. You can just go to the [PR page](https://github.com/ncave/fsharp/pulls), the important ones are listed as long-term open PRs for reference:

- `fable` (see PR [#1](https://github.com/ncave/fsharp/pull/1)) - Modified FCS version that can be compiled with Fable (for REPL use, etc.).
- `service_slim` (see PR [#2](https://github.com/ncave/fsharp/pull/2)) - The FCS version that Fable tool uses. Stock FCS, with a light facade on top.
- `export` (see PR [#3](https://github.com/ncave/fsharp/pull/3)) - a slightly modified FCS version used to produce .NET metadata (for REPL).

There are instructions how to build each one of those, in the PRs.

Technically, you only need to run `sync-fcs-repo` when there is something new in those branches. At the moment, they have already been synched with Fable. These branches are usually rebased from the upstream FCS (`dotnet/fsharp`) from time to time, usually whenever there is some major F# lang revision.

I plan on updating those branches close to the .NET 8.0 release, and I'll sync them with Fable.
The `service_slim` and `export` branches are relatively easy to rebase to the latest F# version, where `fable` is usually a bit more involving.

It all depends on how much churn there was in `dotnet/fsharp` in the mean time, sometimes there is a lot.

If that's unclear or you have other questions, please feel free to ask.

And lest I forget, thanks for picking up the torch and making new releases!

PR #4036, opened 2025-02-08
[JS/TS] PojoDefinedByConsArgs attribute

alfonsogarciacaro:
See https://github.com/fable-compiler/Fable/issues/4025#issuecomment-2636526013

This adapts @MangelMaxime work in #4007 to use a single attribute instead of the combination `Global/ParamObject/Emit`. This simplifies the declaration and allows for the interface to be exported in TS as export rules don't conflict with `Global` attribute.

The name `PojoDefinedByConsArgs` is tentative, please let me know if you prefer to change it.

It's a bit of a tricky feature, as many things can go wrong if the user doesn't declare the type correctly, but let's see how it goes :)


MangelMaxime:
`PojoDefinedByConsArgs` seems like a complex attribute name even if it is quite describe of the internals.

What about using something like:

- `Pojo` as this is what we are trying to generate (I think in Fable 1 or 2 we had a `Pojo` that we removed but because it was a long time ago I think this should be ok)
- `PojoInterface` if we want to avoid re-using an old attribute and have one which seems less intimidating.

If I understand it also means that the old `Global`/`ParamObject` pattern can be fully replaced by the new attribute right? Perhaps, we could generate a warning to ask the user to migrate to the new attribute?

alfonsogarciacaro:
`PojoDefinedByConsArgs` is definitely ugly 😅 `Pojo` could work, I was thinking that we did have something like but I don't remember how it worked.

> Perhaps, we could generate a warning to ask the user to migrate to the new attribute?

ParamObject + Emit will still work, but it is indeed a good idea to emit a warning to encourage the use of the new attribute ...I just hope users don't start decorating everything with Pojo and ask it to support everything (reflection, etc) as it often happens :)

MangelMaxime:
> `PojoDefinedByConsArgs` is definitely ugly 😅 `Pojo` could work, I was thinking that we did have something like but I don't remember how it worked.

If I remember correctly, we could decorated a type with `[<Pojo>]` to emit a Pojo instead of a standard class or something like that. But with the move to Fable 2 or 3, we changed the representation of the some types and they ended up compatible with code expecting Pojo so we were able to remove that attribute. Probably thanks to the inheritance to `Record` class.

> I just hope users don't start decorating everything with Pojo and ask it to support everything (reflection, etc) as it often happens :)

I hope not too.

I think `[<Pojo>]` should be limited to interop code and if someone want to use reflection they should use normal interface/classes because native code will not be able to consume the reflection code (at least not easily).

Also, interfaces are still more powerful in certain usage because you can inherit several interface versus only one class.

MangelMaxime:
@alfonsogarciacaro Do you know why Fable standalone would not be able to see the new attribute added to Fable.Core ?

alfonsogarciacaro:
IIRC it is because fable-standalone gets Fable.Core and other .NET/F# core libraries from fable-metadata/lib so we need to update the Fable.Core.dll copy there (or skip the tests in fable-standalone for now). I don't remember if the Fable.Core.dll copy in metadata was just compiled normally or using @ncave's method to size down dlls #3531

MangelMaxime:
@alfonsogarciacaro That was my suspicion too but because we were having a direct project reference I was not sure.

So in theory, to complete this PR we can just rename the attribute to `[<Pojo>]` or `[<PojoInterface>]` and merge it?

ncave:
@MangelMaxime The `Fable.Core.dll` is just bindings (no code), so you can just drop a fresh version in [here](https://github.com/fable-compiler/Fable/tree/main/src/fable-metadata/lib).

alfonsogarciacaro:
> That was my suspicion too but because we were having a direct project reference I was not sure.

Also iirc, that is what the `--exclude Fable.Core` option is used for, to ignore compilation of project references:
https://github.com/fable-compiler/Fable/blob/5c95930cf4b778956eade0c5bc702070613ea2b1/src/Fable.Build/Test/Standalone.fs#L23-L28

I updated Fable.Core.dll and renamed the attribute. I left `pojoDefinedByConsArgs` in the internal functions as it is a bit more informative. Name choice is tricky, `Pojo` may be too broad, but `PojoInterface` could be confusing (is it only for F# interfaces?). Maybe `TypescriptPojoInterface`, same as we have `TypescriptTaggedUnion`, but it's also a bit long.

Note: Fable.Core.dll compilation with net9.0 didn't work so at the end I used net6.0. I also had to use Debug mode vs Release mode. I guess it is likely related with .dll trimming, probably fable standalone cannot deal with it.

ncave:
@alfonsogarciacaro 
> Fable.Core.dll compilation with net9.0 didn't work

It should work, you just have to turn off the metadata compression using `<CompressMetadata>false</CompressMetadata>`, as current Fable JS standalone cannot handle compressed metadata.

In theory, we could implement partial support for `System.IO.Compression` in JS using [DecompressionStream](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream), but I'm not sure it's worth the effort at the moment, so just disable the compression for now. Also, metadata binaries usually get gzipped when downloaded over the internets anyway, so it's fine.

kerams:
Alfonso is alive! Great to see you here again.

MangelMaxime:
> Name choice is tricky, `Pojo` may be too broad, but `PojoInterface` could be confusing (is it only for F# interfaces?). Maybe `TypescriptPojoInterface`, same as we have `TypescriptTaggedUnion`, but it's also a bit long.

I think `Pojo` is fine because we also specified the attribute target to be used only for classes.

`TypescriptPojoInterface` could also work but this is but raise the same issue as `PojoInterface`, we could also think about `TypescriptInterface` because we could think we want to out this type as a "typescript interface` but the reality is that all the `TypescriptXXX` can also be used when targeting JavaScript.

Pojo is a well known concept in JavaScript, so I think this should be ok.

PR #4037, opened 2025-02-08
[JSX] Print Empty Elem - Self closing tags when elem has no children

shayanhabibi:
I'm trying to enable and enforce the JSX print to self close tags when they don't have children. This way property spreads or children attributes aren't overridden in the Oxpecker.Solid dsl.

Draft:
- ~~I'd really appreciate any direction to where tests for this would be contained and how to further this pull request to completion.~~
- I've done my best to grok the origin of this behaviour and treat it. Apologies if this is incorrect. I would appreciate guidance.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [JS/TS] - Fix anonymous record printing (#4029) (by @alfonsogarciacaro)
 * [Python] - Fix #3998: PhysicalEquality (by @alfonsogarciacaro)
 
+### Changed
+
+* [JS/TS] In `JSX`, generate self closing element when element has no children (#4037) (by @shayanhabibi)
+
 ## 5.0.0-alpha.9 - 2025-01-28
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


alfonsogarciacaro:
I don't remember if there was a specific reason not to close the tag immediately but probably it should be fine. Do we need a space before `/>`? It'd be great to have some JSX tests in `tests/Integration/data`

shayanhabibi:
Well @Lanayx identified that it is not accepted under the HTML5 spec; however it is in JSX/React etc. Without this, spreading properties that contain children are overridden without explicit children between the open and closing tags.

I can try and grok the tests and put something together, but don't know whether they'll be up to scratch

alfonsogarciacaro:
Thanks for the explanation. Could you please add a succinct version as a comment to the code. This way if in the future we want to generate the closing tag to conform with HTML5 we'll remember there's a reason to do it like this.
The CounterJSX component is being called without children in the React tests, so at least that should be a way to test the feature.

shayanhabibi:
Thanks for having a look, your work on this project, and for the helpful comments!

I've added the documentation with some source references but will have to defer writing up the tests to tomorrow for this and #4038 .

For the immediate future I'll use a local build of fable since I can at least be confident the functionalities will be merged in some way.

Issue #4041, opened 2025-02-10
Fable generates javascript that throws an exception in the browser


klofberg:
Having a file named `main.fsx` with the following code and running `dotnet fable`:

```fsharp
type Model = unit

let update (model: Model) =
    model, ()

update ()
```

the fable compiler generates the following javascript:

```javascript 
export function update() {
    return [model, undefined];
}

update();
```

Running this code in the browser will throw an exception `ReferenceError: model is not defined`.

This is a problem for me in Elmish Land (https://elmish.land/docs/core-concepts/pages) where I scaffold elmish pages. The reason for the type alias for the Model is because I want to make it is easy for the user to extend the page.

I'm using fable compiler version 4.24.0 and dotnet version 9.0.101 on macos 15.0.

PR #4042, opened 2025-02-10
Fix removing function argument when argument is unit

klofberg:
This PR fixes the issue: https://github.com/fable-compiler/Fable/issues/4041

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -3370,7 +3370,6 @@ module Util =
         let args, body =
             match args with
             | [] -> args, body
-            | [ arg ] when arg.Type = Fable.Unit -> [], body
             | _ when not isJsx -> args, body
             | _ ->
                 // SolidJS requires values being accessed directly from the props object for reactivity to work properly


klofberg:
Do you think it would be possible to remove the argument from the body instead?

alfonsogarciacaro:
Interesting, I thought we were doing this already but apparently not 😅 
I've sent a naive implementation #4044 but it seems to work

klofberg:
I'm uncertain how to run the tests? If i run them with `dotnet test` all tests passes but I assume that they aren't checked with fable? When I try `./src/Fable.Cli/bin/Debug/net8.0/fable tests/Js/Main --exclude Fable.Core --noCache` I get a lot of errors. 

Is this the best place for me to ask questions or should I use Discord or something else?

MangelMaxime:
@klofberg When working with Fable you should use the `build.sh` and `build.bat` script (run `build.sh --help` to learn more)

They will give you all the commands that you can use to work with Fable.

In this case, it would be something like `build.sh test javascript` 

I made Amplifying F# session showing how Fable works: https://amplifyingfsharp.io/sessions/2023-12-15/

klofberg:
Great, thanks! :-)

alfonsogarciacaro:
@klofberg  I'd be more than happy if you cherry-pick the commit in #4044 (or just copy the code) and continue the PR. An integration test could be a good idea, so we also check that functions with `()` as argument become argument-less in JS.

To add an integration test you just need to add/edit at project in the `tests/Integration/Integration/data` directory with the `.js.expected` files (you can run the tests once first and change the .actual extension). See another project in the same folder as an example.

About Python, it should have the same behavior more or less. The only thing that surprised me is when I checked In Fable repl, Python had `unit` arguments even if they're supposed to be removed, so I need to investigate this: https://github.com/fable-compiler/Fable/blob/5c95930cf4b778956eade0c5bc702070613ea2b1/src/Fable.Transforms/Python/Fable2Python.fs#L3386

klofberg:
@alfonsogarciacaro I've created a new PR #4050 with your code and added a integration test for javascript. I will try to have a look at the Python issue you mention also.

PR #4044, opened 2025-02-11
[JS/TS/Python] Fix #4041: Replace idents with unit type

### LINKED ISSUE #4041: Fable generates javascript that throws an exception in the browser

alfonsogarciacaro:
Fix #4041: Replace idents with unit type

Missing tests, and investigate if unit arg is being removed in Python (what about Rust?)

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -2827,13 +2827,17 @@ module Util =
 
         | _ -> transformDecisionTreeWithExtraSwitch com ctx returnStrategy targets treeExpr
 
-    let transformIdent (com: IBabelCompiler) ctx id =
-        let e = identAsExpr id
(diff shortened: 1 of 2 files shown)


MangelMaxime:
Closing as the original issue has been solved in another PR.

PR #4050, opened 2025-02-17
Fix 4041

klofberg:


klofberg:
@MangelMaxime @alfonsogarciacaro @dbrattli @ncave Hi, Is there any chance for this to be merged into main? Do I need to do anything else? I've tested locally and added tests. (See #4042 for more info)

dbrattli:
The description of this PR is empty. Please add Why and How. Links are good, but the PR should explain exactly what is being done and why.

klofberg:
> The description of this PR is empty. Please add Why and How. Links are good, but the PR should explain exactly what is being done and why.

Sorry. I added info now 👍

klofberg:
> The description of this PR is empty. Please add Why and How. Links are good, but the PR should explain exactly what is being done and why.

@dbrattli 
I’ve added a title and description. Please let me now if it needs to be improved!

dbrattli:
@klofberg Since you are changing `FSharp2Fable.Utils.fs` this PR affects all targets. The code generation for Python looks worse after this fix, so I don't think I would like to have the same fix for Python.

For Python without this PR:

```py
def update(model: None = None) -> tuple[None, None]:
    return (model, None)

ignore(update())
```

With this PR:

```py
def update(__unit: None=None) -> tuple[None, None]:
    return (model, None)  # <--- Undefined name `model`

ignore(update())
```

klofberg:
That's not good. I'll look in to it and try to make the generated python as before. 

Is it ok to have conditionals in that file for different targets or do you have another preferred place to do the fix?

ncave:
@klofberg Thank you for your contribution, I'll close this PR as it is superseded by #4205 (which is basically #4042 + tests).

PR #4052, opened 2025-02-18
 [JS/TS] Fix #4049: decimal/bigint to integer conversion checks

### LINKED ISSUE #4049: Cast from decimal to `uint8`, `uint16`, `uint32`, `uint`, `uint64`, `int`, etc. don't follow .NET behaviour
Running this code in .NET will throw exceptions:

```fs
uint8 -1m |> printfn "%A"
uint16 -1m |> printfn "%A"
uint32 -1m |> printfn "%A"
uint -1m |> printfn "%A"
uint64 -1m |> printfn "%A"
int -18446744073709551615m |> printfn "%A"
```

Exceptions looks like `Value was either too large or too small for a UInt16.`

On Fable, the code succeeds and output:

```text
255
65535
4294967295
4294967295
18446744073709551615
0
```

I am not sure how to approach fixing this.

My first idea was to apply a fix in `toLong` function here:

https://github.com/fable-compiler/Fable/blob/204d30d116a0862bf16080ac3761b87477af35aa/src/Fable.Transforms/Replacements.fs#L299-L303

But I think this would be to wide because in some scenario .NET allows to do that:

```fs
let value = 4294967295UL
let c = uint8 value
printfn "%A" c
// No exception thrown it actually prints: 255uy
```

I believe we should only fix portion of the code when we call `Decimal.op_explicit` at:

https://github.com/fable-compiler/Fable/blob/204d30d116a0862bf16080ac3761b87477af35aa/src/Fable.Transforms/Replacements.fs#L2445-L2454

Should we add function like that to `Decimal.ts` ?

```ts
export function op_Explicit_ToChar(d: Decimal) {
  const n = toNumber(d);
  if (n < 0 || n > 65535 || isNaN(n)) {
    throw new Error("Value was either too large or too small for a character.");
  }

  return String.fromCharCode(n);
}

export function op_Explicit_ToUint64(d: Decimal) {
  const n = toUInt64(fromDecimal(d));
  if (n < 0 || n > 18446744073709551615n) {
    throw new Error("Value was either too large or too small for a UInt64.");
  }

  return n;
}
```

Is it the right approach / place to store them? @ncave @alfonsogarciacaro

ncave:
-  [JS/TS] Fix #4049: decimal/bigint to integer conversion checks

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* [JS/TS] Fix #4049: decimal/bigint to integer conversion checks (by @ncave)
+
 ## 5.0.0-alpha.10 - 2025-02-16
 
 ### Added
--- a/src/Fable.Compiler/CHANGELOG.md
+++ b/src/Fable.Compiler/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* [JS/TS] Fix #4049: decimal/bigint to integer conversion checks (by @ncave)
+
 ## 5.0.0-alpha.10 - 2025-02-16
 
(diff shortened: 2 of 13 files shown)


MangelMaxime:
Thanks for the PR

PR #4065, opened 2025-02-23
[JS/TS] Fix date formatting when repeating a format token more than the known format (example repeating 'd' more than 4 times)

### LINKED ISSUE #4062: Incorrect result for `DateTime.ToString` with more than 3 `g`
```fs
DateTime(2014, 7, 1, 16, 37, 0).ToString("r gg gggggggggggg") |> printfn "%A"
```

Should give `"r A.D. A.D."` but currently gives `"r A.D"`

MangelMaxime:
Fix #4062

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,7 +131,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ["3.10", 3.11, 3.12.8]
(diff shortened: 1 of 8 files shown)


MangelMaxime:
Ok, I don't know why but the Python tests drove me crazy on this one.

It was working locally and failing on the CI but on a untouched tests..

I just needed to rant a bit, it seems to be back to normal now.

Issue #4046, opened 2025-02-11
Numeric format strings silently ignoring unsupported values or producing wrong results


kerams:
```fsharp
let x = 100000
x.ToString "#,#" |> printfn "%s"
x.ToString "N0" |> printfn "%s"
printfn $"{x:N0}"
```

In fsi, this prints 3 identical lines based on the current culture.

In Fable, [the result is](https://fable.io/repl/#?code=DYUwLgBAHhC8EEYAMKUFgBQUB0AVA9gMpgBOAlgHYDmEARAMQA09tEAPgHwQAO5FYAMwp0ApAGdamHAWJ8atAHJJWnHn0HDa4yRl6UNEACS0A3lABcSgL60gA&html=Q&css=Q)

```
100,000
100000
%P(N0)
```

`x.ToString "#,#"` produces the correct string, but I'd appreciate a Fable warning saying that invariant culture will be used (I think I've seen something like that in other places).
`x.ToString "N0"` silently produces an incorrect string. An unsupported format error is warranted in my opinion.
`$"{x:N0}"` creates the compiled representation of the interpolation string, which is complete nonsense. An unsupported format error is warranted in my opinion.

MangelMaxime:
> `x.ToString "#,#"` produces the correct string, but I'd appreciate a Fable warning saying that invariant culture will be used (I think I've seen something like that in other places).

I didn't know Fable was doing that in some cases, but I found this example:

```fs
TimeSpan.Zero.ToString("g") |> ignore
```

Will make Fable compilation fails with this error:

```text
FABLE: TimeSpan.ToString with one argument is not supported, because it depends of local culture, please add CultureInfo.InvariantCulture
```

 @kerams Is it what you are referring too?

kerams:
I think so. Guess it's not a warning but an error.

MangelMaxime:
The warning one that I know about is something like if you pass a CultureInfo parameter Fable says, that it will just be ignored.

I am just discovering the error one today thanks to you.

MangelMaxime:
I am looking at fixing this issue and I am able to fix the first 2 issues.

Looking more into these issues, I have the feeling that we are having a different behavior / expectation depending on what API we are calling.

If we call `(5592405).ToString("x")` or `TimeSpan.ToString("g")`, we expect to generate an error asking the user to provide `CultureInfo.InvariantCulture`.

[...]

PR #4068, opened 2025-02-26
[JS/TS] Fix `N` and `n` numeric format + add `C`, `c`, `B`, `b` numeric format

MangelMaxime:


src/fable-library-ts/String.ts
```diff
@@ -313,23 +313,50 @@ export function format(str: string | object, ...args: any[]) {
...
           break;
         case "p": case "P":
           precision = precision != null ? precision : 2;
           rep = toFixed(multiply(rep, 100), precision) + " %";
           break;
-        case "d": case "D":
-          rep = precision != null ? padLeft(String(rep), precision, "0") : String(rep);
-          break;
+        case "R": case "r":
+          throw new Error("The round-trip format is not supported");
```
ncave, line 359: @MangelMaxime Perhaps throwing an error at runtime may not be best, perhaps just output what we can, and add a warning if possible in the Replacements.

src/Fable.Transforms/Replacements.fs
```diff
@@ -2496,7 +2496,10 @@ let parseNum (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr op
     | "Pow", _ ->
         Helper.GlobalCall("Math", t, args, i.SignatureArgTypes, memb = "pow", ?loc = r)
         |> Some
-    | "ToString", [ ExprTypeAs(String, format) ] ->
+    | "ToString", [ ExprTypeAs(String, format) ]
```
ncave, line 2499: @MangelMaxime Perhaps add another case here for string const formats so we can issue warnings for unsupported formats, something like this:
```fs
| "ToString", [ StringConst format ] ->
```

MangelMaxime: This is similar to the TODO, I put below.

I will have a look at implementing it, I though it would be difficult to do but in theory we just need a regex that match against a single char followed by some digits.

This will also solve the issue from your other comment, as we will be able to capture the "R" format at compile time.

src/fable-library-ts/String.ts
MangelMaxime: I now think throwing at runtime is probably the right behavior see https://github.com/fable-compiler/Fable/issues/4046#issuecomment-2689196864

We are actually missing support for failing for any invalid format currently, I will add it in this PR too.

Regarding `R` format, we could decide to make it fall like any invalid format or generate a custom error saying that Fable don't support it or we welcome contribution (but I am not sure if this adds a lo of value).

MangelMaxime:
This becomes a much bigger PR that I anticipated.

I added tests for all the different numeric types and different format.

This allowed me to fix some errors in the existing implementation.

If a tests is commented this can be because of 2 reasons:

1. I didn't know how to make an implementation for this type of number for example, in JavaScript `bigint` don't supports "unsigned right shift" so I was not able to implements `B` format them
2. We lack information at runtime to generate the exact same code as in .NET. Indeed, depending on the numeric type the default precision when not provided differs.

	The problem is that we don't have this information at runtime, so we can't make the default precision dynamic.

	Because we are generating a "custom string" to identify the format pattern:
	
	```fs
	let format = emitExpr r String [ format ] "'{0:' + $0 + '}'"
	```

	A solution, could be to add the type information to this pattern format example: `{0:N2-0}` where `-0`  means that this is a `byte`. Using a digit to minimize the impact on the bundle size.

	But I don't know if this is worth it. I think the current improvements are already good enough. Can be revisited later, in a same way the current implementation feels difficult to read so perhaps we could rewrite it without using `regex` like we did for the `Date.ts`

src/fable-library-ts/String.ts
ncave: IMO as long as there is a compile time warning for the unsupported types, the runtime behavior can be whatever you think is best.

MangelMaxime:
@ncave @kerams 

This PR includes a mix of the discussion we had at #4046

- It add supports for the missing numeric format I know how to implements
- It fixes existing format which had errors in them
- Try to detect when an invalid/non-supported format is used and generate a compiler error

	Like that people don't rely on a custom/non-defined Fable implementation. If they need to support a different format for .NET and Fable, they can use compiler directives likes done for some others cases when the code needs to be adapted

- The `format` implementation also throw exceptions if the format is invalid or use on a wrong type (for example using `b` on a non-integral type)

	This mimic the behavior of .NET, and this exception throwing mechanism is also a way for us to cover cases when we can't detect to the format used at compile time (because it is not a provided as a `StringConst`)

kerams:
Much obliged, Maxime.

ncave (approved):
I state my [personal opinion](https://github.com/fable-compiler/Fable/issues/4046#issuecomment-2692271376) in #4046, but this is fine.

Issue #2288, opened 2020-11-22
Fable 3 handles <TreatWarningsAsErrors> in referenced projects differently than Fable 2


sasmithjr:
### Description

Fable 3 handles `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` in referenced projects differently than Fable 2.

I have one project, Project A, that has turned on treating warnings as errors. Project B references A, and B contains some code with a warning. A's configuration causes B to treat the warning as an error when using Fable 3.

### Repro code

[Repo here](https://github.com/sasmithjr/nagareyama-008-warnings-as-errors)

1. Clone repo
2. `npm install`
3. `dotnet build ./tests` to see that it builds with a warning
4. `npm run test` to see that Fable 2 runs the test with the warning
5. `npm run test-nagareyama` to see Fable 3 fails to build due to warning-as-an-error
6. Set `<TreatWarningsAsErrors>` to false (or comment the line out) in `./src/Stuff.fsproj`
7. Re-run step 5 to see the test run and display the warning

### Expected and actual results

Not sure what's expected. Saw the Fable3  blog posts say no expected breaking changes, so I thought I'd let y'all know just in case this is an issue.

Found while working to migrate to Fable 3.

### Related information

* Fable version: fable-compiler v2.13.0 and fable 3.0.0-nagareyama-rc-008. Unsure if present in earlier RCs.

alfonsogarciacaro:
Thanks for reporting @sasmithjr! Same as Fable 2, Nagareyama collects all sources and creates a single project with them for compilation. The difference in this case is that, while Fable 2 totally ignored compiler options from refererenced projects, Nagareyama takes some into account. We thought this would be more accurate but it's true in cases like yours this may have unexpected results. Maybe the warnings as errors option should only be set by the main project? What do you think @ncave?

ncave:
@sasmithjr @alfonsogarciacaro This is correct, some options from referenced projects are collected, but the ones from the main project are applied last, so you have the opportunity to turn any of them off, e.g. in this case if you define `<TreatWarningsAsErrors>false</TreatWarningsAsErrors>` in your main project.

We opted to ignore certain options from referenced projects that don't make sense to be aggregated. Perhaps we can consider ignoring `--warnaserror` as well if you think that is breaking existing projects.

[...]

PR #4070, opened 2025-03-05
[JS/TS] add tests for nullness supports

### LINKED ISSUE #3887: Rolling out nullness annotations

### LINKED ISSUE #3991: Support Unchecked.nonNull  for F# 9
### Description

F# 9 provides `Unchecked.nonNull` to supress the nullness check essentially it is a no op or rather acts like an id function.
It would be great if it is supported.

### Repro code

```fsharp
let x = ((obj()).ToString() |> Unchecked.nonNull).ToUpperInvariant()
```

Currently it fails with
``error FABLE: Microsoft.FSharp.Core.Operators.Unchecked.NonNull (static) is not supported by Fable``

### Related information

The API is documented at  https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-fsharp-9/#flow-analysis

* Fable version: 5.0.0-alpha.3
* Operating system: Any

MangelMaxime:
Fix #3887
Fix #3991

@kerams

It seems like Fable already supports "nullness" and its new API thanks to @ncave adding the new `NonNull`, `Null`, APIs.

I tried to create some tests suites to check if "nullness" is working but I think that right now this is more testing if `T | null` and related APIs works with Fable.

I am not sure if we want to enable on the test project `<Nullable>enable</Nullable>`, this generates several new warnings and I am not sure if this adds value for this specific project.

Example of places with warnings:

image

image

image

If we thinks that enabling nullable make sense, I can give a try to fix them.

I am not familiar with nullness API, @kerams do you see a common case that is not covered and you would like to be covered? I tried to look at F# repository for tests regarding but it seems like most of them are related to testing compiler warning.

### DIFF
--- a/.fantomasignore
+++ b/.fantomasignore
@@ -1,3 +1,5 @@
 src/fcs-fable
 tests/**
 tests_external/**
+
+!tests/Js/Main/Nullness.fs
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * [Python] Add support for `nullArgCheck`(by @MangelMaxime)
+* [All] Add support for F# `nullness` (by @MangelMaxime)
 
 ### Fixed
 
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [Python] Add support for `nullArgCheck`(by @MangelMaxime)
 * [All] Add support for F# `nullness` (by @MangelMaxime)
+* [JS/TS] Add support for `Unchecked.nonNull` (by @MangelMaxime)
 
 ### Fixed
 
--- a/src/Fable.Compiler/CHANGELOG.md
+++ b/src/Fable.Compiler/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * [Python] Add support for `nullArgCheck`(by @MangelMaxime)
+* [All] Add support for F# `nullness` (by @MangelMaxime)
 
 ### Fixed
 
--- a/src/Fable.Compiler/CHANGELOG.md
+++ b/src/Fable.Compiler/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [Python] Add support for `nullArgCheck`(by @MangelMaxime)
 * [All] Add support for F# `nullness` (by @MangelMaxime)
+* [JS/TS] Add support for `Unchecked.nonNull` (by @MangelMaxime)
 
 ### Fixed
 
--- a/src/Fable.Compiler/ProjectCracker.fs
+++ b/src/Fable.Compiler/ProjectCracker.fs
@@ -427,10 +427,12 @@ let private extractUsefulOptionsAndSources
                 accSources, line :: accOptions
             else
                 accSources, accOptions
+        // Only forward the nullness flag if it is coming from the main project
+        elif line.StartsWith("--checknulls+", StringComparison.Ordinal) && isMainProj then
+            accSources, line :: accOptions
         elif
             line.StartsWith("--nowarn", StringComparison.Ordinal)
             || line.StartsWith("--warnon", StringComparison.Ordinal)
-            || line.StartsWith("--checknulls+", StringComparison.Ordinal)
         then
             accSources, line :: accOptions
         elif line.StartsWith("--define:", StringComparison.Ordinal) then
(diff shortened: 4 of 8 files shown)


kerams:
Off the top of my head, nullness of input parameters of some Fable.Core functions. More often than not it will be `obj` -> `objnull`.

- `JsInterop.isNullOrUndefined`
- `JsInterop.emitJsExpr` (you can pass `null` or `()` if you don't refer to the parameters)
- `JsInterop.emitJsStatement`
- `JsInterop.jsTypeOf`
- `JS.Constructors.Array.isArray`
- etc.

There's a LOT more, but the functions can be easily adjusted on demand later. The only downside is the signature in Fable.Core has to be hidden behind `#if FABLE_COMPILER_5`.

kerams:
> I am not sure if we want to enable on the test project <Nullable>enable</Nullable>, this generates several new warnings and I am not sure if this adds value for this specific project.

The problem is, as far as I can see F# does not respect `#nullable enable`, so without `<Nullable>enable</Nullable>` you are not checking whether Fable actually shows warnings and enforces `TreatWarningsAsErrors`.

MangelMaxime:
> Off the top of my head, adding nullness to input parameters of some Fable.Core functions. More often than not it will be changing `obj` -> `objnull`.
> 
>     * `JsInterop.isNullOrUndefined`
> 
>     * `JsInterop.emitJsExpr` (you can pass `null` or `()` if you don't refer to the parameters)
> 
>     * `JsInterop.emitJsStatement`
> 
>     * `JsInterop.jsTypeOf`
> 
>     * `JS.Constructors.Array.isArray`
> 
>     * etc.
> 
> 
> There's a LOT more, but the functions can be easily adjusted on demand later. The only downside is the signature in Fable.Core has to be hidden behind `#if FABLE_COMPILER_5 || NUGET_BUILD`.

Here, I think you are not speaking about nullness support directly but more shaping Fable.Core API to takes advantages of nullness.

Fable.Core is a pure binding, meaning that it is served as a `.dll` only not `fable` folder with the sources files. This means that the compiler directives will have no effect.

We will need to check if nullness information are backward compatible in general. But if we move to use `| null` instead of `Option` in the API design this will probably be a major bump because of breaking changes.

> The problem is, as far as I can see F# does not respect `#nullable enable`, so without `<Nullable>enable</Nullable>` you are not checking whether Fable actually shows warnings and enforces them in combination with `TreatWarningsAsErrors`. If you just test it manually for now, I'll believe you:)

You are right neither the warning or error is reported by Fable right now. I will take a look why we don't get the diagnostics (perhaps we are not ignoring it or not passing the configuration to FCS).

kerams:
> We will need to check if nullness information are backward compatible in general

It is, so if Fable.Core is not distributed in source form, we don't need directives.

MangelMaxime:
Nullness warnings are now reported.

image

Supporting `TreatWarningsAsErrors` is a whole other task. 

- https://github.com/fable-compiler/Fable/issues/2521
- https://github.com/fable-compiler/Fable/issues/2288

This is because Fable recreate a "single project" using the user project and dependencies. So when enabling, `TreatWarningsAsErrors` the compilation could fails because of a warning coming from a dependencies project.

kerams:
Damn, that's a good point - we're compiling every library in a null-aware context.

MangelMaxime:
> > Hopefully, if a code use | null syntax it will still be able to be processed by F#
> 
> Only with Fable 5 :/ Sadly, I don't think there's a way around this, and we'll just need to go update these projects.

Oh because otherwise, it will generate a syntax error right? 

Unless, the library author use `#if !NULLABLE` or `#if FABLE_COMPILER_4` to support non null able syntax.

This is unfortunate that this is causing issues in both direction...

MangelMaxime:
If we wanted to avoid this problem, we would probably have to explore the idea of compiling the libraries and the main project in 2 separates projects.

Like mentioned [here](https://github.com/fable-compiler/Fable/pull/2527#issuecomment-913621902)

I think we could have what is is needed to do that now, because Fable has the ability to use `pre-compiled` projects even if this is something I am not familiar with at all... And I am not sure how much work it would be to rework Fable to compile 2 projects instead of one.

Note: I believe I found a solution to support `TreatWarningsAsErrors` at Fable level without too much code change. Need to test it more

MangelMaxime:
@kerams I made a release with support for Nullness and TreatWarningsAsErrors. Feel free to play with it and share your feedback in discussion/issues.

This is more or less experimental but the goal is to understand how this plays out in Fable, what are the requirements for the usage etc.

I am trying to port a real world project to Nullness to see how it plays out and make a Fable blog post for it.

kerams:
Well, as expected, when I enable `Nullable`, I get errors from Fable packages, so the larger ecosystem needs to be updated before I can really enable it in my project.

MangelMaxime:
@kerams Once, I have it working for one of my project I will make a Fable blog post trying to explain the situations. 

So people can refer to it for asking libraries maintainers to make or accept PR adding support for `nullness`

PR #4073, opened 2025-03-12
[All] Add support for `TreatWarningsAsErrors`

MangelMaxime:
This PR tries to re-enable `TreatWarningsAsErrors`, this is really handy feature and becoming even more important with nullness supports to make nullness really strict.

We mimic the behavior at Fable level, because we can pass the option to FCS otherwise it will also report errors from libraries/packages code which the user have no control over.

https://github.com/fable-compiler/Fable/issues/2521

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -1354,7 +1354,7 @@ let startCompilationAsync state =
 
             match compilationResult.ExitCode with
             | 0 -> return Ok(state, logs)
-            | _ -> return Error("Compilation failed", [||])
+            | _ -> return Error("Compilation failed", logs)
 
         with
         | Fable.FableError e -> return Error(e, [||])
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -338,6 +338,7 @@ type ProjectCracked(cliArgs: CliArgs, crackerResponse: CrackerResponse, sourceFi
 
     member _.FableLibDir = crackerResponse.FableLibDir
     member _.FableModulesDir = crackerResponse.FableModulesDir
+    member _.TreatmentWarningsAsErrors = crackerResponse.TreatWarningsAsErrors
 
     member _.MakeCompiler(currentFile, project, ?triggeredByDependency) =
         let opts =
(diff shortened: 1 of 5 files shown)


src/Fable.Cli/Main.fs
```diff
@@ -1139,8 +1140,24 @@ let private compilationCycle (state: State) (changes: ISet<string>) =
...
+                |> fun logs ->
+                    if projCracked.TreatmentWarningsAsErrors then
+                        logs
+                        |> Array.map (fun log ->
+                            match log.Severity with
+                            | Severity.Warning -> { log with Severity = Severity.Error }
+                            | _ -> log
+                        )
+                    else
+                        logs
```
MangelMaxime, line 1160: I think I applied this code too early, we need to ignore warning coming from the `fable_modules` (see for loop below)

MangelMaxime:
If we wanted to add integration tests for this feature we would need to make a `NuGet` package with warnings inside of it to check that do ignore them and don't report them as errors.

The previous implementation could have been tested the same way, so for now for simplicity I think we can avoid going down that path for simplicity. If later, we found issues with the implementation or want to prevent regression we can allocate some time to try crafting a test suit for both warnings and errors output of Fable.

MangelMaxime:
@nojaf @Freymaurer @jkone27

I believe you will need to port this code into `vite-plugin-fable`. 

As support for `TreatWarningsAsErrors` has been tried several times in the past and disabled, if I were you I would wait before porting it to make sure that this implementation is correct and stable.

Freymaurer:
Let me know when! 🙂

src/Fable.Cli/Main.fs
```diff
@@ -1140,31 +1141,49 @@ let private compilationCycle (state: State) (changes: ISet<string>) =
                 }
 
             // Sometimes errors are duplicated
-            let logs = Array.distinct logs
+            let logs = logs |> Array.distinct
```
kerams, line 1144: :)

MangelMaxime: 🙈

This is a left over of my initial implementation, I have it reverted locally

MangelMaxime:
I am going to merge this PR as is, because more "real world testing" is needed to gather feedback.

There are several things that could be improve once we know it works:

1. Avoid doing the array concatenation at the end of `startCompilationAsync` 
2. Adapt the integration tests, and we should probably simplify then if possible

Issue #4075, opened 2025-03-14
[JS] `[<JS.Pojo>]` defined outside file scope creates a non existent import on use


shayanhabibi:
Fable 5.0.0-alpha-11
.net 9.0

# Question of Defined or Undefined Behaviour

**Without** the `[<Global>]` attribute, the example listed below causes an error.
While it is expected to create an interface in TS, there is no codegen, nor should there be, for JS/JSX. However, an import is still defined. My original understanding of `[<Pojo>]` is that it was intended to replace the pattern of `[<Global>]` `[<ParamObject; Emit("$0")>]`.
Is this considered ~~defined~~documented, or ~~undefined~~undocumented behaviour?

Depending on this, I can proceed with a pull when time allows.

A [minimal example](https://github.com/shayanhabibi/Partas.Solid/issues/10):

```fs
# file 1
namespace TestPojoIssue

open Fable.Core.JS

[<Pojo>]
type Toast
    (
        ?id : string,
        ?title : string,
        ?type' : string,
        ?icon : string,
        ?jsx : string,
        ?richColors : bool
    ) =
    member val id = id
    member val title = title
    member val type' = type'
    member val icon = icon
    member val jsx = jsx
    member val richColors = richColors
```

```fs
# file 2
module ConsoleApp2.Program2

open TestPojoIssue

let importsPojo = Toast("Test pojo import")
```

Output:

```js
import { Toast } from "./Program.fs.jsx";

export const importsPojo = {
    id: "Test pojo import",
};
```

No file or export by that name/path exists.

Hurt my head finding this as I originally thought it was my plugin causing the issue hahaha.

MangelMaxime:
This is a bug, when targeting JavaScript we should not emit the `import` statement.

However, if we target TypeScript we want the `import` statement because in this case we will have a concrete type.

PR #4076, opened 2025-03-14
[<>TS] File declaration for POJOs not required outside of TS

shayanhabibi:
@MangelMaxime This works locally, but it's a very blanketed solution. I couldn't get any targeted approaches to work (in trying to treat pojos outside of TypeScript like globals in select transformers). I'm sure alf might have a better idea though of where to target

#4075


MangelMaxime:
Thank you

PR #4096, opened 2025-04-18
[Python] Fix code and format warnings

dbrattli:
- Format and sort `__slots__` and `__all__`
- Replace soon to be deprecated `datetime.utcnow`
- Update ruff
- Better type annotations for the option module
- Use f-strings instead of format specifiers
- Fix dev dependency group in pyproject.toml

No new functionality or bug fixes. Just formatting and fixes for some deprecation warnings. No external changes, so unsure if we need a changelog entry?


MangelMaxime:
Happy to see you back @dbrattli 

If this is all internal stuff and nothing exposed to the user we can skip the changelog entry yes

PR #4090, opened 2025-03-25
[Python] Reenable type aliasing for imports with name "*" #4089

Freymaurer:
Closes #4089 

@MangelMaxime

PR #4100, opened 2025-04-23
Fix import path handling for Python

### LINKED ISSUE #4088: Fable 5.0.0-alpha.10 introduced path bug in python dependencies
I have a fable compatible library with a test project. I encountered an error using this library with the newest fable alpha version, so i wanted to update the library. In this case from `4.19.3` to `5.0.0-alpha.12`. The error i wanted to fix was gone, but instead i got a new error related to **python import paths in fable dependencies**.

The test project has a dependency to `Fable.Pyxpecto` so my `fable_modules` looks like this:

```
fable_modules
|-- fable_library
|-- fable_python
|-- fable_pyxpecto
```

And the error occurs in `fable_modules/fable_pyxpecto/pyxpecto.py`. I copied the relevant parts from each version down below. 

The imports in Fable.Pyxpecto are incorrectly done from the same folder level, instead of from the parent folder (see the code from 4.19.3).

### Fable v5.0.0-alpha.9

```py
from typing import (Any, TypeVar)
from ..fable_library.array_ import (choose, contains as contains_1, append as append_2)
from ..fable_library.async_ import run_synchronously
from ..fable_library.async_builder import (Async, singleton as singleton_1)
from ..fable_library.date import (now, op_subtraction, to_string as to_string_2)
from ..fable_library.fsharp_core import Operators_Using
```

### Fable v5.0.0-alpha.10

```py
from typing import (Any, TypeVar)
# wrong as the import should be from parent folder
from .fable_library.array_ import (choose, contains as contains_1, append as append_2) 
from .fable_library.async_ import run_synchronously
from .fable_library.async_builder import (Async, singleton as singleton_1)
from .fable_library.date import (now, op_subtraction, to_string as to_string_2)
from .fable_library.fsharp_core import Operators_Using
```

dbrattli:
Fix import path fix handling for Python. 

Fixes #4088


Freymaurer:
@dbrattli thanks for looking into this! 

Maybe you can add some tests for this, to ensure correct behavior? I was able to reproduce this with the setup in my pr https://github.com/fable-compiler/Fable/pull/4090/files. The changes done in `tests/Python/TestDependencyImport.fs` show a test, which is commented out ("check correct transpilation of subdependency"), which previously failed. Maybe you could add this test to see if it works

dbrattli:
@Freymaurer This PR works with your test (check correct transpilation of subdependency) but I cannot use it until we have merged your PR since I need that `import json as json_1` fix before I can use it correctly.

MangelMaxime:
@dbrattli There is a test not related to the changes in this PR failing.

This is not the first time, I see this specific test fails on CI and work on local machine. Any ideas?

dbrattli:
@MangelMaxime Looks weird. Should not be related to this PR, but I'll have a look at it 🤔

dbrattli:
Added a fix in #4105 that should help

PR #4104, opened 2025-04-24
allow RegexOptions.NonBacktracking

ieviev:
This lets you use RegexOptions.NonBacktracking in combination with RegexOptions.ECMAScript as the [js equivalent linear `l` flag](https://v8.dev/blog/non-backtracking-regexp) for regex.

`l` is limited but has vastly better performance characteristics for e.g., IsMatch

A small example for comparison:

```fs
open Fable.Core.JS

let linear_regex(s: string) =
    Regex($"(a+)*b$", RegexOptions.NonBacktracking ||| RegexOptions.ECMAScript).Match(s)

let backtracking_regex(s: string) = Regex($"(a+)*b$").Match(s)

let input = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaba"

let measurefn fn =
    let start = Constructors.Date.now ()
    let _ = fn ()
    let stop = Constructors.Date.now ()
    console.log ($"time taken {stop - start} ms")

console.log "linear:"
measurefn (fun v -> linear_regex (input))

console.log "backtracking:"
measurefn (fun v -> backtracking_regex (input))
```
tmp

the `l` flag is not supported together with `u` in JS so i'm unsure if RegexOptions.NonBacktracking should be allowed by itself.

### DIFF
--- a/src/Fable.Build/Test/JavaScript.fs
+++ b/src/Fable.Build/Test/JavaScript.fs
@@ -12,7 +12,7 @@ let private mainTestSourceDir = Path.Resolve("tests", "Js", "Main")
 
 let private mainTestProject = Path.Resolve("tests", "Js", "Main", "Fable.Tests.fsproj")
 
-let private mocha_bin = Path.Resolve("node_modules", ".bin", "mocha")
+let private mochaBin = Path.Resolve("node_modules", ".bin", "mocha")
 
 let private testReact(isWatch: bool) =
     let workingDirectory = Path.Resolve("tests", "React")
@@ -48,7 +48,7 @@ let private testAdaptive(isWatch: bool) =
         CmdLine.empty
         |> CmdLine.appendRaw "node"
         |> CmdLine.appendRaw "--enable-experimental-regexp-engine"
-        |> CmdLine.appendRaw mocha_bin
+        |> CmdLine.appendRaw mochaBin
         |> CmdLine.appendRaw destinationDir
         |> CmdLine.appendPrefix "--reporter" "dot"
         |> CmdLine.appendPrefix "-t" "10000"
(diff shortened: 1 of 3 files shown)


MangelMaxime:
Hello @ieviev,

Thank you for the PR could you please some tests too?

They can be placed in https://github.com/fable-compiler/Fable/blob/main/tests/Js/Main/RegexTests.fs

MangelMaxime:
> This feature is locked behind '--enable-experimental-regexp-engine' for node or a new version of v8, i'm not sure what to do about the tests here

Do you know which version of Node has support for it out of the box?

> This means RegexOptions.NonBacktracking has different unicode behavior in .NET and JS

I am not sure of the exact impact, but in general if the behavior is similar in JS and .NET we are okay with it. In some cases, we are also okay with some small difference and document them on Fable docs for example.

For regex, the [current docs](https://fable.io/docs/javascript/compatibility.html#caveats) says (it could be outdated):

> Regex will always behave as if passed RegexOptions.ECMAScript flag (e.g., no negative look-behind or named groups).

ieviev:
> Do you know which version of Node has support for it out of the box?

According to this `--enable-experimental-regexp-engine` should be available for nodejs versions 16 and up
https://github.com/nodejs/node/issues/38297

Maybe this could be passed into the mocha test runner?

> Regex will always behave as if passed RegexOptions.ECMAScript flag (e.g., no negative look-behind or named groups).

That's very good, if the regex is always assumed to have ECMAScript in js then there should be no problem.

MangelMaxime:
> > Do you know which version of Node has support for it out of the box?
> 
> According to this `--enable-experimental-regexp-engine` should be available for nodejs versions 16 and up [nodejs/node#38297](https://github.com/nodejs/node/issues/38297)

There is no version of Node with support for it out of the box?

> Maybe this could be passed into the mocha test runner?

Because, we don't specify a version for Node in Github CI, I suppose it is using LTS so we should be good and just have to provide the flag to mocha test runner like you proposed.

ieviev:
> There is no version of Node with support for it out of the box?

There's many sources hinting it will be enabled by default some time in both node and chrome as well (https://zenodo.org/records/10806044) but i have not seen it enabled anywhere by default yet, i think `google-chrome --js-flags=--enable-experimental-regexp-engine` or `node --enable-experimental-regexp-engine` is required currently. I'm uncertain when this will become the default

MangelMaxime:
Adding the flag is good enough to me

PR #4107, opened 2025-04-29
[Python] Typematching against an Interface leads to incorrect pattern matches

### LINKED ISSUE #3972: Python: typematching against an Interface leads to incorrect pattern matches
### Description

It seems like, when typematching against an interface, incorrect match result code is generated.

### Repro code

See [REPL](https://fable.io/repl/#?code=C4TwDgpgBAkjB2wICcBmBDAxtAvAWACgpip0AjAZ2GS2CgFsJ6yUoAZAeTagC4oBLRIUIAbCHVCQAsumCYAFgGUA9owBCygB4QAJhzIArCJjoAKZT2WGAlFHxES9WQqjKoAd37B5hElAA+vAD8AohQALQAfFAAjL4kgTwhcIgoGNgR0QBM8cSBAPqZUADMwgRidDp2UKaSEDJySqoQGtp6hsZ0AEQiyiJd1mVgyILAqPBQXQCkAIJdUDpAA&html=Q&css=Q) 

or transpile this code:

```fsharp
type IInterface =
    abstract member LOL : int

let typeMatchSomeBoxedObject (o:obj) =
    match o with
    | :? int -> 1
    | :? IInterface -> 2
    | _ -> 3

let d = (typeMatchSomeBoxedObject "lol")

printfn "%A" d
```

the resulting python code:

```python
from abc import abstractmethod
from typing import (Protocol, Any)
from fable_modules.fable_library.decimal_ import from_parts
from fable_modules.fable_library.string_ import (to_console, printf)
from fable_modules.fable_library.types import uint8

class IInterface(Protocol):
    @property
    @abstractmethod
    def LOL(self) -> int:
        ...

def type_match_some_boxed_object(o: Any=None) -> int:
    if str(type(o)) == "<class \'int\'>":
        return 1
    else: 
        return 2

d: int = type_match_some_boxed_object(from_parts(2, 0, 0, False, uint8(0)))

to_console(printf("%A"))(d)
```

Expected: A type that does not implement the interface does not match the second match case and therefore 3 is returned via wildcard match case. This is what happens when executing the F# code

Actual: 2 is returned, as the wildcard match case is not generated

### Related information

* Fable version: 4.24.0

dbrattli:
## Why

- Typematching against an Interface leads to incorrect pattern matches

Note: This PR does not allow you to correctly type test against interfaces. This still produces a warning (as with JS), and may be fixed in some future PR.

Fixes #3972

## How

- Else-branch was skipped in some cases when transforming if-statement. Fixed by aligning code with `Fable2Babel`.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* [Python] Fixed testing with interfaces leads to incorrect pattern match (#3972) (by @dbrattli)
 * [Python] Fixed import path handling for libraries (#4088) (by @dbrattli)
 * [Python] Reenable type aliasing for imports with name "*" (by @freymauer)
 
--- a/src/Fable.Transforms/Python/Fable2Python.fs
+++ b/src/Fable.Transforms/Python/Fable2Python.fs
@@ -2306,7 +2306,14 @@ module Util =
         let expr, stmts = com.TransformAsExpr(ctx, guardExpr)
 
         match expr with
-        | Constant(value = BoolLiteral value) -> stmts @ com.TransformAsStatements(ctx, ret, thenStmnt)
+        | Constant(value = BoolLiteral value) ->
+            let e =
+                if value then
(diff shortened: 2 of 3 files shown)


MangelMaxime:
Thank you

PR #4115, opened 2025-05-08
[JS/TS] JSX : Alias `empty` CEs list to `null` when encountered in the `children` list

### LINKED ISSUE #4113: [JSX] JSX.create transpiles list CE (similiar to for...loop) when using conditional children, making all children dynamic
# Issue

The following code uses a simple condition to determine if some element should be shown or not. It produces a rather complicated transpilation of the list CE, which in turn makes all children dynamic again (similiar to #4079).

Instead i would expect passing static children as static and creating a small conditional statement with established jsx syntax, as shown below.

```fsharp
open Fable.Core

[<JSX.ComponentAttribute>]
let Compponent2(show: bool) =

  JSX.create "div" [
    "children", [
      if show then
        "I show sometimes!"

      "I show always!"
    ]
  ]
```

Output:

```jsx
export function Compponent2($props) {
    return <div>
        {toList(delay(() => append($props.show ? singleton("I show sometimes!") : empty(), delay(() => singleton("I show always!")))))}
    </div>;
}
```

Expected output:

```jsx
export function Component({show}) {
    return <div>
      {show && "I show sometimes!"} // or `{show ? "I show sometimes!" : null}`
      "I show always!"
    </div>;
}
```

MangelMaxime:
Fix #4113

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [Python] Fable Library for Python is now partially written in Rust (by @dbrattli)
 
+### Fixed
+
+* [JS/TS] JSX : Alias `empty` CEs list to `null` when encountered in the `children` list (by @MangelMaxime)
+
 ## 5.0.0-alpha.13 - 2025-05-04
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


MangelMaxime:
Hello @shayanhabibi, 

I am pinging you because I think in your project you are also missing this case.

shayanhabibi:
Thank you!

Huh, I just realised that I won't need to include the logic anymore with these optimisations right? 👯

MangelMaxime:
> uh, I just realised that I won't need to include the logic anymore with these optimisations right? 👯

Indeed, you won't need it if you output `JSX.create` API.

PR #4121, opened 2025-05-11
[Python] Migrate to uv package manager

dbrattli:
## Why

Poetry can be quite slow so migrating to uv for Python package management. A single tool to replace pip, pip-tools, pipx, poetry, pyenv, twine, virtualenv, and more.

https://docs.astral.sh/uv/


src/Fable.Cli/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* [Python] Use [uv](https://docs.astral.sh/uv/) instead of Poetry for package management (by @dbrattli)
 * [Python] Fable Library for Python is now partially written in Rust (by @dbrattli)
```
MangelMaxime, line 14: @dbrattli 

When updating Changelog entries we also need to update [Fable.Compiler/CHANGELOG.md](https://github.com/fable-compiler/Fable/blob/main/src/Fable.Compiler/CHANGELOG.md).

This is true for features not specific to Fable.Cli so anything not related the CLI parser.

dbrattli: Thanks for the heads-up. I'll add an entry there as well.

PR #4127, opened 2025-06-08
[Python] Do not return None | None for optional unit types

### LINKED ISSUE #4125: [Fable.Python] FsToolkit.ErrorHandling error "TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'"
👋 

Given the following piece of code:

src:

```fsharp
    let ignore<'ok, 'error>
        (value: Async<Result<'ok option, 'error>>)
        : Async<Result<unit option, 'error>> =
        value
        |> map ignore<'ok>

```

tests: 
```fsharp

let ignoreTests =
    testList "AsyncResultOption.ignore tests" [
        testCaseAsync "ignore with Async(Ok Some(x))"
        <| (getUserById sampleUserId
            |> AsyncResultOption.ignore
            |> Expect.hasAsyncOkValue (Some()))

        testCaseAsync "ignore with Async(Ok None)"
        <| (getUserById (UserId(Guid.NewGuid()))
            |> AsyncResultOption.ignore
            |> Expect.hasAsyncOkValue None)

        testCaseAsync "ignore with Async(Error x)"
        <| (getUserById (UserId(Guid.Empty))
            |> AsyncResultOption.ignore
            |> Expect.hasAsyncErrorValue "invalid user id")

        testCase "can call ignore without type parameters"
        <| fun () -> ignore AsyncResultOption.ignore

        testCase "can call ignore with type parameters"
        <| fun () ->
            ignore<Async<Result<int option, string>> -> Async<Result<unit option, string>>>
                AsyncResultOption.ignore<int, string>
    ]
```

Getting the error: 

`TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'`

It generates to this:

```py
def AsyncResultOption_ignore(value: Async[FSharpResult_2[Any | None, Any]]) -> Async[FSharpResult_2[None | None, Any]]:
    def binder(x_0027: FSharpResult_2[_OK | None, _ERROR], value: Any=value) -> Async[FSharpResult_2[None | None, _ERROR]]:
        value_2: FSharpResult_2[None | None, _ERROR]
        input_4: FSharpResult_2[_OK | None, _ERROR] = x_0027
        def _arrow131(__unit: None=None, x_0027: Any=x_0027) -> None | None:
            input_2: _OK | None = input_4.fields[0]
            return None if (input_2 is None) else some(ignore(value_4(input_2)))

        value_2 = FSharpResult_2(1, input_4.fields[0]) if (input_4.tag == 1) else FSharpResult_2(0, _arrow131())
        return singleton.Return(value_2)

    return singleton.Bind(value, binder)

```

It seems the `unit option` is creating a situation where it a generates `None | None` union.

---

Fable dotnet tool: 4.25.0
Fable.Python package: 4.6.1

dbrattli:
Fixes #4125

@MangelMaxime Should we backport this to Fable 4?

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+* [Python] Do not return None | None for optional unit types (#4127) (by @dbrattli)
+
 ### Changed
 
 * [Python] Use [uv](https://docs.astral.sh/uv/) instead of Poetry for package management (by @dbrattli)
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -7,8 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-* [Python] Do not return None | None for optional unit types (#4127) (by @dbrattli)
-
 ### Changed
 
 * [Python] Use [uv](https://docs.astral.sh/uv/) instead of Poetry for package management (by @dbrattli)
@@ -16,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* [Python] Do not return None | None for optional unit types (#4127) (by @dbrattli)
 * [JS/TS] JSX : Alias `empty` CEs list to `null` when encountered in the `children` list (by @MangelMaxime)
 * [JS/TS] JSX : Allow usage of `unbox` when definining properties for `JSX.create` (by @MangelMaxime)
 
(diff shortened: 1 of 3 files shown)


Happypig375:
How will you ensure `Some () <> None`?

MangelMaxime:
@dbrattli In general, we don't backport things to Fable 4 because it is difficult to do.

In the past, when seeing that Fable 5 was taking longer than plan to release I tried to back port all the fixes and features made in Fable 5 and it was not an easy task.

People should be able to use Fable 5, the only reason we keep it as alpha right now is because of `nullness` and the work done for JSX support because they can introduce modification to Fable.AST which means plugins needs to upgrade too

dbrattli:
> How will you ensure `Some () <> None`?

@Happypig375 Fable erases optional types except for `Some ()` that is transpiled and wrapped to `some(None)`. However, Python currently annotates this as `_T | None` but this misses the wrapper case 🙈 I'll start annotating as `Option[_T]` instead. Thanks for the hint 😊

PR #4136, opened 2025-06-18
improve filewatcher error messages

goswinr:
I just had this Exception, It took me a while to figure out what the problem was, so I thought I improve the error message

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -243,7 +243,9 @@ module FileWatcherUtil =
                         dir
                     else
                         match IO.Path.GetDirectoryName(dir) with
-                        | null -> failwith "No common base dir, please run again with --verbose option and report"
+                        | null ->
+                            let dirList = restDirs |> String.concat $"{Environment.NewLine} - "
+                            failwith $"No common base directory was found in:{Environment.NewLine} - {dirList}{Environment.NewLine}This might be because referenced repositories are spread over several local drives, please run again with --verbose option and report"
                         | dir -> getCommonDir dir
 
                 getCommonDir dir


src/Fable.Cli/Main.fs
```diff
@@ -243,7 +243,9 @@ module FileWatcherUtil =
                         dir
                     else
                         match IO.Path.GetDirectoryName(dir) with
-                        | null -> failwith "No common base dir, please run again with --verbose option and report"
+                        | null ->
+                            let dirList = restDirs |> String.concat $"{Environment.NewLine} - "
+                            failwith $"No common base directory was found in:{Environment.NewLine} - {dirList}{Environment.NewLine}This might be because referenced repositories are spread over several local drives, please run again with --verbose option and report"
```
MangelMaxime: What do you think of a more generic message which gives more context:

*Pseudo script I created to look at the output*

```fs
open System

let directories =
    [
        "/home/workspace/projectA"
        "/home/workspace/projectB"
        "/home/workspace/projectC"
        "/dev/workspace/projectSharedA"
    ]

let dir'= "/home/workspace/"
let dirList = directories |> String.concat $"{Environment.NewLine}- "

printfn $"""No common base directory was found was looking for '%s{dir'}' but directories were:

- %s{dirList}

If you think this is a bug, please run again with --verbose option and report"""

```

gives

```text
No common base directory was found was looking for '/home/workspace/' but directories were:

- /home/workspace/projectA
- /home/workspace/projectB
- /home/workspace/projectC
- /dev/workspace/projectSharedA

If you think this is a bug, please run again with --verbose option and report
```

This is because I am not sure if making the error really specific is a good idea, but if there is another reason for the problem we will need to add to the error message and so on.

MangelMaxime (changes requested):

goswinr:
@MangelMaxime is this better? I just saw there was even a TODO comment  "to find outlier paths"

src/Fable.Cli/Main.fs
```diff
@@ -242,8 +241,27 @@ module FileWatcherUtil =
...
+                                )
+
+                            [
+                                "Fable is trying to find a common base directory for all files and projects referenced."
+                                $"But '%s{dir'}' is not a common base directory for these source paths:"
+                                for d in badPaths do
+                                    $" - {d}"
+                                // "These path are OK:"
+                                // for d in goodPaths do
+                                //     $" - {d}"
```
MangelMaxime: Please remove this comment

src/Fable.Cli/Main.fs
```diff
@@ -242,8 +241,27 @@ module FileWatcherUtil =
...
                     else
-                        match IO.Path.GetDirectoryName(dir) with
-                        | null -> failwith "No common base dir, please run again with --verbose option and report"
+                        match IO.Path.GetDirectoryName(dir) with // 'dir' is empty string ?
+                        | null ->
+                            let goodPaths, badPaths =
+                                restDirs
+                                |> List.partition (fun d ->
+                                    (withTrailingSep d).StartsWith(dir', StringComparison.Ordinal)
+                                )
```
MangelMaxime: Because we don't use `goodPaths` this can be simplified to a `List.filter`

src/Fable.Cli/Main.fs
```diff
@@ -242,8 +241,27 @@ module FileWatcherUtil =
...
+
+                            [
+                                "Fable is trying to find a common base directory for all files and projects referenced."
+                                $"But '%s{dir'}' is not a common base directory for these source paths:"
+                                for d in badPaths do
+                                    $" - {d}"
```
MangelMaxime:
```suggestion
                                    $"- {d}"
```

or

```suggestion
                                    $"    - {d}"
```

MangelMaxime:
@goswinr All good to me, I left a few comments to do the clean up.

Logging only the bad paths is good idea to make the error message shorter, even if it can be really long depending on the project set up 😅

goswinr:
I added` |>  List.truncate 20` that should be enough to get an idea about where the errors are

MangelMaxime:
Thank you

PR #4142, opened 2025-06-20
Fix directory delete when building Fable.Libaray

dbrattli:
Fixes error that happens on every other build on MacOS. Trying to build again always works, so there's some timing issue or something that happens when trying to delete the build directory.

```
Cleaning build directory
System.IO.IOException: Directory not empty : '/Users/dbrattli/Developer/Fable-pyo3/temp/fable-library-py'
   at System.IO.FileSystem.RemoveEmptyDirectory(String fullPath, Boolean topLevel, Boolean throwWhenNotEmpty)
   at System.IO.FileSystem.RemoveDirectoryRecursive(String fullPath)
   at Build.FableLibrary.BuildFableLibrary.Run(FSharpOption`1 skipIfExist) in /Users/dbrattli/Developer/Fable-pyo3/src/Fable.Build/FableLibrary/Core.fs:line 63
   at Build.Test.Python.handle(FSharpList`1 args) in /Users/dbrattli/Developer/Fable-pyo3/src/Fable.Build/Test/Python.fs:line 17
   at Build.Main.main(String[] argv) in /Users/dbrattli/Developer/Fable-pyo3/src/Fable.Build/Main.fs:line 126
```

FYI: @MangelMaxime

### DIFF
--- a/src/Fable.Build/FableLibrary/Core.fs
+++ b/src/Fable.Build/FableLibrary/Core.fs
@@ -48,7 +48,8 @@ type BuildFableLibrary
     default _.CopyStage() = ()
 
     /// <summary>
-    /// Robustly delete a directory, handling macOS file system timing issues
+    /// Robustly delete a directory, handling file system issues on MacOS and Windows.
+    /// Ref: https://stackoverflow.com/questions/329355/cannot-delete-directory-with-directory-deletepath-true
     /// </summary>
     member private this.deleteDirectoryRobust(path: string) =
         if Directory.Exists path then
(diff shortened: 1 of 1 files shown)


dbrattli:
Related discussion: https://stackoverflow.com/questions/329355/cannot-delete-directory-with-directory-deletepath-true

MangelMaxime:
Strangely I used MacOS to develop Fable for 2+ years without seeing this issue (or I don't remember).

CI is green and if it works for you I am good with the changes thank you

dbrattli:
Yes, same goes for me. Haven't never seen this error before. But it could be related to the Rust modules and that additional subdir within fable-library-py that triggered the problem

PR #4149, opened 2025-06-23
[Python] Do not run publish-pypi on pull requests

### LINKED ISSUE #4148: `Publish to PyPI` workflows are triggered on each PR
Hello @dbrattli,

It seems like `Publish to PyPI` workflows are triggered on each PR and not just on `main` branch.

Is it normal? I would expect it to be triggered on on `main` branch push because that's where we do release for Fable and I don't think releasing packages before a release is a good idea (for compatibility etc.)

It is possible that the workflow run without pushing anything to PyPi, but it would be better for them to not run at run to not use CI time and make PR CI go green or red faster?

What do you think?

dbrattli:
## Why

Running publish to pypi on every pull-request adds a lot of time for every PR being made. We should only run it on main merge.

Fixes #4148

## How

Remove `pull_request` from `publish-pypi.yml`.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
   workflow_dispatch:
   release:
     types: [published]


MangelMaxime:
Thank you

PR #4151, opened 2025-06-24
Documentation Refactor N1

shayanhabibi:
Iterative process.

Feel free to comment on sections or contribute.

### DIFF
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -41,6 +41,7 @@ open System
 /// <description><c>"mouseover"</c></description>
 /// </item>
 /// </list>
+/// <br/>
 /// You can also use <c>[&lt;CompiledName>]</c> and <c>[&lt;CompiledValue>]</c> to
 /// specify the name or literal of the union case in the generated code:
 /// <code lang="fsharp">
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -12,26 +12,32 @@ open System
 /// <list type="table">
 /// <item>
 /// <term><c>None</c></term>
+/// <term><c>"MouseOver"</c></term>
 /// <description><c>"MouseOver"</c></description>
 /// </item>
 /// <item>
 /// <term><c>LowerFirst</c></term>
+/// <term><c>"mouseOver"</c></term>
 /// <description><c>"mouseOver"</c></description>
 /// </item>
 /// <item>
 /// <term><c>SnakeCase</c></term>
+/// <term><c>"mouse_over"</c></term>
 /// <description><c>"mouse_over"</c></description>
 /// </item>
 /// <item>
 /// <term><c>SnakeCaseAllCaps</c></term>
+/// <term><c>"MOUSE_OVER"</c></term>
 /// <description><c>"MOUSE_OVER"</c></description>
 /// </item>
 /// <item>
 /// <term><c>KebabCase</c></term>
+/// <term><c>"mouse-over"</c></term>
 /// <description><c>"mouse-over"</c></description>
 /// </item>
 /// <item>
 /// <term><c>LowerAll</c></term>
+/// <term><c>"mouseover"</c></term>
 /// <description><c>"mouseover"</c></description>
 /// </item>
 /// </list>
@@ -47,8 +53,9 @@ open System
 /// </code>
 /// Generates:
 /// <code lang="js">
-/// export const eventType = "Abracadabra";
-/// export const magicPower = false;
+/// // Pseudocode. See the Fable docs for your target for actual transpiled code.
+/// eventType = "Abracadabra"
+/// magicPower = false
 /// </code>
 /// </remarks>
 /// <seealso href="https://fable.io/docs/javascript/features.html#caserules">
@@ -73,7 +80,7 @@ type CaseRules =
 /// </summary>
 /// <remarks>
 /// <para>
-/// If you are not planning to use an interface to interact with JS and want to have overloaded
+/// If you are not planning to use an interface to interact with native code and want to have overloaded
 /// members, you can decorate the interface declaration with the <c>Mangle</c> attribute.
 /// </para>
 /// <para>
@@ -96,34 +103,28 @@ type CaseRules =
 ///             failwith "Not implemented"
 /// </code>
 /// <para>
-/// Generates invalid JavaScript code. Note the two methods named <c>Render</c> in the same class.
+/// May be considered invalid code depending on the target. Note the two methods named <c>Render</c> in the same class.
 /// </para>
 /// <code lang="js">
-/// export class Renderer {
-///     constructor() {
-///     }
-///     Render() {
-///         throw new Error("Not implemented");
-///     }
-///     Render(indentation) {
-///         throw new Error("Not implemented");
-///     }
-/// }
+/// // Pseudocode
+/// class Renderer
+///     method Render
+///         Error("Not implemented")
+///
+///     method Render(indentation)
+///         Error("Not implemented")
 /// </code>
 /// <para>
 /// Using <c>Mangle</c>, we would instead generate:
 /// </para>
 /// <code lang="js">
-/// export class Renderer {
-///     constructor() {
-///     }
-///     "Program.IRenderer.Render"() {
-///         throw new Error("Not implemented");
-///     }
-///     "Program.IRenderer.RenderZ524259A4"(indentation) {
-///         throw new Error("Not implemented");
-///     }
-/// }
+/// // Pseudocode
+/// class Renderer
+///     "Program.IRenderer.Render"()
+///         Error("Not implemented")
+///
+///     "Program.IRenderer.RenderZ524259A4"(indentation)
+///         Error("Not implemented")
 /// </code>
 /// </remarks>
 /// <seealso href="https://fable.io/docs/javascript/features.html#mangle">
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -2,6 +2,9 @@ namespace Fable.Core
 
 open System
 
+// Note: The table list renders differently between Ionide on VS Code and Rider. Two term items
+// are used to render correctly on Ionide where the description is ignored. One term item and
+// one description item are used to render correctly on Rider where the second term is ignored.
 /// <summary>
 /// Parameter used in <c>StringEnum</c> and methods/other attributes that
 /// generate strings from unions to change the casing in conversion.
@@ -135,13 +138,74 @@ type MangleAttribute(mangle: bool) =
     inherit Attribute()
     new() = MangleAttribute(true)
 
-/// Used on a class to attach all members, useful when you want to use the class from JS.
+/// <summary>
+/// Used on a class to attach all members without mangling.
+/// </summary>
+/// <remarks>
+/// If you want to have all members attached to a class (as is standard with most target languages)
+/// and not-mangled, use the <c>AttachMembers</c> attribute. But be aware that overloads won't work
+/// in this case.
+/// </remarks>
+/// <seealso href="https://fable.io/docs/javascript/features.html#attachmembers">
+/// Fable Documentation
+/// </seealso>
 [<AttributeUsage(AttributeTargets.Class)>]
 type AttachMembersAttribute() =
     inherit Attribute()
 
-/// Used for erased union types and to ignore modules in JS compilation.
-/// More info: https://fable.io/docs/communicate/js-from-fable.html#erase-attribute
+/// <summary>
+/// Used for erased union types, erasing default constructors, and to ignore modules in Fable compilation.
+/// </summary>
+/// <remarks>
+/// <para>When placed on unions, the union code is erased, leaving the underlying fields as raw values</para>
+/// <code lang="fsharp">
+/// [&lt;Erase>]
+/// type ValueType =
+///     | Number of int
+///     | String of string
+///     | Object of obj
+/// [&lt;Global>] // no code generated for globally accessible methods etc
+/// let prettyPrint (value: ValueType) = nativeCode
+/// prettyPrint (Number 1)
+/// prettyPrint (String "Hello")
+/// prettyPrint (Object {|Name = "Fable"|})
+/// </code>
+/// Generates:
+/// <code lang="js">
+/// // Pseudocode.
+/// prettyPrint(1)
+/// prettyPrint("Hello")
+/// prettyPrint({ Name: "Fable" })
+/// </code>
+/// When used on <c>class</c> type declarations, member bindings et al, it will erase any associated
+/// target code relating to the construction, reflection, and the definition
+/// itself. Be aware that Fable will not stop you from making calls to these
+/// functions which may be undefined without replacement.
+/// <code lang="fsharp">
+/// type AnimationBuilder() = class end
+/// let anime = AnimationBuilder()
+/// [&lt;Erase>]
+/// type TimelineBuilder() = class end
+/// let tl = TimelineBuilder()
+/// </code>
+/// Generates:
+/// <code lang="js">
+/// // Pseudocode
+/// class AnimationBuilder
+/// method AnimationBuilder_$reflection()
+///     return class_type("Test.AnimationBuilder", null, AnimationBuilder)
+/// method AnimationBuilder_$ctor()
+///     return AnimationBuilder()
+/// anime = AnimationBuilder_$ctor()
+/// // no class definition or
+/// // constructor/reflection methods generated
+/// tl = TimelineBuilder_$ctor()
+/// // ^- undefined method potential
+/// </code>
+/// </remarks>
+/// <seealso href="https://fable.io/docs/javascript/features.html#erased-unions">
+/// Fable Documentation
+/// </seealso>
 type EraseAttribute() =
     inherit Attribute()
     new(caseRules: CaseRules) = EraseAttribute()
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -165,7 +165,7 @@ type AttachMembersAttribute() =
 ///     | String of string
 ///     | Object of obj
 /// [&lt;Global>] // no code generated for globally accessible methods etc
-/// let prettyPrint (value: ValueType) = nativeCode
+/// let prettyPrint (value: ValueType) = nativeOnly
 /// prettyPrint (Number 1)
 /// prettyPrint (String "Hello")
 /// prettyPrint (Object {|Name = "Fable"|})
--- a/src/Fable.Core/Fable.Core.Types.fs
+++ b/src/Fable.Core/Fable.Core.Types.fs
@@ -2,6 +2,58 @@ namespace Fable.Core
 
 open System
 
+/// <summary>
+/// Parameter used in <c>StringEnum</c> and methods/other attributes that
+/// generate strings from unions to change the casing in conversion.
+/// </summary>
+/// <remarks>
+/// Conversion of <c>MouseOver</c> with different enums (the qualified access is
+/// ignored for brevity):
+/// <list type="table">
+/// <item>
+/// <term><c>None</c></term>
+/// <description><c>"MouseOver"</c></description>
+/// </item>
+/// <item>
+/// <term><c>LowerFirst</c></term>
+/// <description><c>"mouseOver"</c></description>
+/// </item>
+/// <item>
+/// <term><c>SnakeCase</c></term>
+/// <description><c>"mouse_over"</c></description>
+/// </item>
+/// <item>
+/// <term><c>SnakeCaseAllCaps</c></term>
+/// <description><c>"MOUSE_OVER"</c></description>
+/// </item>
+/// <item>
+/// <term><c>KebabCase</c></term>
+/// <description><c>"mouse-over"</c></description>
+/// </item>
+/// <item>
+/// <term><c>LowerAll</c></term>
+/// <description><c>"mouseover"</c></description>
+/// </item>
+/// </list>
+/// You can also use <c>[&lt;CompiledName>]</c> and <c>[&lt;CompiledValue>]</c> to
+/// specify the name or literal of the union case in the generated code:
+/// <code lang="fsharp">
+/// [&lt;StringEnum>]
+/// type EventType =
+///     | [&lt;CompiledName("Abracadabra")>] MouseOver
+///     | [&lt;CompiledValue(false)>] RealMagic
+/// let eventType = EventType.MouseOver
+/// let magicPower = EventType.RealMagic
+/// </code>
+/// Generates:
+/// <code lang="js">
+/// export const eventType = "Abracadabra";
+/// export const magicPower = false;
+/// </code>
+/// </remarks>
+/// <seealso href="https://fable.io/docs/javascript/features.html#caserules">
+/// Fable Documentation
+/// </seealso>
 type CaseRules =
     | None = 0
     /// FooBar -> fooBar
@@ -15,8 +67,68 @@ type CaseRules =
     /// FooBar -> foobar
     | LowerAll = 5
 
+/// <summary>
 /// Used on interfaces to mangle member names. This allows overloading and prevents conflicts
 /// other interfaces, but will make interop with native code more difficult.
+/// </summary>
+/// <remarks>
+/// <para>
+/// If you are not planning to use an interface to interact with JS and want to have overloaded
+/// members, you can decorate the interface declaration with the <c>Mangle</c> attribute.
+/// </para>
+/// <para>
+/// Interface coming from .NET BCL (like System.Collections.IEnumerator) are mangled by default.
+/// </para>
+/// <para>
+/// For example the following code:
+/// </para>
+/// <code lang="fsharp">
+/// type IRenderer =
+///     abstract Render : unit -> string
+///     abstract Render : indentation : int -> string
+///
+/// type Renderer() =
+///     interface IRenderer with
+///         member this.Render() =
+///             failwith "Not implemented"
+///
+///         member this.Render(indentation) =
+///             failwith "Not implemented"
+/// </code>
+/// <para>
+/// Generates invalid JavaScript code. Note the two methods named <c>Render</c> in the same class.
+/// </para>
+/// <code lang="js">
+/// export class Renderer {
+///     constructor() {
+///     }
+///     Render() {
+///         throw new Error("Not implemented");
+///     }
+///     Render(indentation) {
+///         throw new Error("Not implemented");
+///     }
+/// }
+/// </code>
+/// <para>
+/// Using <c>Mangle</c>, we would instead generate:
+/// </para>
+/// <code lang="js">
+/// export class Renderer {
+///     constructor() {
+///     }
+///     "Program.IRenderer.Render"() {
+///         throw new Error("Not implemented");
+///     }
+///     "Program.IRenderer.RenderZ524259A4"(indentation) {
+///         throw new Error("Not implemented");
+///     }
+/// }
+/// </code>
+/// </remarks>
+/// <seealso href="https://fable.io/docs/javascript/features.html#mangle">
+/// Fable documentation
+/// </seealso>
 type MangleAttribute(mangle: bool) =
     inherit Attribute()
     new() = MangleAttribute(true)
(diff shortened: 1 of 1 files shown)


shayanhabibi:
@MangelMaxime I'm concerned that exhaustive/detailed documentation will just bury the codebase in fluff. However, I do think it is also helpful, and a mark of improvement. Could we side step this issue with `.fsi` files instead? What would the implications there be.

Edit:
From my understanding, we might be able to have documentation show through `.fsi` signature files without having them present in the implementation files, or even have implementation specific documentation for contribution and user/consumer specific documentation through the signature files. But this does make maintenance more difficult.

We would not be able to generate the signature files after the first iteration; I assume this would overwrite the documentation that would be added specifically to the signature files. Of course, this could be staged though, a contributor could generate the sig file, and then copy the new/changed sections to the version controlled sig file.

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -2,6 +2,58 @@ namespace Fable.Core
...
+/// </item>
+/// <item>
+/// <term><c>KebabCase</c></term>
+/// <description><c>"mouse-over"</c></description>
+/// </item>
+/// <item>
+/// <term><c>LowerAll</c></term>
+/// <description><c>"mouseover"</c></description>
+/// </item>
+/// </list>
```
MangelMaxime: I need to check why but the rendering of this element is broken on Ionide.

I think you wanted to render a table but that's not the case for Ionide

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -15,8 +67,68 @@ type CaseRules =
...
+///     constructor() {
+///     }
+///     Render() {
+///         throw new Error("Not implemented");
+///     }
+///     Render(indentation) {
+///         throw new Error("Not implemented");
+///     }
+/// }
+/// </code>
```
MangelMaxime: One potential issue I see with including target code, is that some of the code is not JavaScript specific.

Adding a version of the code for all the target don't seems do-able because of the work needed, and the tooltip becoming way too big ?

Should we consider linking to an external resource?

MangelMaxime:
> Could we side step this issue with `.fsi` files instead? What would the implications there be.

Using `.fsi` does means that the documentation is written in the `.fsi` instead of the `.fs` file, but there are several issues with `.fsi` files.

1. The DX is not good... when you try to navigate the code and do things like `Ctrl+Click` or `Go to definition/implementation`, you often ends up in the `.fsi` when you wanted the `.fs` file and vice versa
2. People are not well versed in `.fsi` and often intimidated which can make contribution harder
3. It means you need to do twice the same work sometime, like if you add a new API, new DU case, you need to write it at both places

I wrote `Fable.Form` using `.fsi` file originally and I decided to revert it because of these reasons.

Having better documentation is always something worth it, we just need to find the right balance.

cc @dbrattli any thought ?

src/Fable.Core/Fable.Core.Types.fs
shayanhabibi: Potentially the website docs having an 'API' like section which would have details and provide example for the behaviors in different targets for the types/interface (like how a JavaScript library would have documentation that targets multiple frameworks by a dropdown). This would be cumbersome to maintain. Can just refer to documentation and direct the user to check their target, but note taken for the target code. Should I avoid it all together? Make it 'pseudocode'?

src/Fable.Core/Fable.Core.Types.fs
shayanhabibi: Rider:

image

Visual Studio:

image

[MSDN Docs about the list xml tag](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#list)

Very confused. It seems like Visual Studio doesn't even support the recommended xml tags for the docs? Or am I missing something

shayanhabibi: If I do the following:

```fs
/// <term><c>None</c></term><term><c>MouseOver</c></term>
/// <description><c>"MouseOver"</c></description>
```

Then we get this on Rider:
image

And this on VSC (I'm not going to bother with Visual Studio)
image

So, if it's not worth fixing, then we can just hack it like this. by duplicating the description in a another term tag

src/Fable.Core/Fable.Core.Types.fs
MangelMaxime: > Potentially the website docs having an 'API' like section which would have details and provide example for the behaviors in different targets for the types/interface (like how a JavaScript library would have documentation that targets multiple frameworks by a dropdown). This would be cumbersome to maintain.

I was not thinking of something that complex. 

`API` like section for Fable, would look something like the one of [Fable.Promise](https://fable.io/fable-promise/reference/Fable.Promise/global-promise.html), I need to find the time to either fix F# Formatting or just do a small highly focused tool to generate nice API documentation from XML doc.

Coming back to the subject, when I said

> Should we consider linking to an external resource?

I was thinking, about redirecting people to Fable.io website which have a `Features` page for each target, with example/documentation about the code generation.

https://fable.io/docs/javascript/features.html#attachmembers

We could in theory, inject an intermediate page which ask the user to choose their `target` language and using JavaScript redirect them to the right page.

But I am not sure if the example would be good enough, plus if we rework the APIs of Fable.Core or something it would be a pain to maintain backward compatibility of the documentation etc.

We could just say something like `Learn more at https://fable.io/docs/` and the user will navigate the website themself

> Make it 'pseudocode'?

I don't think it would work, but the code between the language perhaps be quite different.

shayanhabibi: > > Make it 'pseudocode'?
> 
> I don't think it would work, but the code between the language perhaps be quite different.

Have a look at the recent commits to see how I've tried to provide examples using 'pseudocode'. If you don't reckon that will work, then we'll have to do without target example code and refer to the docs as you say.

I've just realised another annoying difference between Ionide and Rider is that the `seealso` href is not rendered as a link as it is in Rider.

Edit: In the end this is only an issue for the target agnostic types, so it's not a big deal.

Edit2:
Can use anchor tags (`<a>`) in the summary as that will render as a link in both Ionide and Rider

Edit3:
Alternatively, can link to a Fable Repl that has the example code and the user can compile between the different targets themselves

shayanhabibi:
With regards to target example codes for common members, what about linking to a Fable Repl with the example code so that the user can change the target as needed while still seeing how the member works?

An advantage would be that documentation examples et al can be updated independent of whether the fable.io docs have been or not. As an example, the interaction of Erase with anything other than a Union isn't mentioned in the docs on fable.io, but we can easily quickly add a reference to it in the docs and link to a repl with the example.

[An example would be this example code:](https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwChcDaAPAKJZwDOYAfALq4AuAnuFAGpxICuYAKi2CgBeXFDFQAPlAByXALYIwWKMABmUAJYQGo8VIDKDLFoDmK9RSOndYqQHkEAKzABjBuZVPCRAOJJgCJx0uCjuIDgMzAAKxtpQABQAbpw8AFzsKXwCAJTCUBBwDBqJYHYQSEz44WCRTDFa7vGyCkpQAIzZuNW19XHxhrFmAEQAEmBI-kOd3dGxjQ7OblAA3lLScHKCQlBD8MhgQ5IAvp2MAlAAghAacoUawBAAQlwaSAAmSvG52y5IlBRQSBvEI1KBwa6bPJXG53B7PV4fLBffDEMiUGj0ZisXg3cZaMDw96fb5QX7-QEQYGhKAMJB5HGbJD4wmIr5AA&html=DwQgJg9gxgLgngBwKYAIAWMC2AbAfAKGAxwMLSQEMwCUVgYBLGbJXAUQA8olsUBmAGwBWFAwB2KChLQQArgCdgAekbNW+WsExIYFdDBgIAtEgCOshgDcAvAHIAwhDEwkzowBVESWyihOXznYuHDBKxNgA3L5oFPIAzjrWsjAAZkYAHLY0dNq6KGIU2tYARJYMSADuCBDyMMW+-q4wJRUMYDBo1mBIZdxGre1oADSiYkwMFNhGcVCTSNYAjMXZwDPyDAgwKHHyUCUYhnEAXEpKUGBiAHQAVnHd2Fbyl2I6SmIImEpltbKTRpCYAACACZLgswUowAw4qFvjBflMATc4vUoPIIHE4jUGABzcS4ZRrDYwFYwuAsbK0ABGEDAcBQAG8NLQWQgqFCxDijighAAGBAcCLMlkoFL+IwpQoMbBwblxKRxaZIdYpIUigC++GFKF0VJYjO11Jq3Xk0zZUHEXJQ-MFhpQNPkJqMNIMEEw3IWAu2EAeYBQAGIkLyg0G1SL7cblUZ1jiMB6vVjfQGQ8HeWHaJrtTAwEMs2gDeGUAMOty+QL0yyHU6WCkYPGON6k4HU6G7VWozAIAh6422smW2m7WywByrUJy3aIJZlSlsBAKty0G1umIK7RgjAjFOZ3OFygeA8EHFoRXMyLs5cEixYEg-UzC8PR9ybaetee-eIEMkC+Hi2hubM2BQAAFAsvK8gApCgRg8gKACUa7oEguJxigwLAhOhaYLEeJiM+mEspqmhKGSFKEGElDUFqwA0nSKxQpYohgCU2HiMsygMQQyi0XApDKOEBBAA&css=Q)

```fs
open Fable.Core

[<Erase>]
type ValueType =
    | Number of int
    | String of string
    | Object of obj
[<Global>]
let prettyPrint (value: ValueType) = nativeOnly

prettyPrint (Number 1)
prettyPrint (String "Hello")
prettyPrint (Object {| Name = "Fable" |})

type AnimationBuilder() = class end
let anime = AnimationBuilder()

[<Erase>]
type TimelineBuilder() = class end
let tl = TimelineBuilder()
```

src/Fable.Core/Fable.Core.Types.fs
MangelMaxime: > I've just realised another annoying difference between Ionide and Rider is that the seealso href is not rendered as a link as it is in Rider.

I discovered that

Not to give me an excuses, but this is because the first result when searching `XML doc comment seealso` link to a page without support for `href`...

I am looking into adding it to Ionide because when looking at another XML doc comment page `href` is documented... 

So consider it supported

dbrattli:
I'm in favor of better documentation 🥰 However, there are some things about this PR that worries me:

1. Having JS specific examples will be confusing and feel wrong for those of us that use other targets, and pseudo code will most likely just confuse AI to give both incorrect suggestions and completions. Links to https://fable.io/docs/ can be good, but I'm not in favor of language specific links, i.e https://fable.io/docs/javascript/features.html#erased-unions
2. Maintainability. When making changes to Fable.Core it will be significantly more workload for us to keep examples up-to-date. Especially when there are external links that needs to be updated as well, and in multiple locations.

shayanhabibi:
> 1. Having JS specific examples will be confusing and feel wrong for those of us that use other targets, and pseudo code will most likely just confuse AI to give both incorrect suggestions and completions. Links to https://fable.io/docs/ can be good, but I'm not in favor of language specific links, i.e https://fable.io/docs/javascript/features.html#erased-unions

Fair, can just reduce the documentation to provide F# example code of usage and link to the homepage for the docs.

Keep in mind there are already dead links in the docs, and I have mostly just included the already existing links into the xml notation.

> 2. Maintainability. When making changes to Fable.Core it will be significantly more workload for us to keep examples up-to-date. Especially when there are external links that needs to be updated as well, and in multiple locations.

The trouble here in my mind is that there is no fleshed out documentation source to begin with. The Fable Docs are not up to date, or are reduced with regards to the nuances of their usage (such as Erase being applicable to more than just unions).

My preference is to document the source code in full rather than the website, as the onus should be on contributors to update the documentation with their changes so that it is iteratively up-to-date rather than requiring larger efforts at major points.

If we were to instead delegate documentation to being 'in-full' on the website, then we would have the same issue of maintaining multiple pages covering the same material that have to be kept up to date by target language users. We are more likely to have contributions that don't get pushed to the docs, and then I fear there will be a slow build up of hidden features as we invariably miss some changes being added to the docs when it is updated. In saying that, how would the next major release be done with regards to the docs? It would be pretty gruesome if this type of process is heaped only onto the major contributors such as yourselves since the website documentation feels 'separated' from the implementation.

In saying that, I would defer to whatever informed choice is made. I will continue to do small iterations until a larger decision is made on some standard with which to finalise everything with.

Edit: I would also mention that at a certain point of familiarity, we are less likely to refer to the website. At this point, an incorrect tooltip will be more likely to prompt someone to make a small contribution in correcting the documentation, especially if they do not use the website docs. In this sense, I think it might invite more small contributions from seasoned users too. Just a thought

MangelMaxime:
I am sharing the feeling of @dbrattli 

Better documentation is always for the better, but in regards to Fable specificity I am wondering if the XML Doc is the right place for it.

I am wondering if the effort should not be put on https://fable.io/docs/, in the past I put a lot of effort making https://fable.io/docs/ relevant again, and finding a structure which works (it is not perfect of course). 

On the website, we are not limited about what we can do (beside following the current structure for now) as the user can narrow down what he is looking for. It also allows us to have a single place where to redirect users.

Something to note also, is that Fable.Core don't change that often so in general if we document a function / attribute once, it stays relevant for a (very) long time.

----------

If someone is willing to contribute to a PR here, I think they would be ok with sending a PR to Fable Docs repository and that's up to us as maintainer to ask them kindly to do so. If they offer documentation in this PR, we could also port it their manually. Updating the documentation is a straight forward process in general, the difficulty is in being concise enough but explicit too.

For example, so people like to write lengthy documentation but sometimes a short sentences with 1 example is all we need.

> In this sense, I think it might invite more small contributions from seasoned users too

I understand what you are trying to say, but from my experience in OSS we are often lucky if people open an issue. Creating a PRs is rare in general especially on complex project like Fable because it can be intimidating to work with.


------------

I think my stance currently is to keep XML doc really focused on a 1-2 sentences summary and use https://fable.io/docs/ for the complete explanation with edge cases etc.

shayanhabibi:
Sounds good, I'm happy with that, and you both have more experience with OSS so I'll definitely take that reasoning on-board.

I'll make a pull on the website in conjunction with this and iterate over the docs simultaneously.
I'll reduce the remarks to essentials, such as hinting towards related attributes/functions, or relevant considerations/examples. Will append the XML docs with links to the Fable docs:

```fs
/// <seealso href="https://fable.io/docs">
/// Fable Documentation
/// </seealso>
```

Thank you both!

MangelMaxime:
> I'll make a pull on the website in conjunction with this and iterate over the docs simultaneously.

Can you please keep the changes on the docs for this PR and the rest separated please. It will be easier to review as it will be more focused and smaller

shayanhabibi:
I'll ask for this block to be reviewed and see if the style is agreeable before moving forward.

PS: I've left the TypeScriptTaggedUnion attribute as I wasn't sure if we may also take this chance to deprecate the `TypeScript` name, or move it to another namespace such as Fable.Core.JS or JsInterop.

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -15,19 +22,92 @@ type CaseRules =
...
 /// other interfaces, but will make interop with native code more difficult.
+/// </summary>
+/// <remarks>
+/// <para>
+/// If you are not planning to use an interface to interact with native code and want to have overloaded
+/// members, you can decorate the interface declaration with the <c>Mangle</c> attribute.
+/// </para>
+/// <para>
+/// Interface coming from .NET BCL (like System.Collections.IEnumerator) are mangled by default.
+/// </para>
```
MangelMaxime: I don't think this belongs here, it could belong to the different compatibility pages

For example:
- https://fable.io/docs/javascript/compatibility.html

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -15,19 +22,92 @@ type CaseRules =
...
+///         member this.Render() =
+///             failwith "Not implemented"
+///
+///         member this.Render(indentation) =
+///             failwith "Not implemented"
+/// </code>
+/// <para>
+/// May be considered invalid code depending on the target. Note the two methods named <c>Render</c> in the same class.
+/// This would be a primary example case where <c>Mangle</c> would allow F#/Fable to route the call correctly.
+/// </para>
```
MangelMaxime: I find the explanation confusing because we demonstrate a case where `Mangle` could be used but it is not called in the snippet.

I think it is best to have it documented on fable docs because there we can showcase the generated code so people can understand what is happening

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -15,19 +22,92 @@ type CaseRules =
...
+/// type AnimationBuilder() = class end
+/// let anime = AnimationBuilder()
+/// [&lt;Erase>]
+/// type TimelineBuilder() = class end
+/// let tl = TimelineBuilder()
+/// // would fail on runtime as we have erased the constructor from
+/// // the compiled code
+/// </code>
+/// Similar behavior is seen when used on <c>module</c>s. Check the <a href="https://fable.io/docs">
+/// documentation for your target language</a> for more information.
```
MangelMaxime: Can be removed

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -15,19 +22,92 @@ type CaseRules =
...
+/// functions which may be undefined without replacement.
+/// <code lang="fsharp">
+/// type AnimationBuilder() = class end
+/// let anime = AnimationBuilder()
+/// [&lt;Erase>]
+/// type TimelineBuilder() = class end
+/// let tl = TimelineBuilder()
+/// // would fail on runtime as we have erased the constructor from
+/// // the compiled code
+/// </code>
```
MangelMaxime: I still think without the compiled code it is difficult to understand, what do you think?

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -38,39 +118,98 @@ type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
...
     new(tagName: string) = TypeScriptTaggedUnionAttribute(tagName, CaseRules.LowerFirst)
 
-/// Used in place of `CompiledNameAttribute` if the target is not a string.
+/// <summary>
+/// Used in place of <c>CompiledNameAttribute</c> if the target is not a string.
+/// </summary>
+/// <remarks>
+/// Accepts <c>Enum</c>s, <c>int</c>, <c>float</c> or <c>bool</c>.
+/// <br/>
+/// You can also generate <c>undefined</c> by passing <c>null</c>.
```
MangelMaxime: This is target specific

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -38,39 +118,98 @@ type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
...
+/// <summary>
 /// References to the module, type, function... will be replaced by import statements.
-/// Use `[<Import("default", "my-package")>] to import the default member.
-/// Use `[<Import("*", "my-package")>] to import the whole package.
-/// More info: https://fable.io/docs/communicate/js-from-fable.html#type-safety-with-imports-and-interfaces
+/// </summary>
+/// <remarks>
+/// <para>Use <c>[&lt;Import("default", "my-package")>]</c> to import the default member.</para>
+/// <para>Use <c>[&lt;Import("*", "my-package")>]</c> to import the whole package.</para>
+/// <para>You can import proxy-like constructs, or nested submodules as you would natively.</para>
```
MangelMaxime: Seems like a complex sentence to understand to me, can be removed

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -38,39 +118,98 @@ type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
...
 
+/// <summary>
 /// References to the module, type, function... will be replaced by import statements.
-/// Use `[<Import("default", "my-package")>] to import the default member.
-/// Use `[<Import("*", "my-package")>] to import the whole package.
-/// More info: https://fable.io/docs/communicate/js-from-fable.html#type-safety-with-imports-and-interfaces
+/// </summary>
+/// <remarks>
+/// <para>Use <c>[&lt;Import("default", "my-package")>]</c> to import the default member.</para>
+/// <para>Use <c>[&lt;Import("*", "my-package")>]</c> to import the whole package.</para>
```
MangelMaxime: It is actually target specific, I don't think Python has `default` for example

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -38,39 +118,98 @@ type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
...
+/// <seealso href="https://fable.io/docs/">
+/// Fable Documentation
+/// </seealso>
 type ImportAttribute(selector: string, from: string) =
     inherit Attribute()
 
-/// Takes the member name from the value it decorates
+/// <summary>
+/// Similar (or same) to <c>Import</c>, except the <c>selector</c> parameter is automatically filled with the
+/// decorated members name.
```
MangelMaxime: This can decorate other things that `members`, for example functions.

Wording from the documentation is: 

> importMember is used to import a specific member from a JavaScript module, the name is based on the name of the F# value.

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -38,39 +118,98 @@ type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
...
-/// More info: https://fable.io/docs/communicate/js-from-fable.html#type-safety-with-imports-and-interfaces
+/// <summary>
+/// The module, type, function... is globally accessible in the target runtime.
+/// </summary>
+/// <remarks>
+/// Consider a function such as <c>printf</c> in <c>F#</c>, this function is accessible
+/// without any import or declaration. You would decorate a binding for a target construct
+/// using <c>Global</c> to prevent any code being compiled for the declaration, and any
+/// import-like code being generated in usage in other <c>Fable</c> code.
+/// </remarks>
```
MangelMaxime: The wording make sense, but I took me 2 attempts to be sure what it means. (see global comment)

MangelMaxime (changes requested):
I am still wondering if all the `seealso` add something or not.

I mean if people want to learn something about Fable, the first place to go would be Fable documentation.

```fs
/// <seealso href="https://fable.io/docs/">
/// Fable Documentation
/// </seealso>
```

I find the explanations with code samples difficult to interpret because in general. This is because, I think without the compiled code it is difficult to understand what they do.

One of the good example for that is `Erase`, even if I am a user of Fable since 7+ years I have a hard time understanding what the description means.

Where if I look at the [documentation](https://fable.io/docs/javascript/features.html#erased-unions), I can directly see the effect it has.

Something to remember is that often these functions, attributes are used to created bindings so in general people know what the JavaScript or Python code they want look it. Which make it easier to understand by reading Fable docs.

This is at least how I learned it, and based on the popularity of similar blog post that's what works best for most people.

I feel like I saying mostly negative thing in this PR, and I don't like that feeling. If my wording is bad feel free to reach to me, it is possible that I am really conservative too.


---------

I feel like really simple comments like:

```fs
/// <summary>
/// References to the module, type, function... will be replaced by import statements.
/// </summary>
type ImportAttribute(selector: string, from: string) =
    inherit Attribute()
```

Is enough and people need to go to Fable docs in order, to learn special syntax like

```fs
[<Import("default", "my-lib">]
[<Import("*", "my-lib">]
```

for JavaScript, because in Rust, Python it will not be the same.

src/Fable.Core/Fable.Core.Types.fs
shayanhabibi: What about with regards to the description of the behavior?

shayanhabibi:
> I am still wondering if all the `seealso` add something or not.
> 
> I mean if people want to learn something about Fable, the first place to go would be Fable documentation.
> 
> ```fsharp
> /// <seealso href="https://fable.io/docs/">
> /// Fable Documentation
> /// </seealso>
> ```

Agreed, doesn't serve much purpose unless it can route to a specific page/section.

> I find the explanations with code samples difficult to interpret because in general. This is because, I think without the compiled code it is difficult to understand what they do.
> 
> One of the good example for that is `Erase`, even if I am a user of Fable since 7+ years I have a hard time understanding what the description means.

Fair enough. I've reduced that out, but what about the general description? The thing is I think there is difficulty to remember viewing these documents without any context. The very concise documentation seems natural, but it really said nothing/meant nothing when I was trying to learn Fable. 

I would usually have to just see how others were using it and copy paste.
Whether in the xml, or on the website, there should be somewhere that explains how these things work.

Personally, I prefer there to be 0 requirement on external documentation (ie you can learn everything from the documentation on the codebase rather than needing to go to an external website), which is probably what influences my contribution style. This is understandably NOT something that can be done with Fable, but it does inform my style.

> Where if I look at the [documentation](https://fable.io/docs/javascript/features.html#erased-unions), I can directly see the effect it has.
>
Agreed.

> Something to remember is that often these functions, attributes are used to created bindings so in general people know what the JavaScript or Python code they want look it. Which make it easier to understand by reading Fable docs.
> 
> This is at least how I learned it, and based on the popularity of similar blog post that's what works best for most people.

I disagree, I think the information on how to use Fable is pretty scattered across blog posts and libraries/repos that use it because the documentation primarily does not attempt to explore such topics itself, and the advanced usage and nuances of attributes are not entirely discussed or clear.

My opinion comes from recent experience, having just over the past 6-8 months started learning how to use it. I understand that I may just be _slower_, or I was learning in isolation, but I think it is important to not think our learning experience justifies the current status quo.

Some of the first plugins I reach for in an IDE have to do with making it easier for my tooling to show me the documentation, and to consume it within the IDE. This might just be me though.

> I feel like I saying mostly negative thing in this PR, and I don't like that feeling. If my wording is bad feel free to reach to me, it is possible that I am really conservative too.

No, I really appreciate the honesty and feedback, because it helps me improve, learn, or understand others perspectives. I really do appreciate it. Thank you!

> I feel like really simple comments like:
> 
> ```fsharp
> /// <summary>
> /// References to the module, type, function... will be replaced by import statements.
> /// </summary>
> type ImportAttribute(selector: string, from: string) =
>     inherit Attribute()
> ```
> 
> Is enough and people need to go to Fable docs in order, to learn special syntax like
> 
> ```fsharp
> [<Import("default", "my-lib">]
> [<Import("*", "my-lib">]
> ```
> 
> for JavaScript, because in Rust, Python it will not be the same.

Honestly, the simple comment means nothing to me without having seen the docs first, and a lot of the documentation is like this. I think we become complacent with our breadth of experience/knowledge of the codebase informing the level of explanation required for someone without it.

 Without example etc I _HAVE_ to check the docs and hope it is explained further. As mentioned before, I understand this isnt possible in the context of Fable. I still think some example of usage in F# should be required, but I'm happy to leave this for the website with regards to the shared attributes etc.

While I do think that having target specific code is not a good idea, I also would say that it is more helpful to be able to try to infer the behavior from a compiled example in a different language rather than no example at all.

---

I would appreciate if you could have a look at the most recent iteration 🙇‍♂️ 

---

Similarly, please let me know if my comments are inferred as being rude or anything like that. I thoroughly enjoy discussion and debate, but I am aware that text is a very poor format that does not convey the tone in which I intend for some things to be said. Some points I make might just be to invite opinion so that I can learn.

MangelMaxime:
> Whether in the xml, or on the website, there should be somewhere that explains how these things work.

This should be the case on Fable docs, for example https://fable.io/docs/javascript/features.html#mangle

> Personally, I prefer there to be 0 requirement on external documentation (ie you can learn everything from the documentation on the codebase rather than needing to go to an external website), which is probably what influences my contribution style. This is understandably NOT something that can be done with Fable, but it does inform my style.

I agree for standard documentation this what I do also but in this is a special context IHMO.

> My opinion comes from recent experience, having just over the past 6-8 months started learning how to use it. I understand that I may just be _slower_, or I was learning in isolation, but I think it is important to not think our learning experience justifies the current status quo.

I am not saying the current situation is perfect, but that in order for the documentation to be as easy to understand as possible in the context of Fable it needs more information than what the tooltip can delivers.

This goes back to showing the generated code, adding more explanation, etc. and the place for that is Fable docs currently and not the XML doc.

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -2,6 +2,10 @@ namespace Fable.Core
 
 open System
 
+/// <summary>
+/// Parameter used in <c>StringEnum</c> and methods/other attributes that
+/// generate strings from unions or other identifiers to change the casing in conversion.
+/// </summary>
```
MangelMaxime:
```fs
/// <summary>
/// Specifies different naming case transformation rules.
/// </summary>
type CaseRules =
    /// <summary>
    /// No transformation is applied.
    /// 
    /// <example>
    /// Transform <c>FooBar</c> into <c>FooBar</c>
    /// </example>
    /// </summary>
    | None = 0

    /// <summary>
    /// Only the first character is lowercased.
    /// 
    /// <example>
    /// Transform <c>FooBar</c> into <c>fooBar</c>
    /// </example>
    /// </summary>
    | LowerFirst = 1
```

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -15,19 +19,34 @@ type CaseRules =
...
     inherit Attribute()
     new() = MangleAttribute(true)
 
-/// Used on a class to attach all members, useful when you want to use the class from JS.
+/// <summary>
+/// Used on a class to attach all members without mangling.
+/// </summary>
+/// <remarks>
+/// If you want to have all members attached to a class (as is standard with most target languages)
+/// and not-mangled, use the <c>AttachMembers</c> attribute. But be aware that overloads won't work
```
MangelMaxime: This feels like a repeat of the `summary`


I am not confortable with this sentence:

> (as is standard with most target languages)

```fs
/// <summary>
/// Used on a class to attach all members without mangling, be aware that overloads won't work
/// </summary>
```

src/Fable.Core/Fable.Core.Types.fs
```diff
@@ -38,39 +57,73 @@ type TypeScriptTaggedUnionAttribute(tagName: string, caseRules: CaseRules) =
 
     new(tagName: string) = TypeScriptTaggedUnionAttribute(tagName, CaseRules.LowerFirst)
 
-/// Used in place of `CompiledNameAttribute` if the target is not a string.
+/// <summary>
+/// Used in place of <c>CompiledNameAttribute</c> if the target is not a string.
+/// </summary>
+/// <remarks>
+/// Accepts <c>Enum</c>s, <c>int</c>, <c>float</c> or <c>bool</c>.
```
MangelMaxime:
```suggestion
/// Accepts <c>Enum</c>, <c>int</c>, <c>float</c> or <c>bool</c>.
```

MangelMaxime (changes requested):
I feel more comfortable with the new proposition

PR #4159, opened 2025-07-03
[JS/TS] Initial support for Nullable Reference Types

ncave:
- [JS/TS] Initial Support for [Nullable Reference Types](https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-fsharp-9/).
- Not tested yet on larger code bases using NRTs, but seems to address the original issues raised in #4085.
- Supersedes #4085.
- Contains a change to `Fable.AST` which will need to be versioned and released.
 The AST change is intentionally positioned at the end of the `Fable.Type` union to keep it (mostly) backwards compatible with older plugins (to the extent possible).

### DIFF
--- a/src/fable-library-ts/Util.ts
+++ b/src/fable-library-ts/Util.ts
@@ -458,8 +458,14 @@ function equalObjects(x: { [k: string]: any }, y: { [k: string]: any }): boolean
   return true;
 }
 
-export function physicalEquality<T>(x: T, y: T): boolean {
+export function physicalEquals<T>(x: T, y: T): boolean {
   return x === y;
+}
+
+export function nullableEquals<T>(x: Nullable<T>, y: Nullable<T>): boolean {
+  if (x == null) { return y == null; }
+  if (y == null) { return false; }
+  return equals(x, y);
 }
 
 export function equals<T>(x: T, y: T): boolean {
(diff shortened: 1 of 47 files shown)


ncave:
@MangelMaxime This is ready for review, if you have the time. JS/TS only for now, the rest of the languages will come later.

MangelMaxime:
Thank you @ncave , I will have a look at it later this week

ncave:
After some deliberation it looks like storing the nullness as a type in Fable AST is the right way to go, as it needs to be considered in some places where there is limited access to anything else besides AST (e.g. overload suffix calculations).

@MangelMaxime I have merged your #4085 in here, with some changes.

src/fable-library-ts/Util.ts
```diff
@@ -456,10 +458,16 @@ function equalObjects(x: { [k: string]: any }, y: { [k: string]: any }): boolean
   return true;
 }
 
-export function physicalEquality<T>(x: T, y: T): boolean {
+export function physicalEquals<T>(x: T, y: T): boolean {
   return x === y;
 }
```
MangelMaxime, line 463: @ncave Is this rename expected?

Was the name of this function always wrong ?

MangelMaxime (approved):
Overall it looks good to me, thank you for all the work you put into that one.

While testing the `quicktest` project with `nullable` enabled, I found that we will need to upgrade Fable.Core to FSharp.Core 9+ because we need access to `objnull` for some of the APIs.

I will do it in another PR as this one is starting to get really big :)

src/fable-library-ts/Util.ts
ncave: @MangelMaxime It's actually not used right now, it just sounds more better.

Issue #4093, opened 2025-04-09
Inline function resolves static member call on second generic parameter (`'b`) using the implementation of the first (`'a`)


Programmerino:
**Environment:**
*   Fable Version: Fable REPL (4.23.0 as of writing)
*   Target: JavaScript

**Description:**
When using an `inline` function with multiple generic type parameters (`'a`, `'b`) that are both constrained by a static member requirement (e.g., `static member getString`), calling the static member explicitly on the *second* type parameter (`'b.getString`) inside the function body incorrectly resolves to the implementation provided by the *first* type parameter (`'a`).

**REPL Link:**
[...]

PR #4160, opened 2025-07-03
[All] Fix generic parameter resolution in inline functions with static member constraints

Programmerino:
# Fix generic parameter resolution in inline functions with static member constraints

*Note: Claude Code did this for me, so it could be terrible*

## Description

This PR fixes a bug (https://github.com/fable-compiler/Fable/issues/4093) where inline functions with multiple generic type parameters and static member constraints would incorrectly resolve all trait calls to the first type parameter's witness.

## The Issue

When using inline functions with multiple generic parameters that have static member constraints, all static member calls would resolve to the first type parameter:

```fsharp
type A = static member M() = "A"
type B = static member M() = "B"

let inline test<'a, 'b when 'a: (static member M: unit -> string) 
                       and 'b: (static member M: unit -> string)> () =
    'a.M(), 'b.M()  // Both calls incorrectly returned "A"

test<A, B>()  // Would return ("A", "A") instead of ("A", "B")
```

## Root Cause

The issue occurred in two places:
1. During trait call resolution, when multiple witnesses matched the same trait, the code always selected the first one
2. For nested inline functions, generic parameter mappings were being replaced instead of composed

## Solution

1. **Enhanced Witness Resolution**: Added `tryFindWitnessWithSourceTypes` that uses source type information to select the correct witness when multiple witnesses match. It determines which generic parameter is being resolved and selects the corresponding witness based on position.

2. **Fixed Nested Inline Functions**: Modified the generic argument handling in `inlineExpr` to compose mappings rather than replace them, ensuring that inner inline functions can resolve their generic parameters through the outer function's context.

## Changes

- Added `tryFindWitnessWithSourceTypes` in `FSharp2Fable.Util.fs` to handle witness disambiguation
- Updated trait call resolution in `FSharp2Fable.fs` to use source types for witness selection
- Fixed generic argument composition for nested inline functions to preserve outer context mappings

## Testing

The fix has been tested with:
- Basic two-parameter inline functions
- Three or more parameter inline functions  
- Different parameter orders
- Multiple constraints per type
- Nested inline function calls
- Various type combinations

All scenarios now work correctly.

## Impact

This fix enables proper use of inline functions with multiple generic parameters having static member constraints, which is important for generic programming patterns in F#.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -14,9 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [Python] Fix `nonlocal`/`global` declarations generated inside `match/case` bodies causing `SyntaxError` (by @dbrattli)
 * [Python] Fix exception variable captured in deferred closures causing `NameError` (PEP 3110 scoping) (by @dbrattli)
 * [JS/TS] Support format specifiers and single hole in JSX string templates (by @MangelMaxime)
-* [All] Fix generic parameter resolution in inline functions with static member constraints (by @Programmerino
-wants to merge 3 commits into
-and @MangelMaxime)
+* [All] Fix generic parameter resolution in inline functions with static member constraints (by @Programmerino and @MangelMaxime)
 
 ## 5.0.0-rc.2 - 2026-03-03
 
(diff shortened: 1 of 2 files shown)


MangelMaxime:
Hello,

Can you please add some test suits so we can check if the behavior is correct ?

Programmerino:
Let me know if that works!

shayanhabibi:
Not going to lie, I really didn't think Claude could have pulled that off. Actually works. 👀. Nice!

MangelMaxime:
Sorry for taking so long to look at your issue @Programmerino.

I was able to simplify (I think) the code. For that I needed to modify Fable.AST which we try to avoid to do because it can break Fable plugins, but as we are in RC state this is probably fine to do.

PR #4165, opened 2025-07-05
[JS] Adjust Prologue Directive typing

shayanhabibi:
Should allow string enums to be used by default.

The downside is consumers might emit other types accidentally.
The benefit is consumers *can* emit other types also, and make use of the functionality added by `emitJsTopDirectivePrologue` in whatever other way its behavior allows.

### DIFF
--- a/src/Fable.Core/Fable.Core.JsInterop.fs
+++ b/src/Fable.Core/Fable.Core.JsInterop.fs
@@ -54,7 +54,7 @@ let emitJsStatement<'T> (args: obj) (jsCode: string) : 'T = nativeOnly
 /// a specific directive at the top of the file, such as "use client" or "use server".
 /// </summary>
 /// <param name="text">Directive text</param>
-let emitJsTopDirectivePrologue (text: string) : unit = nativeOnly
+let emitJsTopDirectivePrologue (text: obj) : unit = nativeOnly
 
 /// <summary>
 /// Emit a directive prologue at the calling position.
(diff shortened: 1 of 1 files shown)


MangelMaxime:
These changes are not enough because the replacement will only occurs if the argument is of type `string`:

https://github.com/fable-compiler/Fable/blob/ab865570200611b45d54e610d1d389b68f000f70/src/Fable.Transforms/Replacements.fs#L1055-L1057

Plus, I am not sure when the conversion of `StringEnum` to `string` happens.

I don't think `StringEnum` support makes senses here, people can just use module + `[<Literal>]` here in order to not rewrite the same string over and over.

shayanhabibi:
How would I be able to replicate behavior such as this using the emitJsTopDirectivePrologue:

```jsx
// @refresh reload
import { mount, StartClient } from "@solidjs/start/client";
import { attachDevtoolsOverlay } from "@solid-devtools/overlay";
import 'solid-devtools';
```

MangelMaxime:
You can't because what you are showing is not a prologue directive in JavaScript but a comment.

A prologue directive is something like:

```js
"use strict";
"use server";
"use client";
// etc.
```

What you are talking about seems to be named `pragma` on solid-js repo: https://github.com/solidjs/solid-refresh#pragma

TBH some project use the name directive, pragma, etc. for the same thing

shayanhabibi:
Ah yes, but if the typing for prologue is removed, would I not be able to use an Emit macro? That was my end aim :P

MangelMaxime:
It won't do what you want because it will always add the surrounding quotes

If being able to generate comments at the top of the file is a feature needed, it will have to be done via another specialized API.

Nothing prevent us to have both `emitJsTopDirectivePrologue` and `emitJsTopPragma` or `emitJsTopComment`. Would probably go with `pragma`

shayanhabibi:
Yeah I didn't realise originally that this was enclosing it automatically with the quotes, happy to do topPragma when I can. Thanks!

PR #4169, opened 2025-07-06
[Python] Misc fixes

dbrattli:
- Fixed typing issues with option. We now erase types for `'T option` to `T | None` but keep `'T option option` as `Option[T]`. 
- Fixes for array contains, made `cons` optional
- Fixes for IPrintfFormat missing `cont` after migrating to Rust.
- Misc typing issues `int` should be `int32`


ncave:
@dbrattli Not sure why main branch build is failing, seems to work locally 🤔

PR #4171, opened 2025-07-08
Fixes for publish-pypi GH action

dbrattli:
## Why

Looking at the GH action run log I see:

https://github.com/fable-compiler/Fable/actions/runs/16103023239/job/45510132091

Sun, 06 Jul 2025 20:49:34 GMT
Artifact fable-library-py has been successfully uploaded! Final size is 44151193 bytes. Artifact ID is 3472951971

Mon, 07 Jul 2025 21:25:37 GMT
Error: Unable to download artifact(s): Artifact not found for name: fable-library-py

This means that this run took over 24 hours to run, and we have an artifact retention of 1 day. This looks like a issue with GH actions either congestion, rate-limiting, limits, runner availability etc. Thus, I think this was an intermittent issue. If it happens again, we should increase the artifact retention time.

## How

These are just some general fixes to the GH actions workflows I fixed while investigating. It will not fix the issue above if it happens again.

- Fixes for Python publish-pypi GH action, making some dependencies explicit
- Fixed indentation issues in `publish-pypi.yml` GH action
- Upgrades archived `actions-rs/toolchain@v1` to `dtolnay/rust-toolchain@stable` for Rust build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -190,9 +190,7 @@ jobs:
       uses: actions/setup-node@v2.4.1
 
     - name: Setup Rust toolchain
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
+      uses: dtolnay/rust-toolchain@stable
 
     - name: Fable Library - Rust
       run: ./build.sh fable-library --rust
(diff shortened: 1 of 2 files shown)


ncave:
main branch CI build still failing :(

dbrattli:
~~@ncave I don't think so since it should only re-run when another PR is merged 🤔  This merge didn't trigger any build or test since it was only the publish-pypi action that was modified. It will run on the next main merge. https://github.com/fable-compiler/Fable/actions~~

... wait, I see something. Will fix 🏃🏻

PR #4172, opened 2025-07-09
[Python] Fix pyo3 floats hashing

dbrattli:
Fixes pyo3 specific test that broke after refactoring the float hash function


ncave:
@dbrattli Just out of curiosity, do you know why wasn't that test failing when running it locally, and only failing when merging into main?

dbrattli:
@ncave Yes, this is because it's some test I had for verifying the Rust build of fable-library-py (only) before publishing to PyPI. I.e these are not transpiled tests, just plain Python tests. But I need to add them to be run by Fable.Build as well.

PR #4177, opened 2025-07-12
[Python, WIP] Build Python AST using CEs

dbrattli:
- More experimentation of generating Python AST using F# CEs
- Removes a lot of boilerplate code and clutter by hiding the statements "threading" though the code
- How far can we / should we go?


MangelMaxime:
I always have a problem/fear with CEs DSL in F# because the more CEs we have or the more complex they are, the slower F# compiler becomes...

https://github.com/dotnet/fsharp/issues/14429

dbrattli:
Thanks for the warning @MangelMaxime. Then I might not go all-in, and just leave it for the outer boundary to hide the statement "threading" in the expression building of Fable2Python, and instead use "normal" functions/methods on the inside.

MangelMaxime:
Your welcome, I always feel like the "bad guy" when someone create a new CEs DSL as I often point this performance issue 😅

I would love for it to be fixed, but looking at the issue comment it seems like a lot of work ...

PR #4182, opened 2025-07-18
Erase reference nullable types

ncave:
I know, I know, we just added these, and now we're erasing them :(

But hear me out, I had some second thoughts on that subject, for several reasons:

- Technically Nullable Reference Types (NRTs) are only checked at compile time for .NET code,
the IL produced is the same with or without 'Nullable', so the interoperability is good.
(feel free to correct me if I'm wrong, though).

- As much as I would like to have NRTs as actual types in the target language, it introduces potential API friction.
It requires using anonymous unions (for languages that support anonymous sum types, aka anonumous DUs, aka erased DUs), so we can have some interoperability between code compiled with `Nullable` and code compiled without.

- Even using anonymous union types for NRTs only gives you interoperability in one direction:
From code *not* compiled with union types to code compiled *with* `Nullable`, it's fine (except for Rust).
The other direction still breaks the contract, e.g. passing `T` to `T | null` works, but the reverse does not.

- In the case of Rust, there are no anonymous sum types (yet), so neither direction works, so the only option is
to either erase NRTs to the inner type, or hope that everything is compiled with the same `Nullable` setting. 
For now I have opted to erase NRTs for Rust, but this can be revisited.

- Of course, this is *not* an issue *at all*, when all your code is compiled the same, either all `Nullable` or all not `Nullable`.
And Fable more or less always recompiles all source code for all dependencies at once, so it's less of an issue
(but it could be, when mixing already precompiled code bases).

- Nevermind the separate but very likely issue of trying to compile dependencies that have not been retrofitted to be compilable with `Nullable`, so the compilation will have lots of nullability warnings.

So this PR is meant just to stir some conversation, please share your thoughts.

Perhaps I'm overreacting and this PR is completely unnecessary, given that we already have a switch that drives the NRT behavior (i.e. the `Nullable` project option, aka `checknulls+`), and most of the time it will be fine, since all dependencies are compiled the same as the main project, so the issue is only about interoperability between disjoint code bases precompiled with different nullability settings.

Thoughts @MangelMaxime @dbrattli ?

### DIFF
--- a/src/Fable.Transforms/Dart/Fable2Dart.fs
+++ b/src/Fable.Transforms/Dart/Fable2Dart.fs
@@ -194,10 +194,13 @@ module Util =
         let tup = List.length genArgs |> getTupleTypeIdent com ctx
         Type.reference (tup, genArgs)
 
-    let transformNullableType com ctx genArg =
-        transformType com ctx genArg |> Nullable
+    let transformNullableType com ctx isStruct genArg =
+        if isStruct then
+            transformType com ctx genArg |> Nullable
+        else
+            transformType com ctx genArg // nullable reference types are erased
 
-    let transformOptionType com ctx genArg =
+    let transformOptionType com ctx _isStruct genArg =
         let genArg = transformType com ctx genArg
 
         Type.reference (libValue com ctx Fable.MetaType "Types" "Some", [ genArg ])
@@ -652,8 +655,8 @@ module Util =
             | BigInt
             | NativeInt
             | UNativeInt -> Dynamic // TODO
-        | Fable.Nullable(genArg, _isStruct) -> transformNullableType com ctx genArg
-        | Fable.Option(genArg, _isStruct) -> transformOptionType com ctx genArg
+        | Fable.Nullable(genArg, isStruct) -> transformNullableType com ctx isStruct genArg
+        | Fable.Option(genArg, isStruct) -> transformOptionType com ctx isStruct genArg
         | Fable.Array(TransformType com ctx genArg, _) -> List genArg
         | Fable.List(TransformType com ctx genArg) -> Type.reference (getFSharpListTypeIdent com ctx, [ genArg ])
         | Fable.Tuple(genArgs, _) -> transformGenArgs com ctx genArgs |> transformTupleType com ctx
@@ -798,10 +801,10 @@ module Util =
             Expression.invocationExpression (regexIdent.Expr, args, Type.reference regexIdent)
             |> resolveExpr returnStrategy
 
-        | Fable.NewOption(expr, genArg, _isStruct) ->
+        | Fable.NewOption(expr, genArg, isStruct) ->
             let transformOption (com: IDartCompiler) ctx genArg (arg: Expression) =
                 let cons = libValue com ctx Fable.MetaType "Types" "Some"
-                let t = transformOptionType com ctx genArg
+                let t = transformOptionType com ctx isStruct genArg
 
                 let isConst, args =
                     if areConstTypes t.Generics && isConstExpr ctx arg then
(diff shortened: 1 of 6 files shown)


dbrattli:
@ncave, based on what you say, I think it's probably a good idea to take it out for now, but keep it around as a do-not-merge PR, and move the discussion there until we better understand what's the right way forward. 

Could a separate flag like `<FableNullableOutput>preserve</FableNullableOutput>` or `--fable-nullable-output=preserve|erase` make sense for controlling how Fable handles those types in the compiled output? 🤔

ncave:
@dbrattli 
> it's probably a good idea to take it out for now

By "take it out", do you mean merge this PR which uses the inner type for NRTs instead of anonymous sum types like `T | null | undefned`  for JS, or  `T | None` for Python? You can still enable and use `Nullable` in projects, just without the special nullable reference sum types.

Or do you mean postpone this PR until further time? Personally I don't have a preference either way.

Also, I'm not sure we need a separate special switch, since there is already a switch you can easily turn on or off for the main project you're compiling (i.e. the `Nullable` projects setting, aka `checknulls+`). Perhaps we just need to document the effect of it.

dbrattli:
Yes, you are right, for a moment there I was thinking projects (dependencies) used their own `Nullable` setting, but in Fable everything is compiled using the main project `Nullable` setting, so we don't need another flag. You were talking about interop with external code (compiled by Fable) that is not part of the Fable compilation?

Anyways I don't have any strong opinions if we keep it as a PR to remove, or as a PR to add, until we figure out the way forward here 😊

MangelMaxime:
I think following NRTs behavior for .NET which are only checked at compile time is good enough.

When working on Nullable support in the past, and trying to output the "right" TypeScript code I add some issues similar to what you describe here:

> * Even using anonymous union types for NRTs only gives you interoperability in one direction:
>       From code _not_ compiled with `Nullable` to code compiled _with_ `Nullable` union types, it's fine (except for Rust).
>       But the other direction still breaks the contract, e.g. passing `T` to `T | null` works, but the reverse does not.

This was because TypeScript was not able to detect that we were handling `null` check in another function call or something like that.

If I understand correctly, if we erase nullable types, it means that it is handled internally by the F# compiler and we don't need to worry about the capacity of the target runtime to have a way to represent them.

ncave:
@dbrattli 
> You were talking about interop with external code (compiled by Fable) that is not part of the Fable compilation?

Yes, although interoperability with already generated older code from a different code base is potentially dicey at the best of times anyway, unless the same exact Fable version was used to compile both.

If, as it looks like, nobody has strong opinions about it yet, we can leave this PR pending for now, until we get more feedback from users.

MangelMaxime:
> If, as it looks like, nobody has strong opinions about it yet, we can leave this PR pending for now, until we get more feedback from users.

Okay, I will have a look to create a new release this week. This is long due because we have a lot of improvements made since the last one

ncave:
@MangelMaxime 
> If I understand correctly, if we erase nullable types, it means that it is handled internally by the F# compiler and we don't need to worry about the capacity of the target runtime to have a way to represent them.

Correct.

@MangelMaxime @dbrattli I'm starting to think we do need to merge this PR, since the risk of mismatching API signatures is too great when mixing projects compiled with and without the `Nullable` projects setting (aka `checknulls+`) and that can break compatibility, even when using the same exact Fable version.

E.g. In addition to all said above, APIs that have function types with NRTs in them are incompatible in either direction, so it definitely breaks API contract for anything that has function types with NRTs in it.

Also, we can only compile Fable core libraries either with or without nullchecks, not both (unless we want to distribute two versions of each). If we merge this, there will be no difference whether you compile with or without nullchecks, so we can continue as is.

Perhaps we should merge this PR now just to be safe, and take another look in the future. What do you think?

MangelMaxime:
@ncave I am fine with merging this PR, I don't think we want to release 2 versions of each libraries this would be too much work. 🙈

PR #4185, opened 2025-07-20
[Python] Fix python quicktest runscript

dbrattli:
We use `uv` for handling Python and Python dependencies configured in pyproject.toml. Make sure we run Python using `uv´ i.e `uv run python`. This fixes the error you get below you have no global python installed.

```txt
Cannot run: An error occurred trying to start process 'python' with working directory '/Users/dbrattli/Developer/Fable/src/quicktest-py'. No such file or directory
```
It is also important that we run QuickTest using the same version of Python as we used to build Fable Library or the Rust extension will not work since it's built for the specific version of Python. This makes it much easier to test with different versions of Python by using e.g `uv python pin <version>`.

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -964,7 +964,7 @@ let private checkRunProcess (state: State) (projCracked: ProjectCracked) (compil
             match cliArgs.CompilerOptions.Language, runProc.ExeFile with
             | Python, Naming.placeholder ->
                 let lastFilePath = findLastFileRelativePath ()
-                "python", lastFilePath :: runProc.Args
+                "uv", "run" :: "python" :: lastFilePath :: runProc.Args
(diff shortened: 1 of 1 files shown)


MangelMaxime:
@dbrattli I think this change is an issue.

Because you changed code related to `fable` tool itself and not related to the build system of Fable repository.

This means that people are forced to use `uv` in order use `runScript` option of Fable.

I believe these changes should be reverted? Can't `uv` handle control the available version of  `python` when calling `python` CLI?

For example, in JavaScript if people use `fnm`, `nvm` or `node` directly all these tools are able to control the version executed when running `node` CLI.

I hope my explanations are clear 🙈, if not please tell me I try to better rephrase.

dbrattli:
@MangelMaxime Yes of course. My bad 🙈 Will revert. I need to fix this in Build and not in the CLI.

PR #4197, opened 2025-08-01
[JS/TS] Replace Error with Exception

ncave:
- [JS/TS] Replace the remaining native JS `Error` type usage with `Exception` class for consistency.
This fixes some TS exception type mismatch issues, since most exceptions were already derived from `Exception`.


ncave:
@MangelMaxime This is ready for merging, I tagged you in case you want to review and chime in, but do let me know if you don't have time at the moment, and I can merge it.

MangelMaxime:
@ncave This looks fine to me.

I was not active the last 2 weeks because I was on vacation :) 2 weeks with no computer feels good sometimes 😇

PR #4199, opened 2025-08-04
[Python] Add [<ClassAttributes>] attribute to control Python class generation

dbrattli:
## Overview

Introduce a new `[<ClassAttributes>]` attribute that provides Python-specific control over how F# types are transpiled to Python classes. This attribute implies member attachment (similar to `AttachMembers`) while offering Python-specific parameters for better integration with Python frameworks like Pydantic, dataclasses, and SQLAlchemy.

### Without AttachMembers (Default)

```fsharp
type User() =
    member val Name: string = "" with get, set
```

Generates:

```python
class User:
    def __init__(self, __unit: None=None) -> None:
        pass

def User_get_Name(self) -> str:
    return self._Name

def User_set_Name(self, value: str) -> None:
    self._Name = value
```

### With AttachMembers (Current)

```fsharp
[<AttachMembers>]
type User() =
    member val Name: string = "" with get, set
```

Generates:

```python
class User:
    def __init__(self, __unit: None=None) -> None:
        self._Name: str = ""

    @property
    def Name(self) -> str:
        return self._Name

    @Name.setter
    def Name(self, value: str) -> None:
        self._Name = value
```

---

### Basic Syntax

```fsharp
// Replaces [<AttachMembers>] for Python with same behavior (properties style)
[<ClassAttributes>]
type User() = ...

// New: Generate class attributes instead of @property
[<ClassAttributes(Style="attributes")>]
type User() = ...

// With additional Python-specific parameters
[<ClassAttributes(Style="attributes", Slots=true, Frozen=true)>]
type Config() = ...
```

#### F# Input

```fsharp
[<ClassAttributes(Style="attributes")>]
type User() =
    member val Name: string = "" with get, set
    member val Age: int = 0 with get, set
    member val Email: string option = None with get, set
```

#### Generated Output

```python
class User:
    Name: str
    Age: int
    Email: str | None

    def __init__(self, Name: str = "", Age: int = 0, Email: str | None = None):
        self.Name = Name
        self.Age = Age
        self.Email = Email
```

---

## Parameter Reference

|Parameter|Type|Default|Description|
|---|---|---|---|
|`style`|string|`"properties"`|`"properties"` (current behavior) or `"attributes"`|
|`init`|bool|`true`|Generate `__init__` method|
|`slots`|bool|`false`|Add `__slots__` for memory efficiency|
|`frozen`|bool|`false`|Make attributes read-only after init|
|`repr`|bool|`false`|Generate `__repr__` method|
|`eq`|bool|`false`|Generate `__eq__` method|

#### Suppress **init** Generation

```fsharp
[<ClassAttributes(style="attributes", init=false)>]
type Config() =
    member val Debug: bool = false with get, set
    member val Timeout: int = 30 with get, set
```

```python
class Config:
    Debug: bool = False
    Timeout: int = 30
    # No __init__ method generated

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* [Python] Added `ClassAttributes` attribute to control Python class generation (@dbrattli)
+
 ### Fixed
 
 * [TS/Dart] Fixed optional parameter types (by @ncave)
@@ -105,14 +109,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [Python] Fix date formatting when repeating a format token more than the known format (example repeating 'd' more than 4 times) (by @MangelMaxime)
 * [JS/TS] Fix #4010: Supports direct nested types when using `jsOptions` (by @MangelMaxime)
 
-    ```fs
+```fs
     let opts =
         jsOptions<Level1> (fun o ->
             o.level2.level3.valueA <- 10
             o.level2.level3.valueB <- 20
             o.topValueA <- 20
         )
-    ```
+```
 
 * [JS/TS] Fix numeric formats (by @MangelMaxime)
 
--- a/src/Fable.Core/CHANGELOG.md
+++ b/src/Fable.Core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* [Python] Added `ClassAttributes` attribute to control Python class generation (@dbrattli)
+
 ## 5.0.0-beta.1 - 2025-07-25
 
 ### Added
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by MangelMaxime** [comment]
@dbrattli We probably need to update the documentation too.

Also, wouldn't it be better to replace the `string` with a DU, for example something like:

```fs
[<ClassAttributes(style=ClassAttributesStyles.Attributes, init=false)>]
```

this way it avoid potential typo error?

dbrattli:
But attribute values need afaik to be basic types. Then the alternative is to use an enum: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#2224-attribute-parameter-types

MangelMaxime:
I though I remembered something on that line but didn't check 🙈

However, we can use an enum to achieve something similar to DUs. This is what we do for `StringEnum` for example

```fs
[<RequireQualifiedAccess>]
type ClassAttributeStyle =
    | Properties = 0
    | Attributes = 1

[<AttributeUsage(AttributeTargets.Class)>]
type ClassAttributes() =
    inherit Attribute()

    new(style: ClassAttributeStyle) = ClassAttributes()

    new(style: ClassAttributeStyle, init: bool) = ClassAttributes()

[<ClassAttributes(style = ClassAttributeStyle.Properties)>]
type MyClass () =
    class end
```

dbrattli:
Thanks. Will fix

PR #4200, opened 2025-08-05
[WIP/Python] Add Decorate attribute

dbrattli:
## Overview

Introduce a new `[<Decorate>]` attribute that adds Python decorators to generated classes, enabling integration with Python frameworks like dataclasses, attrs, functools, and any other decorator-based libraries.

## Purpose

F# doesn't have a direct equivalent to Python decorators, but they're essential for integrating with the Python ecosystem. The `[<Decorate>]` attribute bridges this gap by allowing F# developers to specify Python decorators that should be applied to the generated class.

## ⚠️ Important: Python Interop Only

**The `[<Decorate>]` attribute is purely for Python interop and does NOT affect F# compilation behavior.**

### What it DOES (Python interop):

- Adds decorators to the **generated Python code**
- Enables integration with Python libraries that require decorators
- Makes the transpiled Python work with Python ecosystem tools

### What it DOESN'T do (F# side):

- Does NOT change how your F# code compiles or behaves
- Does NOT add decorator functionality to F# code
- Does NOT make Python decorator features available in F# compilation

```fsharp
[<Decorate("dataclasses.dataclass")>]
type User() =
    member val Name: string = "" with get, set

// This is still regular F# - no dataclass behavior here
let user = User()
user.Name <- "Alice"  // Standard F# property access

// The @dataclass behavior only exists in the generated Python
```
#### F# Input

```fsharp
[<Decorate("dataclasses.dataclass")>]
type User() =
    member val Name: string = "" with get, set
    member val Age: int = 0 with get, set
```

#### Generated Output

```python
from dataclasses import dataclass

@dataclass
class User:
    Name: str = ""
    Age: int = 0
```

#### F# Input

```fsharp
[<Decorate("dataclasses.dataclass")>]
[<Decorate("functools.total_ordering")>]
type User() =
    member val Name: string = "" with get, set
```

#### Generated Output

```python
from dataclasses import dataclass
from functools import total_ordering

@total_ordering
@dataclass
class User:
    Name: str = ""
```

**Note:** Decorators are applied in **reverse order** (bottom to top), following Python's standard decorator stacking behavior.

---

#### F# Input

```fsharp
[<Decorate("functools.lru_cache", "maxsize=128")>]
type CachedClass() =
    member val Value: string = "" with get, set
```

#### Generated Output

```python
from functools import lru_cache

@lru_cache(maxsize=128)
class CachedClass:
    Value: str = ""
```

#### F# Input

```fsharp
[<Decorate("attrs.define", "auto_attribs=True, slots=True")>]
type OptimizedClass() =
    member val Data: string = "" with get, set
```

#### Generated Output

```python
import attrs

@attrs.define(auto_attribs=True, slots=True)
class OptimizedClass:
    Data: str = ""
```

---

### Dataclasses

```fsharp
[<Decorate("dataclasses.dataclass")>]
type Point() =
    member val X: float = 0.0 with get, set
    member val Y: float = 0.0 with get, set
```

```python
from dataclasses import dataclass

@dataclass
class Point:
    X: float = 0.0
    Y: float = 0.0
```

### Attrs Library

```fsharp
[<Decorate("attrs.frozen")>]
type ImmutablePoint() =
    member val X: float = 0.0 with get, set
    member val Y: float = 0.0 with get, set
```

```python
import attrs

@attrs.frozen
class ImmutablePoint:
    X: float = 0.0
    Y: float = 0.0
```

### DIFF
--- a/src/Fable.Core/Fable.Core.Py.fs
+++ b/src/Fable.Core/Fable.Core.Py.fs
@@ -27,6 +27,30 @@ module Py =
 
         abstract Decorate: fn: Callable * info: Reflection.MethodInfo -> Callable
 
+    /// <summary>
+    /// Adds Python decorators to generated classes, enabling integration with Python
+    /// frameworks like dataclasses, attrs, functools, and any other decorator-based
+    /// libraries.
+    /// </summary>
+    /// <remarks>
+    /// <para>The [&lt;Decorate&gt;] attribute is purely for Python interop and does NOT
+    /// affect F# compilation behavior.</para>
+    /// <para>Multiple [&lt;Decorate&gt;] attributes are applied in reverse order
+    /// (bottom to top), following Python's standard decorator stacking behavior.</para>
+    /// <para>Examples:</para>
+    /// <para>[&lt;Decorate("dataclasses.dataclass")&gt;] - Simple decorator</para>
+    /// <para>[&lt;Decorate("functools.lru_cache", "maxsize=128")&gt;] - Decorator with
+    /// parameters</para>
+    /// </remarks>
+    [<AttributeUsage(AttributeTargets.Class, AllowMultiple = true)>]
+    type DecorateAttribute(decorator: string) =
+        inherit Attribute()
+
+        new(decorator: string, parameters: string) = DecorateAttribute(decorator)
+
+        member val Decorator: string = decorator with get, set
+        member val Parameters: string = "" with get, set
+
     // Hack because currently Fable doesn't keep information about spread for anonymous functions
     [<Emit("lambda *args: $0(args)")>]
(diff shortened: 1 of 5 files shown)


MangelMaxime:
@dbrattli We probably need to update the documentation on [Fable website too](https://fable.io/docs/python/features.html)

dbrattli:
@MangelMaxime Yes, I was wondering if it should be done before or after Fable 5 is released, but I can probably do it now and tag it with (New in Fable 5)?

MangelMaxime:
@dbrattli IHMO we can document it now, so we don't forget about it.

If you want to you can add a "tag" with the next version in it:

image

dbrattli:
Thanks, I'll add it later today

PR #4201, opened 2025-08-05
[JS/TS/Python] Fixed eq comparer mangling

ncave:
- [JS/TS/Python] Fixed eq comparer mangling

### DIFF
--- a/src/fable-library-py/fable_library/util.py
+++ b/src/fable-library-py/fable_library/util.py
@@ -121,13 +121,13 @@ class IEquatable(Protocol):
         raise NotImplementedError
 
     @abstractmethod
-    def __hash__(self) -> int:
+    def __hash__(self) -> int32:
         raise NotImplementedError
 
 
 class IComparable(IEquatable, Protocol):
     @abstractmethod
-    def __cmp__(self, __other: Any) -> int:
+    def __cmp__(self, __other: Any) -> int32:
         raise NotImplementedError
 
     @abstractmethod
@@ -137,7 +137,7 @@ class IComparable(IEquatable, Protocol):
 
 class IComparable_1[T_in](IEquatable, Protocol):
     @abstractmethod
-    def __cmp__(self, __other: T_in) -> int:
+    def __cmp__(self, __other: T_in) -> int32:
         raise NotImplementedError
 
     @abstractmethod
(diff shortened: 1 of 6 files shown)


ncave:
@dbrattli I made some minor changes in `util.py`, do you mind taking a quick look just to make sure you don't object? Thanks!

ncave:
@dbrattli I made some minor changes in `util.py`, do you mind taking a quick look just to make sure you don't object? Thanks!

src/fable-library-py/fable_library/util.py
```diff
@@ -162,48 +162,44 @@ class IComparer_1[T_in](Protocol):
...
+    @abstractmethod
     def GetHashCode(self) -> int32:
         raise NotImplementedError
 
 
 class IStructuralEquatable(Protocol):
     @abstractmethod
     def Equals(self, other: Any, comparer: IEqualityComparer) -> bool:
-        return NotImplemented
+        return NotImplementedError
```
dbrattli:
```suggestion
        raise NotImplementedError
```

dbrattli:
@ncave Looks great, thanks for fixing. Only have one fix. But thinking about it, I'm mixing two concepts here 🙈  1) Abstract base classes (ABC) and 2) Protocols. ABC uses abstract methods, protocols do not. We should only use protocols (`Protocol`) for this i.e structural subtyping. So, these methods should not be `@abstractmethod` at all, and should just return `...`, not raise anything. They should be just empty i.e `...`. But if this compiles and passes the tests, then I can fix it in follow-up PR.

dbrattli:
PS: The reason I used `@property` which I know looks weird is that F#/Fable generates code like this which will not match the protocol. But I agree that the protocol should be a method and not a callable. I should probably find another way of fixing it in e.g the generated code or find a way for the protocol to recognize both forms 🤔 :

```py
def test_map_is_empty_works(__unit: None=None) -> None:
    class ObjectExpr1340:
        @property
        def Compare(self) -> Callable[[IComparable, IComparable], int32]:
            return compare

    Testing_equal(True, is_empty_1(of_seq(to_enumerable([]), ObjectExpr1340())))
    class ObjectExpr1341:
        @property
        def Compare(self) -> Callable[[int32, int32], int32]:
            return compare_primitives

    Testing_equal(False, is_empty_1(of_seq(to_enumerable([(int32(1), int32(1))]), ObjectExpr1341())))
```

ncave:
@dbrattli
> But if this compiles and passes the tests, then I can fix it in follow-up PR.

Ok, merging then.

Issue #2423, opened 2021-04-01
Automatic bundling/dev server?


alfonsogarciacaro:
So far, Fable has tried to be a bridge between F# and JS ecosystems. And as such, it didn't try to do too much and instead let the user pick their JS tools to bundle, debug the frontend, etc... However, since Fable 3 moved to a be a "pure" dotnet tool and with the appearance of libraries that don't require npm dependencies like Sutil or Elmish.Snabbdom (or even React if you use it from a CDN) it may be helpful particularly for newcomers if Fable could do (or at least invoke) the basic operations to write the frontend of a web app: bundling and dev server. For that we would include new CLI args like `--serve` and/or `--bundle`.

Writing our own bundler is out of hand. At first I thought we could embed a bundler in the Fable package, but probably it's easier to just create a hidden `.fable-tools` dir, install there the needed tools with npm and invoke them as necessary. The ideal would be to rely on a bundler with good defaults but that can be customized if necessary. The question is, what bundler to use?

- Webpack: the most extended and probably the preferred option when user needs to customize. Defaults won't likely be enough but we could generate an automatic webpack.config and if necessary let the user customize it. If I'm not mistaken that's how create-react-app works.
- Parcel: this would be a very good candidate, as it has many good defaults, but they seem to be stuck in the v2 upgrade forever.
- Rollup: we've used it sometimes in Fable for its ease of use, but recently I find it lacking and difficult to customize, particularly the dev server.
- Other? I've lost track of the new bundlers like Snowpack or whatever they are called now 😅 

Any opinions? Remember this will only be optional and users will still be able to choose any other bundler as it's now.

MangelMaxime:
I don't think this is a good idea. It will made the boundary between F# and JavaScript world blur again. Similar to how it was in Fable 2 where people had a hard time understanding what was responsible for what.

Fable 3 is awesome because once you have called Fable to generate the JavaScript files then you are in JavaScript world directly. So you can find all the documentation you want on the subject made by the JavaScript community.

I would largely prefer a well made documentation explaining how Fable 3 work start and stop (compiling F# to JS) and then explain to the people from here you can use the tool you prefer. And have documentation simple documentation explaining how to do it and then to delegate the explanation to the chosen tool documentation.

IHMO, people need to understand the ecosystem they are working in/with. Yes, for really simple showcase you can ignore how JavaScript works and so on. But, soon enough you will need to understand the big picture in order to be able to write your application.

[...]

PR #4206, opened 2025-08-09
[Discussion] Add flag to format generated output

dbrattli:
## Why

It would be nice to generate nicely formatted output so code looks the way you expect it to be, and do not give warnings by linters in code editors. For many languages these days it the code has a default recommended formatting e.g `cargo fmt`, `ruff format`. 

image

## How

Add a flag `--format` so the target languages can perform "standard" formatting to the generated output. Not sure this is the right place, or if we could do it in a single step for all the generated files at once. 

What do you think? @MangelMaxime @ncave

### DIFF
--- a/src/Fable.AST/Plugins.fs
+++ b/src/Fable.AST/Plugins.fs
@@ -38,6 +38,7 @@ type CompilerOptions =
         FileExtension: string
         TriggeredByDependency: bool
         NoReflection: bool
+        Format: bool
     }
 
 type PluginHelper =
--- a/src/Fable.Build/Quicktest/Core.fs
+++ b/src/Fable.Build/Quicktest/Core.fs
@@ -53,6 +53,7 @@ let genericQuicktest (config: QuicktestConfig) (args: string list) =
         |> CmdLine.appendRaw "--noCache"
         |> CmdLine.appendRaw "--verbose"
         |> CmdLine.appendRaw "--watch"
+        |> CmdLine.appendRaw "--format"
         |> appendRunMode,
         workingDirectory = projectDir
     )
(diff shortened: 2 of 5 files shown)


ncave:
@dbrattli
This is just my personal opinion, so take it for what is worth, and by all means feel free to ignore or disagree.

Although I absolutely agree that default proper formatting is great and necessary, I think it's unnecessary to add it in Fable, even as an optional flag. I'm sure many arguments can be made in favor of, but here are some of the reasons against it, IMO:

- It adds unnecessary dependencies on other tools. We'll have to document what other tools need to be available in the system when you run Fable, so the build won't crash.
- It hard-codes specific formatters, making it strongly opinionated.
- Formatters sometimes expect or require configuration settings or options that potentially may have to be passed in.
- It is just as easy, and almost the same amount of characters, to just chain with '&&` the formatter of your choice after the Fable build step.
- Even if you want to avoid an additional build step chaining for some reason, there is already a `--run` flag in Fable that can be used for that.
- I agree very much with other transpilers' philosophy that we should keep Fable a simple pipeline and allow other tools to do their job afterwards (e.g. see TypeScript's stated [non-goals # 4](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals#non-goals), see also discussion in #2423).

TL;DR
Anyway, IMO formatting can be very easily added as a subsequent step, it creates less dependencies and is more flexible this way (as you can add whatever arguments to the formatter you desire).

dbrattli:
Hey @ncave , thanks for good feedback. I'll think of alternative ways of handing this outside Fable, e.g. create some tooling around Fable that I can use instead.

MangelMaxime:
@dbrattli Perhaps you can tell the linter to ignore the generated files? Because, in theory users should not need to look at the generated files.

PR #4209, opened 2025-08-10
[Dart] Format Set.fs

dbrattli:
Fantomas is complaining about this file. Tried to format it


nojaf:
Hmm, no immediate clue.
Line endings might be Windows versus Unix, although I believe we compare without that.
Could be wrong on that.

Not sure, you could ignore that file for now to get unblocked.

Maybe adding ` --verbosity d` shows something more but I'm not sure.

ncave:
@dbrattli I noticed that happening too in CI, but cannot reproduce locally.
Re-running the job somehow makes it go away, not sure why.

ncave:
@dbrattli I updated the `build.yml` in #4210, ~seems to work, I don't see that issue anymore~.

Update: Nevermind, still [happening](https://github.com/fable-compiler/Fable/actions/runs/16862927794/job/47765796000).
It's a crapshoot, sometimes it works, sometimes it doesn't.
Re-running the failed job makes it go away, but not always.

PR #4214, opened 2025-08-14
Remove utf-8 bom

ncave:


nojaf:
@ncave that resolved it?

ncave:
@nojaf
> that resolved it?

We'll see, maybe. It's happening quite inconsistently, so we need more data points (builds).
I don't see anything else in that file that could possibly explain or cause it (not that this explains it in any way, really).
It could still be something completely external (environmental).

ncave:
@nojaf
> that resolved it?

No, looks like it's still [failing](https://github.com/fable-compiler/Fable/actions/runs/17084109396/job/48444650131) from time to time, no idea why.

PR #4218, opened 2025-08-26
Cli Rewrite PoC

shayanhabibi:
Files changed:
Fable.Cli/Entry.fs has been removed.
New files:
Spec.fs
Settings/*.fs


shayanhabibi:
@MangelMaxime @dbrattli @nojaf 

There's the option of doing the rewrite with Fargo instead, so that we can generate autocompletion scripts.

I'll be able to progress this PR and refine it if we are happy to use the Spectre approach. I'll be able to provide help to a pull with Fargo, but I'm time starved at the moment, and enough time has been sunk into this that I can't afford to pivot atm (but happy to assist if someone does a pull).

I don't see why we can't make a script generation of our own from Spectre, but it's just a matter of time.

This is a circumstance where I actually would value the aesthetic appeal of the tool more than the autocompletion (in my opinion), as it is sorely dated, and looks 'beta-ish'. The features with interactivity and what not in spectre can also lend themselves to developing nice to have QoL features down the line (I'm thinking project setup with multi selection, or file stats using stacked bar graphs).

But, the usability of the tool should be priority. I do agree.
With the number of commands and options that fable has, and the tendency to abstract it away with scripts, I don't see as much value to autocompletion though.

I'd appreciate a comment if you would like this PR progressed. I'll leave it as it is for the time being.

shayanhabibi:
Actually it seems System.CommandLine will be the way to go. I believe the author of Spectre also contributes to that.

Provides the autocompletion, and enough access to the rendering to allow us to style with spectre.

shayanhabibi:
Superseded by #4220

PR #4227, opened 2025-09-20
[TS] Enable typedArrays flag usage for TypeScript

ncave:
- [TS] Enable `--typedArrays` flag usage for TypeScript (default = true)

### DIFF
--- a/src/Fable.Transforms/Fable2Babel.fs
+++ b/src/Fable.Transforms/Fable2Babel.fs
@@ -711,11 +711,17 @@ module Annotation =
         |> TupleTypeAnnotation
 
     let makeArrayTypeAnnotation com ctx genArg kind =
-        match genArg with
-        | JS.Replacements.TypedArrayCompatible com kind name -> makeAliasTypeAnnotation com ctx name
-        | _ ->
+        // match genArg with
+        // | JS.Replacements.TypedArrayCompatible com kind name -> makeAliasTypeAnnotation com ctx name
+        // | _ ->
+        //     // makeNativeTypeAnnotation com ctx [genArg] "Array"
+        //     makeTypeAnnotation com ctx genArg |> ArrayTypeAnnotation
+        match kind with
+        | Fable.ResizeArray ->
             // makeNativeTypeAnnotation com ctx [genArg] "Array"
             makeTypeAnnotation com ctx genArg |> ArrayTypeAnnotation
+        | Fable.MutableArray
+        | Fable.ImmutableArray -> makeFableLibImportTypeAnnotation com ctx [ genArg ] "Util" "MutableArray"
 
     let makeListTypeAnnotation com ctx genArg =
         makeFableLibImportTypeAnnotation com ctx [ genArg ] "List" "FSharpList"
@@ -1423,6 +1429,14 @@ module Util =
             | args when info.HasSpread ->
                 let args, lastArg = List.splitLast args
                 let args = args |> List.map (fun a -> a, ParameterFlags())
+
+                let lastArg =
+                    match lastArg.Type with
+                    | Fable.Array(genArg, Fable.MutableArray) ->
+                        // modify the last argument's type if it's a spread parameter
+                        { lastArg with Type = Fable.Array(genArg, Fable.ResizeArray) }
+                    | _ -> lastArg
+
                 args @ [ lastArg, ParameterFlags(isSpread = true) ]
 
             | thisArg :: args when info.IsInstance && List.sameLength args parameters ->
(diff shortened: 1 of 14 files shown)


src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -711,11 +711,17 @@ module Annotation =
...
 
     let makeArrayTypeAnnotation com ctx genArg kind =
-        match genArg with
-        | JS.Replacements.TypedArrayCompatible com kind name -> makeAliasTypeAnnotation com ctx name
-        | _ ->
+        // match genArg with
+        // | JS.Replacements.TypedArrayCompatible com kind name -> makeAliasTypeAnnotation com ctx name
+        // | _ ->
+        //     // makeNativeTypeAnnotation com ctx [genArg] "Array"
+        //     makeTypeAnnotation com ctx genArg |> ArrayTypeAnnotation
```
MangelMaxime: Should we clean up this code?

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -711,11 +711,17 @@ module Annotation =
...
+        match kind with
+        | Fable.ResizeArray ->
             // makeNativeTypeAnnotation com ctx [genArg] "Array"
```
MangelMaxime, line 716: Should we clean up this code?

src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -4083,24 +4097,31 @@ but thanks to the optimisation done below we get
 
             let args =
                 args
-                |> Array.mapi (fun i a ->
-                    let name = defaultArg a.Name $"arg{i}"
+                |> Array.mapi (fun i arg ->
+                    let name = defaultArg arg.Name $"arg{i}"
```
MangelMaxime:
```suggestion
                    let name = defaultArg arg.Name $"arg%i{i}"
```

src/fable-library-ts/Array.fs
```diff
@@ -741,16 +745,16 @@ let chunkBySize (chunkSize: int) (array: 'T[]) : 'T[][] =
         invalidArg "size" "The input must be positive."
 
     if array.Length = 0 then
-        [| [||] |]
+        ResizeArray<'T[]>() |> asArray
     else
-        let result: 'T[][] = [||]
+        let result = ResizeArray<'T[]>()
```
MangelMaxime: I have a doubt here

```fs
[| [||] |]
```

seems to be an array with another empty array in it

```fs
ResizeArray<'T[]>() |> asArray
```

seems to be only an empty array so equivalent to `[||]`.

Am I missing something? Or perhaps this is not problematic even if it is different

src/fable-library-ts/Array.fs
```diff
@@ -1079,9 +1083,9 @@ let splitInto (chunks: int) (array: 'T[]) : 'T[][] =
         invalidArg "chunks" "The input must be positive."
 
     if array.Length = 0 then
-        [| [||] |]
+        ResizeArray<'T[]>() |> asArray
     else
-        let result: 'T[][] = [||]
+        let result = ResizeArray<'T[]>()
```
MangelMaxime: Same here

src/fable-library-ts/CollectionUtil.ts
```diff
@@ -29,7 +29,7 @@ export function isReadOnly<T>(col: Iterable<T>): boolean {
     return (col as any)["System.Collections.Generic.ICollection`1.get_IsReadOnly"](); // collection
   } else {
     if (isArrayLike(col)) {
-      return ArrayBuffer.isView(col); // true for typed arrays, false for other arrays
+      return false; // array, resize array
```
MangelMaxime, line 32: Why enabling typed arrays for TS change this function behavior? It also impact JavaScript no?

src/Fable.Transforms/Fable2Babel.fs
ncave: Sure

src/fable-library-ts/Array.fs
ncave: Not a problem, cause the other branch was already returning `let result: 'T[][] = [||]` in some cases.
But I can simplify it a bit.

src/fable-library-ts/Array.fs
ncave: Same as the other one

src/fable-library-ts/CollectionUtil.ts
ncave: Arguably `false` is the [correct value for arrays](https://learn.microsoft.com/en-us/dotnet/api/system.array.isreadonly?view=net-9.0) for both JS and TS.

ncave:
@MangelMaxime Thanks for reviewing, I cleaned it up as suggested.

src/Fable.Transforms/Fable2Babel.fs
ncave: Fixed

MangelMaxime:
Looks good to me :+1:

PR #4237, opened 2025-10-15
Update FCS to latest

ncave:
- Updated FCS to latest (commit [cfda5f6](https://github.com/dotnet/fsharp/commits/cfda5f68f781f1b336b7d6c1689d69f2c34da751/))
- Updated TFM to .NET 9.0 (to avoid package warnings)
- FCS built with [.NET 10.0.100](https://dotnet.microsoft.com/en-us/download/dotnet/10.0)

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -24,15 +24,15 @@
     </ItemGroup>
     <ItemGroup>
         <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
-        <PackageReference Include="FSharp.Analyzers.Build" Version="0.3.0">
+        <PackageReference Include="FSharp.Analyzers.Build" Version="0.5.0">
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>build</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="G-Research.FSharp.Analyzers" Version="0.17.0">
+        <PackageReference Include="G-Research.FSharp.Analyzers" Version="0.20.0">
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>analyzers</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="Ionide.Analyzers" Version="0.14.6">
+        <PackageReference Include="Ionide.Analyzers" Version="0.14.10">
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>analyzers</IncludeAssets>
         </PackageReference>
--- a/src/Fable.Cli/Fable.Cli.fsproj
+++ b/src/Fable.Cli/Fable.Cli.fsproj
@@ -47,8 +47,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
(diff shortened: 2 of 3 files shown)


ncave:
@MangelMaxime This is ready for merging, but perhaps we want to create another alpha release before merging this, to not mix it with all the other changes.

MangelMaxime:
@ncave Sorry I missed your notification.

I think we should wait for the crackers to be fixed #4257 and #4252 and released. 

Then, we can include this PR in the main branch and see how it goes. I should have some time available tomorrow to look at the PR, and see if we can easily use a local version of the buildalyzer using `NuGet.config` or something like that.

MangelMaxime:
@ncave I think we are now ready to merge this PR.

- Should we also update the `global.json` from .NET 9 or .NET 10?
- Should we also update the `fable-metadata` dlls?

ncave:
@MangelMaxime 
> Should we also update the global.json from .NET 9 or .NET 10?

Perhaps, sure. But it's not directly related to this PR.
What we can try though is perhaps try again the Fable CLI with TFM 8.0, now that the Buildalyzer is fixed.
The TFM change is a separate commit so you can easily reverse it if needed, but I'm not sure if the issue is related though.

>  Should we also update the fable-metadata dlls?

No, that would be a different PR, for the repl. There are still some runtime issues with it that need looking into.

MangelMaxime:
> What we can try though is perhaps try again the Fable CLI with TFM 8.0, now that the Buildalyzer is fixed.
> The TFM change is a separate commit so you can easily reverse it if needed, but I'm not sure if the issue is related though.

Sorry, I am not sure to understand this part. Do you means to revert to `<TargetFramework>net8.0</TargetFramework>` in `Fable.CLI.fsproj`? Or should we target `net10.0`?

ncave:
@MangelMaxime Never mind, I thought that the Buildalyzer fix can help somehow, but it doesn't.

For Fable CLI, we can use TFM 9.0 (as this PR already does), or go to TFM 10.0 if you prefer, your choice.

If we stay with TFM 8.0, it kind of works, but we get build warnings for package version clashing at build time.
The issue stems from latest FCS picking up some v9.0.0 dependencies that clash with TFM 8.0:
```
".NETStandard,Version=v2.0/": {
     "FSharp.Compiler.Service/43.11.200-dev": {
        "dependencies": {
          "FSharp.Core": "10.0.200-dev",
          "FSharp.DependencyManager.Nuget": "10.0.200-dev",
          "System.Buffers": "4.6.0",
          "System.Collections.Immutable": "9.0.0",
          "System.Diagnostics.DiagnosticSource": "9.0.0",
          "System.Memory": "4.6.0",
          "System.Reflection.Emit": "4.7.0",
          "System.Reflection.Metadata": "9.0.0",
          "System.Runtime.CompilerServices.Unsafe": "6.1.0"
        },
```

MangelMaxime:
@ncave Yes, the Buildalyzer fix was only to fix the resolution of the source files.

> For Fable CLI, we can use TFM 9.0 (as this PR already does), or go to TFM 10.0 if you prefer, your choice.

I think we already had a similar discussion in the past, but I can't remember if we decided to go with the latest version available or the latest LTS version. I think we should go with the lastest LTS as this is probably aligned with how often people update their own projects.

ncave:
@MangelMaxime 
> I think we should go with the latest LTS

Sure, we can switch to TFM 10.0 (LTS) in a separate PR. I can make one after this gets merged.

PR #4247, opened 2025-11-02
Replace the depracated marked substr from TS

Thorium:
Replace the depracated marked substr from TS


MangelMaxime:
Thank you @Thorium 

I will make a new release later this week

PR #4255, opened 2025-11-13
[All] Fix MSBuildResolver to include define when restoring the project

MangelMaxime:


src/Fable.Compiler/MSBuildCrackerResolver.fs
```diff
@@ -34,7 +34,12 @@ module private MSBuildCrackerResolver =
 
             psi.WorkingDirectory <- Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)
 
-            psi.Arguments <- $"msbuild \"%s{fsproj}\" %s{args}"
+            let defineArgs =
+                defines
+                |> List.map (fun constant -> $"/p:%s{constant}=True")
```
cr3wdayt5p, line 39:
```suggestion
                |> List.map (fun constant -> $"/p:%s{constant}=true")
```

The legacy cracker set the values to `true` (not `True`).

Example from Fable 4 project:
```
dotnet restore Frontend.fable-temp.csproj -p:FABLE_COMPILER=true -p:FABLE_COMPILER_4=true -p:FABLE_COMPILER_JAVASCRIPT=true
```

MangelMaxime: I don't think it matters I tested it with 

```xml
<Compile Condition="'$(FABLE_COMPILER)'=='true'" Include="TupleTypeTests.fs" />
```

You can verify with Fable 4.27.0 which uses the same code as in this PR

MangelMaxime: I found the documentation section about that:

image

The string comparison is case insensitive

cr3wdayt5p: I can confirm it is case insensitive.
It works fine with `'$(FABLE_COMPILER)'=='tRUe'` on `4.27.0`.

Issue #4257, opened 2025-11-16
`MSBuildCracker` does NOT respect `global.json` (Fable 4 + 5)


cr3wdayt5p:
Hi

I am looking to upgrade our setup from Fable 4 + .NET 8 to Fable 5 + .NET 10, and in this process I got results that I could not easily explain.

But I have now pinned down the reason for some of the confusing results:
**`MSBuildCracker` does NOT respect `global.json`** (and hence does NOT use the SDK version specified for the repo)

This can result in a project failing to build with `dotnet build` but not with `dotnet fable` (or the other way around).

This is confusing when trying to investigate issues related to version upgrades 😊

<br>

This is a breaking change from the legacy cracker in Fable 4 which does respect `global.json`.

<br>

**Steps to verify**
1. Install both .NET SDK 8 and .NET SDK 10 on the machine (verify with `dotnet --info`).
2. Fix the SDK version to .NET 8 in `global.json` in your repo, e.g.: `{"sdk": {"version": "8.0.416","rollForward": "latestFeature"}}` 
3. Set the `TargetFramework` to `net42.0` (or some other non-released version) in your project file. This is just to provoke an error in the restore process.
4. Run `dotnet build`. Observe that the error is produced by SDK 8, e.g.: `/usr/local/share/dotnet/sdk/8.0.416/Sdks/Microsoft.NET.Sdk/...`.
5. Run `dotnet fable`. Observe that the error is produced by SDK 10, e.g.:  `/usr/local/share/dotnet/sdk/10.0.100/Sdks/Microsoft.NET.Sdk/...`.

This affects both Fable `4.27.0` (using `--test:MSBuildCracker`) and Fable `5.0.0-alpha.14`.

<br>

**Where to look**
It is likely caused by this line: https://github.com/fable-compiler/Fable/blob/main/src/Fable.Compiler/MSBuildCrackerResolver.fs#L35
```fsharp
psi.WorkingDirectory <- Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)
```
But there is no comment in the surrounding code explaining why `dotnet msbuild` should run in user profile directory rather than in the directory in where `dotnet fable` was called from (and hence would pick up the correct `global.json` file).

<br>
<br>

<details>
<summary>Machine Info</summary>

- OS: macOS 26.1
- .NET SDK: `8.0.416` and `10.0.100`
- Fable: `4.27.0` and `5.0.0-alpha.14`
</details>

[...]

PR #4261, opened 2025-11-19
Switch `MSBuildCracker` to use CWD (reproducible builds)

cr3wdayt5p:
Fix https://github.com/fable-compiler/Fable/issues/4257

**Before**
`MSBuildCracker` used `Environment.SpecialFolder.UserProfile` as the working directory for running `dotnet msbuild`. This makes the build results non-reproducible as the .NET SDK version available in the user profile is machine dependent, i.e. an arbitrary .NET SDK version is used to crack the project.

**After**
`MSBuildCracker` now uses the current working directory where `dotnet fable` was invoked (`cliArgs.RootDir`). This makes the build results reproducible since the selected .NET SDK version can be pinned with `global.json` in the repo. This is consistent with `dotnet build`.

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * [All] Fix `--cwd` with quoted paths (by @MangelMaxime)
 
+### Changed
+
+* [All] Invoke MSBuild from Fable current working directory (by @cr3wdayt5p)
+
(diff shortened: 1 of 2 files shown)


cr3wdayt5p:
Marked the PR as draft. See https://github.com/fable-compiler/Fable/issues/4257#issuecomment-3553946583. @MangelMaxime suggest we add better error messages – especially for the Fable 4 case.

cr3wdayt5p:
Also awaiting this fix to the integration tests before this PR (or another one like it) can be merged: https://github.com/fable-compiler/Fable/pull/4262

PR #4262, opened 2025-11-19
Fix use of `--cwd` argument in integration test

cr3wdayt5p:
The integration tests in https://github.com/fable-compiler/Fable/pull/4261 fail due to a misconfigured value in the `--cwd` argument.

This bug surfaced with https://github.com/fable-compiler/Fable/pull/4261 since it actually uses the value passed by the `--cwd` argument. It appears that the other tests never used the `--cwd` argument for anything (or at least not in way that would fail with the misconfigured value).

The errors were like:
```
System.ComponentModel.Win32Exception (2): An error occurred trying to start process '/usr/share/dotnet/dotnet' with working directory
'/home/runner/work/Fable/Fable/tests/Integration/Integration/$"{testCaseDir}"'. No such file or directory
```

I.e. `$"{testCaseDir}"` is used as the argument rather than the actual value of `testCaseDir`.

This PR fixes that to use the `testCaseDir` path directly. The value is enclosed in single quotes to allow spaces in the path (but not double quotes to avoid acidental interpretation).

### DIFF
--- a/tests/Integration/Integration/CompilationTests.fs
+++ b/tests/Integration/Integration/CompilationTests.fs
@@ -20,7 +20,7 @@ let tests =
 
                 // Compile project
                 let exitCode =
-                    Fable.Cli.Entry.main [| project; "--cwd"; "$\"{testCaseDir}\""; "-e"; ".jsx.actual" |]
+                    Fable.Cli.Entry.main [| project; "--cwd"; $"'%s{testCaseDir}'"; "-e"; ".jsx.actual" |]
 
                 Expect.equal exitCode 0 "Expected exit code to be 0"
 


MangelMaxime:
This fix does not work when applied to #4261 

The generate path provided to `cwd` is invalid:

> Running MSBuild in /home/mmangel/Workspaces/Github/fable-compiler/Fable/scratchpad/tests/Integration/Integration/'/home/mmangel/Workspaces/Github/fable-compiler/Fable/scratchpad/tests/Integration/Integration/data/jsxListOptimisation'

MangelMaxime:
Additional to the misplaced `$`, I believe there is an error in:

https://github.com/fable-compiler/Fable/blob/42816d62f073e0f83ed7b2291c65ad62d1870b68/src/Fable.Compiler/Util.fs#L248-L265

Because given `"/home/mmangel/Workspaces/Github/fable-compiler/Fable/scratchpad/tests/Integration/Integration/data/eraseAttribute"` it seems to return `/home/mmangel/Workspaces/Github/fable-compiler/Fable/scratchpad/tests/Integration/Integration/"/home/mmangel/Workspaces/Github/fable-compiler/Fable/scratchpad/tests/Integration/Integration/data/eraseAttribute"`

MangelMaxime:
It seems like the issue is caused by the quotes on the input.

I think at the time, I added the quotes so we could support path with spaces in them. I think for now, I will just simply remove the quotes and expects people to not have a space in their path when working with Fable. It seems to only impact the integration tests which few people touch in general.

So it should be an ok compromise

MangelMaxime:
Issue is caused by `Path.GetFullPath`

```fs
Path.GetFullPath "\"/home/maxime\""
// "/home/mmangel/Workspaces/Github/Experiad/motivity/websharper-9/"/home/maxime""
```

Closing this PR as I included the fix in #4261 directly.

PR #4272, opened 2025-11-26
[Python] Use fable-library from PyPI instead of bundle

dbrattli:
Python: Distribute fable-library via PyPI instead of bundling in NuGet. I somehow thought that fable-library would be recompiled, but when using `dotnet fable ...` then its's only copied from the NuGet. This makes the binary files (Rust / PyO3) wrong both for wrt Python version and architecture. So we must either 1) use `fable-library` from PyPI all the time 2) Add Rust files to NuGet and recompile when fable-library is copied (need dependency on Rust tool chain + Maturin). The simplest solution is 1), this PR:

## Changes

- The fable-library for Python now uses PyPI distribution by default instead of being bundled in the NuGet package. This fixes binary compatibility issues with the PyO3/Rust native extensions (.so files) that were compiled for a specific Python version.
- Remove bundled `fable-library` from NuGet since binary files and Python version will most likely not match the users version or architecture.
- Make sure fable-library contains all Python files (not just the core files)

Users need to install fable-library separately, e.g:
- `pip install fable-library`
- `poetry add fable-library`
- `uv add fable-library`

Custom --fableLib paths still work as before.

### DIFF
--- a/src/Fable.Build/Quicktest/Python.fs
+++ b/src/Fable.Build/Quicktest/Python.fs
@@ -12,7 +12,6 @@ let private fableLibraryBuildDir = Path.Resolve("temp", "fable-library-py")
 let handle (args: string list) =
     // Install local fable-library as editable package for testing
     // This ensures quicktest uses the locally built version, not PyPI
-    // Note: genericQuicktest will build fable-library first via FableLibBuilder
     if not (args |> List.contains "--skip-fable-library") then
         BuildFableLibraryPython().Run(false)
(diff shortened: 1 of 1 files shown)


Freymaurer:
@dbrattli please remember to add this to docs when fable 5 releases 🙂

dbrattli:
@Freymaurer Yes, there is some text already, but I'll update it to say this is now the default also when building locally https://fable.io/docs/python/build-and-run.html

Freymaurer:
@dbrattli by the way i messaged you on discord about the new changes in python. Only if you have the time for it of course ❤️

PR #4277, opened 2025-11-29
[Python] Python erased interfaces should not generate any code

dbrattli:
- Fixed regression since v4 that erased interfaces should not generate any code. Not a bug since the code would not do anything e.g `IExports` for bindings etc, but we should not emit it.
- Not adding to changelog since this issue was not present in Fable v4 and is a regression from one of the v5-alpha changes.


MangelMaxime:
We should still add a change-log entry for this kinds of changes because this is a `fix` in regards of Fable 5.

dbrattli:
Ok, I will fix that.

MangelMaxime:
Ooops, I already fixed the changelog @dbrattli

PR #4282, opened 2025-12-02
[Python] fix regression for record member method naming

dbrattli:
- Fix regression for record member method naming
- Update changelogs


dbrattli:
FYI: @Freymaurer this will fix another issue in Pyxpecto

PR #4289, opened 2025-12-04
[Python] Fix regression for `%A` format specifier to output booleans as lowercase

dbrattli:
- Fix regression %A format specifier to output booleans as lowercase true/false (F# semantics) instead of Python's True/False
- Found when testing Pyxpecto


MangelMaxime:
@dbrattli 

I didn't pay attention to the last commits format you submitted, but for now please keep using the old commit format for features. This is so we have some kind of consistency in the history.

I know I used `chore: ...` for some of them but this is because they are not in the changelog entry.

Once release-please is set up we can all move on to using the conventional commit format

PR #4308, opened 2025-12-24
Sync changelogs

ncave:
@MangelMaxime I assume the changelogs divergence was intentional, so feel free to just close this PR.
But if we manage to keep the versions the same and always release all components at the same time, we might be able to have just one changelog for everything. What do you think?


MangelMaxime:
I believe the changes were intentional because some of the features are related to Fable.CLI only.

Plus, it can happens that Fable.CLI gets a release when Fable.Compiler doesn't. This is why there is a 1 version offset between both the projects...

We should keep it as it is I think, and with the automation of release we will not need to worry about manually updating the changelogs anymore. I am looking into it.

PR #4325, opened 2026-01-05
[Python] Type checking in CI

dbrattli:
Testing out reporting type annotation errors in the generated Python library and test code, so we can keep the number under control and know if a PR reduces the number of type errors in the generated code or not. 

- The GH action will comment on the PR, but not fail the build (for now).
- Will only trigger if Python related files are touched:

```
- 'src/Fable.Transforms/Python/**'
- 'src/fable-library-py/**'
- 'tests/Python/**'
- 'pyrightconfig.ci.json'
- 'pyproject.toml'
```

### DIFF
--- a/.github/workflows/python-type-checking.yml
+++ b/.github/workflows/python-type-checking.yml
@@ -81,7 +81,7 @@ jobs:
           issue-number: ${{ github.event.pull_request.number }}
           edit-mode: replace
           body: |
-            ## Python Type Checking
+            ## Python Type Checking (Pyright) Results
 
             ${{ steps.pyright.outputs.result }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/python-type-checking.yml
```diff
@@ -0,0 +1,105 @@
...
+          pipx install maturin
+
+      - name: Build Fable Library - Python
+        run: ./build.sh fable-library --python
+
+      - name: Build Python tests (compile only)
+        run: |
+          uv sync
+          uv pip install -e temp/fable-library-py
+          dotnet fable tests/Python --outDir temp/tests/Python --lang python --exclude Fable.Core --noCache
```
MangelMaxime: We should probably use the existing `Fable.Build` for that and we can add a `--compile-only` flag to not execute the tests if you think it can slow down things.

dbrattli: Yes, just discovered that. Will fix!

PR #4339, opened 2026-01-29
[All] Fix `StringBuilder.Chars` getter and setter

### LINKED ISSUE #4338: [JS-lib] BUG: stringBuilder.Chars(idx) returns wrong char
I belive there is a bug in `fable-library-js/System.Text.js`

```fsharp
open System.Text

let sb = StringBuilder()

sb.Append "ABCDE" |> ignore 

// printfn $"c0 = {sb.Chars(0)}" // throws index out of bounds exception in Fable
printfn $"c1 = {sb.Chars(1)}" // D but B in .NET runtime 
printfn $"c2 = {sb.Chars(2)}" // C but C in .NET runtime 
printfn $"c3 = {sb.Chars(3)}" // B but D in .NET runtime 
printfn $"c4 = {sb.Chars(4)}" // A but E in .NET runtime 
```

see https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYDoAqYAeSBQ+ANmElAgEZQC8sSATgJYQDmAQgK6NEAmY9ACgCUhSugCCIcBB5QAROLYBhACIBROVAA+APiiMWEYPTBRCAenNQQTCEgBm0ACRyAxgAYaUAN5ilACwBDegQBdyEAX01LKCR-emAAdwR9GVwoYA4yYHsoCkyZFNxXMBAkRmBoZigAMUCKEnwbZgdnNwBGL18KdADg0PbI6KsVPKyoNlSodAA5NUwoeg47RlRTJttWqBdXACYuvyCQgV2hqBilMbJL6tn5xeXytbNmu0dttwBmA56+48+zjFJhRxqNbnMFksVs8Ni13jsACw-XpHUIIwFWcRXKBqKZ3SGPVbrQj4IA&html=Q&css=Q

Image

MangelMaxime:
Fix #4338

### DIFF
--- a/src/fable-library-py/fable_library/System.Text.fs
+++ b/src/fable-library-py/fable_library/System.Text.fs
@@ -15,11 +15,11 @@ type StringBuilder(value: string, capacity: int) =
     new() = StringBuilder("", 16)
 
     member x.Append(s: string | null) =
-        buf.Add(string s)
+        buf.Add(string<string | null> s)
         x
 
     member x.Append(s: string | null, startIndex: int, count: int) =
-        buf.Add((string s).Substring(startIndex, count))
+        buf.Add((string<string | null> s).Substring(startIndex, count))
         x
 
     member x.Append(c: char) =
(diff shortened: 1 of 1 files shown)


dbrattli:
We also need to update `src/fable-library-dart/System.Text.fs` and `src/fable-library-rust/src/System.Text.fs`. Not sure about the Python test since it works locally. In CI it seems to use the old `System.Text.fs` somehow. Let's see if it fixes itself when the other files are updated.

MangelMaxime:
> We also need to update `src/fable-library-dart/System.Text.fs` and `src/fable-library-rust/src/System.Text.fs`. 

Indeed, I though I tested it locally but I think I didn't save the tests files at that moment :facepalm: 

> Not sure about the Python test since it works locally. In CI it seems to use the old `System.Text.fs` somehow. Let's see if it fixes itself when the other files are updated.

Python.Library seems to be referencing the `System.Text.fs` from `../../fable-library-ts/System.Text.fs` which explains why it is fixed.

I am looking into it

MangelMaxime:
Strangely Python test is failing on CI but working for me locally 😅

But like you said @dbrattli, it seems to be using the previous version of the code 🤔

The test should be

```py
def StringBuilder__get_Chars_Z524259A4(x: StringBuilder, index: int32) -> str:
    len_1: int32 = int32.ZERO
    i: int32 = int32.ZERO
    while ((len_1 + len(x.buf[i])) <= index) if (i < len(x.buf)) else False:
        len_1 = len_1 + len(x.buf[i])
        i = i + int32.ONE
    if True if (index < int32.ZERO) else (i >= len(x.buf)):
        raise Exception("Index was outside the bounds of the array")

    else:
        pos: int32 = index - len_1
        return x.buf[i][pos]
```

but it is using:

```py
     def StringBuilder__get_Chars_Z524259A4(x: StringBuilder, index: int32) -> str:
        len_1: int32 = int32.ZERO
        i: int32 = int32.NEG_ONE
        while (len_1 < index) if ((i + int32.ONE) < int32(len(x.buf))) else False:
            i = i + int32.ONE
            len_1 = len_1 + get_length(x.buf[i])
        if True if (True if (index < int32.ZERO) else (i < int32.ZERO)) else (i >= int32(len(x.buf))):
>           raise Exception("Index was outside the bounds of the array")
E           Exception: Index was outside the bounds of the array
```

Any idea why ?

dbrattli:
@MangelMaxime This branch seems quite old which could be an issue. Perhaps try to rebase on main to get it up-to-date to see if that has something to do with the old code being used?

MangelMaxime:
> @MangelMaxime This branch seems quite old which could be an issue.

Right, I changed my git config which broke `git fetch origin && git rebase origin/main` for me.

Let's try again with the latest version of `main` now

dbrattli:
Ok, that explains it now. So Python already had a separate `System.Text.fs`, but did not see it in your older PR so that explains why it used the old code. The new PR shows the file and I have updated it so think that this should work now when GitHub resolves their action issues.

MangelMaxime:
Thank you for looking into it and indeed I missed the Python specific files.

PR #4357, opened 2026-02-23
[All] Fix ~45 code scanning warnings in Fable.Transforms

dbrattli:
## Summary
- Address four categories of code scanning analyzer warnings across 15 files in `src/Fable.Transforms/`
- **GRA-STRING-001/002/004**: Add explicit `StringComparison.Ordinal` to `StartsWith`, `EndsWith`, and `LastIndexOf` string overloads
- **IONIDE-007**: Replace cons-nil list patterns (`x :: []`) with list literal patterns (`[x]`), also improving readability of complex multi-element patterns in `Rust/Replacements.fs`
- **IONIDE-005**: Replace `s = ""` conditionals with `String.IsNullOrEmpty(s)` for more efficient empty string testing
- **GRA-TYPE-ANNOTATE-001**: Add type annotations to ambiguous `string` function calls using `string<type>` syntax

Note: GRA-INTERPOLATED-001 (`:s`/`:d` format specifiers) was attempted but reverted — the Standalone tests compile the compiler to JavaScript via Fable, which doesn't handle .NET-style format specifiers correctly.

## Test plan
- [x] `dotnet build src/Fable.Transforms/Fable.Transforms.fsproj` — 0 errors, 0 warnings
- [x] `./build.sh test javascript --skip-fable-library` — 2,603/2,607 passed (4 failures are pre-existing locale-dependent Convert format tests)
- [x] CI pipeline passes

### DIFF
--- a/src/Fable.Transforms/BabelPrinter.fs
+++ b/src/Fable.Transforms/BabelPrinter.fs
@@ -447,7 +447,7 @@ module PrinterExtensions =
                 |> List.iter (
                     function
                     | _, NullOrUndefinedOrVoid -> ()
-                    | key, StringConstant value -> printProp (fun () -> printer.Print($"{key:s}=\"{value:s}\""))
+                    | key, StringConstant value -> printProp (fun () -> printer.Print($"{key}=\"{value}\""))
                     | key, value ->
                         printProp (fun () ->
                             printer.Print(key + "={")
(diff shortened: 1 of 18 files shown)


dbrattli:
The though with this is to reduce the number of code scanning security warnings for Fable, starting with Fable.Transforms. Currently 726 warnings which doesn't look too good. What do you think? @MangelMaxime @ncave

MangelMaxime:
I think it is a good idea to reduce the number of warnings.

Especially, because right now even if we have the F# analyzers in place personally I don't look at it because their are too many warnings. 🙈

If we get the number of warning down, or even reach zero we can then only merge a PR when it does not increase it or add new warnings 🙂

ncave:
@dbrattli But I miss my `cons-nil list patterns (x :: [])` /jk

Joking aside, it was only done just to avoid the horrible source formatting that list literals have, so the source can stay human readable. But since human readability is no longer important anyway, nobody cares, so it's all good. The less warnings, the better. 👍

PR #4366, opened 2026-02-26
[Python] Fix PyPI version pattern to support rc tags

dbrattli:
## Summary
- The `dunamai` pattern in the PyPI publish workflow used `theta` instead of `rc` for release candidates
- This caused version extraction to fail for tags like `5.0.0-rc.1`
- Replace `theta` with `rc`, which is already a valid PEP 440 pre-release identifier (no conversion needed)

## Test plan
- [ ] Verify the workflow runs successfully on the next rc release tag


MangelMaxime:
Thank you

Issue #4368, opened 2026-02-27
How to test fable-library-ts ?


goswinr:
How can I test `fable-library-ts` ? 
By running `tsc` in its folder?

@MangelMaxime I am picking up on https://github.com/thoth-org/Thoth.Json/issues/217 and https://github.com/fable-compiler/Fable/issues/3586   

## Is it possible that `fable-library-ts` compiles with `tsc`  only  for Node but not for the browser?

For example [this line](https://github.com/fable-compiler/Fable/blob/7d08d8475520e24e2ae1ffdb382036dbf05e2ef4/src/fable-library-ts/Encoding.ts#L139) fails to compile because TS doesn't know what `Buffer` is in a browser context.

<img width="445" height="172" alt="Image" src="https://github.com/user-attachments/assets/91ed05a4-682e-4e6f-8ed2-d4a5dfd61b1c" />

Could I manually just edit these `fable-library-ts` files so that `tsc` works?

(I found some context here: https://github.com/fable-compiler/Fable/pull/3420 )

MangelMaxime:
That's a good question.

I think there are several solutions:

1. Remove `Buffer` usage and rely only on `TextEncoder` and the polyfill
2. When targeting browser specify that people need mark `Buffer` as external in their bundler (this can perhaps do the trick)

	However, I don't think `tsc` as a setting for that out of the box
3. Use a `global.d.ts` file say "I know what I am doing" 😅
	```ts
	// global.d.ts
	declare const Buffer: any;
	```

Thinking about it and looking at support for `TextEncoder` API I think we should go with solution 1.

[TextEncoder & TextDecoder as 96.34% supports](https://caniuse.com/?search=textencoder)

@ncave What do you think?

goswinr:
Thanks for looking at this @MangelMaxime and @ncave.
So editing the ts files here would be the right way? 
I think we should make sure all of [@fable-org/fable-library-ts](https://www.npmjs.com/package/@fable-org/fable-library-ts?activeTab=code) can be compiled with `tsc`  and a specified `tsconfig.json` ( currently there are two. at [src/fable-library-ts](https://github.com/fable-compiler/Fable/blob/main/src/fable-library-ts/tsconfig.json) and at [src/fable-library-ts/ts](https://github.com/fable-compiler/Fable/tree/main/src/fable-library-ts/ts) .  Does such a test exist? 
[...]

PR #4372, opened 2026-02-28
[JS/TS] Allows compiling `fable-library-ts` for Browser environment

goswinr:
addressing https://github.com/fable-compiler/Fable/issues/4368
where can I add tests?

### DIFF
--- a/src/fable-library-ts/Encoding.ts
+++ b/src/fable-library-ts/Encoding.ts
@@ -98,8 +98,8 @@ class UTF16LE {
     } else if (index != null) {
       str = str.substring(index);
     }
-	// Allow to compile for Browser environment without @types/node being installed
-	// See https://github.com/fable-compiler/Fable/issues/4368
+    // Allow to compile for Browser environment without @types/node being installed
+    // See https://github.com/fable-compiler/Fable/issues/4368
     const g = globalThis as any;
     if (typeof g.Buffer !== "undefined") {
       const bytes = g.Buffer.from(str, "utf16le");
@@ -117,6 +117,8 @@ class UTF16LE {
     } else if (index != null) {
       buffer = buffer.subarray(index);
     }
+    // Allow to compile for Browser environment without @types/node being installed
+    // See https://github.com/fable-compiler/Fable/issues/4368
     const g = globalThis as any;
     if (typeof TextDecoder !== "undefined") {
       return new TextDecoder("utf-16le").decode(buffer);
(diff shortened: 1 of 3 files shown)


src/fable-library-ts/Encoding.ts
```diff
@@ -98,8 +98,9 @@ class UTF16LE {
     } else if (index != null) {
       str = str.substring(index);
     }
-    if (typeof Buffer !== "undefined") {
-      const bytes = Buffer.from(str, "utf16le");
+    const g = globalThis as any;
```
MangelMaxime, line 103: We should probably add a note:

```suggestion
	// Allow to compile for Browser environment without @types/node being installed
	// See https://github.com/fable-compiler/Fable/issues/4368
    const g = globalThis as any;
```

src/fable-library-ts/Encoding.ts
```diff
@@ -134,10 +136,11 @@ class UTF8 {
     } else if (index != null) {
       str = str.substring(index);
     }
+    const g = globalThis as any;
```
MangelMaxime, line 145: We should probably add a note:

```suggestion
	// Allow to compile for Browser environment without @types/node being installed
	// See https://github.com/fable-compiler/Fable/issues/4368
    const g = globalThis as any;
```

src/fable-library-ts/Encoding.ts
```diff
@@ -152,10 +155,11 @@ class UTF8 {
     } else if (index != null) {
       buffer = buffer.subarray(index);
     }
+    const g = globalThis as any;
```
MangelMaxime, line 166: We should probably add a note:

```suggestion
	// Allow to compile for Browser environment without @types/node being installed
	// See https://github.com/fable-compiler/Fable/issues/4368
    const g = globalThis as any;
```

MangelMaxime:
> where can I add tests?

Hum, we don't have a setup for that.

I don't know if creating the setup required for it is worth it or not.

If I was to add the setup for that, I think we would want to create a new `src/Fable.Build/FableLibrary/TypeScriptBrowser.fs` that file would be based on https://github.com/fable-compiler/Fable/blob/main/src/Fable.Build/FableLibrary/JavaScript.fs.

Where we would replace `PostFableBuildStage` body with a code doing the 

1. Apply first the `copy package.json` related task
1. Adapt the `package.json` to remove the `@types/node` (run `npm unsintall @types/node`)
1. Build the library using `tsc`

We would also replace the output dir to something like `fable-library-js-browser`.

This should cover the initial goal of trying to compile `fable-library` without `@types/node` installed. If needed, we could even configure `tsconfig.json` to target the browser environment but I think the current config is generic enough for that.

If you are not sure of what I am talking about, I can have a look to send a commit to this PR showcasing the change and ask you to review it.

goswinr:
thanks @MangelMaxime  you can merge if you like it. I would make a seperate PR for the test. Will also try Thoth.Json.Core with TS first..

PR #4390, opened 2026-03-07
Update Python type-checking action

ncave:
- Update comment actions version to latest
- Skip PR comments update step if the PR event is triggered by a fork to avoid permission issues
(see https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#restrictions-on-repository-forks)


dbrattli (approved):
Thanks for fixing

PR #4389, opened 2026-03-07
Update dev dependencies

ncave:
- Update dev dependencies

PR #4394, opened 2026-03-12
[CI] Downgrade fsharp-analyzers to 0.35.0

dbrattli:
## Summary
- Downgrade `fsharp-analyzers` from 0.36.0 to 0.35.0 to fix CI build failures

## Test plan
- [ ] Verify CI build passes with the downgraded version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v5
         with:
-          dotnet-version: "10.0.x"
+          global-json-file: global.json
 
       - name: Restore tools
         run: dotnet tool restore
@@ -42,7 +42,7 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v5
         with:
-          dotnet-version: "10.0.x"
+          global-json-file: global.json
 
       - name: Restore project
         run: dotnet restore Fable.sln
@@ -60,7 +60,7 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v5
         with:
-          dotnet-version: "10.0.x"
+          global-json-file: global.json
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v5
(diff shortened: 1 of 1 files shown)


dbrattli:
This fixes the build but I think this is may have been a result of fsharp-analyzers tool was bumped from 0.34.1 → 0.36.0 in  #4389 and that it does not work well with FSharp.Analyzers.Build 0.5.0, so not sure the analyzers works anymore. Any ideas? @ncave @MangelMaxime

MangelMaxime:
If the problem is that `FSharp.Analyzers.Build 0.5.0` broke because of the new version of `fsharp-analyzers` we should report it in https://github.com/ionide/FSharp.Analyzers.SDK.

I don't think this is normal that we need to do such work around in the CI pipeline

dbrattli:
Should we then merge this until we can resolve the analyzer issue to get the builds passing again?

MangelMaxime:
After looking at it again, I am able to run the analyzers locally on my machine.

So it seems like the issue is related to .NET setup perhaps in the CI. I am trying to investigate in this direction

ncave:
@dbrattli 
>  it does not work well with FSharp.Analyzers.Build 0.5.0

Can this be reproduced locally?

dbrattli:
Yes, I can reproduce it locally:

```
MSBuild version 18.0.2+b0f34d51f for .NET                                                                                                                        
    [15:31:02.259] [FSharp.Analyzers.Cli] error: Trying to load /Users/dbrattli/.nuget/packages/ionide.analyzers/0.14.11/analyzers/dotnet/fs/Ionide.Analyzers.dll  
  which was built using SDK version 0.35.0.0. Expect 0.36.0.0 instead. Assembly will be skipped.                                                                   
    [15:31:02.260] [FSharp.Analyzers.Cli] warn: Analyzer path /Users/dbrattli/Developer/fable/fix-build/src/Fable.Analyzers/bin/Release/net10.0 does not exist     
    [15:31:04.340] [FSharp.Analyzers.Cli] error: Because we failed to load some assemblies to obtain analyzers from them, exiting (failure count: 1) 
```

So the fix should be to downgrade fsharp-analyzers back to 0.35.0

MangelMaxime:
With these changes now, it seems like we are able to run all the Analyzers against our project.

PR #4398, opened 2026-03-14
[Python/Beam] Add F# quotation support (<@ ... @>) 

dbrattli:
## Summary

FYI: This is just an investigation of what it would take to add Quotation support to Fable [WIP]

- Adds F# code quotation support (`<@ ... @>`) to the **Python** and **Beam** targets
- Quotation ASTs can be **constructed**, **pattern matched** (via `Microsoft.FSharp.Quotations.Patterns`), and **evaluated** (via `LeafExpressionConverter.EvaluateQuotation`)
- Shared `QuotationEmitter.fs` transforms quoted `Fable.Expr` into runtime library calls at compile time
- JS/TS support will be added in #4474

### Supported quotation nodes

Value, Lambda, Let, IfThenElse, Application, Call, Sequential, NewTuple, Operation (all arithmetic/comparison/logical), TypeCast, Get (TupleIndex, UnionField, UnionTag, FieldGet), Set (ValueSet, FieldSet), NewUnion, NewRecord, NewOption, NewList, DecisionTree (simple cases).

### Files changed

| Area | Files |
|------|-------|
| **AST** | `Fable.fs` — adds `Quote of quotedExpr: Expr * isTyped: bool * range: SourceLocation option` |
| **Frontend** | `FSharp2Fable.fs`, `FableTransforms.fs`, `Transforms.Util.fs` — capture and propagate `Quote` |
| **Shared engine** | `QuotationEmitter.fs` (new) — transforms quoted Fable.Expr → runtime lib calls |
| **Shared replacements** | `Replacements.Util.fs` — `Quotations` module with shared `tryQuotationCall` for all targets |
| **Python** | `Fable2Python.Transforms.fs`, `Python/Replacements.fs`, `quotation.py` (new) |
| **Beam** | `Fable2Beam.fs`, `Beam/Replacements.fs`, `fable_quotation.erl` |
| **Other targets** | `Fable2Babel.fs`, `Fable2Dart.fs`, `Fable2Rust.fs`, `Fable2Php.fs` — error stubs for `Quote` |
| **Tests** | `TestQuotation.fs` (Python, 24 tests), `QuotationTests.fs` (Beam, 18 tests) |
| **Changelogs** | `Fable.Cli/CHANGELOG.md`, `Fable.Compiler/CHANGELOG.md` |

## Design note: AST approach vs original PR #1839

The [original quotation PR (#1839)](https://github.com/fable-compiler/Fable/pull/1839) added a complex `ExprData` record to the AST:

```fsharp
// PR #1839
| Quote of typed: bool * data: ExprData * range: SourceLocation option

type ExprData = {
    typ: Type; variables: VarData[]; values: ValueData[]
    literals: Expr[]; types: Type[]
    members: array<FSharpEntity * Type * MemberInfo * Type[]>
    data: byte[]
}
```

This PR takes a **minimal AST approach** instead:

```fsharp
// This PR
| Quote of quotedExpr: Expr * isTyped: bool * range: SourceLocation option
```

The quoted `Expr` is kept as-is in the Fable AST. All decomposition into runtime calls happens in the transform layer (`QuotationEmitter.fs`), not the AST. This means:

- **+** Minimal AST footprint (no new types, no new `Type` variants)
- **+** Source location is preserved for error reporting and source maps (consistent with other `Expr` cases)
- **+** No serialization/deserialization layer needed
- **+** `QuotationEmitter` can evolve without breaking the AST contract
- **−** `Quote _ -> Any` is less precise than `Expr(Some typ)` / `Expr None`

If a more precise type is desired, a `| Quotation of gen: Type option` could be added to the `Type` union — but that would be a separate AST change worth discussing.

## Test plan

- [x] All 2103 Python tests pass (`./build.sh test python`)
- [x] All 2287 Beam tests pass (`./build.sh test beam`)
- [x] .NET tests pass for both Python and Beam test projects
- [x] Verify no regressions in other targets

### DIFF
--- a/src/Fable.Cli/CHANGELOG.md
+++ b/src/Fable.Cli/CHANGELOG.md
@@ -10,10 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * [Python/Beam] Add F# quotation support — construction, pattern matching, and evaluation via `LeafExpressionConverter.EvaluateQuotation` (by @dbrattli)
-* [All] Add support for `Guid.CreateVersion7()` and `Guid.CreateVersion7(DateTimeOffset)`
-* [Rust] Add missing `System.Random` implementation and tests (by @ncave)
-* [Rust] Add missing `Array`, `List` and `Seq` module members and tests: `randomChoice`, `randomChoiceBy`, `randomChoiceWith`, `randomChoices`, `randomChoicesBy`, `randomChoicesWith`, `randomSample`, `randomSampleBy`, `randomSampleWith`, `randomShuffle`, `randomShuffleBy`, `randomShuffleWith` (by @ncave)
-* [Beam] Implement missing DateTimeOffset members, add DateOnly and TimeOnly support
 * [All] Add support for `Guid.CreateVersion7()` and `Guid.CreateVersion7(DateTimeOffset)` (by @OnurGumus)
 * [All] Add missing `Array`, `List`, and `Seq` random choice/shuffle/sample members and tests (by @ncave)
 * [Dart/Rust] Add missing `System.Random` implementations and tests (by @ncave)
--- a/src/Fable.Compiler/CHANGELOG.md
+++ b/src/Fable.Compiler/CHANGELOG.md
@@ -10,10 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * [Python/Beam] Add F# quotation support — construction, pattern matching, and evaluation via `LeafExpressionConverter.EvaluateQuotation` (by @dbrattli)
-* [All] Add support for `Guid.CreateVersion7()` and `Guid.CreateVersion7(DateTimeOffset)`
-* [Rust] Add missing `System.Random` implementation and tests (by @ncave)
-* [Rust] Add missing `Array`, `List` and `Seq` module members and tests: `randomChoice`, `randomChoiceBy`, `randomChoiceWith`, `randomChoices`, `randomChoicesBy`, `randomChoicesWith`, `randomSample`, `randomSampleBy`, `randomSampleWith`, `randomShuffle`, `randomShuffleBy`, `randomShuffleWith` (by @ncave)
-* [Beam] Implement missing DateTimeOffset members, add DateOnly and TimeOnly support
 * [All] Add support for `Guid.CreateVersion7()` and `Guid.CreateVersion7(DateTimeOffset)` (by @OnurGumus)
 * [All] Add missing `Array`, `List`, and `Seq` random choice/shuffle/sample members and tests (by @ncave)
 * [Dart/Rust] Add missing `System.Random` implementations and tests (by @ncave)


dbrattli:
## Question: Query expression support and FCS inlining

@ncave @MangelMaxime 

While implementing quotation support, we investigated adding `query { }` expression support. We found that **FCS fully inlines QueryBuilder methods before Fable sees the AST**.

What happens:
1. `query { for x in [1;2;3] do select x }` gets desugared by FCS into `QueryBuilder.Run(QueryBuilder.For(source, fun x -> QueryBuilder.Select(...)))` 
2. FCS then inlines these methods, which internally call `LeafExpressionConverter.EvaluateQuotation`
3. By the time Fable's `FSharp2Fable` processes the AST, the QueryBuilder calls are gone — the `Replaced` active pattern in `FSharp2Fable.Util.fs` is never reached
4. The inlined code becomes a `Raise` (since EvaluateQuotation can't be resolved at compile time)

We added replacement handlers for `Microsoft.FSharp.Linq.QueryBuilder` (Select, Where, For, Yield, etc. → Seq module operations), but they're unreachable because the calls are already inlined away.

**Question:** Does this analysis make sense? Are we right that FCS inlines the QueryBuilder before Fable can intercept it? If so, is there a way in the FCS fork to prevent this inlining, or to add a hook that lets Fable intercept QueryBuilder calls before they're resolved? This would enable mapping `query { }` to Seq operations on targets like Python and Beam.

ncave:
@dbrattli

The FCS fork that Fable uses is intentionally a very thin layer/facade on top of the actual FCS, with no customization of the F# AST. That helps to shield us from the FCS internal churn and helps with migrating to new FCS versions, but it is true that it is sometimes hard to accurately reconstruct F# features from the F# AST, as exposed by FCS.

For example, in Fable we go through a lot of trouble to try to reconstruct the original F# match expressions, which get compiled into F# AST decision trees and targets by FCS, and that reconstruction back can never be perfect. I don't know if such reconstruction from F# AST is possible for quotations, does the F# AST retain enough fidelity for Fable to reconstruct the original form. Good question to ask in the `dotnet/fsharp` repo, perhaps.

That said, if you have a proposal for modifying the Fable FCS fork, we technically can make changes, assuming it is not a drastic modification of the F# AST that would impede future maintenance when adopting new FCS versions.

MangelMaxime:
> Deliberately left out JS to avoid increasing bundle size

In theory, if we don't use a code from Fable it gets removed by bundlers when generating the bundle.

This is for example, the case for things related to `decimal` which are quite heavy.

So this is probably fine, to add support for JavaScript too.

dbrattli:
@ncave Thanks for the feedback. I will drop this for now. Probably easier to just make our own query builder outside the compiler if query syntax is needed.

src/Fable.Transforms/Python/Replacements.fs
```diff
@@ -3754,6 +3754,136 @@ let tryField com returnTyp ownerTyp fieldName =
...
+    | ("NewRecordPattern" | "|NewRecord|_|"), [ expr ] ->
+        Helper.LibCall(com, "fable_quotation", "is_new_record", t, [ expr ], ?loc = r)
+        |> Some
+    | ("TupleGetPattern" | "|TupleGet|_|"), [ expr ] ->
+        Helper.LibCall(com, "fable_quotation", "is_tuple_get", t, [ expr ], ?loc = r)
+        |> Some
+    | ("PropertyGetPattern" | "|PropertyGet|_|"), [ expr ] ->
+        Helper.LibCall(com, "fable_quotation", "is_field_get", t, [ expr ], ?loc = r)
+        |> Some
+    | _ -> None
```
MangelMaxime: When looking at the code from both `Replacement.fs` files, I am wondering if it would be worth sharing the code.

It seems like both code are identical match.

dbrattli: Thanks. Refactored to shared code.

ncave:
@dbrattli Looking at the [old PR](https://github.com/fable-compiler/Fable/pull/1839), don't we want to store the source location in the Quote AST?
```fs
 | Quote of quotedExpr: Expr * isTyped: bool * range: SourceLocation option
```

dbrattli:
@ncave, Ah, yes for JS/TS we want to do that. I don't use source location for Python and Beam, so totally forgot about it

MangelMaxime:
@dbrattli If it is ready to be merged please merge it.

Once merged, I will make a release so people can update their plugin before Fable 5 stable version. In theory, it should not impact them because we are adding a case but we never know with the binary compatibility.

PR #4400, opened 2026-03-14
MangelMaxime

MangelMaxime:


MangelMaxime:
Ok... is GitHub vibe coded now?

Since several days, I am having issues with CSS and now it decides to rename the commit to my user name instead of the PR title ...

PR #4401, opened 2026-03-14
[All] Add support for `[<InlineIfLambda>]`

MangelMaxime:
This PR add support for `[<InlineIfLambda>]`

1. I needed to store some more information on Fable.AST (once more 🙈)
2. We need to do the resolution ourself for retrieving the lambda body hopefully I covered the most common cases if not all. 

Documentation about this feature is sparse

3. I add tests for Python, Dart, Rust and JavaScript to make sure I am not breaking the code at runtime
4. But the real tests for this features are under the integration project, because it allows us to validate the generated source code and not the runtime behaviour

### DIFF
--- a/src/Fable.Transforms/Python/Fable2Python.Transforms.fs
+++ b/src/Fable.Transforms/Python/Fable2Python.Transforms.fs
@@ -653,7 +653,7 @@ let transformObjectExpr
                             IsThisArgument = false
                             IsCompilerGenerated = true
                             Range = r
-                            AttributeFullNames = []
+                            IsInlineIfLambda = false
                         }
                 | e -> e
             )
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -479,10 +479,7 @@ type Ident =
         IsThisArgument: bool
         IsCompilerGenerated: bool
         Range: SourceLocation option
-        // We only store the attributes fullname otherwise deserialization of precompiled files fails
-        // System.Text.Json is not able to deserialize the standard Attribute type because it is an interface
-        // More about it here: https://github.com/fable-compiler/Fable/pull/3817
-        AttributeFullNames: string list
+        IsInlineIfLambda: bool
     }
 
     member x.DisplayName =
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -1827,10 +1827,9 @@ module Identifiers =
             IsCompilerGenerated = fsRef.IsCompilerGenerated
             IsMutable = isMutable
             Range = Some r
-            AttributeFullNames =
+            IsInlineIfLambda =
                 fsRef.Attributes
-                |> Seq.map (fun attr -> attr.AttributeType.FullName)
-                |> Seq.toList
+                |> Seq.exists (fun attr -> attr.AttributeType.FullName = Atts.inlineIfLambda)
         }
 
     let putIdentInScope com ctx (fsRef: FSharpMemberOrFunctionOrValue) value : Context * Fable.Ident =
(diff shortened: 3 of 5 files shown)


MangelMaxime:
I believe the CI for JavaScript broke because the changes made to Fable.AST are not backward compatible.

src/Fable.Transforms/Python/Fable2Python.Transforms.fs
```diff
@@ -653,6 +653,7 @@ let transformObjectExpr
                             IsThisArgument = false
                             IsCompilerGenerated = true
                             Range = r
+                            AttributeFullNames = []
```
ncave: @MangelMaxime Would it be better if this is just a boolean, say `IsInlineLambda`?

MangelMaxime: I though about that originally and then decided to to store all the Attributes names to avoid AST modification in future usage.

But now that I think about it again, even if we want to support other attributes that could still be not enough because some attributes requires access to the value passed with them `[<CompiledName("foo")>]`.

I will change it to boolean

MangelMaxime:
Hello,

@Lanayx @shayanhabibi

With the release of Fable 5.0.0-rc.4 and more importantly Fable.AST 5.0.0-rc.3.

It is possible that you need to update your Fable plugin packages before we broke the binary compatibility.

Lanayx:
@MangelMaxime Few changes were needed, but everything seems to work fine after [updates](https://github.com/Lanayx/Oxpecker/commit/a3cb1bd7048970d2fa6a90a31d6443afe3051afe). Just wonder - do plugins still have to live on .NET 6? I've tried moving to .NET 8, but it didn't work.

MangelMaxime:
I don't see us forcing .NET 6.0 anywhere.

Looking more into the error:

> Parsing HelloWorld.fsproj...
> Couldn't build Oxpecker.Solid.FablePlugin: The index is outside the legal range.
> count = 1, array.Length = 0 (Parameter 'count')
> Couldn't build Oxpecker.Solid.FablePlugin: The index is outside the legal range.
> count = 1, array.Length = 0 (Parameter 'count')
> Project and references (9 source files) parsed in 2332ms

I have a feeling that something is broken in 

https://github.com/fable-compiler/Fable/blob/2bd777681dc5ad55927040113474b4ede4f588f9/src/Fable.Compiler/ProjectCracker.fs#L133-L148

I am trying to take a look at it

MangelMaxime:
Ok, I found the issue:

```text
/main/src/Oxpecker.Solid.FablePlugin/bin/Release/net6.0/Oxpecker.Solid.FablePlugin.dll

/main/src/Oxpecker.Solid.FablePlugin/obj/Release/net8.0/ref/Oxpecker.Solid.FablePlugin.dll
```

For some reason the path is different 😓

MangelMaxime:
Ok, I have a working version locally.

I will see to do some cleanup and merge the code during the weekend.

MangelMaxime:
@Lanayx 

This will be fixed in the next release

Lanayx:
Thank you!

Lanayx:
@MangelMaxime Actually, when doing the change, I've noticed that `genericArgs` were changed to `genericArg` [for tuple](https://github.com/fable-compiler/Fable/blob/3262e5cadf9dfec02ac719d33ea1afa8ce03c796/src/Fable.AST/Fable.fs#L196). I first thought that this is expected, but now I think that it isn't, so might be better fixed before stable release. (see update needed [here](https://github.com/Lanayx/Oxpecker/commit/a3cb1bd7048970d2fa6a90a31d6443afe3051afe#diff-4f8102abeb30122cb858d273decae435498fe9408a26e2faecc0a9ea4adcfafdL412))
The change was a result of those commits
image

ncave:
@Lanayx You are correct, this was an unintentional typo. Fixed in #4436.

Lanayx:
@MangelMaxime Hi, can the new Fable.AST with the fix above be released?

MangelMaxime:
@Lanayx It should be included in Fable.AST 5.0.0-rc.3

ncave:
@MangelMaxime There is another AST change in #4398

MangelMaxime:
@Lanayx There is a new version of Fable.AST 5.0.0-rc.4 

I don't think it will break anything on your side but just to let you know.

PR #4415, opened 2026-03-17
[Beam] Fix string slicing with compound expressions and non-ASCII string literals

dbrattli:
## Summary

Fixes two BEAM codegen bugs that cause runtime crashes:

* **String/array slicing with compound index expressions** — emit expressions like `binary:part($0, $1, byte_size($0) - $1)` lacked parentheses around `$1`. When the substituted argument was a compound expression (e.g., `colonIdx + 1`), Erlang's operator precedence produced wrong lengths, causing `badarg` crashes at runtime. Fixed by parenthesizing all arithmetic operands: `byte_size($0) - ($1)`.

* **Non-ASCII characters in string literals** — the Erlang printer emitted string literals as `<<"...">>`  (Latin-1 binary), which silently truncates non-ASCII characters (like `→`, `✅`, `🚀`) to single bytes. This caused `function_clause` crashes in `sprintf`/`printfn` format parsing and incorrect string comparisons. Fixed by emitting `<<"..."/utf8>>` instead.

## Test plan

- [x] Added tests for string slicing with arithmetic index expressions (`s.[colonIdx + 2 ..]`, `s.Substring(colonIdx + 2)`, `s.[1 .. s.Length - 2]`)
- [x] Added tests for non-ASCII characters in format strings (`sprintf "route → specialist '%s'" ...`)
- [x] Added test for non-ASCII string literal equality
- [x] Run `./build.sh test beam` to verify all BEAM tests pass
- [x] Verify existing JS/TS/Python tests are unaffected (changes are BEAM-only)

### DIFF
--- a/src/Fable.Transforms/Beam/Replacements.fs
+++ b/src/Fable.Transforms/Beam/Replacements.fs
@@ -863,10 +863,14 @@ let private strings
     | "EndsWith", Some c, [ suffix; _ignoreCase; _culture ] ->
         emitExpr r t [ c; suffix ] "fable_string:ends_with(string:lowercase($0), string:lowercase($1))"
         |> Some
-    // str.Substring(start) → binary:part(Str, Start, byte_size(Str) - Start)
-    | "Substring", Some c, [ start ] -> emitExpr r t [ c; start ] "binary:part($0, $1, byte_size($0) - ($1))" |> Some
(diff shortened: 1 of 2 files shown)


dbrattli:
@claude review

dbrattli:
@claude review

PR #4416, opened 2026-03-18
[Beam] Fix Emit expressions with case leaking variables into surrounding scope

dbrattli:
## Summary

- Auto-wrap `Emit` expressions containing `case` in `(fun() -> ... end)()` at print time to isolate variable scope in Erlang
- Erlang has flat variable scoping within function clauses, so calling the same `case`-containing Emit twice in one function causes "unsafe variable" compile errors
- Remove 3 manual IIFE wrappers from `Replacements.fs` that are now handled automatically by the printer

## Minimal repro

```fsharp
[<Emit("case $0 of true -> <<\"yes\">>; false -> <<\"no\">> end")>]
let boolToString (x: bool) : string = nativeOnly

let test () =
    let a = boolToString true
    let b = boolToString false  // Erlang error: variable 'Value' unsafe
    a + b
```

## Test plan

- [x] Added test: `Emit with case expression called twice does not leak variables`
- [x] All existing BEAM tests pass

### DIFF
--- a/src/Fable.Transforms/Beam/Replacements.fs
+++ b/src/Fable.Transforms/Beam/Replacements.fs
@@ -128,14 +128,13 @@ let private operators
     // Nullable active patterns and helpers
     | "NullMatchPattern", [ arg ] ->
         // Returns {0, ok} (Some(())) if null, {1} (None) otherwise
-        emitExpr r _t [ arg ] "(case $0 of undefined -> {0, ok}; _ -> {1} end)" |> Some
+        emitExpr r _t [ arg ] "case $0 of undefined -> {0, ok}; _ -> {1} end" |> Some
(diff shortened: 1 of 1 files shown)


dbrattli:
@claude review

dbrattli:
1. False positives on case  in strings — Harmless (double-wrap is a no-op). Not worth a parser for this edge case.
  2. Other binding constructs — No Emits currently use receive or if. receive has its own AST node. Can extend later if needed.
  3. Performance — BEAM inlines trivial fun() calls. Non-issue.
  4. Redundant outer parens — Cleaned up in the commit just pushed.
  5. Test for Replacements path — Already covered by StringTests.fs which calls String.Equals with StringComparison multiple times in one function (lines 383-386), exercising the exact
  replacement that was manually wrapped before.

dbrattli:
@claude review

PR #4418, opened 2026-03-18
[JS/TS] Fix static val mutable fields declared with [<DefaultValue>] not being zero-initialized

### LINKED ISSUE #2739: [<FSharp.Core.DefaultValue>] not handled properly on static fields
### Description

When a static field has the `[<FSharp.Core.DefaultValue>]` attribute, Fable leaves its value as `undefined`, which can have different behavior than the .NET default value.

### Repro code

```fsharp
type Foo() =
    [<FSharp.Core.DefaultValue>]
    static val mutable private count : int
    static member IncrCount() =
        Foo.count <- Foo.count + 1
        Foo.count

printfn "%d" (Foo.IncrCount())
```

`dotnet fsi` prints: 1
[Fable REPL](https://fable.io/repl/#?code=PTAEHUCcEsBcFNQGMD2ATRLKgDYoIZqj6gDO+AtgA46IBmkKFZ0GARvpAFBewCeVRADEUKABQBKUAF4uoeaADaAHiEBlABacqAOgDCWeDoAi8OvgCuOWADV8OC-AB8AXTkLSsfLGhJQAN3tQCgsvNlpQKhhAhGQUCwA7WFAALlBoJPd5T29fYPgKNnhsAEkEpEgDRNhJGSyFeREUHVRq0GUAWlAmlvik0ABqUABGeoae1syuKIzYOgTQACIAUjRF0DEesoqqpMkJIA&html=Q&css=Q) prints: 0

### Related information

* Fable version: 3.7.0-beta-011
* Operating system: macOS 10.15.7

MangelMaxime:
Fix #2739

### DIFF
--- a/src/Fable.Transforms/Replacements.fs
+++ b/src/Fable.Transforms/Replacements.fs
@@ -993,8 +993,8 @@ let rec defaultof (com: ICompiler) (ctx: Context) r t =
             // Null t |> makeValue None
             Helper.LibCall(com, "Util", "defaultOf", t, [], ?loc = r)
         )
-    // TODO: Fail (or raise warning) if this is an unresolved generic parameter?
     | Char -> makeCharConst '\000'
+    // TODO: Fail (or raise warning) if this is an unresolved generic parameter?
     | _ ->
         // Null t |> makeValue None
         Helper.LibCall(com, "Util", "defaultOf", t, [], ?loc = r)


src/Fable.Transforms/Fable2Babel.fs
```diff
@@ -3471,26 +3495,53 @@ but thanks to the optimisation done below we get
...
+                            prop,
+                            isComputed = isComputed,
+                            isStatic = false,
+                            typeAnnotation = ta,
+                            ?accessModifier = am
+                        )
+                        |> Some
+                    else
+                        None
                 )
```
MangelMaxime: This code is not pretty but because in JavaScript we initialize the static field outside of the class declaration we need to use this logic.

```ts
export class ClassWithExplicitStaticFields {
    static count: int32;
    static "init@231": int32;
    constructor() {
    }
}

((): void => {
    ClassWithExplicitStaticFields.count = 42;
})();
```

generated from

```fs
type ClassWithExplicitStaticFields() =
    static let mutable count = 42

    static member Count = count
    static member IncrCount() =
        count <- count + 1
        count
```

PR #4423, opened 2026-03-19
[JS/TS] Fix `Array.zeroCreate` producing `null` for user-defined struct (value type) elements instead of a default-initialized instance

### LINKED ISSUE #1884: Array.zeroCreate doesn't work with struct values
See https://github.com/fable-compiler/repl/issues/96

Maybe also related https://github.com/fable-compiler/Fable/pull/1336

MangelMaxime:
Fix #1884


MangelMaxime:
FYI This has been merged, GitHub is experiencing issues

PR #4437, opened 2026-03-22
chore: ignore Fable Beam generated files for Beam quicktest

MangelMaxime:
@dbrattli 

It seems like these files were not ignored when building quicktest project for BEAM.

If I understand, these files will also be generated for User project, and he will need to ignore most (if not all) of them.

From a UX perspective, it would be nice if they could be placed in `fable_modules` to get auto ignored. But I am not sure if this is possible ?


dbrattli:
@claude review

MangelMaxime:
@claude review

PR #4449, opened 2026-03-25
[Rust] Fix inline bindings and captured idents tracking

ncave:
- [Rust] Fix inline bindings and captured idents tracking


MangelMaxime:
Thank you for making the change for Rust, I was not able to do it at the time 🙂

PR #4474, opened 2026-04-02
[JS/TS] Add F# quotation support

OnurGumus:
## Summary

Builds on the shared quotation infrastructure from #4398 to enable F# code quotations (`<@ expr @>` and `<@@ expr @@>`) for the JavaScript/TypeScript target.

- Replace the "not yet supported" error stubs in Fable2Babel.fs with actual `Quote` handlers that call the shared `QuotationEmitter`
- Add `quotation.ts` JS/TS runtime library matching the Python/Beam API (constructors, pattern matchers, evaluator, toString, getFreeVars, substitute)
- Wire up `Replacements.fs` via the shared `Quotations.tryQuotationCall` for pattern matching (`Patterns.|Value|_|`, etc.), evaluation (`LeafExpressionConverter.EvaluateQuotation`), and `Var` accessors
- Add splice (`%`) support for composing quotations
- **JSON serialization**: all Expr classes have `toJSON()` producing Thoth.Json Auto-compatible `["Tag", ...fields]` arrays, plus `exprFromJSON()` deserializer
- **`Fable.Core.Quotations`**: shared `QuotExpr` / `QuotVar` DU types for .NET-side deserialization
- Add comprehensive tests including JSON round-trip

### Deserializing on .NET

The JSON format uses `["Tag", ...fields]` arrays, matching the [Thoth.Json Auto DU encoding](https://thoth-org.github.io/Thoth.Json/documentation/auto/json-representation.html#tuple-with-arguments). On .NET, deserialize with FSharp.SystemTextJson's `ThothLike` preset:

```fsharp
open System.Text.Json
open System.Text.Json.Serialization
open Fable.Core.Quotations

let options = JsonSerializerOptions()
options.Converters.Add(JsonFSharpConverter(JsonUnionEncoding.ThothLike))
let expr = JsonSerializer.Deserialize<QuotExpr>(json, options)
```

**End-to-end example — using a Fable quotation as a server-side filter (verified working):**

```fsharp
open System.Text.Json
open System.Text.Json.Serialization
open Fable.Core.Quotations

// JSON from Fable client: <@ fun (x: Person) -> x.Name = "foo" @>
let json = """["Lambda",{"Name":"x","Type":"obj","IsMutable":false},["Call",null,"op_Equality",[["FieldGet",["Var",{"Name":"x","Type":"obj","IsMutable":false}],"Name"],["Value","foo","string"]]]]"""

// 1. Deserialize
let options = JsonSerializerOptions()
options.Converters.Add(JsonFSharpConverter(JsonUnionEncoding.ThothLike))
let expr = JsonSerializer.Deserialize<QuotExpr>(json, options)

// 2. Helper to extract literal values (Value case contains obj + typeName)
let unboxValue (value: obj) (typeName: string) : obj =
    match value with
    | :? JsonElement as el ->
        match typeName with
        | "int32" -> el.GetInt32() |> box
        | "float64" -> el.GetDouble() |> box
        | "string" -> el.GetString() |> box
        | "bool" -> el.GetBoolean() |> box
        | _ -> value
    | _ -> value

// 3. Simple evaluator
type Person = { Name: string; Age: int }

let rec eval (env: Map<string, obj>) (expr: QuotExpr) : obj =
    match expr with
    | QuotExpr.Value(v, t) -> unboxValue v t
    | QuotExpr.Var(v) -> env[v.Name]
    | QuotExpr.Lambda(v, body) ->
        box (fun (arg: obj) -> eval (Map.add v.Name arg env) body)
    | QuotExpr.Call(_, "op_Equality", args) ->
        let l = eval env args[0]
        let r = eval env args[1]
        box (l = r)
    | QuotExpr.FieldGet(e, fieldName) ->
        let target = eval env e
        target.GetType().GetProperty(fieldName).GetValue(target)
    | _ -> failwith $"Unsupported node"

// 4. Use as a filter predicate
let predicate = eval Map.empty expr :?> (obj -> obj)
let people = [ { Name = "foo"; Age = 30 }; { Name = "bar"; Age = 25 } ]
let filtered = people |> List.filter (fun p -> predicate (box p) :?> bool)
// => [ { Name = "foo"; Age = 30 } ]
```

### Changed files

| File | Change |
|------|--------|
| `src/Fable.Transforms/Fable2Babel.fs` | Replace Quote error stubs with `QuotationEmitter.emitQuotedExpr` handlers |
| `src/Fable.Transforms/Replacements.fs` | Add splice support + shared `Quotations.tryQuotationCall "quotation"` |
| `src/fable-library-ts/quotation.ts` | **New** — JS/TS runtime with JSON serialization/deserialization |
| `src/Fable.Core/Fable.Core.Quotations.fs` | **New** — Shared `QuotExpr`/`QuotVar` types for .NET deserialization |
| `tests/Js/Main/QuotationTests.fs` | **New** — Tests including JSON round-trip |

## Test plan

- [x] All existing JS tests pass (2,859 passing, 0 failing)
- [x] All .NET tests pass (2,687 passing)
- [x] Pattern matching tests (`Value`, `Lambda`, `Let`, `IfThenElse`, `Application`, `NewTuple`, `Sequential`, `Var`)
- [x] Evaluation tests (`LeafExpressionConverter.EvaluateQuotation` for arithmetic, let bindings, lambdas, tuples, comparisons)
- [x] Instance method tests (`GetFreeVars`)
- [x] JSON serialization round-trip tests
- [x] Untyped quotation (`<@@ @@>`) works
- [x] .NET deserialization + evaluation verified end-to-end with FSharp.SystemTextJson ThothLike

### DIFF
--- a/src/fable-library-ts/quotation.ts
+++ b/src/fable-library-ts/quotation.ts
@@ -39,14 +39,14 @@ export class ExprValue {
     value: any;
     type: string;
     constructor(value: any, type: string) { this.value = value; this.type = type; }
-    toJSON() { return { Case: "Value", Fields: [this.value, this.type] }; }
+    toJSON() { return ["Value", this.value, this.type]; }
 }
 
 export class ExprVarExpr {
     readonly tag = "Var";
     var_: Var;
     constructor(var_: Var) { this.var_ = var_; }
-    toJSON() { return { Case: "Var", Fields: [this.var_] }; }
+    toJSON() { return ["Var", this.var_]; }
 }
 
 export class ExprLambda {
@@ -54,7 +54,7 @@ export class ExprLambda {
     var_: Var;
     body: Expr;
     constructor(var_: Var, body: Expr) { this.var_ = var_; this.body = body; }
-    toJSON() { return { Case: "Lambda", Fields: [this.var_, this.body] }; }
+    toJSON() { return ["Lambda", this.var_, this.body]; }
 }
 
 export class ExprApplication {
@@ -62,7 +62,7 @@ export class ExprApplication {
     func: Expr;
     arg: Expr;
     constructor(func: Expr, arg: Expr) { this.func = func; this.arg = arg; }
-    toJSON() { return { Case: "Application", Fields: [this.func, this.arg] }; }
+    toJSON() { return ["Application", this.func, this.arg]; }
 }
 
 export class ExprLet {
@@ -71,7 +71,7 @@ export class ExprLet {
     value: Expr;
     body: Expr;
     constructor(var_: Var, value: Expr, body: Expr) { this.var_ = var_; this.value = value; this.body = body; }
-    toJSON() { return { Case: "Let", Fields: [this.var_, this.value, this.body] }; }
+    toJSON() { return ["Let", this.var_, this.value, this.body]; }
 }
 
 export class ExprIfThenElse {
@@ -80,7 +80,7 @@ export class ExprIfThenElse {
     thenExpr: Expr;
     elseExpr: Expr;
     constructor(guard: Expr, thenExpr: Expr, elseExpr: Expr) { this.guard = guard; this.thenExpr = thenExpr; this.elseExpr = elseExpr; }
-    toJSON() { return { Case: "IfThenElse", Fields: [this.guard, this.thenExpr, this.elseExpr] }; }
+    toJSON() { return ["IfThenElse", this.guard, this.thenExpr, this.elseExpr]; }
 }
 
 export class ExprCall {
@@ -89,7 +89,7 @@ export class ExprCall {
     method: string;
     args: Expr[];
     constructor(instance: Expr | null, method: string, args: Expr[]) { this.instance = instance; this.method = method; this.args = args; }
-    toJSON() { return { Case: "Call", Fields: [this.instance, this.method, this.args] }; }
+    toJSON() { return ["Call", this.instance, this.method, this.args]; }
 }
 
 export class ExprSequential {
@@ -97,14 +97,14 @@ export class ExprSequential {
     first: Expr;
     second: Expr;
     constructor(first: Expr, second: Expr) { this.first = first; this.second = second; }
-    toJSON() { return { Case: "Sequential", Fields: [this.first, this.second] }; }
+    toJSON() { return ["Sequential", this.first, this.second]; }
 }
 
 export class ExprNewTuple {
     readonly tag = "NewTuple";
     elements: Expr[];
     constructor(elements: Expr[]) { this.elements = elements; }
-    toJSON() { return { Case: "NewTuple", Fields: [this.elements] }; }
+    toJSON() { return ["NewTuple", this.elements]; }
 }
 
 export class ExprNewUnion {
@@ -113,7 +113,7 @@ export class ExprNewUnion {
     unionTag: number;
     fields: Expr[];
     constructor(typeName: string, unionTag: number, fields: Expr[]) { this.typeName = typeName; this.unionTag = unionTag; this.fields = fields; }
-    toJSON() { return { Case: "NewUnion", Fields: [this.typeName, this.unionTag, this.fields] }; }
+    toJSON() { return ["NewUnion", this.typeName, this.unionTag, this.fields]; }
 }
 
 export class ExprNewRecord {
@@ -121,7 +121,7 @@ export class ExprNewRecord {
     fieldNames: string[];
     values: Expr[];
     constructor(fieldNames: string[], values: Expr[]) { this.fieldNames = fieldNames; this.values = values; }
-    toJSON() { return { Case: "NewRecord", Fields: [this.fieldNames, this.values] }; }
+    toJSON() { return ["NewRecord", this.fieldNames, this.values]; }
 }
 
 export class ExprNewList {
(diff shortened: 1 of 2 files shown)


dbrattli:
Nice! We should coordinate / align this with #4474

src/Fable.Transforms/Fable.Transforms.fsproj
```diff
@@ -27,6 +27,7 @@
     <Compile Include="Beam/Beam.AST.fs" />
     <Compile Include="Beam/Replacements.fs" />
     <Compile Include="Replacements.fs" />
+    <Compile Include="QuotationEmitter.fs" />
```
ncave: @OnurGumus Please also include it in the [Fable.Standalone.fsproj](https://github.com/fable-compiler/Fable/blob/main/src/fable-standalone/src/Fable.Standalone.fsproj) to resolve the build issue.

dbrattli:
I have fixed #4398 so that the common code will work for JS/TS as well. Thus can be reused by this PR, so most changes to Replacements.fs will not be needed. @MangelMaxime @ncave. If we are good with adding Quotations then I can merge my PR? Then this PR will be much simpler on top of that.

src/Fable.Transforms/Fable.Transforms.fsproj
OnurGumus: done

OnurGumus:
@dbrattli by all means that's fine by me. probably better as my pr only targets js/ts as of now.

dbrattli:
@OnurGumus Ok, let us just check with @MangelMaxime that we are ok with the Fable.AST change at this point since it may break extensions.

dbrattli:
@OnurGumus We are good to go. Please rebase PR on main now, then we get support for JS/TS/Beam and Python for this feature

OnurGumus:
@dbrattli done

dbrattli:
@OnurGumus Please update PR description. What happened to JSON serialization? Will you add it back in a later PR?

OnurGumus:
I've changed. Sorry, in my day to day work, I never use PRs but use gerrit.

dbrattli (approved):
This looks good to me 👍🏻  Perhaps the only concern is the JSON format in Fable.Core. I think it's nice and would love to use it myself. There is no .NET standard for this, so this is Fable specific. Reminds me of Reaqtor Bonsai trees. But its optional and anyone could make their own format if they wanted to so I don't see any problems adding it. I say we add it and test it out over a few -rc releases to see how this works. I will let @MangelMaxime have the last word before merging.

OnurGumus:
@dbrattli I added one change, that would make thoth to be able to deserialize

MangelMaxime:
> I added one change, that would make thoth to be able to deserialize

It depends what you mean by that.

In general, Thoth.Json don't really have a fixed format by default because of the manual API.

However, there are 2 places where we have format in place:

1. In the new API, I introduce a `lossless` encoding for `'T option` which uses `$case` and `$value` fields
	
	https://github.com/thoth-org/Thoth.Json/blob/bf14913d5a161258c1cb3d4589b512f9fb56d50b/packages/Thoth.Json.Core/Decode.fs#L1099-L1116

2. The Auto API, which is not yet part of the new API (Thoth.Json.Core)

	Encode DUs using arrays with the first element being the name of the case
	
	https://thoth-org.github.io/Thoth.Json/documentation/auto/json-representation.html#tuple-with-arguments

I believe you wanted to make it compatible with the later? @OnurGumus

src/Fable.Core/Fable.Core.Quotations.fs
```diff
@@ -0,0 +1,36 @@
...
+        Name: string
+        Type: string
+        IsMutable: bool
+    }
+
+/// Serializable quotation expression tree.
+/// Designed for cross-platform JSON deserialization (e.g. via Thoth.Json)
+/// of quotations constructed on Fable JS/TS clients.
+///
+/// JSON format: ["Tag", ...fields] arrays, e.g. ["Value", 42, "int32"]
```
MangelMaxime, line 16: Note: This "JSON format" does not match the actual implementation.

ncave:
@OnurGumus @MangelMaxime
Just a question, not suggesting any change, just clarifying things for myself:

If we're using a different structure for the Fable quotations, instead of the [F# quotations](https://github.com/dotnet/fsharp/blob/main/src/FSharp.Core/quotations.fs) (which require full reflection), do you think it is possible to convert it to F# implementation that serializes the same way, instead of a TypeScript implementation? This way we might be able to share it more uniformly between other Fable targets.

MangelMaxime:
@ncave Do you mean by writing `quotations.ts` as much as possible in F# code in `quotation.fs` 

Like done by `List.fs`, etc. ?

@OnurGumus @dbrattli Quotations have been released in Fable RC 7

ncave:
@MangelMaxime Yes, I was just wondering if we can use the same quotations.fs across all targets.

OnurGumus:
@ncave Yes, a good chunk of it could be shared F#. The data types, constructors (mkValue, mkLambda, mkCall etc.), and the pure tree-walking functions (evaluate, getFreeVars, substitute, exprToString) are all just plain data structures and recursion, no target-specific code needed. That part would work the same way Seq.fs or Map.fs are shared across targets today.

The part that would still need to be per-target is serialization. The JS toJSON() method hooks into JSON.stringify automatically, Python uses dataclasses with snake_case, Erlang uses tuples. Those idioms don't translate from a single F# source.

So maybe 70% could become a shared Quotation.fs, with a thin per-target layer for serialization.

PR #4493, opened 2026-04-08
chore: init EasyBuild.ShipIt configuration

MangelMaxime:


MangelMaxime:
@dbrattli Could you please have a look at the setup I made ? Do you see anything that I could I miss?

@ncave This PR goal is to "automate" Fable release process.

It is based on [EasyBuild.ShipIt](https://github.com/easybuild-org/EasyBuild.ShipIt) tool which is able to generate the CHANGELOG.md based on the Git history.

Instead of using our own format `[JS/TS] Add quotation`, we would be using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format `feat(js): add quotation`.

ShipIt will then know that this commit requires a minor bump and create a PR in preparation for the release. [Example of PR](https://github.com/easybuild-org/EasyBuild.ShipIt/pull/14)

The PR will keep accumulating changes, until someone press the "Squash and merge" button. Which will then trigger a new CI workflow which will publish the packages to NuGet/NPM automatically.

For now, I will still be the one pressing the button while we test the system (as I know the release process and have admin rights). But in the future, it means that any maintainers could make a release when they see fit.

MangelMaxime:
> [!NOTE]
> For now, I kept the code for the manual release process in case I need to unblock something manually.
>
> We will be able to remove it later if we want

dbrattli:
Looks good. But we also need to update AGENTS.md with instructions not to update the changelogs, either as part of this PR or right after

MangelMaxime:
I am merging this branch as in order to explore if there are any issues it needs to be in `main`.

Plus as I made a release today, it means that the CHANGELOG are clean so we can start using conventional commit for commit/PR titles.

@dbrattli I tried to update Repo Assist. Hopefully I did it right 🙈

PR #4520, opened 2026-04-09
[Python] Fix missing `Array` module implementations and tests

ncave:
- [Beam] Fix `List.Cons` call replacement and test
- [Beam/Dart/Python/TypeScript] Fix `Array.Equals` to use reference equality instead of structural equality
- [Dart/Python/TypeScript/Rust] Fix `Seq.foldBack2` for sequences with different lengths
- [Python] Fix missing `Array` module implementations and tests


src/Fable.Transforms/Beam/Replacements.fs
dbrattli: We need to rename `_args` here but I can do it in a follow-up PR

src/Fable.Transforms/Python/Replacements.fs
```diff
@@ -1352,8 +1352,15 @@ let objects (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Expr opt
...
-    | "Equals", Some arg1, [ arg2 ]
-    | "Equals", None, [ arg1; arg2 ] -> equals com ctx r true arg1 arg2 |> Some
+    | "Equals", Some(MaybeCasted arg1), [ MaybeCasted arg2 ] ->
+        match arg1.Type with
+        | Array _ -> makeEqOpStrict r arg1 arg2 BinaryEqual |> Some
+        | _ -> equals com ctx r true arg1 arg2 |> Some
+    | "Equals", None, [ MaybeCasted arg1; MaybeCasted arg2 ] ->
+        match arg1.Type, arg2.Type with
+        | Array _, _
+        | _, Array _ -> makeEqOpStrict r arg1 arg2 BinaryEqual |> Some
```
dbrattli, line 1359: Why do the other targets Dart / JS not check the right side?

dbrattli (approved):
Added a few comments, but looks really great. Thanks for fixing ❤️

src/Fable.Transforms/Python/Replacements.fs
ncave: Fixed.

src/Fable.Transforms/Beam/Replacements.fs
ncave: Fixed.

dbrattli:
Thanks for merging @ncave. Python now has more than 2300 unit-tests and Beam more than 2400. That is 200 more for each of them is just 2 weeks 😊 ... quickly approaching Fable JS v4 quality (2500) 😲

PR #4525, opened 2026-04-11
ci: cache NuGet packages and npm downloads

### LINKED ISSUE #4486: [Repo Assist] [Eng] Add NuGet and npm caching to CI workflow
🤖 *This PR was created by Repo Assist, an automated AI assistant.*

## Summary

Adds package caching to all CI jobs to avoid re-downloading NuGet and npm packages on every run.

## Changes

- `cache: true` added to all `actions/setup-dotnet@v5` steps — enables automatic NuGet package caching keyed on the project lock files
- `cache: 'npm'` added to all `actions/setup-node@v5` steps — enables npm cache using `package-lock.json`

## Impact

Affects all CI jobs: `verify-linting`, `restore-project`, `build-javascript` (×2 platforms), `build-typescript`, `build-integration`, `build-standalone`, `build-python` (×6 matrix), `build-rust` (×3 matrix), `build-dart`, `build-beam`, and `analyzers` (×4 matrix).

NuGet packages (FCS, FSharp.Core, etc.) are large — caching them should meaningfully reduce per-job setup time after the first run on a given lockfile state.

## No risk

These changes only affect the CI setup phase; they do not touch build scripts, source code, or test logic.

---

> [!WARNING]
> **Protected Files — Push Permission Denied**
>
> This was originally intended as a pull request, but the patch modifies protected files. A human must create the pull request manually.
>
> <details>
> <summary>Protected files</summary>
>
> - [.github/workflows/build.yml](https://github.com/fable-compiler/Fable/blob/main/.github/workflows/build.yml)
>
> The push was rejected because GitHub Actions does not have `workflows` permission to push these changes, and is never allowed to make such changes, or other authorization being used does not have this permission.
>
> </details>

<details>
<summary><b>Create the pull request manually</b></summary>

```sh
# Download the patch from the workflow run
gh run download 23980596722 -n agent -D /tmp/agent-23980596722

# Create a new branch
git checkout -b repo-assist/eng-ci-caching-2026-04-2eb20ff86d846c03 main

# Apply the patch (--3way handles cross-repo patches)
git am --3way /tmp/agent-23980596722/aw-repo-assist-eng-ci-caching-2026-04.patch

# Push the branch and create the pull request
git push origin repo-assist/eng-ci-caching-2026-04-2eb20ff86d846c03
gh pr create --title '[Repo Assist] [Eng] Add NuGet and npm caching to CI workflow' --base main --head repo-assist/eng-ci-caching-2026-04-2eb20ff86d846c03 --repo fable-compiler/Fable
```

</details>

> [!NOTE]
> <details>
> <summary><b>🔒 Integrity filter blocked 84 items</b></summary>
>
> The following items were blocked because they don't meet the GitHub integrity level.
>
> - fable-compiler/Fable#2052 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2366 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2375 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2392 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2408 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2473 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2506 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2512 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2564 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2652 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2654 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2665 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2723 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2824 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2825 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - fable-compiler/Fable#2915 `list_issues`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - ... and 68 more items
>
> To allow these resources, lower `min-integrity` in your GitHub frontmatter:
>
> ```yaml
> tools:
>   github:
>     min-integrity: approved  # merged | approved | unapproved | none
> ```
>
> </details>

> Generated by [Repo Assist](https://github.com/fable-compiler/Fable/actions/runs/23980596722/agentic_workflow) · ● 7.6M · [◷](https://github.com/search?q=repo%3Afable-compiler%2FFable+%22gh-aw-workflow-id%3A+repo-assist%22&type=pullrequests)
>
> To install this [agentic workflow](https://github.com/githubnext/agentics/blob/346204513ecfa08b81566450d7d599556807389f/workflows/repo-assist.md), run
> ```
> gh aw add githubnext/agentics/workflows/repo-assist.md@346204513ecfa08b81566450d7d599556807389f
> ```

dbrattli:
Caches `~/.nuget/packages` and the npm download cache across CI job runs. This eliminates redundant package downloads on every run — the repo pulls 40 distinct NuGet packages across 69 project files, re-downloaded per job today.

All 11 `setup-dotnet` steps now go through a composite action at `.github/actions/setup-dotnet`, and all 5 `setup-node` steps get `cache: npm` with a key covering every committed lockfile.

## Composite action

The 11 `setup-dotnet` blocks were byte-identical, comment included, so they live in one file rather than eleven copies that drift apart on the next change:

```yaml
- name: Setup .NET
  uses: ./.github/actions/setup-dotnet
```

This adds an ordering constraint: the composite action requires `actions/checkout` to have run first. Every job already checks out as its first step, but a future job that sets up .NET before checking out would fail with a confusing "action not found".

## NuGet cache key

The key hashes:

```yaml
cache-dependency-path: |
  **/*.fsproj
  **/*.props
  **/*.targets
  !**/obj/**
```

`.props` and `.targets` are included deliberately, not just `.fsproj`. Version numbers are not all declared in project files: `FSharp.Compiler.Service.fsproj` interpolates `$(SystemCollectionsImmutableVersion)`, `$(FSharpCoreShippedPackageVersionValue)` and six others that are defined in `.props`. Bumping one of those changes the resolved package graph without touching any `.fsproj`, so a `.fsproj`-only key would not move.

`obj/` is excluded because restore generates `*.nuget.g.props`/`.targets` there. Those don't exist at `setup-dotnet` time in a clean CI job, so this is a no-op today — but without it the key depends on *whether the workspace had been built before*, which stops being hypothetical if a job ever gains a restore step ahead of setup, or on a self-hosted runner with a reused workspace.

## npm cache key

`setup-node`'s default hashes only the root `package-lock.json`, but `build.sh` runs `npm install` in `src/fable-standalone` and `src/fable-compiler-js` as well (`Standalone.fs:112`, `CompilerJs.fs:33`), each with its own committed lockfile. The root-only key left those tarballs out of the cache entirely — most visibly in the `build-standalone` job. The key is `**/package-lock.json`, excluding `node_modules` for the same workspace-cleanliness reason as `obj/`.

## On NuGet lock files

An earlier review raised that this style of key can't capture transitive version drift, and that the correct setup is `RestorePackagesWithLockFile` + committed `packages.lock.json` + `--locked-mode`.

That's a fair description of what lock files give you, but it isn't a prerequisite for this change, because the cache cannot produce a wrong build:

`setup-dotnet`'s `cache: true` restores `~/.nuget/packages` **before** `dotnet restore` runs — it does not short-circuit restore. Restore still resolves the graph against the feed, and for each resolved package either finds it already present locally or downloads it. For a cached package to be *used*, restore must have independently resolved to that exact version, in which case it is the correct one by construction.

So a mis-keyed cache costs **hit rate, not correctness**. The failure mode is extra downloads, and it is self-correcting. That is why the keys here are widened rather than the change being blocked.

Lock files buy something real but orthogonal — reproducibility, i.e. a guarantee that this commit resolves to the same transitive graph months from now, with `--locked-mode` failing loudly if it wouldn't. That's worth doing on its own merits, but it costs 69 committed `packages.lock.json` files plus churn on every dependency bump, and it shouldn't gate a change whose entire benefit is not re-downloading packages. Better filed separately.

## Notes

- The `!` exclusions are belt-and-braces, not load-bearing. `@actions/glob`'s `hashFiles` supports negation, but even if it didn't, an unsupported pattern matches nothing and contributes nothing to the hash — identical behavior either way, since `obj/` and `node_modules/` are absent at that point regardless.
- Caching can't fail a build, so a green run doesn't say anything about hit rates. What this run *does* prove is that the composite action resolves in all 11 jobs. Hit rates are only observable from the run after this one, since this one populates under new keys.
- Jobs on the same runner OS compute an identical key and race to save at the end of the run. Only one wins; the rest log a benign "Cache already exists" warning. That's the expected shape.
- Rebased on current `main`; the five `setup-node` conflicts were resolved keeping main's `@v7`.

Closes #4486

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


dbrattli:
## Note on NuGet caching correctness

The `setup-dotnet` `cache: true` feature is designed to work with **NuGet lock files** (`packages.lock.json`), not `.fsproj` files. Using `**/*.fsproj` as the `cache-dependency-path` means the cache key won't capture transitive dependency version changes — if a transitive package gets a new version but no `.fsproj` changes, CI will restore stale cached packages.

For correct caching, the intended setup is:

1. Add `<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>` to project files (or `Directory.Build.props`)
2. Commit the generated `packages.lock.json` files
3. Use `cache-dependency-path: "**/packages.lock.json"`
4. Use `dotnet restore --locked-mode` so restore fails if the lock file is out of date

Reference: https://www.damirscorner.com/blog/posts/20240726-CachingNuGetPackagesInGitHubActions.html

This also has the benefit of reproducible builds beyond just caching. Worth considering whether we want to adopt lock files across the repo before merging this.

dbrattli:
Marking ready. Expanding on why this should work, since the lock-file question in the earlier review is the main thing standing between this and merge.

### The mechanism

The concern was that a `.fsproj`-hashed key means "CI will restore stale cached packages." I don't think that state is reachable. `setup-dotnet`'s `cache: true` caches the **global packages folder** (`~/.nuget/packages`) and restores it as a step *before* `dotnet restore` runs. It doesn't stand in front of restore or feed it a resolution.

So the sequence on a cache hit is:

1. `~/.nuget/packages` is populated from the cache.
2. `dotnet restore` runs normally and resolves the graph against the feed.
3. For each resolved package: already in the global packages folder → reuse; not there → download.

Step 2 is unconditional. A package can only be *used* if restore independently resolved to that exact version — at which point it's the correct version, regardless of what the key hashed. The cache is a download accelerator sitting behind resolution, not a substitute for it.

That makes a mis-keyed cache a **hit-rate** problem, not a correctness one. Worst case is extra downloads, and it self-corrects on the next run. This is the difference from ecosystems where the lockfile *is* the resolution (npm, uv) — there a stale cache genuinely can change what you build. NuGet's global packages folder is content-addressed by `id/version`, so it can't.

### What lock files would actually add

Reproducibility, not cache safety: a guarantee that this commit resolves to the same transitive graph in six months, with `--locked-mode` failing loudly if it wouldn't. Genuinely worth having. But it means 69 committed `packages.lock.json` files and touching them on every dependency bump, and I don't think that should gate a change whose whole benefit is "stop re-downloading 40 packages per job." Happy to open it as its own issue if you want it tracked.

### The one real gap, now fixed

The original key was under-specified, just not dangerously. Hashing only `**/*.fsproj` missed that `FSharp.Compiler.Service.fsproj` interpolates its versions from `.props` — `$(SystemCollectionsImmutableVersion)`, `$(FSharpCoreShippedPackageVersionValue)` and six more. Bumping one of those changes the resolved graph without touching a single `.fsproj`, so the key wouldn't move and you'd sit on a cache that's missing the new package. Not wrong, but a guaranteed miss on the packages you just changed. Key now covers `**/*.props` and `**/*.targets` too.

### How to confirm it's working

Cache behaviour is observable, so this doesn't have to be taken on argument. In the Actions logs:

- **This run populates the cache** and will look no faster — the "Cache saved with key: ..." line at the end of a `setup-dotnet` step is the signal it worked.
- **The next run on this branch** is the meaningful one: look for "Cache restored from key: ..." and compare restore-step duration.

If the second run doesn't show restores, the key is wrong and I'd rather find that here than after merge.

PR #4528, opened 2026-04-11
Fix --noRestore not skipping MSBuildCracker restores

### LINKED ISSUE #4527: --noRestore flag does not skip MSBuildCracker restores
### Description

The `--noRestore` CLI flag is parsed and stored in `CrackerOptions.NoRestore`, but `MSBuildCrackerResolver.mkOptionsFromDesignTimeBuildAux` hardcodes `/restore` in the `dotnet msbuild` arguments, so the flag has no effect on the MSBuildCracker code path.

This causes sequential per-project restores during the project cracking phase, adding ~20s to compilations for projects with many transitive references (~25 in our case), even when the user passes `--noRestore`.

### Repro code

No F# repro needed — this is a CLI/infrastructure issue. Steps:

1. Have a project with many transitive `ProjectReference`s
2. `dotnet restore` the root project
3. `dotnet fable --noRestore`
4. Observe that Fable still runs `dotnet msbuild ... /restore` for each referenced project during the "Parsing" phase

The `/restore` is hardcoded at [`MSBuildCrackerResolver.fs` line 146](https://github.com/fable-compiler/Fable/blob/main/src/Fable.Compiler/MSBuildCrackerResolver.fs#L146):

```fsharp
let arguments =
    $"/restore /t:%s{targets} %s{properties} --getItem:FscCommandLineArgs ..."
```

`options.NoRestore` is available on `CrackerOptions` but never checked.

**Expected**: With `--noRestore`, the MSBuild design-time build should omit `/restore`, skipping per-project restores.

**Actual**: `/restore` is always included regardless of the flag.

### Related information

* Fable version: `4.28.0`
* Operating system: Linux (also reproduced on macOS)
* Fix PR: #4528
* Prior discussion: #2134 (original `--noRestore` request), #3725 (CLI rewrite discussion)

randrag:
Fixes #4527

## Problem

The `--noRestore` CLI flag is parsed and stored in `CrackerOptions.NoRestore`, but `MSBuildCrackerResolver.mkOptionsFromDesignTimeBuildAux` hardcodes `/restore` in the `dotnet msbuild` arguments, ignoring the flag.

This causes sequential per-project restores during project cracking, adding ~20s to every Fable compilation for projects with many transitive references (~25 in our case), even when the user explicitly passes `--noRestore`.

## Fix

One-line change: conditionally include `/restore` based on `options.NoRestore`.

```fsharp
// Before:
let arguments =
    $"/restore /t:%s{targets} ..."

// After:
let restoreArg = if options.NoRestore then "" else "/restore"
let arguments =
    $"%s{restoreArg} /t:%s{targets} ..."
```

## Testing

Tested on a monorepo with 1577 source files across ~25 transitive project references:

- **Before**: Parsing phase ~44s (includes ~22s of sequential restores)
- **After with `--noRestore`**: Parsing phase ~28s (restores skipped)
- Default behavior (without `--noRestore`) is unchanged

### DIFF
--- a/src/Fable.Compiler/MSBuildCrackerResolver.fs
+++ b/src/Fable.Compiler/MSBuildCrackerResolver.fs
@@ -142,8 +142,10 @@ Exception:
             let targets =
                 "ResolveAssemblyReferencesDesignTime,ResolveProjectReferencesDesignTime,ResolvePackageDependenciesDesignTime,FindReferenceAssembliesForReferences,_GenerateCompileDependencyCache,_ComputeNonExistentFileProperty,BeforeBuild,BeforeCompile,CoreCompile"
 
+            let restoreArg = if options.NoRestore then "" else "/restore"
+
             let arguments =
-                $"/restore /t:%s{targets} %s{properties} --getItem:FscCommandLineArgs --getItem:ProjectReference --getProperty:OutputType -warnAsMessage:NU1608"
+                $"%s{restoreArg} /t:%s{targets} %s{properties} --getItem:FscCommandLineArgs --getItem:ProjectReference --getProperty:OutputType -warnAsMessage:NU1608"
 
             let! json =
                 dotnet_msbuild_with_defines options.RootDir fsproj.FullName arguments options.FableOptions.Define


ncave:
We can technically save a space there, not that it matters in any way:
```fs
let restoreArg = if options.NoRestore then "" else "/restore "
let arguments =
    $"%s{restoreArg}/t:%s{targets} ..."
```

MangelMaxime:
The changes from this PR have been included in main in #4548

I needed a new PR because I could not push against the fork. I don't know why ...

Thank you @randrag

randrag:
Thank you Maxime, sorry for the extra work - next time I will set maintainerCanModify correctly.

MangelMaxime:
No worries

PR #4558, opened 2026-04-22
fix(js/ts): Fix datetime custom format off by one year

SimenLK:
When you use a custom format like "yyyy-MM-dd HH:mm", you can get an edge-case where you get back off by one year string.

So, if you have `localizedDate = Wed Jan 01 2025 00:00:00 GMT+0100 (Central European Standard Time)`, you'll get `2025-12-31 23:00:00`. This is because the functions used for month, day, hour, etc, use the correct helper functions that take `kind` into account.

Using the `year` function should solve it.


MangelMaxime:
Thank you

PR #4561, opened 2026-04-24
fix(js/ts): Fix Decimal.GetBits returning incorrect mantissa after round arithmetic result

NaserAnjum21:
## Problem
`Decimal.GetBits` returned incorrect bits for any decimal value produced by arithmetic that yields a round result (e.g. 1M * 10M, 10M + 10M). Reconstructing a decimal from these bits via `Decimal(GetBits(d))` would give a value off by some power of 10.

**Root cause:** big.js normalizes its internal representation by stripping trailing zeros from `d.c` and compensating via `d.e`. The old implementation read `d.c` directly as the mantissa digit array, so for 10 (stored as `c=[1], e=1`), it would extract mantissa 1 instead of 10.

## Repro

Following code snippet can be used for repro.

```fsharp
open System

let check (d: decimal) =
    let bits = Decimal.GetBits(d)
    let d2 = Decimal(bits)
    printfn "original=%A reconstructed=%A match=%A" d d2 (d = d2)

check (10M + 10M) // Expected 20, gets 2
check (100.5M - 0.5M) // Expected 100 gets 1
check (-0.1M * 100M) // Expected -10 gets -1 

```
Also, [Repl](https://fable.io/repl/#?code=PYBwpgdgBAygngZwC5gLYCh0BsxKgYwAsx8BrKACgBMAuKKkgS1QEMsBKKAXnSj6hx4ARoyQJuUACJNWWAHQBxXACFRCau179B9AEwTp+Zmwoixm-lBAAnRhCQAzaACJgtgOZ22XAKQBBKGsSYAhkawBXfBQqXwDWJCJY53o9SioJKl1NdCIScgoARgAGAFkoAGooYpLNXLJKYqK5AFYygFooJtba4nqKNqaCsoAqKqLS9iA&html=Q&css=Q)

## Fix

Derive the mantissa from `d.toString()` instead of `d.c`. Strip the sign and decimal point from the string representation to get the mantissa digits, and compute scale from the dot position. This approach should be less prone to big.js internal representation changes.

## Notes
- The inverse direction `(fromParts → big.js)` seems to be unaffected; only reading back via getBits was broken
- Searched for other use cases of `d.c` part of big.js representation where it could be prone to the same normalization issue, did not find any.

### DIFF
--- a/src/fable-library-ts/Decimal.ts
+++ b/src/fable-library-ts/Decimal.ts
@@ -248,14 +248,16 @@ export function fromParts(low: number, mid: number, high: number, isNegative: bo
 
 export function getBits(d: Decimal) {
   const bitSize = 96;
-  const decDigits = Uint8Array.from(d.c);
+  const decStr = d.toString();
+  const absStr = decStr[0] === "-" ? decStr.slice(1) : decStr;
+  const dotPos = absStr.indexOf(".");
+  const scale = dotPos < 0 ? 0 : absStr.length - dotPos - 1;
+  const mantissaStr = dotPos < 0 ? absStr : absStr.slice(0, dotPos) + absStr.slice(dotPos + 1);
+  const decDigits = Uint8Array.from(mantissaStr, ch => ch.charCodeAt(0) - 48);
   const hexDigits = decimalToHex(decDigits, bitSize);
   const low = getInt32Bits(hexDigits, 0);
   const mid = getInt32Bits(hexDigits, 8);
   const high = getInt32Bits(hexDigits, 16);
-  const decStr = d.toString();
-  const dotPos = decStr.indexOf(".");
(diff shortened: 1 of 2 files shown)


NaserAnjum21:
Alternate solution was to skip string/char operations if granular performance is a potential concern. It would be something like below.

```typescript
const trailingZeros = Math.max(0, d.e - (d.c.length - 1));
const decDigits = trailingZeros > 0
  ? Uint8Array.from([...d.c, ...new Array(trailingZeros).fill(0)])
  : Uint8Array.from(d.c);
  
....
```

Benefit of the above would be no string alloc and no charCode conversion, it's pure numeric array ops. However, that would depend more on big.js internals and can break with change in normalization rules. I'd be happy to discuss the trade-offs.

PR #4565, opened 2026-04-25
ci(repo-assist): trigger downstream workflows on bot-created PRs

dbrattli:
## Summary

- Wire up `github-token-for-extra-empty-commit` under `safe-outputs.create-pull-request` and `push-to-pull-request-branch` so Repo Assist pushes one extra empty commit under a PAT after creating/updating a PR. That extra commit triggers `pull_request` / `pull_request_target` workflows that GitHub deliberately skips when the triggering actor is `GITHUB_TOKEN`.
- Pinned to `\${{ secrets.GH_AW_CI_TRIGGER_TOKEN }}` rather than the `app` shorthand because the repo doesn't currently have a GitHub App configured (no `APP_ID` / `APP_PRIVATE_KEY`); the magic-secret path works without one.
- Includes regenerated `repo-assist.lock.yml` from `gh aw compile` (also picks up some unrelated container-image digest pinning that the newer compiler emits).

## Why

PR #4560 was a Repo Assist PR with the title \`[Repo Assist] fix(python): fix regex lookbehind patterns being incorrectly converted\`. The \`Lint PR\` workflow (conventional-commit check) and the \`build\` workflow never ran on it — the title prefix \`[Repo Assist]\` is not a valid Conventional Commits type, so had \`Lint PR\` actually fired it would have blocked the merge. Because no workflow ran, the malformed title slipped to \`main\` (commit \`43bf94835\`).

Root cause is GitHub's documented behaviour: events triggered by `GITHUB_TOKEN` do not start new workflow runs, to prevent recursion. The fix is documented at https://github.github.com/gh-aw/reference/triggering-ci/.

## Required follow-up — set the secret

This PR alone is a no-op until the maintainer creates the secret:

\`\`\`
gh secret set GH_AW_CI_TRIGGER_TOKEN --body "<fine-grained PAT>"
\`\`\`

Use a fine-grained PAT scoped to this repository with \`Contents: write\` permission. Once set, the next Repo Assist PR will receive an extra empty commit, and \`Lint PR\` / \`build\` will run on it as on any human PR.

## Test plan

- [ ] Set \`GH_AW_CI_TRIGGER_TOKEN\` repo secret
- [ ] Trigger Repo Assist (e.g. via \`workflow_dispatch\`) and let it open a draft PR
- [ ] Confirm the PR has two commits (the agent's commit + an empty CI-trigger commit) and that \`Lint PR\` and \`build\` both ran

## Out of scope (potential follow-ups)

- Tighten triggers on \`repo-assist.md\` (drop \`reaction: "eyes"\`, gate slash-command on maintainer role) to reduce the public-repo abuse surface that the docs warn about.
- Reconcile the \`[Repo Assist]\` title prefix with Conventional Commits — Task 6 / Task 11 currently rely on the prefix to identify bot PRs, so dropping it would need code-path updates.

### DIFF
--- a/.github/aw/actions-lock.json
+++ b/.github/aw/actions-lock.json
@@ -5,10 +5,10 @@
       "version": "v9.0.0",
       "sha": "d746ffe35508b1917358783b479e04febd2b8f71"
     },
-    "github/gh-aw-actions/setup@v0.69.1": {
+    "github/gh-aw-actions/setup@v0.68.3": {
       "repo": "github/gh-aw-actions/setup",
-      "version": "v0.69.1",
-      "sha": "b2d32ac2825a5c57cb99851c2ca00094d5e131a6"
+      "version": "v0.68.3",
+      "sha": "ba90f2186d7ad780ec640f364005fa24e797b360"
     },
     "github/gh-aw/actions/setup@v0.69.1": {
       "repo": "github/gh-aw/actions/setup",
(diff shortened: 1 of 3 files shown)


dbrattli:
Superseded by #4567, which uses a GitHub App rather than a personal PAT — better attribution (bot identity instead of a maintainer's account), no expiry, and not tied to anyone's personal credentials.

PR #4592, opened 2026-05-04
fix(all): duplicate LetRec bindings during inline expansion

dbrattli:
## Summary

`resolveInlineExpr` (`src/Fable.Transforms/FSharp2Fable.fs`) seeded its fold accumulator with the original (unresolved) bindings list instead of `[]`. For an `N`-binding `let rec ... and ...` group inside an inline body, it produced **2N** bindings: the unresolved versions followed by the resolved ones.

## Why only BEAM surfaced this

Other targets silently masked the duplicate bindings via last-write-wins lowering:
- **JS / TypeScript**: `function` redeclaration; later definition shadows.
- **Python**: `def` rebinding; later definition shadows.

On **BEAM**, `let rec ... and ...` lowers to a single named-fun with pattern-matching dispatch (`fun Mk_rec(a) -> ... ;Mk_rec(b) -> ... end`). Erlang's match-dispatch picks the **first** matching clause, so callers got the broken (unresolved) body — captures replaced by `undefined`, which then crashed at runtime as `error:{case_clause,undefined}`. The Erlang compiler also warned `this clause cannot match because a previous clause always matches` for the (correct) later clauses.

## Triggers required (all of)

1. Outer function is `inline`
2. Nested non-inline helper, defined inside the inline outer
3. Innermost group is `let rec ... and ...` (mutual recursion)
4. The mutually recursive functions reference outer-scope captures

## Minimal repro

```fsharp
let inline outer<'T> (capture: int) =
    let inner () =
        let rec a () = if capture = 0 then "ok" else b ()
        and b () = a () + "!"
        a ()
    inner ()

let callsite () = outer<int> 0
```

Generated Erlang before the fix (4 clauses, first two broken, captures replaced by `undefined`):

```erlang
Mk_rec_0 = fun Mk_rec_0(b) ->                  %% broken
    fun(UnitVar_2) -> undefined end
;Mk_rec_0(a) ->                                %% broken
    fun(UnitVar_1) -> case undefined of true -> <<"ok">>; false -> ... end end
;Mk_rec_0(a) ->                                %% correct, but unreachable
    fun(UnitVar_1_1) -> case fable_comparison:equals(0, 0) of ... end end
;Mk_rec_0(b) ->                                %% correct, but unreachable
    fun(UnitVar_2) -> erlang:iolist_to_binary([(Mk_rec_0(a))(ok), <<"!">>]) end
end
```

## The fix

```diff
 | Fable.LetRec(bindings, b) ->
     let ctx, bindings =
-        ((ctx, bindings), bindings)
+        ((ctx, []), bindings)
         ||> List.fold (fun (ctx, bindings) (i, e) ->
```

## Test plan

- [x] Minimal repro fails before the fix (`error:{case_clause,undefined}` plus erlc warning); succeeds after.
- [x] BEAM: full suite green (2440 passed). Three new regression tests in `tests/Beam/NonRegressionTests.fs` covering nested mutual rec capture (basic, with intermediate `let`, and the `inline` outer + nested non-inline helper trigger).
- [x] Python: full suite green (2348 passed).
- [x] JavaScript: 2912 passing, 1 failing — `DateTimeOffset.ToString('s')` is a pre-existing timezone-dependent flake (confirmed identical failure on clean main without this change).
- [x] TypeScript: hit a parallel-build collision locally; will rely on CI.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -2227,7 +2227,7 @@ let resolveInlineExpr (com: IFableCompiler) ctx info expr =
 
     | Fable.LetRec(bindings, b) ->
         let ctx, bindings =
-            ((ctx, bindings), bindings)
+            ((ctx, []), bindings)
             ||> List.fold (fun (ctx, bindings) (i, e) ->
                 let i = resolveInlineIdent ctx info i
                 let e = resolveInlineExpr com ctx info e
--- a/tests/Beam/NonRegressionTests.fs
+++ b/tests/Beam/NonRegressionTests.fs
@@ -142,3 +142,25 @@ let ``test string StartsWith char from tuple element`` () =
     let pair = (1, 'H')
     let result = "Hello".StartsWith(snd pair)
     equal true result
+
+// Inline outer + nested non-inline helper + innermost `let rec ... and ...`
+// referencing an outer-scope capture: the inliner used to emit each binding
+// twice, with the first (unresolved) copy losing the capture.
+module IssueInlineMutualRec =
+    let inline outer<'T> (capture: int) =
+        let inner () =
+            let rec a () =
+                if capture = 0 then
+                    "ok"
+                else
+                    b ()
+
+            and b () = a () + "!"
+
+            a ()
+
+        inner ()
+
+[<Fact>]
+let ``test inline outer with nested mutual recursion captures outer scope`` () =
+    IssueInlineMutualRec.outer<int> 0 |> equal "ok"


dbrattli:
To explain it a bit better. Here is the transpiled JS from the above example, before and after the fix:

Before:
```js
export function callsite() {
    const b = () => {
    };
    const a = () => {
        if (null) {
            return "ok";
        }
        else {
            return b();
        }
    };
    const a = () => "ok";
    const b = () => (a() + "!");
    return a();
}
```


After:
```js
export function callsite() {
    const a = () => "ok";
    const b = () => (a() + "!");
    return a();
}
```

We see that before the fix, the non-inlined versions are also added to the closure, but ignored at runtime since they are shadowed.

PR #4604, opened 2026-05-15
fix: [JS/TS/Tests] Remove mocha to resolve package vulnerabilities

ncave:
- [JS/TS/Tests] Remove mocha to resolve package vulnerabilities
- Update npm package dependencies

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -151,8 +151,8 @@
     //     "type": "node",
     //     "request": "launch",
     //     "name": "Run bench-compiler JS tests",
-    //     "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
-    //     "args": ["out-tests"],
+    //     "program": "out-tests/Main.js",
+    //     "args": ["--test"],
     //     "cwd": "${workspaceRoot}/src/fable-standalone/test/bench-compiler"
     // },
     // {
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,57 +6,16 @@
     "": {
       "name": "fable-compiler",
       "dependencies": {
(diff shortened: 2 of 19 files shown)


ncave:
A bit strange (cause it works fine locally), I had to bump Node version to be able to disable test isolation to make this one test pass in CI (technically, test order randomization also makes it pass somehow).

Ready for merging now.

.github/workflows/build.yml
```diff
@@ -202,6 +210,8 @@ jobs:
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v6
+        with:
+          node-version: 26
```
MangelMaxime: If we want to make future update easier we could set the engines in `package.json`

```json
    "engines": {
        "node": ">=22.12.0"
    },
```

and use 

```yml
	  - name: Setup Node.js environment
        uses: actions/setup-node@v6
        with:
          node-version-file: 'package.json'
```

in the workflows

MangelMaxime (commented):
I had one idea to make future update smoother, otherwise this looks good to me

.github/workflows/build.yml
ncave: @MangelMaxime  We probably don't want to prevent people from running older Node.js versions. Technically using newer properties like `--test-isolation` or `--test-randomize` to run the tests would also require using the latest Node.js version, so that's not ideal either.

Fortunately just manually reordering the problematic test and allowing a bit more generous timeout fixes the CI issue in a backwards-compatible way, so I rolled back the bump in Node.js version, and we'll continue running the CI on whatever Node.js version the `setup-node` action defaults to.

Thanks for reviewing!

PR #4636, opened 2026-06-09
fix(python): make [<AttachMembers>] union static members work (#4634)

### LINKED ISSUE #4634: Fable Python: `[<AttachMembers>]` union static property is emitted eagerly before union cases exist
Fable 5.1.0 emits Python for a static F# property on an `[<AttachMembers>]` discriminated union as a class-body `StaticProperty(...)` assignment. When the property value constructs a union case, Python evaluates that constructor call while defining the hidden union base class, before the generated case class exists. The generated file fails at import time with `NameError`.

F# property getters are evaluated when the property is invoked, not while the enclosing type is being defined. Fable's Python output also fails before any user code uses the property.

## Environment

- OS: Windows
- Fable tool: 5.1.0
- Fable.Core: 5.0.0
- fable-library: 5.0.0
- Python: 3.13.7 in local `.venv`

## Minimal Reproduction

`FableStaticPropertyRepro.fsproj`:

```xml
  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Fable.Core" Version="5.0.0" />
    <Compile Include="Library.fs" />
  </ItemGroup>
</Project>
```

`Library.fs`:

```fsharp
namespace StaticPropertyRepro

open Fable.Core

[<AttachMembers>]
type Demo =
    | A of string
    | B

    static member propDefault = A "prop"
    static member methDefault() = A "meth"

module Use =
    let x = Demo.propDefault
    let y = Demo.methDefault()
```

Compile and import in PowerShell:

```powershell
dotnet tool restore
dotnet tool run fable ./ -o ./py --lang Python
$env:PYTHONPATH = "./py"
.\.venv\Scripts\python.exe -c "import library"
```

## Generated Python Code

The official Fable REPL and local Fable 5.1.0 emit the same failing shape:

```python
class _Demo(Union):
    @staticmethod
    def cases() -> list[str]:
        return ["A", "B"]

    @staticmethod
    def meth_default(__unit: Unit=UNIT) -> _Demo:
        return Demo_A("meth")

    prop_default = StaticProperty["Demo"](Demo_A("prop"))

@tagged_union(0)
class Demo_A(_Demo):
    Item: str
```

### Actual Result

The file fails to import:

```text
NameError: name 'Demo_A' is not defined
```

The usage site does not need to run. Python raises while evaluating the class body because `StaticProperty["Demo"](Demo_A("prop"))` evaluates `Demo_A("prop")` before `Demo_A` has been defined.

### Expected Result

The generated module should import successfully. The static property should behave like an F# property getter and construct `A "prop"` when the property is accessed, not during the hidden union base class definition.

## Additional Observations

An explicit getter does not avoid the issue:

```fsharp
static member propDefault with get() = A "prop"
```

Fable still emits an eager `StaticProperty(...)` class-body assignment.

Wrapping through a method changes the first failure:

```fsharp
static member propDefault = Demo.methDefault()
```

This can emit `StaticLazyProperty(lambda: Demo.meth_default())`, avoiding the immediate `Demo_A` `NameError`. However, generated use sites still reference the public union alias:

```python
type Demo = Demo_A | Demo_B
Use_x: Demo = Demo.prop_default
```

That then fails because `Demo` is a Python type alias, while the attached member lives on the hidden `_Demo` base class:

```text
AttributeError: 'typing.TypeAliasType' object has no attribute 'prop_default'
```

## Workaround

One possible workaround is to skip `[<AttachMembers>]` when compiling to Python:

```fsharp
#if !FABLE_COMPILER_PYTHON
[<AttachMembers>]
#endif
type Demo =
    | A of string
    | B

    static member propDefault = A "prop"
    static member methDefault() = A "meth"
```

With `[<AttachMembers>]` disabled for Python, Fable no longer emits the failing class-body descriptor:

```python
prop_default = StaticProperty["Demo"](Demo_A("prop"))
```

Instead, it emits top-level functions and updates generated use sites to call those functions:

```python
def Demo_get_propDefault(__unit: Unit=UNIT) -> Demo:
    return Demo_A("prop")

Use_x: Demo = Demo_get_propDefault()
```
This is not ideal, since it changes the shape of the generated API.

dbrattli:
Fixes #4634

## Problem

A discriminated union with `[<AttachMembers>]` is emitted in Python as three pieces:

- a private base class `_Demo(Union)` that **holds the attached static members**,
- the case classes `Demo_A`, `Demo_B`,
- a public **type alias** `type Demo = Demo_A | Demo_B`.

Given:

```fsharp
[<AttachMembers>]
type Demo =
    | A of string
    | B
    static member propDefault = A "prop"
    static member methDefault() = A "meth"
```

this shape was broken in two ways (plus a latent third):

1. **Eager evaluation, raising `NameError` at import.** The static property was emitted as a class-body assignment inside `_Demo`:

   ```python
   prop_default = StaticProperty["Demo"](Demo_A("prop"))
   ```

   Python evaluates `Demo_A("prop")` *while defining* `_Demo`, before the case class exists, so the module failed to import.

2. **Use sites hit the type alias.** `Demo.propDefault` / `Demo.methDefault()` compiled to `Demo.prop_default` / `Demo.meth_default()`, but `Demo` is a `TypeAliasType` that does not carry the members, raising `AttributeError`. This broke **both** static methods and static properties.

3. **(Latent) bad type annotation.** When the union lived in a module, the descriptor annotation was emitted as a bare `StaticLazyProperty[Issue4634_Demo]` instead of a string forward reference, also raising `NameError`. The eager-evaluation bug masked it.

## Fix

- **`getInitialValue` / `transformStaticProperty`** (`Fable2Python.Transforms.fs`): defer a read-only union static property value with a lambda, producing `StaticLazyProperty["Demo"](lambda: Demo_A("prop"))`. This avoids the forward reference and matches F# getter semantics (re-evaluated per access). Scoped to getter-only to preserve settable-field semantics.
- **`transformStaticProperty`**: emit the descriptor type annotation as a string forward reference whenever the property type is the enclosing union (compare by `FullName`, not only `DisplayName`).
- **`makeCallWithArgInfo`** (`FSharp2Fable.Util.fs`): when referencing a union to reach a **static** member in Python, target the `_Demo` base class instead of the `Demo` type alias — the same underscore convention already used for reflection. This change is gated to `Python` + `IsFSharpUnion` and is a **no-op for every other target**.

## Verification

- New regression test `Issue4634` in `tests/Python/TestNonRegression.fs`.
- Full Python suite: **2351 passed**.
- Pyright: no new errors introduced (the pre-existing 34 are in unrelated files).
- Generated module now imports and runs: `x = A "prop", y = A "meth"`.


dbrattli:
## Follow-up: cross-module static member access was still broken

While adding a cross-module regression test, I found the original fix only handled the **same-file** case. `entityIdent` yields a bare `IdentExpr` for a same-file union but an `Import` for a union defined in another file — and the `Import` case fell through to the type alias, so static member access across modules still raised:

```
AttributeError: 'typing.TypeAliasType' object has no attribute 'prop_default'
```

### Changes (commit `7733d0594`)

- **`FSharp2Fable.Util.fs`** — added an `Import` arm alongside the `IdentExpr` arm that rewrites the import `Selector` to the underscore-prefixed base class (`_Demo`). `transformImport` uses the selector string as the imported name, so cross-file references now resolve to the base class that actually holds the static members.
- **`MiscTestsHelper.fs` + `TestNonRegression.fs`** — added a cross-module regression test: `CrossModuleDemo` is defined in a separately-compiled module and its static property/method are referenced from `TestNonRegression.fs`. This exercises the `Import` path; the prior commit's test only covered same-file references.
- **`Fable2Python.Transforms.fs`** — minor cleanup: replaced the dense inline `let … in` inside the `when` guard with a named `isSelfReference` binding.

### Generated output (cross-module)

```python
from misc_tests_helper import (CrossModuleDemo_A, _CrossModuleDemo)
...
Testing_equal(CrossModuleDemo_A("prop"), _CrossModuleDemo.prop_default)
Testing_equal(CrossModuleDemo_A("meth"), _CrossModuleDemo.meth_default())
```

### Verification

- Full Python suite: **2352 passed** (was 2351; +1 for the new cross-module test, now green).

> [!NOTE]
> `build.sh test python` returned exit code `0` even while the new test was failing — the failure only showed in the pytest summary line, not the process exit. Worth keeping in mind for CI.

PR #4646, opened 2026-06-10
fix: add type specifiers to interpolated strings

dbrattli:
## Summary

Resolves all **127 `GRA-INTERPOLATED-001`** code-scanning alerts — _"missing type specifiers in interpolated strings"_ — across 17 source files in `Fable.Compiler` and `Fable.Transforms`.

Each flagged interpolation hole gets the correct `%`-specifier so the type is checked at compile time. **Output is unchanged**; this only adds compile-time type safety:

- **`%s`** for strings (names, full names, formatted types, paths, error/warning messages) — the majority
- **`%d`** for ints (generated identifiers like `arg{i}`, `FSharpChoice${n}`, union tags, counters, arities)
- **`%O`** for objects whose default formatting already called `ToString()` (`DateTime` timestamps, `Fable.Type`, `obj` values, `string option`)

One site (`FSharp2Fable.Util.fs`) had a multi-line `if/else` inside the hole that went offside once the prefix widened, so it was hoisted into a small `staticSuffix` binding.

## Verification

`dotnet build src/Fable.Cli/Fable.Cli.fsproj -c Release` succeeds with **0 warnings, 0 errors**. The build itself is the safety net: a wrong specifier (`%s` on a non-string, `%d` on a non-int) is a hard compile error, so a clean build confirms every choice is type-correct. The identifier-generating strings (which feed into *generated code*, not just messages) produce byte-identical output, so there is no behavior change.

### DIFF
--- a/src/Fable.Compiler/ProjectCracker.fs
+++ b/src/Fable.Compiler/ProjectCracker.fs
@@ -706,7 +706,7 @@ let getFableLibraryPath (opts: CrackerOptions) (shouldCopy: bool) =
                 |}
             |> Option.defaultWith (fun () ->
                 Fable.FableError
-                    $"Cannot find [temp/]{buildDir} from {baseDir}.\nPlease, make sure you build {buildDir}"
+                    $"Cannot find [temp/]%s{buildDir} from %s{baseDir}.\nPlease, make sure you build %s{buildDir}"
                 |> raise
             )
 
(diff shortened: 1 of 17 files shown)


dbrattli:
This should be safe since a wrong specifier should trigger a compile error. But touches many files, so please have a look and approve the PR if it looks OK with you @MangelMaxime @ncave

ncave (approved):

MangelMaxime (approved):

PR #4648, opened 2026-06-11
Fix fable clean command crash when output directory doesn't exist

goswinr:
The clean command would crash with DirectoryNotFoundException if the specified output directory (via -o flag) didn't exist. Now it gracefully handles this case by checking if the directory exists before attempting to clean it.

### DIFF
--- a/src/Fable.Cli/Entry.fs
+++ b/src/Fable.Cli/Entry.fs
@@ -453,7 +453,10 @@ let clean (args: CliArgs) language rootDir =
                 recClean subdir
         )
 
-    recClean cleanDir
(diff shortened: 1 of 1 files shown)


goswinr:
@MangelMaxime how to fix the lint issue ?

MangelMaxime:
@goswinr You need to follow the commit convention from [Shipit](https://github.com/easybuild-org/EasyBuild.ShipIt#commit-conventions).

goswinr:
in https://github.com/fable-compiler/Fable/pull/4651 now @MangelMaxime

PR #4651, opened 2026-06-12
fix(all): cli fable clean command crash when output directory doesn't exist

goswinr:
The cli clean command would crash with DirectoryNotFoundException if the specified output directory (via -o flag) didn't exist. Now it gracefully handles this case by checking if the directory exists before attempting to clean it.


dbrattli:
@goswinr You cannot update the CHANGELOG since that will automatically be taken care of by ShipIt based on the commit message.

goswinr:
@dbrattli I reverted

PR #4657, opened 2026-06-13
chore: use postfix array syntax to satisfy PostfixGenericsAnalyzer

dbrattli:
## Summary

Resolves all **52 open code-scanning alerts** of type _"Detect if generic type should be in the postfix position"_ — emitted by `Ionide.Analyzers`' `PostfixGenericsAnalyzer` (rule `IONIDE-002`).

The analyzer wants generic collection/cell types written in word-postfix form. This PR converts every flagged annotation:

| Before | After |
| --- | --- |
| `Foo[]` | `Foo array` |
| `ref<Regex>` | `Regex ref` |

51 of the 52 alerts were the `Foo[]` bracket form; one was the `ref<Regex>` prefix form (`Globbing.fs`).

## Notes

- These are **purely syntactic** changes to type annotations — the compiled types are identical, so there is **no behavioral or public-API change**.
- This includes the single annotation in `src/Fable.AST/Fable.fs` (`string[]` → `string array`). Although `AGENTS.md` flags `Fable.AST` as off-limits because changes there are usually breaking, this one is not: `string[]` and `string array` are the same type. Happy to revert that one line and dismiss the alert instead if preferred.
- Each fix was applied by exact source column so no unrelated `[]` usages were touched.

## Verification

`dotnet build src/Fable.Cli` (transitively covers `Fable.AST`, `Fable.Compiler`, and all of `Fable.Transforms`): **succeeds, 0 warnings, 0 errors.**

### DIFF
--- a/src/Fable.AST/Fable.fs
+++ b/src/Fable.AST/Fable.fs
@@ -537,7 +537,7 @@ type ValueKind =
     | NewList of headAndTail: (Expr * Expr) option * typ: Type
     | NewTuple of values: Expr list * isStruct: bool
     | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list
-    | NewAnonymousRecord of values: Expr list * fieldNames: string[] * genArgs: Type list * isStruct: bool
+    | NewAnonymousRecord of values: Expr list * fieldNames: string array * genArgs: Type list * isStruct: bool
     | NewUnion of values: Expr list * tag: int * ref: EntityRef * genArgs: Type list
 
     member this.Type =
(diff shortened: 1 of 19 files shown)


src/Fable.AST/Fable.fs
```diff
@@ -537,7 +537,7 @@ type ValueKind =
     | NewList of headAndTail: (Expr * Expr) option * typ: Type
     | NewTuple of values: Expr list * isStruct: bool
     | NewRecord of values: Expr list * ref: EntityRef * genArgs: Type list
-    | NewAnonymousRecord of values: Expr list * fieldNames: string[] * genArgs: Type list * isStruct: bool
+    | NewAnonymousRecord of values: Expr list * fieldNames: string array * genArgs: Type list * isStruct: bool
```
dbrattli, line 540: This should be safe right @MangelMaxime ? It will not be a breaking change for Fable.AST?

PR #4663, opened 2026-06-15
chore: regenerate the `uv.lock` file

MangelMaxime:
According to Claude:

> Someone bumped the lower bounds in the [dependency-groups] dev list of pyproject.toml in an earlier commit but never regenerated uv.lock.


dbrattli:
Probably dependabot

PR #4692, opened 2026-06-26
feat(all): add `Compiler.is*` target detection flags for conditional branching

MangelMaxime:
Fable's conditional compilation relies on compiler directives. It is working fine but has a huge limitations.

In IDE, you often end up with no IDE support most of the compiler directives branches as the compiler directives is not set in the `fsproj`. And even when setting it up in the `fsproj` the experience is not always great depending on the IDEs.

This PR is building on top of the existing foundation in Fable used for `Compiler.debugMode` or `Compiler.triggeredByDependency`.

The idea is that in Fable, if a `if ... then ... else ...` structure is using boolean constant in their condition then Fable auto erase unreachable branches.

This now allow us to write things like

```fs
open Fable.Core

let greet name =
    if Compiler.isJavaScript then
        $"Hello from JS, {name}!"
    elif Compiler.isPython then
        $"Hello from Python, {name}!"
    else
        $"Hello from .NET, {name}!
```

which compiles to

**JavaScript**

```js
 export function greet(name) {
      return `Hello from JS, ${name}!`;
  }
```

**Python**

```py
def greet[_A](name: _A=UNIT) -> str:
      return ("Hello from Python, " + str(name)) + "!"
```

The only impact is on .NET where there will be a real runtime check against the boolean. This should be more than fast enough and people can use `Compiler.isDotnet` to avoid having to check all the conditions.

## Summary

Thanks to these changes, we are now able to provide a better experience for interop/universal code.


MangelMaxime:
Hello @ncave,

my memory is failing.

In order to make, `fable-standalone` tests to pass I need to regenerate the `fable-metadata` for `Fable.Core`.

I don't remember how to do that, is it as simple as copying the result of `dotnet build -c Release` over their ?

ncave:
@MangelMaxime 
`Fable.Core` doesn't actually have code, just bindings/metadata, so you can just use it as is. Just replace the one in `fable-metadata` with latest (as trimming it will not save much anyway).

PR #4694, opened 2026-06-27
feat(js/ts): map task { } to Promise<T>

MangelMaxime:


dbrattli:
So now I will soon be able to run Giraffe on Node? 🥰🤩 https://github.com/dbrattli/Fable.Giraffe

MangelMaxime:
You should indeed :)

I wonder how easy it would be to integrate into a express router. Asking, the express-like API is something that is also compatible with server-less functions like Vercel, etc.

But that can be left out to experiment by users :)

PR #4728, opened 2026-07-06
perf(python): rewrite hot Seq combinators as native Python generators

MangelMaxime:
Python's `Seq` combinators built on `generate`/`Enumerator.generateWhileSome`: an `Option` allocated per element, plus a `MoveNext`/`Current` call per step - a faithful port of F#'s `IEnumerable`/`IEnumerator` model, but on CPython the per-element object churn dominates.

## Benchmark

Compiled a pipeline exercising several of the converted combinators together (`map`, `choose`,
`map2`, `takeWhile`, `tryItem`, `fold` - not just `map`/`filter`/`collect`), before vs after,
5,000 calls:

| before | after | speedup |
| ------ | ----- | ------- |
| 4.60s  | 2.91s | 1.58x   |

Results identical both runs: `69360`.

A simpler `map`/`filter`/`collect`/`fold` pipeline measured separately shows ~1.7x.


dbrattli:
## Review

Rewrites ~17 hot `Seq` combinators from the `generate`/`generateWhileSome` protocol (an `Option` + `MoveNext`/`Current` per element) to plain Python generators in a new hand-written `seq_native.py`, with `Replacements.fs` redirecting the call sites and `Seq.fs` either keeping `[<Import> … nativeOnly]` stubs or deleting the F# bodies. Python-only — other targets link `fable-library-ts/Seq.fs`, so there's no cross-target leakage.

### Correctness — verified end-to-end

I copied the module into the built library and exercised every function:

- **Error messages match .NET/Fable byte-for-byte**, including the `(Parameter 'source')` / `(Parameter 'index')` suffix that `invalidArg` generates (`take` insufficient, `head` empty, `item`/`tryItem` OOB).
- **Restartability preserved** — a source thunk with side effects re-runs on each enumeration. The `NativeSeq`-holds-a-thunk (not a live generator) design is the crux of matching F# `seq` semantics and it's correct.
- **`map2`/`map3`/`mapIndexed2` stop at the shorter sequence** (F# `Seq` semantics, unlike `List`/`Array` which throw) — `zip` gives exactly this.
- **The `_source_iter` `Mapping` bridge is necessary and correct**: `FSharpMap` is a `collections.abc.Mapping`, whose `__iter__` yields keys only, so it's routed through `to_iterator(get_enumerator(xs))` to recover `KeyValuePair` tuples. Verified: bare `iter(map)` → `[1,2,3]`, `_source_iter(map)` → `[(1,'a'),(2,'b'),(3,'c')]`.
- **`int32(i)` in `map_indexed` is the correct index type** — matches the transpiled `list.py` `map_indexed` (`Callable[[int32, T], U]`), so no plain-`int` regression.
- **Array/List fast-paths and negative-index handling** in `try_item`/`try_head` mirror the original behavior exactly.
- **`injectArg` omission is safe** — none of the redirected methods appear in the `"Seq"` map in `ReplacementsInject.fs`.

No dangling references (deleted F# bodies aren't referenced elsewhere; the internally-used ones keep `[<Import>]` stubs), and no circular import (`seq_native` doesn't import `seq`).

### Broadest-blast-radius change

The `util.py` `Enumerator.Dispose` change now calls `close()` on **every** wrapped iterator (all `IEnumerator`s, not just native seq). It's a net-positive fix — runs generator `finally` blocks via `GeneratorExit`, correctly guarded with `hasattr` — but worth being aware it's global. Validated by the full suite below.

### Test results

- `./build.sh test python` → **2373 passed**, including `test_seq.py` / `test_map.py` / `test_list.py`.
- `--type-check` → same 2373 pass; Pyright's remaining errors are all **pre-existing baseline** in test files, **none** in the files this PR touches (`fable_library` itself reports 0).

### Follow-up changes pushed (c05ede255)

Two minor points from the review are now addressed in-branch, plus a small style cleanup:

- **Completed the `list.pyi` stub** with `try_head`/`try_item` signatures — `seq_native` imports them, and without this the source-tree stub was out of sync with what the module imports.
- **Added `__all__`** to `seq_native.py` (at the bottom, per the convention in `util.py`/`array_.py`/`bases.py`).
- **Converted the `isinstance` chains** in `_source_iter`/`try_head`/`try_item` to `match` statements, restoring the shape of the F# originals they replace (`match xs with | :? array<'T> -> … | :? list<'T> -> …`). Zero perf cost (same `isinstance` machinery). In `try_item` the `index < 0` guard moved into `case _:` since it only applies to the generic path.

### Verdict

Solid, well-reasoned change — the two documented invariants (thunk-not-generator for restartability; `close()` for disposal) are exactly the right things to get correct, and both hold up under testing. No correctness blockers. 👍

dbrattli (approved):
Very nice improvements. I added some additional fixes directly to the PR

PR #4731, opened 2026-07-07
perf(python): emit union field reads via `narrow` helper instead of `typing.cast`

dbrattli:
## Summary

Follow-up to #4725, targeting its branch.

#4725 speeds up union case field reads by emitting `cast(Case, x).field` instead of the `.fields`-rebuild property. To make Pyright accept those casts it disables `reportUnnecessaryCast` **globally**, because some sites (single-case unions, already-narrowed values) make the cast redundant from Pyright's point of view.

This PR moves that "cast" into a tiny library helper so generated code stays `cast`-free and the global rule can stay on.

## Changes

- **`fable_library/union.py`** — add `narrow[T](case: type[T], value: object) -> T`, a typed identity function (documented, exported in `__all__`). It plays the same role as `typing.cast` but is scoped to unions.
- **`Fable2Python.Transforms.fs`** — union field reads now emit `narrow(Case, x).field` instead of `cast(Case, x).field`.
- **`pyrightconfig.json`** — revert `reportUnnecessaryCast` back to `true`; `narrow` is not `typing.cast`, so the rule no longer fires on these reads (and still guards the rest of the generated code).

Generated output before → after:
```python
# before (#4725)
return cast(Tree_Node, t).left_
# after (this PR)
return narrow(Tree_Node, t).left_
```

## Why not `cast(T, value)` inside `narrow`?

`narrow`'s body is `return value  # pyright: ignore[reportReturnType]` rather than `return cast(T, value)`. `typing.cast` is a plain Python function (`return val`), so routing through it would make **two** calls per field read instead of one — a needless regression on the exact hot path this optimization targets (~8M reads in the #4725 benchmark). The ignore is targeted, and `reportUnnecessaryTypeIgnoreComment` (enabled) confirms it's genuinely necessary.

## Verification

- Pyright on `union.py`: 0 errors/warnings; the ignore is not flagged as unnecessary.
- Call-site check: `narrow(Tree_Leaf, x).value_` resolves to the field's real type (`int`).
- Quicktest compiles and emits `narrow(...)` for both user unions and library unions (`Result`/`Choice`); the only remaining `cast` in output is a legitimately-necessary one that `reportUnnecessaryCast: true` accepts.

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -9,9 +9,9 @@
   "reportConstantRedefinition": false, // Fable preserves casing for variables starting with uppercase
   "reportUnnecessaryIsInstance": false, // Not a problem in generated code (look into this later)
   "reportUnnecessaryComparison": false, // Generated tests do this on purpose
-  "reportUnnecessaryCast": false, // Generated code for union case field need addional cast otherwise Pyrigh cannot infer the resolved typed
   "reportImportCycles": true,
   "reportMissingImports": true,
+  "reportUnnecessaryCast": true,
   "reportDuplicateImport": true,
   "reportOverlappingOverload": true,
   "reportInconsistentConstructor": true,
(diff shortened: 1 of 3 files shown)


MangelMaxime:
I tried to find a way to not disable `reportUnnecessaryCast` and could not come up with a trick for that. Great finding

PR #4738, opened 2026-07-08
fix: Misc 9 fixes

Thorium:
Misc 9 fixes. You can merge all or pick what you need:
- endsWith regression for empty pattern with non-ordinal comparison
- NaN comparison to match .NET total ordering
- async cancellation propagation and continuation double-resolve
- MailboxProcessor dropping falsy messages and unit replies
- Regex.Replace count/substitutions and Regex.Split semantics
- BigInteger byte-array corruption, checked conversions, gcd, log2
- int64 non-decimal parsing and BitConverter .NET mismatches
- date arithmetic: PM parsing, AddMonths day clamp, AddYears offset
- Observable/Event dropping null/undefined/unit values

Here is a Copilot generated text-wall:

This pull request introduces a number of improvements and bug fixes to the Fable library's TypeScript runtime, focusing on correctness, compatibility with .NET/F#, and improved error handling. The most significant changes are grouped below by theme.

### Async/Await and Continuation Handling

* Improved cancellation token propagation in `Async` combinators (`startChild`, `parallel`, `sequential`), ensuring child computations are cancelled when their parent is cancelled, matching .NET behavior. Also, added handling to prevent unhandled promise rejections in child computations. [[1]](diffhunk://#diff-5f596b7377441d97336fab37716e6d80f4b776857aae021a1139497d0be1c92eL63-R61) [[2]](diffhunk://#diff-5f596b7377441d97336fab37716e6d80f4b776857aae021a1139497d0be1c92eR75-R84) [[3]](diffhunk://#diff-5f596b7377441d97336fab37716e6d80f4b776857aae021a1139497d0be1c92eL135-R156)
[...]

PR #4740, opened 2026-07-08
fix(ts/js): MailboxProcessor dropping falsy messages and unit replies

Thorium:
- Dequeued messages were tested for truthiness, so 0, false, "", null and undefined (unit) messages were silently dropped; the queue now wraps dequeued values so emptiness is distinguishable
- postAndAsyncReply gated completion on result !== undefined, so AsyncReplyChannel<unit> never completed and the caller hung forever; use an explicit replied flag

(splitted from #4738)


MangelMaxime:
Thank you

PR #4742, opened 2026-07-08
fix(js/ts): Observable/Event dropping null/undefined/unit values

Thorium:
- Observable.filter passed the raw value to choose instead of wrapping with some(), dropping null/undefined/unit values that satisfied the predicate (Event.filter already did this correctly)
- Observable.pairwise used a null check instead of a have-last flag, suppressing the pair after any null element
- Event.Trigger disambiguated its overloads by value === undefined, misrouting Trigger(sender, undefined) so the handler received the sender as the event value; now disambiguated by arity

(split from #4738)


MangelMaxime:
Thank you

PR #4743, opened 2026-07-08
fix(js/ts): BigInteger byte-array corruption, checked conversions, gcd, log2

Thorium:
- toSignedBytes computed the two's-complement width from fractional Math.log2, undersizing the buffer for some negatives so toByteArray/fromByteArray round-trips corrupted values (e.g. -62837 -> +2699); rewritten via byte-wise iteration
- Checked toIntN conversions used abs(x) >> bits as the bounds test, silently wrapping out-of-range values instead of throwing
- greatestCommonDivisor was wrong for negative arguments
- getBitLength was wrong for zero and negatives; ilog2 crashed on non-powers-of-two; both now derive from the exact bit length
- toDecimal silently truncated to the low 96 bits instead of throwing on overflow

(split from #4738)

### DIFF
--- a/src/fable-library-ts/BigInt.ts
+++ b/src/fable-library-ts/BigInt.ts
@@ -135,12 +135,12 @@ export function toIntN_unchecked(bits: number, x: bigint, signed: boolean): bigi
 }
 
 export function toIntN(bits: number, x: bigint, signed: boolean): bigint {
-    let higher_bits = abs(x) >> BigInt(bits);
-    if (higher_bits !== 0n) {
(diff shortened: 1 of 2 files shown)


dbrattli:
@Thorium Thanks for the PRs. Could you please add the right scope to the PR titles i.e fix(js/ts): for PRs that handles js/ts and not affect other targets.

Thorium:
Sure, I think my PRs were all focused on (js/ts)

MangelMaxime:
Thank you

PR #4745, opened 2026-07-08
fix(js/ts): endsWith regression for empty pattern with non-ordinal comparison

Thorium:
str.slice(-pattern.length) becomes slice(-0) === slice(0) (the whole string) when pattern is empty, so EndsWith("") returned false for culture/ignore-case comparisons. .NET returns true (every string ends with ""). Use slice(str.length - pattern.length) instead.

(split from #4738)


MangelMaxime:
Thank you

PR #4747, opened 2026-07-08
fix(js/ts): NaN comparison to match .NET total ordering

Thorium:
comparePrimitives returned 1 for any NaN operand, violating antisymmetry and corrupting sort order of float collections containing NaN. .NET Double.CompareTo: NaN equals NaN and is less than any other value.

(split from #4738)


MangelMaxime:
Thank you

PR #4748, opened 2026-07-08
fix(js/ts): int64 non-decimal parsing and BitConverter .NET mismatches

Thorium:
- Long.parse never applied two's-complement wrap for hex/octal/binary, returning values outside the int64 range (0xFFFFFFFFFFFFFFFF parsed to 2^64-1 instead of -1); now mirrors Int32's sign-extension via BigInt.asIntN with proper range checks
- parse crashed with a raw JS error on negative prefixed input like -0xFF (valid .NET input, expect -255)
- BitConverter.toBoolean returned true only for byte 1; .NET treats any nonzero byte as true
- BitConverter.toString emitted lowercase hex; .NET is uppercase

(split from #4738)


MangelMaxime:
Thank you for the PR

Issue #4751, opened 2026-07-08
Use Temporal for dates and times in JS


kerams:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

Temporal will be part of ES2027, but Chrome and Firefox already support the APIs, plus there are polyfills for Safari and others. Getting rid of `Date` is an observable breaking change, so we'll want a flag and a define, since Remoting or serialisation libraries might depend on the underlying representation of `DateTime`, `DateOnly`, etc.

MangelMaxime:
Hello,

Up to now, the rules regarding API usage in Fable has been to use widely support API when available.

Right now Temporal is only covering [67% of the devices](https://caniuse.com/?search=temporal).

One solution could be indeed to integrate a polyfills in Fable but for now I would prefer to not do that.

Outside of serialization using `Temporal` instead `Date` could also break bindings or libraries in a more subtle way because up to now it was assumes that `DateTime` being a `Date` at runtime, it was possible to pass it directly to JavaScript libraries and that contract would not hold anymore.

Is there anything that the current DateTime implementation is missing for your use case?

I have a POC on my machine allowing Fable plugin to also register `Replacement` API. I still need to refined it, but it could be a way for people to experiment with porting `DateTime` to `Temporal` for example.

The original usage for that POC was for cross-target libraries who need to write code for some of the `System.*` API that we don't want to include support for in Fable. But it could feat nicely in experimenting in this issue.

kerams:
> integrate a polyfills in Fable

Why in Fable?

> Outside of serialization using Temporal instead Date could also break bindings or libraries

That's why it would be opt-in, and a define will allow libraries to react accordingly.

I don't have any pressing need for this, but .NET date and time types map much more cleanly onto Temporal primitives without needing to abuse `Date`. Once `DateTimeOffset` becomes `Instant` for example, you unlock types of manipulation that is only possible with brittle arithmetic over `Date` (akin to what is done in `DateOffset.ts`). `ToString` also immediately becomes more usable out of the box.

So there's no rush, but I'd like to see this become available, much like when Fable started making use of native typed arrays and `BigInt`, although the latter has been switched without a compiler flag.

MangelMaxime:
> > integrate a polyfills in Fable
> 
> Why in Fable?

I think people using Fable expect code generated by Fable to just work.

[...]

PR #4759, opened 2026-07-10
feat: add Temporal dates and times (opt-in)

kerams:
Implements #4751.

Opt-in with `--test:js-temporal`, which defines `FABLE_COMPILER_JAVASCRIPT_TEMPORAL`.

- [x] - `DateOnly` as [`PlainDate`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDate)
- [x] - `TimeOnly` as [`PlainTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainTime)
- [x] - `DateTime` as [`PlainDateTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDateTime) extended with kind metadata
- [x] - `DateTimeOffset` as [`ZonedDateTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime) with an offset-based time zone rather than IANA
- [x] - `TimeSpan` as [`Duration`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration)
- [x] - tests

All these types have nanosecond precision, whereas the old representation truncates everything below milliseconds (partly due to `Date`), which is something that bit me in the past. However, the precision in .NET is a tick (100ns), so the current implementation is truncating nanosecond values on construction and value access.


src/fable-library-ts/DateOnlyTemporal.ts
```diff
@@ -0,0 +1,205 @@
...
+    namespace Now {
+      function plainDateISO(): PlainDate;
+    }
+  }
+}
+
+export type DateOnly = Temporal.PlainDate;
+
+export const PlainDate = Temporal.PlainDate;
+export type PlainDate = Temporal.PlainDate;
```
kerams: This will go (likewise in the other file) once we upgrade to Typescript 6 or 7.

src/fable-library-ts/DateOnlyTemporal.ts
```diff
@@ -0,0 +1,205 @@
...
+      try {
+        return new Temporal.PlainDate(y, m, d);
+      } catch {
+        return fail();
+      }
+    }
+  }
+
+  return fail();
+}
```
kerams, line 165: This `parse` (likewise in the other file) is only necessary if we want to closely follow .NET's parse behaviour. We could decide to remove it and instead delegate to [`from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDate/from), which should still be close enough and acceptable in the vast majority of cases.

MangelMaxime:
Note to my self, this PR is changing Fable.AST this needs to be think through.

src/fable-library-ts/DateOnlyTemporal.ts
MangelMaxime: IHMO keeping `parse` is the right path 

This is because Fable promise is to be close to .NET behavior.

MangelMaxime:
Thank you @kerams

kerams:
No, thank *you* for plugging the remaining holes, Maxime.

PR #4774, opened 2026-07-13
feat(all): F# quotations type improvements and Rust target

Thorium:
Improves quotations in the shared emitter (with better type-info) and adds quotation support to the Rust and PHP targets. Motivated by compiling F# query (<@ @>) expressions, e.g. toward a future TypeProviders (like SQLProvider) port to Python/Rust.

Shared emitter (QuotationEmitter.fs) — benefits all targets:
- typeToString: render all numeric kinds, char/regex/guid/datetime/ datetimeoffset/timespan, option/voption/list/array/seq/nullable, declared-type names, generics and delegates instead of "obj".
- Call nodes record the declaring type; Get OptionValue/ListHead/ListTail map to property-getter calls.
- DecisionTree inlined into IfThenElse/Let; Test nodes (union/option/list/ type) represented via union-tag comparison / getter calls.
- mkNull constructor for null instances/values (no raw null literal, so statically typed languages compile); is_call maps it back to None.

mk_call kept backward-compatible (3 and 4 params on Beam); mkNull added to each.

Rust target (new): route Quote through the emitter (Fable2Rust), wire Replacements to tryQuotationCall, and add a transpiled F# runtime (QuotationTypes.fs + Quotation.fs) exposing FSharpExpr/FSharpVar at Microsoft.FSharp.Quotations and the quotation_ module. Construction + deconstruction verified via cargo; arithmetic evaluate works (let/lambda eval pending an upstream getZero fix for Rc<dyn Any>).

Check the tests for more use-case examples.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -537,6 +537,10 @@ type Context =
         InlinePath: Log.InlinePath list
         CaptureBaseConsCall: (FSharpEntity * (Fable.Expr -> unit)) option
         Witnesses: Fable.Witness list
+        // When true we are translating the body of a code quotation (<@ @>). The body is
+        // data, never executed, so member calls are kept as-is (not replaced/emitted/inlined)
+        // to preserve their .NET member metadata for QuotationEmitter.
+        CapturingQuotation: bool
     }
 
     static member Create(?usedRootNames) =
@@ -555,6 +559,7 @@ type Context =
             InlinePath = []
             CaptureBaseConsCall = None
             Witnesses = []
+            CapturingQuotation = false
         }
 
 type IFableCompiler =
@@ -3058,6 +3063,13 @@ module Util =
         (callInfo: Fable.CallInfo)
         =
         match memb, memb.DeclaringEntity with
+        // Inside a quotation the body is data, never executed. Preserve the original member
+        // call (skip replace/emit/import/inline) so its .NET member metadata survives for
+        // QuotationEmitter. Skipping inlining here also matches real F# quotation semantics.
+        | _ when ctx.CapturingQuotation ->
+            let membTyp = makeType ctx.GenericArgs memb.FullType
+            memberIdent com r membTyp memb membRef |> makeCall r typ callInfo
+
         | Emitted com ctx r typ (Some callInfo) emitted, _ -> emitted
         | Imported com ctx r typ (Some callInfo) imported -> imported
         | Replaced com ctx r typ callInfo replaced -> replaced
--- a/src/Fable.Transforms/FSharp2Fable.fs
+++ b/src/Fable.Transforms/FSharp2Fable.fs
@@ -1503,7 +1503,9 @@ let private transformExpr (com: IFableCompiler) (ctx: Context) appliedGenArgs fs
                     |> addErrorAndReturnNull com ctx.InlinePath (makeRangeFrom fsExpr)
 
         | FSharpExprPatterns.Quote quotedExpr ->
-            let! body = transformExpr com ctx [] quotedExpr
+            // Translate the quoted body in "capturing" mode so member calls keep their
+            // .NET metadata (not replaced/emitted/inlined) — the body is data, not code.
+            let! body = transformExpr com { ctx with CapturingQuotation = true } [] quotedExpr
             let exprType = fsExpr.Type
             let isTyped = exprType.GenericArguments.Count > 0
             return Fable.Quote(body, isTyped, makeRangeFrom fsExpr)
--- a/src/Fable.Transforms/Php/Fable2Php.fs
+++ b/src/Fable.Transforms/Php/Fable2Php.fs
@@ -936,7 +936,10 @@ let rec tryFindMethod methodName (phpType: PhpType) =
 let rec convertExpr (com: IPhpCompiler) (expr: Fable.Expr) =
     match expr with
     | Fable.Extended _ -> failwith "TODO: Extended instructions"
-    | Fable.Quote _ -> failwith "Quotations are not yet supported for PHP target"
+    | Fable.Quote(body, _isTyped, _r) ->
+        // Lower the quoted body to runtime calls that build the quotation AST,
+        // then convert like any other expression (mirrors JS/Python/Beam/Rust).
+        convertExpr com (QuotationEmitter.emitQuotedExpr com body)
 
     | Fable.Unresolved _ -> failwith "Unexpected unresolved expression"
 
--- a/src/Fable.Transforms/QuotationEmitter.fs
+++ b/src/Fable.Transforms/QuotationEmitter.fs
@@ -115,18 +115,22 @@ let rec emitQuotedExpr (com: Compiler) (expr: Expr) : Expr =
         let instanceExpr =
             match info.ThisArg with
             | Some thisArg -> emitQuotedExpr com thisArg
-            | None -> Value(Null Any, None)
+            // A static/operator call has no instance. Emit a runtime-built null node
+            // (not a raw null literal) so every target — notably statically typed ones
+            // like Rust — sees an Expr in the instance position.
+            | None -> mkNullExpr com "null"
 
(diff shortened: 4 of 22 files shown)


MangelMaxime (changes requested):
Looking roughly at the diff, I see several issues.

Not all the changes are covered for example I see changes made for JavaScript but no tests added for that target.

Also, in general we try to keep the tests between target in sync and I don't think this is the case. For example, I see 

```fs
let private lambdaVarName (e: Expr) =
    match e with
    | Lambda(v, _body) -> v.Name
    | _ -> "?"

let private letVarName (e: Expr) =
    match e with
    | Let(v, _value, _body) -> v.Name
    | _ -> "?"

let private isIfThenElseNode (e: Expr) =
    match e with
    | IfThenElse(_g, _t, _e) -> true
    | _ -> false
```

In `tests/Rust/tests/src/QuotationTests.fs` but unless I am mistaken I don't see that in other tests for example Python ones. So there is a variation here somewhere. Plus seems to have gained quotations support in this PRs but didn't get all the Python tests migrated.

Thorium:
Feedback addressed: Rust seems to be a bit rusty, so I think it needs a separate PR to fix some of its Todos in code, but the other test-suites should now be consistent across the targets.

_Edit: I have done Rust fixes to https://github.com/Thorium/Fable/tree/rust-todo-fixes but as I know you prefer small commits, it'll be a separate PR later._

dbrattli:
## Overview

Three things bundled into one PR:

1. **Shared quotation emitter** (`QuotationEmitter.fs`): richer `typeToString`, declaring-type on `Call` nodes, `DecisionTree` inlining, `Test` node support, and a `mkNull` constructor replacing raw `null` literals.
2. **New `CapturingQuotation` flag** in `FSharp2Fable.Context` that disables replacements/emit/import/inlining for member calls inside `<@ @>`.
3. **New targets**: Rust (F#-authored runtime, transpiled) and PHP (hand-written `quotation.php`), plus an unrelated `Map.find`/`tryFind` fix in `fable-library-rust`.

The direction is good — routing PHP/Rust through the shared emitter is exactly the "keep targets aligned" convention in AGENTS.md, and the F#-authored Rust runtime is a nice trick. But there are correctness gaps and a couple of things I don't think should ship as-is.

## Blocking / high-confidence issues

**`declaringType` is plumbed everywhere but unreachable from F#.** `call_declaring_type` / `callDeclaringType` accessors were added to all four runtimes, but `Replacements.Util.fs`'s `quotationPatterns` has no case that maps to them — nothing in `tryQuotationCall` references `declaringType`. So the whole point of the change (SQLProvider-style translation needs to know *which* `List.map` it's looking at) is not yet achievable from user code. Either wire it into the `Call` pattern (real F# `Patterns.Call` yields a `MethodInfo` with `DeclaringType`) or drop the accessors until there's a consumer.

**Rust `Expr.Call(...)` is broken.** `quotationExprs` emits `mkCall` with **3** args (`Replacements.Util.fs:1553`), but the Rust runtime's `mkCall` takes 4 curried params with no default. Beam solved this with `mk_call/3` → `mk_call/4`; TS/Python/PHP use default parameters. Rust will produce a partial application, not an `FSharpExpr`. Add a 3-arg overload, or make the emitter always pass 4.

**Rust `evaluate` drops the tuple index:** `| ExprTupleGet(inner, _) -> eval env inner` returns the whole tuple instead of element `index`. Compare PHP (`$eval(...)[$e['index']]`). Not covered by the new Rust tests.

**Rust `isCall` may not compile the way you expect.** The comment in `applyOperator` explicitly says *"Fable-Rust would emit string literals as match patterns, which isn't valid Rust; an if-chain compiles to string == comparisons"* — yet `isCall` uses exactly that: `| ExprValue(_, "null") -> None`. If the `applyOperator` note is accurate, this pattern is either invalid or silently never matches, which would make every static call on Rust report a non-`None` instance. Could you confirm what Rust this actually emits? If it doesn't match, use an if-chain here too.

**PHP is entirely untested.** There is no PHP test runner in `src/Fable.Build/Test/` and no PHP CI job, so `tests/Php/TestQuotation.fs` never runs and `quotation.php` is unverified. Concretely visible gaps in `quotation.php`:
- No `exprToString`, so `q.ToString()` (wired in `quotationExprs`) calls an undefined function.
- `substitute` doesn't recurse into `Call.instance` (TS does).
- `applyOperator` uses `intdiv` for `op_Division` (breaks on floats), and the `OPERATORS` const is dead.

I'd either verify the PHP path by hand against generated output and say so in the PR, or split PHP out.

## Design concerns

**`CapturingQuotation` is a broad change to a shared file.** The guard sits *ahead of every other case* in `makeCallWithArgInfo`, so it also intercepts constructors, interface/abstract dispatch, delegate `Invoke`, and module values — for module values the old path returned `funcExpr` directly, the new one wraps in `makeCall` unconditionally. It happens to be harmless today because `emitQuotedExpr` discards the callee and only reads `info.MemberRef`, but that's an implicit coupling worth a comment. It also means the emitter's `Operation` branch is now largely dead code for user quotations (operators arrive as `Call` nodes instead). Also worth checking: `FableTransforms` still runs over the `Quote` body, and uncurrying/beta-reduction on calls to members that no longer resolve to real output could rewrite them. A quicktest on `<@ List.map (fun x -> x + 1) [1] @>` for JS/Python would be reassuring.

**`Test` nodes abuse the `declaringType` slot.** `TypeTest` emits `mkCall(target, "op_TypeTest", [], typeToString typ)` — a fake method name with the tested type smuggled into the declaring-type field. Real F# quotations have a `TypeTest` node. A dedicated `mkTypeTest` would cost one constructor per runtime and avoid a landmine for anyone who later reads `declaringType` generically.

**DecisionTree inlining duplicates shared targets.** Every `DecisionTreeSuccess` reference re-inlines the whole target body, so an N-way match with a shared default emits N copies of the mk*-call tree into generated code. Fine for small quotations; worth a comment noting the blow-up. Also `List.item idx targets` inside the visitor is O(n²) — use an array.

## Smaller notes

- The `Option` case in `typeToString` renders as a Fantomas-mangled multi-line interpolation. Hoist it: `let suffix = if isStruct then "voption" else "option"`.
- `DelegateType` and curried `LambdaType` both stringify to `a -> b -> c` — indistinguishable.
- Rust `mkNull` stores `box 0` with a type tag, so `isValue` on a genuine `null`/`None`/`[]` literal hands back `0`. Documented in the comment, but it's a real semantic hole.
- Rust `applyOperator` `unbox<int>` for all comparisons — `<@ "a" = "b" @>` or float comparison will fail at runtime.
- The `Map.fs` fix is a genuine, valuable bug fix (`Unchecked.defaultof` → `getZero`/`mem::zeroed` panicking for `Rc`-backed values) with a good regression test — but it's unrelated to quotations and would land better as its own PR. Note `tryGetValue` still has the same defect for external callers.
- Adding `QuotationTypes.fs`/`Quotation.fs` unconditionally to `Fable.Library.Rust.fsproj` grows every Rust binary. Is there a feature-gate convention in that lib worth following?

## Test coverage

Structural tests (`Lambda(_, IfThenElse ...)`) are added consistently across Beam/JS/Python/PHP, which is the right instinct. Gaps: nothing exercises the new `declaringType` (unreachable — see above), nothing exercises `TypeTest`/`ListTest`, and no test covers `Get(ListHead/ListTail/OptionValue)` producing `get_Head`/`get_Value` calls. The Rust suite skips `Value` deconstruction entirely, which is where the `mkNull`/`box 0` design is weakest.

Thorium:
Thanks for the feedback. I'll address the issues and split these into multiple smaller features again.
They will be along the lines of:
- Quotations (shared emitter)
- Quotations and tests (Rust)
- Quotations and tests (Dart)
- Quotations and tests (PHP)
- Reflection (Rust)
- Other Rust bug fixes and missing features
- Environment on Dart/Erlang/PHP as well

Thorium:
Your PR feedback is now addressed. I am sorry I couldn't make the PRs independent; there are dependencies. If that feels messy, this picture hopefully helps: 

image

PR #4776, opened 2026-07-14
chore(beam): drop the inlined assertion special case

dbrattli:
Follow-up to #4775 (merged).

`Fable2Beam` intercepted every call site whose **import selector was named** `assertEqual`, `assertNotEqual`, `Testing_equal` or `Testing_notEqual`, and replaced it with an inlined assertion:

```erlang
test_infix_add_can_be_generated() ->
    case fable_comparison:equals(999, 6) of
        true  -> ok;
        false -> erlang:error({assert_equal, 6, 999})
    end.
```

Three reasons to drop it:

**It matched a name, not an entity.** Any imported function called `Testing_equal` (etc.), from any module, had its body replaced by an assertion — a user with their own `Testing.equal` helper would find their function silently never called. That is exactly what made `tests/Beam/Util.fs`'s own `equal` dead code: `fable_tests_beam_util:testing_equal/2` was emitted but no call site ever reached it.

**It is now redundant.** After #4775, `Fable.Core.Testing.Assert.AreEqual`/`NotEqual` lower to `fable_utils:assert_equal`/`assert_not_equal`, which raise. Any wrapper around `Assert` — including the test suite's `equal` — therefore raises on its own.

**Its rationale never held.** `FABLE-BEAM.md` justified the inlining as "no runtime dependency on a util module", but the inlined code called `fable_comparison:equals` anyway, so fable-library was already a dependency. The doc entries are updated accordingly: the Phase 5 checklist item, and a bullet documenting the `Assert_actual_N`/`Assert_expected_N` temporaries emitted by the now-deleted `storeIfComplex`.

## Effect on generated code

Call sites now go through the function that was always meant to run:

```erlang
test_infix_add_can_be_generated() ->
    fable_tests_beam_util:testing_equal(6, 6).

testing_equal(Expected, Actual) ->
    fable_utils:assert_equal(Actual, Expected).
```

⚠️ **Assertion failures now raise the library's map rather than a bare tuple.** The reason is readable, and `e.Message` reads cleanly in `try ... with`, but anything downstream pattern-matching on the old tuple needs updating:

```erlang
%% before: {assert_equal, 6, 999}
%% after:  #{message => <<"Expected: 999 - Actual: 6">>, actual => 6, expected => 999}
```

## Verification

Full Beam suite: **2586 passed, 0 failed**.

Assertions demonstrably still fire without the special case: with a deliberately corrupted expectation (`4 + 2 |> equal 999`) the suite reports exactly one failure, that test, with the map above — so the ~2,500 call sites are asserting through the library call, not silently passing.


dbrattli:
Retargeted to `main` so CI actually runs — `build.yml` only triggers on `pull_request: branches: [main]`, so while this was based on the #4775 branch it got no build.

⚠️ **Merge #4775 first.** This PR is stacked on it and its diff currently includes #4775's commit; the assertion inlining removed here is only safe once `Assert.AreEqual`/`NotEqual` lower to the throwing `fable_utils:assert_equal` (that PR). Once #4775 merges, this diff collapses to its own two files.

PR #4779, opened 2026-07-14
feat(all): auto-exclude plugin project references

MangelMaxime:


MangelMaxime:
Closing because this is actually trying to fix an issue that does not exist and make the situation worth.

PR #4782, opened 2026-07-14
feat(php): System.Environment

Thorium:
System.Environment for PHP.


Thorium:
Lint fails for php missing. Fixed in another PHP PR already.

PR #4783, opened 2026-07-14
feat(beam): support setting `Environment.CurrentDirectory`

Thorium:
System.Environment for Erlang/Beam


MangelMaxime:
Thank you, I reworked the PR a little because without knowing I had made a partial implementation of the features in another PR.

PR #4784, opened 2026-07-14
feat(dart): quotations

Thorium:
Quotations for Dart. This is dependent on https://github.com/fable-compiler/Fable/pull/4780 (so that should be merged first).

### DIFF
--- a/src/Fable.Transforms/Dart/Fable2Dart.fs
+++ b/src/Fable.Transforms/Dart/Fable2Dart.fs
@@ -220,6 +220,20 @@ module Util =
 
         match entRef.FullName, genArgs with
         | Types.enum_, _ -> Integer
+        // F# Quotation types are modelled dynamically by the quotation runtime,
+        // so map them to `dynamic` (the runtime returns plain tagged objects).
+        | Types.fsharpExpr, _
+        | Types.fsharpExprGeneric, _
+        // The NewUnionCase active pattern's first slot is typed as UnionCaseInfo in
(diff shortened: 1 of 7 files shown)


MangelMaxime:
Similar question as in the PHP PR

https://github.com/fable-compiler/Fable/pull/4785#issuecomment-4995816794

Thorium:
A typical F# developer is not a junior, but someone who has spent 10 years in C# and then got bored and thought things could be done better. Part of that 10-year journey has been trying to build overly generic solutions, learning complex systems and architectures, etc., which has led F# programmers to develop a good understanding of, for example, metaprogramming. 
I think F# shines these kinds of complex things.

And for that, reflection and quotations are important.

So I would want to see Fable + quotations + reflection (+typeproviders) possibility on different business domains:
- Web-frontend (TS/JS target)
- Web-backend (well, you can just use traditional F#)
- Mobile apps (I've seen react.js hacks, etc, but here, as far as I understand, Dart support could be useful)
- Performance critical code (Rust target)
- Data analysis (Python target)
- And so on. So I don't care so much about the target "do we need tech X", but rather, can I (or agents) do an impressive business-Y solution with F#.

MangelMaxime:
OK thank you for the explanations.

I guess we can accept the Dart quotation support as for Mobile development we don't have other good alternative.

The difficulty is where to draw the line with new targets supports etc. as we are a small set of developers and don't know all too well some of the target languages. But that's a problem for the future and there are technical solution to help us here.

I will review this PR later today or in the coming days.

PR #4785, opened 2026-07-14
feat(php): quotations

Thorium:
Quotations for PHP. This is dependent on https://github.com/fable-compiler/Fable/pull/4780 (so that should be merged first).


MangelMaxime:
Hello @Thorium,

Are you adding support for Quotations in PHP because you need it or for parity with other targets?

Asking because to be honest PHP target is more or less abandoned right now because it does not have any active maintainer. Kind of like Dart target. 

If this is for the later reason, I think we can close this PR as it will add more work to the maintainers and we don't even run the tests for that target on the CI.

It was ported originally, because someone made a fork of Fable to support it and it was easy enough to include. But the target never got any traction from the community.

Edit: I closed the PR by mistake instead of sending my comment 😅. I re-opened it waiting for your answer

Thorium:
This was for parity: I don't know if anyone uses PHP nowadays.

MangelMaxime:
For the reason explained above I am going to close this PR.

If PHP get traction from the community we can always cherry pick the commits

PR #4787, opened 2026-07-14
feat(rust): reflection (PR 4 of 7 Rust improvements) 

Thorium:
Adds reflection to Rust.
Part of the dependency chain.

Also makes the PropertyGet quotation pattern work: System.Reflection.PropertyInfo now erases to a shared FSharpPropertyInfo carrier (as UnionCaseInfo/MethodInfo already do), used by both reflection and quotations, so pi.Name works from either. The field getter is resolved from the type registry by name, like getValue(pi, v) => v[pi[0]] in JS/TS.


ncave:
Closed with #4790

PR #4791, opened 2026-07-14
feat(rust): add tests to task computation expressions

Thorium:
Turns on the `task { }` computation expression tests for the Rust target, and lets a computation omit a trailing `return`.

The `Task_` / `TaskBuilder_` runtime already existed in `Async.rs`, but `TaskTests.fs` was commented out of the test project, so none of it was exercised. Most of the core turned out to work already, so the first commit is tests only; the second is a small fix described below.

Like `AsyncTests.fs`, the module is gated with `OuterAttr("cfg", "feature = \"threaded\"")`, since the Task runtime (thread pool + futures) is behind the `threaded` feature.

### Supported (14 tests)

`return`, `let!` (including several bindings in one computation), `do!`, `Task.FromResult`, `.Result`, `Async.AwaitTask` and `Async.StartAsTask`, at concrete types.

Three of the tests cover work that does not complete immediately, rather than only tasks that are already finished:

- a task awaiting a 300ms computation observes its side effect, proving it really waits
- `.Result` blocks until the work completes, and the work is queued rather than run inline
- a delay awaited from inside a computation, via `let! _ = Async.Sleep ms |> Async.StartAsTask`

### No trailing `return` (second commit)

A body that does not end in `return` is just `return ()`, but it compiles to `TaskBuilderBase.Zero`, whose declared type is FSharp.Core's `TaskCode<'TOverall, unit>`. The Rust backend emits generic arguments for a library import called without args (`Fable2Rust`, "library imports without args need explicit genArgs") taking them from the return type, so the resumable-code internals leaked into the call:

```rust
zero::<&MutCell<ResumableStateMachine_1<_>>, bool>()
```

Giving the `Zero` mapping a `unit` return type keeps those generics out of the emitted call, which is then simply `zero()`, and `Task_::zero` loses a generic parameter it never used. No `ResumableStateMachine` reference remains in the generated code. This makes the common `task { let! x = f (); useIt x }` shape work.

### Not yet supported

The remaining tests are kept in the file, commented, each with the reason. They cannot be marked with `OuterAttr("ignore")` instead: Fable rejects these builder members outright, so they fail the build rather than compiling into a skipped test.

- **`while` / `for` / `try-with` / `try-finally` / `use!`** — `TaskBuilderBase.While`, `For`, `TryWith`, `TryFinally`, `Using` and `Combine` are not mapped in the Rust `Replacements`, and `Task_` has no matching runtime combinators. (`AsyncBuilder_` lacks these too; the async tests just avoid those constructs.) A loop body would also need to be re-runnable, whereas a `Task` here is hot-started and caches its result.
- **`Task.Delay`** — not implemented; `do! Task.Delay ...` also needs binding of the non-generic `Task`, which goes through `TaskBuilderExtensions.LowPriority.Bind` and `Task.GetAwaiter`.
- **`GetAwaiter().GetResult()`** — not supported, so `.Result` is used to wait.
- **Generic helpers over `Task<'T>`** — e.g. `let run (t: Task<'T>) = t.Result` emits `<T: Clone + 'static>` while `Task<T>` requires `T: Send`, so bounds are missing. Concrete types are unaffected.
- **`TaskCompletionSource`** — not implemented.

### Testing

- `./build.sh test rust --threaded` → 2516 passed, 14 task tests running
- `./build.sh test rust` → 2487 passed, module gated out as expected
- `./build.sh test rust --no_std` → module gated out as expected

### DIFF
--- a/src/Fable.Transforms/Rust/Replacements.fs
+++ b/src/Fable.Transforms/Rust/Replacements.fs
@@ -3181,7 +3181,10 @@ let taskBuilderB (com: ICompiler) (ctx: Context) r t (i: CallInfo) (thisArg: Exp
         Helper.LibCall(com, "Task", "delay", t, args, i.SignatureArgTypes, ?loc = r)
         |> Some
     | "Zero", _, _ ->
-        Helper.LibCall(com, "Task", "zero", t, args, i.SignatureArgTypes, ?loc = r)
+        // A body that does not end in `return` is just `return ()`. Its declared
+        // type is TaskCode<_,_>, whose generics are FSharp.Core resumable-code
+        // internals, so use unit here to keep them out of the emitted call.
+        Helper.LibCall(com, "Task", "zero", Fable.Unit, args, i.SignatureArgTypes, ?loc = r)
         |> Some
     | _ -> None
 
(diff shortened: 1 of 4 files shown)


Thorium:
Many "task" features are still missing.
If you prefer to show the library's current status with unit tests, then keep this PR.
If you prefer to implement the whole task CE as a bigger feature and not to have partial implementations, then close this PR.

PR #4805, opened 2026-07-16
fix(python): make union case classes hashable again

### LINKED ISSUE #4804: [Python] Union values are unhashable — `dataclass()` sets `__hash__ = None` on case classes
Hashing an F# union value that contains another union as a field raises `TypeError: unhashable type` at runtime on the Python target. The same code works on .NET (all F# unions are structurally hashable) and on the JS target.

This breaks common operations like `List.distinct`, `Set.ofList` or using unions as dictionary keys whenever the union is nested inside another union.

The exception is also easy to miss: if it fires inside an `async` continuation it doesn't fail cleanly but deadlocks the running event loop.

## Minimal reproduction

`Repro.fsproj`:

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
  </PropertyGroup>
  <ItemGroup>
  </ItemGroup>
  <ItemGroup>
  </ItemGroup>
</Project>
```

`Program.fs`:

```fsharp
module Program

type Inner = | A
type Outer = | Wrap of Inner

// works on .NET, crashes on Python
printfn "hash: %d" ((Wrap A).GetHashCode())

// real-world trigger: any structural hashing of nested unions
[ Wrap A; Wrap A ] |> List.distinct |> printfn "distinct: %A"
```

```sh
dotnet run                                  # .NET: prints a hash and [Wrap A]
pip install fable-library==5.9.0
python out/program.py
```

## Expected behavior (same as .NET):

```
hash: <some number>
distinct: [Wrap A]
```

## Actual behavior:

```
TypeError: unhashable type: 'Inner_A'
```

## Root cause

Union case classes are decorated with `@tagged_union` (`fable_library/union.py`), which applies `dataclass()` internally. Per the Python data model, a dataclass that generates `__eq__` (the default) and is not frozen sets `__hash__ = None` on the class. That shadows the `__hash__` inherited from the `Union` base class (`HashableBase.__hash__`, which delegates to `GetHashCode`), so every case class is explicitly unhashable.

`Union.GetHashCode` itself computes `hash((self.tag, *self.fields))` — the moment a field is another union, the built-in `hash()` hits the `__hash__ = None` of the inner case class and raises. The `tagged_union` docstring says "`__eq__`, `__repr__`, `__hash__` are generated", but dataclass only generates a hash for `frozen=True`, so the hash was never actually there.

Restoring the intended hash after the `dataclass()` call fixes it:

```python
dc_cls.__hash__ = HashableBase.__hash__
```

This is consistent with the dataclass-generated `__eq__` (equal values have equal `(tag, *fields)`, hence equal hashes).

caroott:
Fixes #4804

Hashing a union value that contains another union as a field raises TypeError: unhashable type at runtime, e.g.
```
type Inner = | A
type Outer = | Wrap of Inner

(Wrap A).GetHashCode() // TypeError: unhashable type: 'Inner_A'
```
The same code works on .NET and JS, and breaks things like `List.distinct` or using unions as dictionary keys on Python.

The `tagged_union` decorator applies `dataclass()` to every case class. Since that generates `__eq__` without `frozen=True`, Python sets `__hash__ = None` on the class, shadowing the `__hash__` the Union base inherits from `HashableBase`. So case classes were explicitly unhashable, and `Union.GetHashCode (hash((self.tag, *self.fields)))` blows up as soon as a field is itself a union. The decorator's docstring already claims `__hash__` is generated. This restores it after the `dataclass()` call. It's consistent with the dataclass `__eq__`: equal values have equal `(tag, *fields)`, hence equal hashes.

Added tests for the hash protocol, hash/eq consistency, nested unions and set membership.

One note: type checkers still consider the case classes unhashable statically (`dataclass_transform` models them as non-frozen dataclasses), even though `__hash__` now exists at runtime.

### DIFF
--- a/src/fable-library-py/fable_library/union.py
+++ b/src/fable-library-py/fable_library/union.py
@@ -126,11 +126,13 @@ def tagged_union(tag: int):
     Uses @dataclass_transform() so type checkers understand:
     - Field annotations become constructor parameters
     - __match_args__ is generated
-    - __eq__, __repr__, __hash__ are generated
+    - __eq__, __repr__ are generated
 
     Additionally sets:
     - cls.tag = tag (numeric case discriminator)
     - cls.fields property (list of field values for backwards compat)
+    - cls.__hash__ from HashableBase (dataclass sets it to None when
+      generating __eq__, which would make union values unhashable)
     """
 
     def decorator[T](cls: type[T]) -> type[T]:
@@ -148,6 +150,12 @@ def tagged_union(tag: int):
             return Array[Any]([getattr(self, name) for name in field_names])
 
         dc_cls.fields = fields
+
+        # dataclass() generates __eq__ and therefore sets __hash__ to None.
+        # Restore the hash protocol so union values stay hashable via
+        # GetHashCode, e.g. when a union is a field of another union.
+        dc_cls.__hash__ = HashableBase.__hash__
+
         return dc_cls
 
     return decorator
--- a/src/fable-library-py/tests/test_union.py
+++ b/src/fable-library-py/tests/test_union.py
@@ -1,0 +1,45 @@
+from fable_library.union import Union, tagged_union
+
+
+class _Shape(Union):
+    @staticmethod
+    def cases() -> list[str]:
+        return ["Circle", "Boxed"]
+
+
+@tagged_union(0)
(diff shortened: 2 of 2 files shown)


dbrattli (approved):
Thanks for fixing!

PR #4807, opened 2026-07-16
fix(python): make class declarations and references agree on type names

### LINKED ISSUE #4806: [Python] Generated class names don't match their references (lowercase type names, name collisions)
When an F# type has a lowercase name, Fable declares the Python class in PascalCase (`class Shape`) but references to it still use the original name (`shape`). Importing such a type from another module fails with an `ImportError`.

There is a second problem with type names that collide across modules. The import gets aliased (`Direction as Direction_1`), but type annotations still use the original name, so they either point at the wrong class or raise a `NameError`.

Both work on .NET and JS.

### Minimal reproduction

`Repro.fsproj`:

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="Lib.fs" />
  </ItemGroup>
  <ItemGroup>
  </ItemGroup>
</Project>
```

`Lib.fs`:

```fsharp
module Lib

type shape() =
    member _.Size = 1
```

`Program.fs`:

```fsharp
module Program

open Lib

let s = shape()
printfn "size: %d" s.Size
```

```sh
dotnet run                                  # .NET: size: 1
pip install fable-library==5.9.0
python out/program.py
```

### Expected behavior (same as .NET)

```
size: 1
```

### Actual behavior

```
from lib import (shape__ctor, shape, shape__get_size)
ImportError: cannot import name 'shape' from 'lib'. Did you mean: 'Shape'?
```

`lib.py` declares `class Shape`, `program.py` imports `shape`.

### Second reproduction, name collision

`Shapes.fs`:

```fsharp
module Shapes

type Direction =
    | Up
    | Down
```

`Program.fs`:

```fsharp
module Program

open Fable.Core

[<AttachMembers>]
type Direction() =
    static member Up = Shapes.Direction.Up

printfn "%A" Direction.Up
```

.NET prints `Up`. Python:

```
  File "out/program.py", line 15, in Direction
    Up = StaticProperty[Direction](Direction_Up.singleton)
NameError: name 'Direction' is not defined. Did you mean: '_Direction'?
```

The union is imported as `Direction as Direction_1` because the name collides with the local class, but the `StaticProperty[...]` annotation still says `Direction`, which isn't bound yet while the class body runs.

### Root cause

Class declarations apply `Naming.toPascalCase` (`Fable2Python.Transforms`, since #4134). References come from `getEntityDeclarationName`, which doesn't apply it, so the two only match if the F# name is already PascalCase.

The annotation problem is separate: `makeEntityTypeAnnotation` in `Fable2Python.Annotation.fs` registers the import via `com.GetImportExpr(...)` but ignores the returned local identifier, so the alias never ends up in the annotation.

caroott:
Fixes #4806

Class declarations are PascalCased on the Python target, but references to them are not, so a type with a lowercase name gets declared as class Shape while other modules import shape:
```
from lib import (shape__ctor, shape, shape__get_size)
ImportError: cannot import name 'shape' from 'lib'. Did you mean: 'Shape'?
```
First commit: all entity names funnel through `getEntityDeclarationName`, so the PascalCasing is applied there (Python branch only) and declarations and references agree by construction. The same commit fixes the equivalent mismatch for member names: reference sites re-case the full composed name, so a lowercase member with an overload suffix was declared as `shape__ctor_Z524259A4` but imported as `shape__ctor_z524259a4`. The composed declaration name now gets the same casing convention.

Second commit: when an imported union's name collides with a local declaration, the import is correctly aliased (`Direction as Direction_1`) but `makeEntityTypeAnnotation` discarded the returned local identifier and emitted the original name, which fails with a `NameError` inside `StaticProperty[...]` annotations. Annotations now use the actual (possibly aliased) import identifier.

Only types whose names aren't already PascalCase produce different output than before, and references to those were broken one way or another.
Note that `toPascalCase` only upper-cases the first character, which is why types nested in modules (mangled to `Module_name`) never hit this.

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -676,7 +676,11 @@ module Helpers =
 
         let sanitizedName =
             match com.Options.Language with
-            | Python -> Fable.Py.Naming.sanitizeIdent Fable.Py.Naming.pyBuiltins.Contains name part
+            | Python ->
+                // Python classes are declared with PascalCase names (PEP 8), so entity
+                // references must use the same casing as the declaration
+                let name = Fable.Py.Naming.toPascalCase name
+                Fable.Py.Naming.sanitizeIdent Fable.Py.Naming.pyBuiltins.Contains name part
             | Rust -> (entityName |> cleanNameAsRustIdentifier)
             | Dart -> Naming.sanitizeDartIdent (fun _ -> false) name part
             | _ -> Naming.sanitizeJsIdent (fun _ -> false) name part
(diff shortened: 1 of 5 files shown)


dbrattli (approved):
Thanks for fixing!

Issue #4737, opened 2026-07-08
No error or warning on duplicate CompiledName


kerams:
https://fable.io/repl/#?code=C4TwDgpgBAGlC8BYAUFNUA+UDaAeAwgPYC2YAlgDYQAmAcgIbHQBEAZq8wHwC6UAginSYcBEuSp1GLdl14AhIA&html=Q&css=Q

```fsharp
type X =
    | [<CompiledName "ff">] A
    | [<CompiledName "ff">] B
```

Fable, unlike .NET, honours this attribute on union cases. However, it doesn't verify that the names are unique. At the very least I would like to see a warning if there are valid use cases for this, otherwise an error.

MangelMaxime:
I am almost certain that I saw some JavaScript libraries that used different value of enum to map to the same value.

In those scenarios it is normal case and does not need a warning or error message as they would be false positive.

Because of that I am unsure if this is something we can improve.

kerams:
You can disable a warning, but you can't conjure one out of thin air:). I am starting to make heavy use of `CompiledName` and I want compile-time assurance that I didn't make a mistake in duplicating labels.

MangelMaxime:
> You can disable a warning, but you can't conjure one out of thin air

Fable does not have this feature (yet)

Also, I confused `CompiledName` with `CompiledValue` which is something else.

And indeed, I think have 2 enum cases name being the same is probably strange and should be reported as error. If there are issues with that in the future, we can reduce it to a warning.

MangelMaxime:
While working on the fix for this issue I found one Fable library using this feature:

[Fable.React.Props.fs](https://github.com/fable-compiler/fable-react/blob/b3dc782001c6db4d24c526ff401a99ab571f1d8d/src/Fable.React/Fable.React.Props.fs#L31-L37)

```fs
type Prop =
    | Key of string
    | Ref of (Element->unit)
    /// To be used in combination with `useRef` hook
    | [<System.Obsolete("Use RefValue")>] [<CompiledName("ref")>] RefHook of IRefHook<Element option>
    | [<CompiledName("ref")>] RefValue of IRefValue<Element option>
    interface IHTMLProp
```

I am still doing more investigation

PR #4808, opened 2026-07-16
feat(all): support suppressing warnings via `// fable-disable` comments

MangelMaxime:
Hello,

In Fable we added support for `TreatWarningsAsErrors`: some warnings Fable generates as informational signals can be treated as errors when this is enabled.

With this PR, I'm adding support for ESLint-style suppression comments (we can't reuse `#nowarn` - it only accepts numeric codes and is resolved by FCS before Fable's own warnings even exist - and a new compiler-level directive would mean patching FCS itself).

This adds support for:

- Disabling the current line: `// fable-disable-line`
- Disabling the next line: `// fable-disable-next-line`
- Disabling a block: `// fable-disable` ... `// fable-enable`

For example, a user recently [requested a warning for duplicate `CompiledName`s on union cases](https://github.com/fable-compiler/Fable/issues/4737). I originally thought this should be an error, but some libraries (e.g. Fable.React) use duplicate names on purpose to represent JS binding "overloads" - so it could only ever be a warning.

By adding suppression comments, people can keep `TreatWarningsAsErrors` enabled for safety, while still deciding, case by case, to suppress the specific warnings they judge harmless.

---

The current codes in `src/Fable.Transforms/Global/WarningCodes.fs` are here to serve as an example.


ncave (commented):
@MangelMaxime Looks good, just some thoughts in no particular order (non-blocking, feel free to disagree or ignore):
- Since files can be compiled in parallel, would the non-thread safe warnings cache work as expected?
- I may be missing something, but almost all `Directive` union cases look the same, what's the difference between them, and can they be simplified?
- It's nice to standardize the warnings with codes, instead of using free form, absolutely. But that also adds some complexity and maintenance on its own.
- Now that everyone heavily uses AI, perhaps we can eliminate some classes of warnings by just implementing the missing methods or overrides.

dbrattli (commented):
Overall this is a solid design — suppressing at `AddLog` (so it composes with `TreatWarningsAsErrors` and errors stay unsuppressible) and using real comment tokens instead of text matching are both the right calls. A few things worth addressing before merge.

## 🔴 Blockers

**1. Directives inside `#if FABLE_COMPILER` are silently ignored.**
There are two independent tokenizations here. FCS compiles the project *with* `FABLE_COMPILER` defined (`Entry.fs:317-331`), so code inside those blocks is active and its warnings fire. But `WarningSuppression.compute` runs a *second*, independent pass with **no defines**:

```fsharp
// WarningSuppression.fs
let sourceTok = FSharpSourceTokenizer([], None, None, None)
```

With `defines = []`, every token inside a `#if FABLE_COMPILER` block comes back as `InactiveCode`, not `Comment`, so `scanLineComments` drops the directive. Net effect: the warning fires but the suppression comment sitting right next to it is invisible — and `#if FABLE_COMPILER` is exactly where Fable-specific code (and its warnings) lives.

Reproduced against FCS 43.12.203: `defines = []` → the comment tokenizes as `InactiveCode`; `defines = ["FABLE_COMPILER"]` → `Comment`.

Fix: thread the real define list (`options.Define`, ideally plus the project's `DefineConstants`) into the tokenizer, the way `fable-standalone/src/Lexer.fs:52-62` already does. Needs a regression test.

**2. The warning code is never shown to the user.**
`LogEntry.Code` is stored but `formatLog` (`Main.fs:65-84`) never renders it, so there's no way to discover that a warning is `FABLE0001` in order to suppress it. Rendering it MSBuild-style (`warning FABLE0001: …`) also makes Fable warnings parseable by IDEs/MSBuild loggers. Worth a test asserting the code appears in the formatted output.

## 🟠 Users will trip on these

**3. Malformed directives fail silently.** The regex `^fable-(disable-next-line|disable-line|disable|enable)(?:\s+(.+))?$` means:
- `// fable-disable-line: FABLE0001` (colon) — no match, silently suppresses nothing. People coming from `# noqa: E501` / `@ts-ignore` will write this.
- `// fable-disable-line FABEL0001` (typo) — parses, suppresses nothing, no signal.
- `// fable-disable-line FABLE0001 because …` — the prose words all become "codes".

Suggest tolerating `:`/`--` separators, and reporting unused/unknown directives. The registry is closed (`WarningCodes.fs`), so validating codes is cheap — ESLint (`reportUnusedDisableDirectives`), Rust (`#[expect]`), and TS (`@ts-expect-error`) all converged on this. This is the highest-value follow-up.

**4. Only `range.start.line` is consulted.** `IsSuppressed(r.start.line, code)` means a trailing `// fable-disable-line` on a multi-line / formatted expression has no effect:
```fsharp
"abc".StartsWith(
    "a", true, CultureInfo.InvariantCulture) |> ignore // fable-disable-line FABLE0001  // ← ignored
```
Consider "suppressed if the directive falls anywhere in `[start.line, end.line]`". At minimum, pin current behaviour with a test and document it.

## 🟡 Correctness / robustness

**5. Source is re-read from disk at warning time, and the content hash is discarded.** `State.fs` does `(snd (read fileName)).Value`; `File.ReadSource()` (`File.fs:22-29`) returns `lazy readAllTextNonBlocking path` once the hash is set — a fresh disk read each time. In watch mode the on-disk text can be newer than what was compiled, so line numbers are computed against the wrong source. `SourceReader` returns `int * Lazy<string>` where the `int` is a content hash — an ideal cache key that's currently thrown away, and `suppressionsCache` is per-`CompilerImpl` (per compiled file), so nothing is shared. Also `try … with _ ->` swallows everything, including `OperationCanceledException` and the `KeyNotFoundException` for files outside the project.

**6. Inline-call warnings can only be suppressed in the definition file.** `addLogWithCode` uses `actualFile = ToFile` (= `inExpr.FileName`, `FSharp2Fable.Util.fs:2966`), so a warning from an inlined library function is suppressible only by editing the *library's* source — and not at all for a precompiled dll (file not in `fileDic` → exception → `Empty`). Defensible default; needs a test and a doc line, since it's the case users will file issues about.

## ⚪ Minor

- Codes compared case-sensitively; `fable0001` won't match. Normalize.
- `if lines.Length = 0` in `compute` is dead — `String.Split` never returns an empty array.
- Bare `// fable-disable` (no codes) blanket-disables all Fable warnings to EOF. Consider requiring codes for the block form, or flagging it.
- `WarningCodes.stringSecondArgumentIgnored`'s doc says "with a `StringComparison` argument on the Dart target", but the call site fires on any extra arg — the comment is narrower than the behaviour.
- `cultureInfoIgnored ()` could just be a value; the `()` reads as if it did work.
- `fable-standalone` compiles `State.fs` but never passes a `sourceReader`, so suppression is a silent no-op in the REPL. Fine, but worth a comment so it isn't mistaken for a bug.
- The test comment mentioning "Python's Replacements.fs" is confusing — these integration tests compile to JS and never touch Python's replacements.
- `Compiler.AddLog` gaining a parameter is source-breaking for external implementers of that interface (Fable.Compiler is on NuGet). Probably fine, but worth a release note.

## Tests to add

`#if FABLE_COMPILER` directive (catches #1); **errors are not suppressed** (deliberate invariant, currently unpinned); `(* fable-disable-line *)` block-comment form (implemented in `stripCommentMarkers`, untested); multi-line expression with a trailing directive; `fable-disable` with no matching `fable-enable`; a malformed/typo'd directive; CRLF input; and a warning raised from an inlined function.

PR #4825, opened 2026-07-21
fix(python): avoid UTF-8 boundary panic in integer parsing

### LINKED ISSUE #4823: [Python] Int32.TryParse panics on valid multi-byte Unicode input
`System.Int32.TryParse` panics in the Python target when the input starts with a multi-byte Unicode character. Instead of returning `false`, the native Rust runtime raises a `pyo3_runtime.PanicException`.

The same input works correctly on .NET.

### Minimal reproduction

`Repro.fsproj`:

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
  </PropertyGroup>
  <ItemGroup>
  </ItemGroup>
  <ItemGroup>
  </ItemGroup>
</Project>
```

`Program.fs`:

```fsharp
module Program

let success, value = System.Int32.TryParse "–"
printfn "%b, %d" success value
```

```sh
dotnet run
pip install fable-library==5.11.0
python out/program.py
```

### Expected behavior (same as .NET)

```text
false, 0
```

### Actual behavior

```text
thread '<unnamed>' panicked at src\ints.rs:1069:22:
end byte index 2 is not a char boundary; it is inside '–' (bytes 0..3 of string)

Traceback (most recent call last):
  File "out/program.py", line ...
pyo3_runtime.PanicException: end byte index 2 is not a char boundary; it is inside '–' (bytes 0..3 of string)
```

Other nonnumeric multi-byte inputs such as `�`, `é`, and `𝟙` trigger the same problem.

### Probable root cause

`determine_radix` in `src/fable-library-py/src/ints.rs` checks `string.len() >= 2` and then slices the Rust string by byte index:

```rust
match &string[..2] {
    "0x" | "0X" => 16,
    "0b" | "0B" => 2,
    "0o" | "0O" => 8,
    _ => default_radix,
}
```

`str::len()` returns the number of bytes, but Rust string slices must end on a UTF-8 character boundary. For `–`, which occupies three bytes, byte index 2 lies inside the character.

`remove_prefix` contains the same unsafe `&string[..2]` pattern.

Both locations could use checked slicing instead:

```rust
match string.get(..2) {
    Some("0x") | Some("0X") => 16,
    Some("0b") | Some("0B") => 2,
    Some("0o") | Some("0O") => 8,
    _ => default_radix,
}
```

This preserves ASCII prefix detection while treating other valid Unicode input as a normal parse failure.

caroott:
Fixes #4823

The Python integer parser checks for `0x`, `0b`, and `0o` prefixes by slicing the first two bytes of a Rust `str`. This is safe for ASCII, but byte index 2 can fall inside a multi-byte UTF-8 character. In that case, `Int32.TryParse` raises a `PanicException` instead of returning `false`.

Use `str::get(..2)` in `determine_radix` and `remove_prefix`, so an invalid character boundary is treated like a non-prefix. The remaining `&string[2..]` is safe because it is only reached after matching a two-byte ASCII prefix. Prefix handling is otherwise unchanged.

Add native regression coverage for the `NumberStyles.Any` and `NumberStyles.HexNumber` paths, which exercise the two checked-slice locations.

### DIFF
--- a/src/fable-library-py/src/ints.rs
+++ b/src/fable-library-py/src/ints.rs
@@ -1064,16 +1064,13 @@ fn determine_radix(string: &str, style: i32, default_radix: i32) -> i32 {
         return 16;
     }
 
-    // Single length check followed by pattern match
-    if string.len() >= 2 {
-        match &string[..2] {
-            "0x" | "0X" => 16,
-            "0b" | "0B" => 2,
-            "0o" | "0O" => 8,
-            _ => default_radix,
-        }
-    } else {
-        default_radix
+    // `get` returns None when the string is shorter than two bytes or the
+    // boundary falls inside a multi-byte character, so this never panics
+    match string.get(..2) {
+        Some("0x") | Some("0X") => 16,
+        Some("0b") | Some("0B") => 2,
+        Some("0o") | Some("0O") => 8,
+        _ => default_radix,
     }
 }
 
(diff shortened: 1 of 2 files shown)


dbrattli (approved):
Thanks for fixing!

PR #4836, opened 2026-07-22
fix(js/ts,python): inline Option/ValueOption combinators to avoid closures and allocations

MangelMaxime:
By building the replacement using AST directly, we allow the code to be candidate to `InlineIfLambda` which improves performances and can be critical for some library chasing that path.

### DIFF
--- a/src/Fable.Transforms/FableTransforms.fs
+++ b/src/Fable.Transforms/FableTransforms.fs
@@ -230,10 +230,14 @@ let noSideEffectBeforeIdent identName expr =
         | Let(_, v, b) -> findIdentOrSideEffect v || findIdentOrSideEffect b
         | TypeCast(e, _)
         | Test(e, _, _) -> findIdentOrSideEffect e
-        | IfThenElse(cond, thenExpr, elseExpr, _) ->
-            findIdentOrSideEffect cond
-            || findIdentOrSideEffect thenExpr
-            || findIdentOrSideEffect elseExpr
+        // Only `cond` always runs; `thenExpr`/`elseExpr` are conditional, so treat them like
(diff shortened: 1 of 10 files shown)


MangelMaxime:
This PRs surfaced a bug in Fable that is fixed in the second commit

dbrattli (approved):
Inlines Option/ValueOption combinators to AST instead of library calls — removes a closure allocation + lib call per combinator and opens the door to `InlineIfLambda`. Nicely done.

**Correctness** — the two traps of this optimization are handled:
- Each option arg is evaluated exactly once (`bindOnce`/`matchOption` binds it before the test + unwrap); `map2`/`map3` bind all args eagerly up front, matching F#'s eager argument semantics.
- The function arg is evaluated eagerly even when its branch isn't taken. The `noSideEffectBeforeIdent` change (treating `IfThenElse` branches as opaque/unsafe) is the load-bearing piece here — it stops `canInlineArg` from sinking a side-effecting, singly-referenced value into a conditional branch. It's strictly more conservative, so it can only suppress inlinings, never introduce a wrong one. Well covered by the "evaluate their function-producing argument exactly once" and "doesn't double-evaluate" tests.

Removed `Option.ts` exports are only defined, never imported internally; other targets use their own libraries and are untouched. Verified all target suites green in CI (incl. Dart, Beam, Rust, integration fixture).

**Minor (non-blocking):** `retType` in `map`/`map2`/`map3`/`bind` destructures `mapping.Type`/`binder.Type` with a whole-type fallback; `innerType t` (the known result option type) yields the same value more robustly and would be simpler.

dbrattli (approved):

PR #4837, opened 2026-07-22
fix(all): eliminate dead bindings left by CE/inline lowering

MangelMaxime:
Improve DCE, as those could not be optimized by JavaScript bundlers.

```fs
let dceTest (log: string -> unit) =
    el () {   // CE builder, all members inline and discard their argument
        log "a"
        log "b"
    }
```

**Before**

```fs
export function dceTest(log_1) {
    let _arg;
    log_1("a");
    log_1("b");
    _arg = undefined;
}
```

**After**

```
export function dceTest(log_1) {
    log_1("a");
    log_1("b");
}
```

### DIFF
--- a/src/Fable.Transforms/Babel/BabelPrinter.fs
+++ b/src/Fable.Transforms/Babel/BabelPrinter.fs
@@ -353,21 +353,28 @@ module PrinterExtensions =
 
         /// True when the expression's leftmost token would otherwise be parsed as the start of
         /// a block/function/class declaration, e.g. `{ A: 1 };` parses as an invalid block.
+        ///
+        /// Sub-expressions that the printer already wraps via `ComplexExpressionWithParens` when
+        /// `IsComplex` (e.g. a `CallExpression`'s callee) are skipped: they're safely parenthesized
+        /// regardless of statement position, so recursing into them would add a redundant paren pair.
         member printer.NeedsParensAsExpressionStatement(expr: Expression) =
+            let recurseUnlessAutoWrapped subExpr =
+                not (printer.IsComplex(subExpr))
+                && printer.NeedsParensAsExpressionStatement(subExpr)
+
             match expr with
             | CommentedExpression(_, e) -> printer.NeedsParensAsExpressionStatement(e)
             | ObjectExpression _
             | ClassExpression _
             | FunctionExpression _ -> true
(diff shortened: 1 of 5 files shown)


dbrattli (approved):
Nicely scoped fix, and the three-part shape is well-motivated: the shared transform now surfaces a discarded value at statement position on every target, and it's exactly JS and Rust that need printer-level support for it.

**Correctness — verified against the surrounding code**

- **`FableTransforms.fs`** — new case is correctly ordered before the existing erasing case and precisely guarded (`countReferencesUntil 1 … = 0` + `canHaveSideEffects`), which is exactly the gap `tryInlineBinding` couldn't cover (0 uses → nothing to inline into → dead `let` retained). Evaluation order and result type are preserved, and the `Sequential exprs -> Sequential(value :: exprs)` branch avoids needless nesting.
- **`BabelPrinter.fs`** — checked against the printers: `ObjectExpression` is *not* `IsComplex`, so member/callee/operand sub-exprs aren't auto-wrapped, meaning `recurseUnlessAutoWrapped` correctly re-adds parens for `{a:1}.b` / `{}.f()` / `{a:1} + 1` while skipping already-wrapped complex sub-exprs — no double wrapping. `UpdateExpression(false, …)` = postfix, so recursing on the argument is the right leftmost-token choice.
- **`Fable2Rust.fs`** — `transformExprsAsStmts` keeps the tail via the unchanged `transformAsStmt` and forces non-tail to `Semi`, matching rustc's `Expr`-vs-`Semi` semantics; behavior-preserving for the tail, fixing only the block-like non-tail case.

Also confirmed the reasoning for why Python/Dart/BEAM need no printer changes (no statement-start grammar ambiguity, no static tail-type rejection) — consistent with the build being green on all targets.

**Minor, non-blocking**

- `NeedsParensAsExpressionStatement` diverges from `PrintConditionalExpression`'s constant-fold: for a `ConditionalExpression` with a literal-boolean test the printer emits a branch directly, but the check inspects `test`. Requires an unfolded constant conditional with an object-literal branch at a discarded position — almost certainly unreachable, but a `ConditionalExpression(_, c, a)` fallthrough would make it airtight.
- `SequenceExpression` falls through to `false`; Fable only emits these inside ternary branches, not at statement position, so it's a theoretical gap only.

Comments are excellent (each explains the *why*), and the tests are pure additions targeting the exact shapes the fixes address. LGTM.

PR #4870, opened 2026-08-03
fix(all): recompile when a source file is added or its output is missing

dbrattli:
## Problem

Adding a `<Compile Include>` through an imported `.props` could leave a build broken while still reporting success — `Skipped compilation because all generated files are up-to-date!`, exit code 0, new module never generated. The failure surfaced far from its cause (Python: `TypeError: exceptions must derive from BaseException`; Beam: `'Scalars' is not defined`). Bad for CI: a pipeline that adds a file and reuses a warm cache green-lights a build whose new code was never compiled.

Two independent defects, both reproduced from source.

### 1. The project options cache ignored imported .props/.targets

Cache validity compared only `.fsproj` and reference timestamps. A `.props` adds `<Compile Include>` items without any `.fsproj` being touched, so the stale source list was reused. Adding the file directly to the `.fsproj` always worked.

Fixed by also invalidating on the files taking part in a project's MSBuild evaluation: transitive `<Import Project>`, plus `Directory.Build.props`/`.targets` and `Directory.Packages.props` found upwards. Only `$(MSBuildThisFileDirectory)` and `$(MSBuildProjectDirectory)` are expanded — an import path built from any other property is skipped, since missing an import costs a stale cache while guessing wrong costs a bogus path.

### 2. A missing output file counted as up-to-date

`getFilesToCompile` selects a source whose output does not exist (`Main.fs:945`), then `areCompiledFilesUpToDate` reported that same file as up-to-date (`Main.fs:981`) and skipped the compilation it had just asked for. No `.props` needed: delete any generated file and rebuild → skipped, exit 0, file still missing.

## Trade-off

A file whose generated code is empty is never written, so it is indistinguishable from one that was never compiled. Projects containing such a file lose the skip-compilation shortcut — real, not hypothetical (`module Empty` emits no output on JS/Dart/Beam), and measured at ~1.2s recompile instead of a skip. Normal projects still skip (0.23–0.28s), so the added import walk costs nothing measurable. Losing a shortcut is recoverable; silently shipping a half-generated build is not.

## Verification

Integration tests in `CacheInvalidationTests.fs` for both cases, checked in both directions — they fail without this change and pass with it. The deleted-output fixture has two source files on purpose: with one, the existing "no compiled files found" guard masks the bug.

Checked by hand on Python, JavaScript and Beam, which now behave identically: add via `.fsproj`, add via imported `.props`, a `.props` shared by two projects built in turn, remove from the compile list, deleted output regenerated.

Suites: Python 2458 passed, Integration 42 + 116 passed, fantomas clean.

## Not addressed

Found alongside, out of scope here. With the two cache defects fixed, both only surface on a compile that already exits non-zero and prints the error — noisy rather than silent.

- **Output files are written even when the compile fails** (reproduced standalone: exit 1, `a.py` written anyway). This is what poisons the output tree so the *next* run takes the skip path. Fable compiles each file as it type-checks, and project-wide diagnostics only arrive at `FSharpCompilationFinished`, so gating writes on "no errors" means buffering output or restructuring that pipeline.
- **The Python backend emits `raise 1`** — a non-exception — from F#'s error-recovery AST. A symptom of the above: generating code from an AST that failed type-checking.

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -979,12 +979,9 @@ let private areCompiledFilesUpToDate (state: State) (filesToCompile: string[]) =
 
                     upToDate
                 else
-                    // A source with no output must be compiled: it may be newly added, or
-                    // its output may have been deleted. `getFilesToCompile` already selected
-                    // it for that reason, so reporting it as up-to-date here would skip the
-                    // compilation it just asked for and leave the output missing.
-                    // (A file whose generated code is empty is never written, so it also
-                    // lands here and costs this project the skip-compilation shortcut.)
+                    // A missing output means the source is new or its output was deleted, so it
+                    // must be compiled. A file generating empty code is never written and also
+                    // lands here, costing that project the skip-compilation shortcut.
                     Log.verbose (lazy $"Output file {File.relPathToCurDir outPath} does not exist")
 
                     false
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -965,7 +965,7 @@ let private areCompiledFilesUpToDate (state: State) (filesToCompile: string[]) =
             )
             |> Array.forall (fun source ->
                 let outPath = getOutPath state.CliArgs pathResolver source
-                // Empty files are not written to disk so we only check date for existing files
+
                 if IO.File.Exists(outPath) then
                     foundCompiledFile <- true
 
@@ -979,7 +979,15 @@ let private areCompiledFilesUpToDate (state: State) (filesToCompile: string[]) =
 
                     upToDate
                 else
-                    true
+                    // A source with no output must be compiled: it may be newly added, or
+                    // its output may have been deleted. `getFilesToCompile` already selected
+                    // it for that reason, so reporting it as up-to-date here would skip the
+                    // compilation it just asked for and leave the output missing.
+                    // (A file whose generated code is empty is never written, so it also
+                    // lands here and costs this project the skip-compilation shortcut.)
+                    Log.verbose (lazy $"Output file {File.relPathToCurDir outPath} does not exist")
+
+                    false
             )
         // If we don't find compiled files, assume we need recompilation
         upToDate && foundCompiledFile
--- a/src/Fable.Compiler/ProjectCracker.fs
+++ b/src/Fable.Compiler/ProjectCracker.fs
@@ -837,23 +837,17 @@ let loadPrecompiledInfo (opts: CrackerOptions) otherOptions sourceFiles =
         Some info, otherOptions, sourceFiles
     | None -> None, otherOptions, sourceFiles
 
-/// Files besides the project file itself that take part in its MSBuild evaluation
-/// and so can change its `<Compile Include>` list: explicit `<Import Project="..."/>`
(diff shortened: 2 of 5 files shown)


src/Fable.Cli/Main.fs
```diff
@@ -979,7 +979,15 @@ let private areCompiledFilesUpToDate (state: State) (filesToCompile: string[]) =
...
                     upToDate
                 else
-                    true
+                    // A source with no output must be compiled: it may be newly added, or
+                    // its output may have been deleted. `getFilesToCompile` already selected
+                    // it for that reason, so reporting it as up-to-date here would skip the
+                    // compilation it just asked for and leave the output missing.
+                    // (A file whose generated code is empty is never written, so it also
+                    // lands here and costs this project the skip-compilation shortcut.)
+                    Log.verbose (lazy $"Output file {File.relPathToCurDir outPath} does not exist")
```
MangelMaxime: Can you please Claude to be more concise and generate a 1-2 (3 max) lines comment.

I found that in general, concise comments were much easier to read and pollute less the code base.

src/Fable.Compiler/ProjectCracker.fs
```diff
@@ -837,6 +837,79 @@ let loadPrecompiledInfo (opts: CrackerOptions) otherOptions sourceFiles =
...
     | None -> None, otherOptions, sourceFiles
 
+/// Files besides the project file itself that take part in its MSBuild evaluation
+/// and so can change its `<Compile Include>` list: explicit `<Import Project="..."/>`
+/// targets (transitively) and the implicit Directory.Build.props/targets and
+/// Directory.Packages.props found by walking up from the project directory.
+///
+/// Without these, adding a source file through a shared .props leaves every
+/// .fsproj timestamp untouched, and the cached project options — including the
+/// now-stale source list — are reused.
```
MangelMaxime: Could perhaps do the same here

src/Fable.Cli/Main.fs
dbrattli: Done in 52e4d47 — this one is 3 lines now.

src/Fable.Compiler/ProjectCracker.fs
dbrattli: Done — trimmed this doc comment to 3 lines, and applied the same to the other comments the PR adds (`expandMacros`, the cache-key comment, and the four in `CacheInvalidationTests.fs`).

src/Fable.Compiler/ProjectCracker.fs
```diff
@@ -837,6 +837,79 @@
...
+let private getProjectImports (projFile: string) : string list =
+    let imports = ResizeArray()
+    let visited = HashSet<string>(StringComparer.OrdinalIgnoreCase)
+
+    // Only the directory macros are expanded. An import path built from any other
+    // property cannot be resolved without evaluating MSBuild, so it is skipped:
+    // missing an import costs a stale cache, guessing wrong costs a bogus path.
+    let expandMacros (fileDir: string) (path: string) =
+        let withSep = fileDir + string IO.Path.DirectorySeparatorChar
```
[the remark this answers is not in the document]
dbrattli: Fixed: `string<char> IO.Path.DirectorySeparatorChar`.

PR #4915, opened 2026-08-27
fix(all): Allow private inline functions to use private values

### LINKED ISSUE #14: Compile error with Fable >=5.5.0
Compiling with 5.5.0+ results in the following errors when running `npm run test`

```
.\src\XParsec\OperatorParsing.fs(315,31): (315,38) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(315,52): (315,59) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(321,31): (321,38) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(326,31): (326,38) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(316,40): (316,47) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(321,66): (321,73) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(326,66): (326,73) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(327,46): (327,53) error FABLE: The value 'failure' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(377,14): (377,58) error FABLE: The value 'parseLhsInternal' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(381,12): (381,95) error FABLE: The value 'parseRhsInternal' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
.\src\XParsec\OperatorParsing.fs(382,27): (382,50) error FABLE: The value 'mergeWithError' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible
Compilation failed
```

It appears the issue is caused by https://github.com/fable-compiler/Fable/pull/4701 being too restrictive about inlining. i.e. a private function should be able to inline into another private function in the same module.

roboz0r:
Allows a `private inline` function to call other `private` functions and values in the same file.

- A refinement to the restriction added in https://github.com/fable-compiler/Fable/pull/4701 
- Fixes https://github.com/roboz0r/XParsec/issues/14

### DIFF
--- a/src/Fable.Transforms/FSharp2Fable.Util.fs
+++ b/src/Fable.Transforms/FSharp2Fable.Util.fs
@@ -2445,9 +2445,6 @@ module Util =
             else
                 Some(entityIdent com ent.Ref)
 
-    /// An inline function that is not visible outside of the file declaring it can only be called
-    /// from that file, so `resolveInlineExpr` turns the references in its body back into plain
-    /// identifiers instead of imports. This makes it safe for its body to use private members.
     let private isInlinedInDeclaringFileOnly (ctx: Context) =
         match ctx.PrecompilingInlineFunction with
         | Some inlineMemb -> not (isNotPrivate inlineMemb)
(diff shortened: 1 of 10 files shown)


MangelMaxime:
Looking at this PR, please don't merge it as it seems like there are holes in the implementations.

roboz0r:
Happy to take another look at this. I'm not very familiar with the codebase so I was partly trusting Claude to help me find the relevant functions to change, and the tests looked reasonable. Was there a specific gap you noticed or is it a wider structural issue?

PR #4917, opened 2026-08-27
fix(rust): F# exceptions, non-exhaustive matches and obj type tests

Thorium:
- makeRecord now supplies the synthetic __base__ value, so constructing an F# exception no longer crashes the compiler with "The lists had different lengths"
- get_Message skips that base field, whose null innerException faulted at runtime
- MatchFailureException is declared in the runtime library, the way the other targets already get it, so a non-exhaustive match compiles
- try_downcast returns a value rather than a reference, so `:? string as s` binds an s that is usable where a string is expected


src/Fable.Transforms/Rust/Fable2Rust.fs
```diff
@@ -2002,10 +2002,23 @@ module Util =
...
+        // value, and the placeholder is always overwritten before it can be read.
+        | Fable.String -> Fable.Value(Fable.StringConstant "", None) |> transformExpr com ctx
+        | Fable.Option(genArg, isStruct) ->
+            Fable.Value(Fable.NewOption(None, genArg, isStruct), None)
+            |> transformExpr com ctx
+        | Fable.Array(genArg, kind) ->
+            Fable.Value(Fable.NewArray(Fable.ArrayValues [], genArg, kind), None)
+            |> transformExpr com ctx
+        | Fable.List genArg -> Fable.Value(Fable.NewList(None, genArg), None) |> transformExpr com ctx
+
```
ncave: Strings, options, arrays and lists already have `NullableRef` implementations, so perhaps just call `makeNull com ctx typ`,
or add them to the `isConcreteLrcRef`

Thorium: Thanks, done.

PR #4919, opened 2026-08-28
feat(all): Quotations: DerivedPatterns, and captured locals as Value nodes

Thorium:
- AndAlso, OrElse and SpecificCall are mapped, with an implementation in each
  runtime that has a quotation module. They were the last patterns still
  reporting "not supported by Fable", and && and || desugar into IfThenElse, so
  without them a consumer has to re-derive that shape by hand
- a local captured by a quotation is now spliced in as a Value holding its
  value, the way .NET does, instead of a Var carrying only a name. That needed
  three changes together: the emitter tracks which names the quotation binds,
  getSubExpressions lets analysis see into a Quote, and canInlineArg stops
  removing a binding whose only reference lives inside one

### DIFF
--- a/src/Fable.Transforms/FableTransforms.fs
+++ b/src/Fable.Transforms/FableTransforms.fs
@@ -255,40 +255,56 @@ let noSideEffectBeforeIdent identName expr =
 
     findIdentOrSideEffect expr && not sideEffect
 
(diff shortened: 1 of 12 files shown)


ncave (approved):

ncave:
LGTM

PR #4923, opened 2026-08-29
fix: report Fable's own logs from the Fable.Compiler API

### LINKED ISSUE #4922: Fable.Compiler's compile APIs drop the compiler's own logs, so Fable errors are lost
## Summary

`CodeServices.compileMultipleFilesToJavaScript` and `CodeServices.compileFileToJavaScript` fill `CompileResult.Diagnostics` from the FCS type-check results only. Errors that Fable itself raises while translating a file (`com.AddLog(..., Severity.Error, ...)`, so every `addErrorAndReturnNull` among others) are never returned to the caller, and the emitted JavaScript comes back as if nothing went wrong.

Code that type-checks but that Fable cannot translate therefore compiles "successfully" into JavaScript that silently does the wrong thing, and a host built on `Fable.Compiler` has no way to find out.

## Where

Links are against `616f81bc`.

The compiler is upcast to the `Compiler` interface as soon as it is constructed:

https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Compiler/Library.fs#L223-L231

`Logs` is a member of `CompilerImpl` ([State.fs#L307](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Transforms/State.fs#L307)), not of the `Compiler` interface, so nothing can be read from that binding, and the instance is discarded once `compileFileToJs` returns. What comes back instead is:

- [Library.fs#L247](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Compiler/Library.fs#L247): `Diagnostics = typeCheckProjectResult.ProjectCheckResults.Diagnostics`
- [Library.fs#L338](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Compiler/Library.fs#L338): `Diagnostics = checkProjectResult.Diagnostics`

`Fable.Cli` does not have the problem, because it keeps the concrete type ([Main.fs#L228](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Cli/Main.fs#L228), `compileFile (com: CompilerImpl)`), reads [`com.Logs`](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Cli/Main.fs#L248), merges them with the FCS diagnostics and fails the run on any `Severity.Error` ([Main.fs#L1523](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Cli/Main.fs#L1523)).

## Reproduction

```fsharp
module Math

let blocking () = Async.RunSynchronously (async { return 42 })
```

`RunSynchronously` maps to `None` in [Replacements.fs#L3956](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Transforms/Replacements.fs#L3956), so the call is reported as "... is not supported by Fable" through `addErrorAndReturnNull` ([FSharp2Fable.Util.fs#L2772](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Transforms/FSharp2Fable.Util.fs#L2772)).

Compiled through `compileMultipleFilesToJavaScript`, `Diagnostics` is empty and `CompiledFiles` holds:

```js
export function blocking() {
	return null;
}
```

I ran into this through [vite-plugin-fable](https://github.com/fable-compiler/vite-plugin-fable), which drives that function from a daemon: `vite build` printed no diagnostic, exited 0, and put `return null` in the production bundle.

## Suggested fix

Keep the concrete `CompilerImpl` in both functions and surface what it collected after `compileFileToJs`.

The shape is your call, which is why this is an issue rather than a PR: `CompileResult.Diagnostics` is `FSharpDiagnostic array`, while `com.Logs` is `LogEntry array` ([State.fs#L265](https://github.com/fable-compiler/Fable/blob/616f81bcc678131f95ef808e4ce06b382499d880/src/Fable.Transforms/State.fs#L265)), so it is either a new field on `CompileResult` or a conversion into `FSharpDiagnostic` at the boundary. Happy to send a PR once you have a preference.

nojaf:
`CodeServices.compileMultipleFilesToJavaScript` and `compileFileToJavaScript` filled `CompileResult.Diagnostics` from the FCS type-check results only. The `CompilerImpl` that holds Fable's own logs was upcast to the `Compiler` interface at the binding and then discarded, so every error Fable raised while translating a file was dropped. Code that type-checks but that Fable cannot translate compiled "successfully" into JavaScript that silently does the wrong thing, and a host built on `Fable.Compiler` had no way to find out.

This surfaced while working on [vite-plugin-fable](https://github.com/fable-compiler/vite-plugin-fable), which drives `compileMultipleFilesToJavaScript` from a compiler daemon. On the repro in the issue, `vite build` printed no diagnostic, exited 0, and put `return null` in the production bundle. Silent wrong output rather than a crash, which is why it went unnoticed for so long and why it is worth a breaking change to fix properly.

Both functions now keep the concrete compiler and return everything that went wrong as `LogEntry array`, tagged `FABLE` or `FSHARP`, which is the representation `Fable.Cli` already uses. `compileFileToFableAST` reports the same way. The conversion from FCS diagnostics to `LogEntry` moves out of `Fable.Cli` into `Fable.Compiler`, so there is one implementation of it rather than two that can drift.

Logs are collected per file and concatenated afterwards, because `AddLog` takes no lock and the files are compiled under `Async.Parallel`.

Nothing drove `CodeServices` from a test, which is a large part of why this went unnoticed, so this adds coverage that calls the library API directly: the `Async.RunSynchronously` snippet from the issue (it type-checks, but Fable has no translation for it) has to come back as a `Severity.Error` tagged `FABLE`, and an ordinary F# type error as one tagged `FSHARP`.

### Breaking change

`CompileResult.Diagnostics: FSharpDiagnostic array` is replaced by `CompileResult.Logs: LogEntry array`, and `compileFileToFableAST` now returns `FableASTResult` instead of `AST.Fable.File`.

Adding `Logs` alongside `Diagnostics` would have been non-breaking, and that is its whole appeal, but it leaves the default wrong: a consumer reading `Diagnostics`, a field named as though it were the complete answer, would still see nothing. "Nobody knew there was somewhere else to look" is precisely the bug being fixed.

Since this is marked `fix!`, `shipit` will bump the affected packages to a new major. Happy to reshape this into a non-breaking `Logs` addition first and drop `Diagnostics` in the next major instead, if that suits the release plan better.

### Follow-up: `Fable.Cli`

`Fable.Cli` still collects `com.Logs` in its own compile pipeline. Only the duplicated FCS-diagnostic conversion moved out of it here.

Making the CLI consume `CodeServices` outright would leave one implementation of "compile a project and report what went wrong" rather than two, and the CLI would keep that implementation honest. It is not in this PR because `CodeServices` would first have to grow watch mode, precompiled info, inline expression collection, watch dependencies and the non-JavaScript targets, which is a much larger change than this fix. Worth discussing before anyone starts on it.

### Verification

`dotnet fable` output is unchanged by this PR for both a Fable error and an F# error: the reported diagnostics and the emitted JavaScript were compared before and after, since the CLI is being rewired to a new source for the same data. `./build.sh test javascript` and the `tests/Integration/Compiler` suite are green.

Fixes #4922.

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -12,6 +12,7 @@ open Fable
 open Fable.AST
 open Fable.Transforms
 open Fable.Transforms.State
+open Fable.Compiler
 open Fable.Compiler.ProjectCracker
 open Fable.Compiler.Util
 
@@ -87,35 +88,6 @@ module private Util =
         logs
         |> Seq.filter (fun log -> log.Severity = Severity.Error)
         |> Seq.iter (fun log -> Log.error (formatLog rootDir log))
-
-    let getFSharpDiagnostics (diagnostics: FSharpDiagnostic array) =
-        diagnostics
-        |> Array.map (fun er ->
-            let severity =
-                match er.Severity with
-                | FSharpDiagnosticSeverity.Hidden
-                | FSharpDiagnosticSeverity.Info -> Severity.Info
-                | FSharpDiagnosticSeverity.Warning -> Severity.Warning
-                | FSharpDiagnosticSeverity.Error -> Severity.Error
-
-            let range =
-                SourceLocation.Create(
-                    start =
-                        {
-                            line = er.StartLine
-                            column = er.StartColumn + 1
-                        },
-                    ``end`` =
-                        {
-                            line = er.EndLine
-                            column = er.EndColumn + 1
-                        }
-                )
-
-            let msg = $"%s{er.Message} (code %i{er.ErrorNumber})"
-
-            LogEntry.Make(severity, msg, fileName = er.FileName, range = range, tag = "FSHARP")
-        )
 
     let getOutPath (cliArgs: CliArgs) pathResolver file =
         match cliArgs.CompilerOptions.Language with
@@ -672,7 +644,7 @@ and FableCompiler(checker: InteractiveChecker, projCracked: ProjectCracked, fabl
 
                         let state =
                             { state with
-                                FSharpLogs = getFSharpDiagnostics results.Diagnostics
+                                FSharpLogs = CodeServices.getFSharpDiagnostics results.Diagnostics
                                 HasFSharpCompilationFinished = true
                             }
 
@@ -1590,7 +1562,7 @@ let private compilationCycle (state: State) (changes: ISet<string>) =
 
                             match result with
                             | Some ex ->
-                                getFSharpDiagnostics diagnostics |> logErrors cliArgs.RootDir
+                                CodeServices.getFSharpDiagnostics diagnostics |> logErrors cliArgs.RootDir
                                 Log.error (ex.Message)
                                 return 1
                             | None ->
--- a/src/Fable.Compiler/Library.fs
+++ b/src/Fable.Compiler/Library.fs
@@ -17,12 +17,24 @@ type TypeCheckProjectResult =
         ProjectCheckResults: FSharpCheckProjectResults
     }
 
+type FableASTResult =
+    {
+        /// The current file transformed into the Fable AST
+        FableAST: AST.Fable.File
+        /// Everything the compiler reported: the diagnostics of the F# project as checked up to
+        /// the current file, tagged "FSHARP", followed by the logs Fable raised while
+        /// transforming that file, tagged "FABLE".
+        Logs: LogEntry array
+    }
+
 type CompileResult =
     {
         /// A map of absolute file path to transpiled JavaScript code
         CompiledFiles: Map<string, string>
-        /// Diagnostics of the entire checked F# project
-        Diagnostics: FSharpDiagnostic array
+        /// Everything the compiler reported: the diagnostics of the entire checked F# project,
+        /// tagged "FSHARP", followed by the logs Fable raised while translating the compiled
+        /// files, tagged "FABLE".
+        Logs: LogEntry array
     }
 
 type BabelWriter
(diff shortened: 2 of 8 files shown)


nojaf:
@MangelMaxime one thing to flag before this gets merged: the release impact.

This is marked `fix!`, so `shipit` will read it as a breaking change. Going by `feat(standalone)!: reference a precompiled library` (#4899, #4900), which produced `fable-standalone 3.0.0`, merging this will open a release PR bumping Fable, `Fable.Compiler` and the other packages sharing that commit pool to **6.0.0**. A major bump triggered by a fix is a bit unusual, so I would rather you decide that than have it show up as a surprise release PR.

The break itself is small and, as far as I know, I am the only consumer of `CodeServices` today (through [vite-plugin-fable](https://github.com/fable-compiler/vite-plugin-fable)), so nobody else should be affected in practice.

If a major is not what you want right now, the alternative is to land `Logs` alongside `Diagnostics` in a patch release so downstream is unblocked, then drop `Diagnostics` in the next major. I did not go that way because it leaves the default wrong (`Diagnostics` reads as though it were the complete answer and still would not be), but it is an easy change to make here if you prefer it. Your call.

MangelMaxime:
@nojaf I didn't review the PR yet but to answer your comment about major bump.

Because of how Fable versioning works, I think we should keep that change as a minor bump.

There are 3 solutions:

1. Remove the `!` from the PR title
2. Keep `!` from the PR title, because you find it indicative and add a [`force_version`](https://github.com/easybuild-org/EasyBuild.ShipIt#force_version) to the CHANGELOG which will be wrongly computed. This is for example, something I needed to in the past release of Fable because `fable-standalone` changes were also bumping Fable CLI which I didn't want.
3. Or we make it so the code does not break, but because Fable.Compiler is not a widely used library I think this is ok to introduce breaking changes on minor version if it avoid working twice on the same issue.

nojaf:
@MangelMaxime thanks, I removed the `!` from the title.

MangelMaxime:
Thank you

nojaf:
Thanks @MangelMaxime, this worked out nicely!

https://github.com/fable-compiler/vite-plugin-fable/releases/tag/v0.4.0

PR #4924, opened 2026-08-30
fix: preserve project FSharp.Core reference

dbrattli:
## Summary
- preserve the project's resolved FSharp.Core reference when cracking projects
- avoid injecting the compiler's bundled FSharp.Core into netstandard projects
- add a netstandard2.0 Python task-expression fixture covering let! and try/with

## Regression origin
This was introduced in Fable 5.14 by [#4900](https://github.com/fable-compiler/Fable/pull/4900) (`feat(standalone)!: reference a precompiled library`). Its new Thoth.Json dependencies brought FSharp.Core 10.1.203 into the CLI (5.13 used 10.0.110). ProjectCracker had long discarded the project's FSharp.Core reference and appended the compiler's own reference. That was incompatible with a netstandard2.0 project resolving FSharp.Core 6.0.0: FCS lowered the task expression to its recovery `Raise(1)` form, which the Python backend printed as `raise 1` without a diagnostic. The source-less-member guard in #4894 is not on this path.

## Validation
- dotnet build src/Fable.Cli/Fable.Cli.fsproj --no-restore -m:1
- fixture compiles to Python task-builder code; with the pre-fix cracker it emits raise 1
- Fable.Python AsyncContextManager tests with the local compiler: 4 passed
- Fable.Python just build

### DIFF
--- a/src/Fable.Compiler/ProjectCracker.fs
+++ b/src/Fable.Compiler/ProjectCracker.fs
@@ -1002,7 +1002,6 @@ let getFullProjectOpts (resolver: ProjectCrackerResolver) (opts: CrackerOptions)
             let ignoredRefs =
                 HashSet
                     [
-                        "FSharp.Core"
                         "WindowsBase"
                         "Microsoft.Win32.Primitives"
                         "Microsoft.VisualBasic"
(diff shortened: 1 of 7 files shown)


MangelMaxime:
It turns out JavaScript had the same issue, so we can use the existing snapshot testing to write the test.

One day, we will see how we can modernize our system 🤞

Letting CI run and I will merge after

PR #4927, opened 2026-08-31
fix(rust): wait for the other files instead of sleeping before the last one.

Thorium:
Current 1 second is not enough on a big project.

The Rust target's last file re-exports the namespaces every other file registers into a shared registry while it compiles, so it has to be compiled after all of them. The fixed one-second sleep only shrank that race: on a project whose files take longer than a second to compile, the entry file still won it, and the crate's merged namespaces silently lost whichever modules had not registered yet, non-deterministically.

The last file is now held in the compiler state when it type-checks early, counted as pending so the finished-check keeps waiting, and released for compilation by the completion handler once the results of the other files are in.

### DIFF
--- a/src/Fable.Cli/Main.fs
+++ b/src/Fable.Cli/Main.fs
@@ -502,6 +502,9 @@ type FableCompilerState =
         FilesCheckedButNotCompiled: Set<string>
         FableFilesToCompileExpectedCount: int
         FableFilesCompiledCount: int
+        /// For Rust: the project's last file, held back until every other file
+        /// has compiled. See the note where it is set.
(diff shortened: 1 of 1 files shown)


Thorium:
If you wonder what I'm using this for: https://github.com/Thorium/SQLProvider.Fable

Write F# Type-provider database queries, get Rust (or Erlang or Node.Js) executing SQL on whatever PostgreSql/MySql/Sqlite database. Not as convenient as the real F# SQLProvider (which has years of work behind it and is tested in large systems); rather, it's its little sister as a tech concept research.

PR #4939, opened 2026-09-02
fix(rust): Fix interface casting, move reflection into declarations

ncave:


ncave:
@Thorium For performance reasons, the `type_info` was moved to the type declaration, and using it requires enabling the `reflection` feature.

PR #4985, opened 2026-09-21
fix(js/ts): Fix Dictionary.Remove and HashSet.Remove to remove empty buckets

### LINKED ISSUE #4984: [JS/TS] `Dictionary.Remove` / `HashSet.Remove` leave an empty bucket per key hash ever held
Both `Remove` members splice the entry out of its bucket and never drop the bucket from `hashMap` when it empties:
e.g. in https://github.com/fable-compiler/Fable/blob/50a37f78f31394d61349b09a025c1713529d8ec0/src/fable-library-ts/MutableSet.fs#L60-L65

A long-lived table whose keys churn keeps one `Map` entry and one empty array for every key hash it has ever held, while `Count` stays small.

Each empty bucket cost ~100B in my personal use case.

`Clear` resets `hashMap` and is the only way to get the memory back.

ncave:
- Fixes #4984


ncave:
We can't directly tests the inner dictionary size through the public API, but hopefully this fixes it.

